Issue #10, opened 2020-04-01
add setup.py, setuptools


cmoussa1:
In order to build this project and run tests, we need to add a `setup.py` file and use setuptools.

PR #11, opened 2020-04-01
repo: Add setup.py, setuptools

cmoussa1:
As discussed in #6 and #10, in order to run unit tests, build a package, wheel, etc, there is a need for a `setup.py` file in the repository. This PR adds `setup.py` and the **build/dist/test** directories to the project. 

It also combines the two original source code directories, **src** and **cmd** into just one directory, **accounting**. 

Also, in order to use Python's [test discovery](https://docs.python.org/2/library/unittest.html#test-discovery) with this new directory structure, the unit test files should be renamed to `test_*.py`. This PR renames the single unit test file currently in this repo, **fs_unittest.py** ➡️ **test_fairshare.py**. This will allow us to run the unit test files like the following:

```
> python -m unittest discover

Ran 5 tests in 0.001s
```


cmoussa1:
Ah, just realized that **test_fairshare.py** is referencing an outdated .db file 🤦‍♂️. I'm going to refactor this unit test file so it doesn't use it anymore.

grondo:
I would say it would also be useful to add a github workflow to run your tests.

It is pretty straightforward, but let me know if you need any help.

If you add that, might as well add the `pr-validator` action, mergify, etc.

cmoussa1:
Thanks @grondo! I'll let you know if I need any help. Would you recommend I just add the **.yml** to this PR?

grondo:
Yes, I'd just add it here. Then that is a nice verification that everything is setup correctly and the tests pass.

cmoussa1:
Okay, just pushed a couple of changes/additions to this PR that I think will really clean up this repo for future work:

- the **fairshare.py** and **fs_unittest.py** files were really just template source code files and didn't really contain anything significant in them, so I've removed them from the directory for now. I will add those back in once I have something meaningful pertaining to fairshare. This outdates the original comment in the beginning of this PR:

> Also, in order to use Python's test discovery with this new directory structure, the unit test files should be renamed to test_*.py. This PR renames the single unit test file currently in this repo, fs_unittest.py ➡️ test_fairshare.py. This will allow us to run the unit test files like the following:
`python -m unittest discover` since that test file is no longer in the directory. 

- per @grondo's suggestion of adding a Github workflow, a workflow file has been added that is configured to run on PRs. It uses the `validate commit` check from **flux-core**, as well as runs the Black formatter and runs the unit tests in the directory (which, for now is just one example unit test file to ensure that the unit tests run correctly).

grondo:
Don't mean to add more work for you here, but you might as well also add a `.mergify.yml`. Something like this should work:

```yaml
pull_request_rules:
  - name: rebase and merge when passing all checks
    conditions:
      - base=master
      - status-success="validate commits"
      - status-success="Lint with flake8"
      - status-success="Test with unittest"
      - label="merge-when-passing"
      - label!="work-in-progress"
      - "approved-reviews-by=@flux-framework/core"
      - "#approved-reviews-by>0"
      - "#changes-requested-reviews-by=0"
      - -title~=^\[*[Ww][Ii][Pp]
    actions:
      merge:
        method: merge
        strict: smart
        strict_method: rebase

```

cmoussa1:
Ugh, I pushed the commit that adds `.mergify.yml` to this branch while Github was acting up, and I don't see the commit showing up on this page, but I see the commit on my branch on my forked repo.

grondo:
Use the `--force` @cmoussa1. Trust in your local git branch...

grondo:
Looks like a couple binary files were added in commit ffb3ce7e3898d4bfbf2bfca610f553fcb469ce7a. Perhaps if you remove those and force push again, GitHub will finally update this PR page?

cmoussa1:
Looks like the force push kicked the tires! I can see the almighty mergify commit now 🙏  thanks @grondo !

grondo:
I still see `dist/flux-accounting-cmoussa1-0.0.1.tar.gz` committed in ffb3ce7e3898d4bfbf2bfca610f553fcb469ce7a

cmoussa1:
yeah, I only removed one at a time and was moving kind of slow, sorry! both binary files should be removed from `dist/` now

cmoussa1:
Also pushed a change to ignore both `__pycache__` directories in **test/** and **accounting/**.

grondo:
It appears some `*.pyc` files crept into some commits.

grondo:
Also, is the `flux_accounting_cmoussa1.egg-info/` supposed to be checked in? If it is indeed to be included under version control, having a username in the directory is slightly odd, should it just bee `flux_accounting.egg-info/`? (note: I have no idea what the `egg-info` directory is for so I apologize if this is a dumb question)

cmoussa1:
Thanks @grondo - just squashed a commit on top of adding the **__pycache__** directories to get rid of those `.pyc` files.

Per your question about the `.egg-info`, after some quick googling: 

> A "Python egg" is a logical structure embodying the release of a specific version of a Python project, comprising its code, resources, and metadata. There are multiple formats that can be used to physically encode a Python egg, and others can be developed. However, a key principle of Python eggs is that they should be discoverable and importable. That is, it should be possible for a Python application to easily and efficiently find out what eggs are present on a system, and to ensure that the desired eggs' contents are importable.
> The .egg format is well-suited to distribution and the easy uninstallation or upgrades of code, since the project is essentially self-contained within a single directory or file, unmingled with any other projects' code or resources. It also makes it possible to have multiple versions of a project simultaneously installed, such that individual programs can select the versions they wish to use.

Looks like the `.egg-info` directory contains some meta data about the project, so I agree with you, it should probably just contain **flux-accounting** without my username (I set this directory up by following a tutorial, which is why I included my username in the name).  I can go ahead and make this change now.

cmoussa1:
OK, pushed a change to **setup.py** to build `.egg-info` using the name **flux-accounting** instead of **flux_accounting_cmoussa1**.

cmoussa1:
Thanks @grondo! That plan sounds good to me. One question I have is for some advice for the **build/** in this PR. I believe this directory gets created after running `python3 setup.py sdist bdist_wheel` (this command was copied from the Python [packaging tutorial](https://packaging.python.org/tutorials/packaging-projects/) I followed). Should this directory also be added to **.gitignore**? I'll ping @trws and @SteVwonder for some expertise if they don't mind.

build/lib/accounting/write_jobs.py
```diff
@@ -0,0 +1,135 @@
...
+###############################################################
+# Copyright 2020 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+from __future__ import print_function
```
SteVwonder, line 12: Since we've gone python3-only, we don't need this compatibility import anymore.

build/lib/accounting/write_jobs.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+        print("insertion into inactive table successful")
+
+    # commit changes
+    conn.commit()
+
+    # make sure the writes to DB were successful and we can
+    # retrieve them
+    sel_statement = "SELECT * FROM inactive"
+    print(pd.read_sql_query(sel_statement, conn))
```
SteVwonder, line 63: Is the plan going forward to use pandas to execute the sql queries and interact with the results?  That actually seems totally reasonable to me.  Working with the returned values via Pandas will be much faster than with Python directly.  If that is the plan, the Pandas requirement should be documented somewhere and included in the `requirements.txt` (which looks empty right now).

If this is just a convenience thing for debugging, and you plan to stick with the Python standard library for the main code, I'd wrap this print up in a conditional import:
```suggestion
    try:
        import pandas as pd
        print(pd.read_sql_query(sel_statement, conn))
    except ImportError:
        print("Failed to import pandas, skipping debug print of results")
```

build/lib/accounting/write_jobs.py
```diff
@@ -0,0 +1,135 @@
...
+
+import flux.job
+import flux.constants
+import flux.util
+from flux.core.inner import raw
+
+
+def write_to_db(inactive):
+    # open connection to database
+    print("Opening JobCompletion DB...")
```
SteVwonder, line 27: It doesn't have to be in this PR, but I would look into converting these prints into calls to `logging`: https://docs.python.org/3/library/logging.html

Down the road, as more and more stuff is logged, it will give us fine-grained control over how much gets printed to the terminal and how much goes to a file.  There are some examples of it being used in the python front-end commands in flux-core, like `flux-mini.py`.  Unfortunately, the only examples appear to be `.error` while there are plenty of other options, like `.debug`, `.info`, and `.warn`.

build/lib/accounting/write_jobs.py
```diff
@@ -0,0 +1,135 @@
...
+    print("write complete\n")
+
+    # <demo> show contents of last job's job id
+    print("showing contents of tmp file...")
+    f = open("tmp_id.txt", "r")
+    for line in f:
+        print(line)
+
+
+def main():
```
SteVwonder, line 105: I recommend decorating this main function with our `util.CLIMain` decorator, which captures uncaught exceptions and pretty-prints them (i.e., avoids dumping a massive stack-trace unless you set a specific environment variable).

Here is an example of that in `flux-mini.py`: https://github.com/flux-framework/flux-core/blob/a06041b4a3e981a11e5b46c219beb7853e00a214/src/cmd/flux-mini.py#L297

setup.py
```diff
@@ -0,0 +1,22 @@
...
+    author_email="moussa1@llnl.gov",
+    description="User accounting interface for Flux framework",
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/flux-framework/flux-accounting",
+    packages=setuptools.find_packages(),
+    classifiers=[
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: GNU Lesser General Public License",
+        "Operating System :: OS Independent",
```
SteVwonder, line 19:
```suggestion
        "Operating System :: POSIX :: Linux",
```

build/lib/accounting/write_jobs.py
SteVwonder: Also note, that if you want to use pandas to drive this script, we'll need to settle on a minimum support version.  Pandas recently tagged 1.0 (which I think has some breaking changes from 0.23), but I doubt that is available on TOSS4 (which I think is built on top of RHEL 8......I've lost track).

cmoussa1:
Thanks for your feedback @SteVwonder - that was all really helpful! The suggestions you made for **write_jobs.py** will definitely be used, but I'm wondering if the code changes you suggested for that file might be better suited for #6, where I am heavily refactoring **create_db.py** as well - I could then make those changes there to keep the scope of work separated (this PR to implement setuptools, and #6 to write a database creation script and job retrieval script). This PR could definitely use the additional changes you suggested - adding the `pandas` dependency and changing the OS supported in **setup.py**.

Actually, now that I think about it, much of the code in **write_jobs.py** probably will be changed anyways because my understanding is flux-accounting will now instead be querying the job archive DB in flux-core/flux-framework PR [#2880](https://github.com/flux-framework/flux-core/pull/2880), instead of using the `list-inactive` Python binding. 

Then again, maybe it's better to just make those changes here since this PR is setting up the directory to be built. What do you think?

build/lib/accounting/write_jobs.py
cmoussa1: I am planning on using pandas - it provides a lot of nice formatting options with SQLite queries along with executing queries. Right now, I am using `0.25.3`. I should have added that to the requirements page, thank you for pointing that out!

trws:
I’d say keep build out of version control. As to the rest, I need to take some more time to look this over, but it would probably be a good idea to add a pyproject.toml (a slightly more modern way to specify what tools are needed to build a project) and it might help reduce some of the duplication, or even make the setup.py unnecessary if it’s simple enough.

trws:
Always happy with pandas. Best data frame this side of R, good choice.

cmoussa1:
Thanks again for your feedback and suggestions @SteVwonder - I went ahead and removed the compatibility import from **write_jobs.py**, added `pandas` to **requirements.txt**, and changed the OS classifier in **setup.py**. I was thinking the other two suggestions you made in **write_jobs** could be saved for the other outstanding PR.

SteVwonder:
Thanks @cmoussa1!  Only other thing I think is the `flux-accounting.egg-info` directory.  That all gets auto-generated from the `setup.py` script, right?  If so, I'd leave that out of version control.

dongahn:
Just a quickie. I still see 83be3130054dd1b5bd12e681cbc45ff924c43627 with *.pyc files in this PR. Is it only me?

cmoussa1:
No, you were right @dongahn! I saw them too. I went ahead and pushed a change so they should be removed from the commit history, as well as removed the `.egg-info` directory from version control.

trws:
It’s not just you, a rebase or squash to excise those would be good.

dongahn:
Great! Thank you @cmoussa1 and @trws!

cmoussa1:
Thanks for all the help and reviewing! 😁 Are there any other things I should take care of in this PR before the big green merge button is pressed?

SteVwonder:
Ah. This repo does have mergify yet, that's why there is no label. LOL.  I'm happy to push the button :)

cmoussa1:
Thanks @SteVwonder !

PR #14, opened 2020-05-04
repo: add install instructions, Makefile to install dependencies and run unit tests 

### LINKED ISSUE #12: Add dependencies into README.md etc
It would be good to add dependencies needed to play with these tools. I created a docker off of flux-sched, which seems to bring in most but not all: like pandas.

I was a bit lost in my attempt to play with these tools. If these tools can be used to some extent at this point, it will be good to add them to README.md.

Finally, a typo in the last sentence:

< ...  retrieve user account information amd job record history and 
> ... retrieve user account information and job ...

cmoussa1:
Noted by @dongahn in #6 and #12, the install and unit test instructions in the **README.md** were not clear/non-existent. This PR adds a **Makefile** that installs the dependencies listed in **requirements.txt** as well as runs the unit tests, so both can be run with `make check`.

```
$ make check
pip3 install -r requirements.txt
Collecting pandas==0.25.3
Installing collected packages: pandas
Successfully installed pandas-0.25.3
python3 -m unittest discover
....
Ran 4 tests in 0.021s

OK
```

**README.md** has also been updated to reflect these instructions. 

Fixes #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,14 +2,43 @@ _NOTE: The interfaces of flux-accounting are being actively developed and are no
 
 ## flux-accounting
 
-Development for a bank/accounting interface for the Flux resource manager. Writes and saves job step history, user account information, and priority calculation to persistent storage using Python's SQLite3 API. 
+Development for a bank/accounting interface for the Flux resource manager. Writes and saves user account information to persistent storage using Python's SQLite3 API.
 
-##### job step history
+### Build Requirements
 
-Running as a cron job, `write_jobs.py` uses a flux-core RPC to retrieve job data that have recently transitioned to an INACTIVE state in the past _x_ minutes. It parses the job record data and writes it to a SQLite database, which can be queried at a later time. 
+flux-accounting requires the following packages to build:
+
+| centos8       | ubuntu      | version |
+| ------        | --------    | ------- |
+| python3-devel | python3-dev | >= 3.6  |
+| python3-pip   | python3-pip | 20.0.2  |
+| tox           | tox         | 3.15.0  |
+
+### Install Instructions
+
+You can install the dependencies required by flux-accounting (located in **requirements.txt**) with the following command:
+
+```
+$ pip3 install -r requirements.txt
+```
+
+### Test Instructions
+
+Run the unit tests with `tox` to ensure the correctness of this package on your platform::
+
+```
+$ tox
+python3.6 run-test: commands[0] | python -m unittest discover
+....
+----------------------------------------------------------------------
+Ran 4 tests in 0.008s
+
+OK
+_______________________________summary _______________________________
+  python3.6: commands succeeded
+  congratulations :)
+```
 
 ##### user account information
 
-Users have banks that they can charge their jobs against; this will affect their priority when submitting future jobs in order to maintain a fair balance of resources between users running jobs on a single cluster. Multiple factors play a role in determining a user's job priority, but in general, the amount of resources used vs. the resource limits initially allocated will either lower or heighten a user's job priority. 
-
-Priority will be calculated by calling another Python file, `fairshare.py`, which will retrieve user account information amd job record history and perform a "fairshare" calculation to determine a user's job priority on a cluster. 
+The accounting table in this database stores information like user name and ID, the account to submit jobs against, an optional parent account, the shares allocated to the user, as well as static limits, including max jobs submitted per user at a given time and max wall time per job per user.
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ accounting/FluxAccounting.db
 accounting/db_creation.log
 test/FluxAccounting.db
 build/
+.tox
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,6 @@
+# minimal Makefile to run Python unit tests
+# for flux-accounting
+
+check:
+	pip3 install -r requirements.txt
+	python3 -m unittest discover
--- a/accounting/test/docker/centos8/Dockerfile.centos8
+++ b/accounting/test/docker/centos8/Dockerfile.centos8
@@ -1,0 +1,20 @@
+# Docker file for a slim Ubuntu-based Python3 image
+
+FROM centos:8
+
+LABEL maintainer="Christopher Moussa <moussa1@llnl.gov>"
+
+# add flux-accounting/ to src/ in container
+ADD . src/
+
+RUN yum -y update \
+  && yum install -y python3-pip python3-devel git \
+  && cd /usr/local/bin \
+  && ln -s /usr/bin/python3 python \
+  && pip3 install --user --upgrade pip \
+  && pip3 install tox
+
+# go into flux-accounting, install dependencies, and run unit tests
+RUN cd src/ \
+  && pip3 install --user -r requirements.txt \
+  && tox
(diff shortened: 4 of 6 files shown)


README.md
```diff
@@ -2,14 +2,26 @@ _NOTE: The interfaces of flux-accounting are being actively developed and are no
...
 ## flux-accounting
 
-Development for a bank/accounting interface for the Flux resource manager. Writes and saves job step history, user account information, and priority calculation to persistent storage using Python's SQLite3 API. 
+Development for a bank/accounting interface for the Flux resource manager. Writes and saves user account information to persistent storage using Python's SQLite3 API.
 
-##### job step history
+##### Install instructions
 
-Running as a cron job, `write_jobs.py` uses a flux-core RPC to retrieve job data that have recently transitioned to an INACTIVE state in the past _x_ minutes. It parses the job record data and writes it to a SQLite database, which can be queried at a later time. 
+This repository uses `pandas` for data display information. To install this dependency, along with any other packages specified in **requirements.txt**, and also run unit tests, you can run `make check`:
```
dongahn, line 9: Thanks @cmoussa1. Strictly speaking, run tests and make check are not part of Install instructions? 

Would it make sense to have "build instructions", "test instructions" and "install instructions"? And discuss the step for each in different section? 

Is `pandas` a test program dependency or a dependency for the core flux-accounting?

README.md
```diff
@@ -2,14 +2,26 @@ _NOTE: The interfaces of flux-accounting are being actively developed and are no
...
 
-##### user account information
+```
```
dongahn, line 11: Are there any other dependencies? Does yours depend on flux-core or flux-sched at this point? I think I saw one of the module depends on flux-core's python binding. If so, do you want to say "flux-sched requires an installed flux-core package"?

README.md
```diff
@@ -2,14 +2,26 @@ _NOTE: The interfaces of flux-accounting are being actively developed and are no
...
+$ make check
```
dongahn, line 12: I am not sure if I don't know if adding Makefile support for `make check` in an ad hoc fashion would be a long term solution. Did you add this as a short term stopgap?

@SteVwonder or @trws: is there any good python project that we can model after in terms of how to build, how to test and how to install etc?

README.md
```diff
@@ -2,14 +2,26 @@ _NOTE: The interfaces of flux-accounting are being actively developed and are no
...
+pip3 install -r requirements.txt
```
dongahn, line 13: When I used flux-sched docker, pip3 wasn't there so I had to manually install it. So you may as well mention pip3 as a dependency.

README.md
cmoussa1: Thanks for the feedback @dongahn. I believe what's required as far as install instructions after pulling down this repository are the following (let me know if you had to do anything else on your end):

- have `python3` installed in your environment (I used `brew install python3`)
- have `pip3` installed in your environment (which should be included when installing `python3`)
- install the dependencies listed in **requirements.txt** (which for now is just `pandas`) with `pip3 install -r requirements.txt`

`pandas` is a dependency for the core flux-accounting.

README.md
cmoussa1: Thanks for catching that. At this point, flux-accounting doesn't depend on flux-core's python binding. I was going to decorate the main function with flux-core's `util.CLIMain` decorator, but I decided to just include my own logger for now, which uses Python's standard `logging` module.

dongahn: Ah. Cool. This can make the testing of your core logic easier!

README.md
SteVwonder: I think this python guide is a good data point, in particular the section on using a Makefile for kicking off the dependency installation and the test suite: https://docs.python-guide.org/writing/structure/#makefile

The python setuptools documentation mentions that the `test` command is deprecated and recommends the python package tox: https://setuptools.readthedocs.io/en/latest/setuptools.html#test-build-package-and-run-a-unittest-suite.  I think that tox is probably overkill right now for this project.

In terms of installation and distribution, `setup.py sdist` will create a source distribution that can be uploaded to PyPI.  `setup.py install` and `pip install .` will both install locally from the source tree.  `setup.py develop` and `pip install -e .` will install into develop mode (makes it look installed, but you can see edits to source files without having to re-install).  Relevant Docs:
- https://packaging.python.org/tutorials/installing-packages/
- https://setuptools.readthedocs.io/en/latest/setuptools.html#development-mode

dongahn: Thanks. 

At first glance, it looks like autotools aren't best practices for python projects...

It will be a good test whether we can install our flux-accounting package properly with `flux-core` and `flux-sched` packages including support for both co-installation and standalone  installation) without GNU autotools and testing these configurations

`flux-sched` benefited heavily from mirroring various GNU autotools features.

cmoussa1:
Thanks for your suggestions and feedback @SteVwonder and @dongahn - I've gone ahead and pushed a couple of changes:

- I've added two Dockerfiles to this repo, one for `centos8` and one for `ubuntu`. As of now, they sit in the top level directory and are launched with the following command:

`docker build -f Dockerfile.ubuntu .`

I've only been able to get them to work when they are both sitting in the top level directory. Is it better to have them sitting in a `docker_files/` directory or something? If it is, I'm unsure how to structure it. 

- I added the dependency table to **README.md** for ubuntu and centos8, which includes `python3.6`, `pip3`, and `tox`.

> At first glance, it looks like autotools aren't best practices for python projects...

Okay! For now I have left the `Makefile` in this branch and added the Python-specific build and install commands to **README.md**. Would you prefer we leave the Makefile out of this repo entirely?

dongahn:
> I've only been able to get them to work when they are both sitting in the top level directory. Is it better to have them sitting in a docker_files/ directory or something? If it is, I'm unsure how to structure it.

How about mirroring `flux-core` and `flux-sched` and put those Dockerfiles in new directories 

- `src/test/docker/bionic`
- `src/test/docker/centos7`
- `src/test/docker/centors8`


This way, you will be able to reuse many of travis testing infrastructure from these repos. 

You won't have much inside `src` directories but my guess is it will grow, going forward. As you add front end commands like reporting tool, it will likely go into `src/cmd`.

cmoussa1:
Thanks @dongahn. I went ahead and moved the Dockerfiles to their respective directories inside `accounting`, so the path is: `flux-accounting/accounting/test/docker/<os>/Dockerfile.<os`, and they should each be able to be run with: `docker build -f <file name> ../../../../`.

README.md
```diff
@@ -2,14 +2,43 @@ _NOTE: The interfaces of flux-accounting are being actively developed and are no
...
+You can install the dependencies required by flux-accounting (located in **requirements.txt**) with the following command:
+
+```
+$ pip3 install -r requirements.txt
+```
+
+### Test Instructions
 
-Users have banks that they can charge their jobs against; this will affect their priority when submitting future jobs in order to maintain a fair balance of resources between users running jobs on a single cluster. Multiple factors play a role in determining a user's job priority, but in general, the amount of resources used vs. the resource limits initially allocated will either lower or heighten a user's job priority. 
+Make sure all is right correctly by running the unit tests with `tox`:
```
dongahn, line 27: < Make sure all is right correctly by running the unit tests with `tox`:

> Run the unit tests with `tox` to ensure the correctness of this package on your platform:

dongahn:
@cmoussa1: I just made one minor editing suggestion. Once that's addressed, this can go in from my perspective.

README.md
cmoussa1: Thanks @dongahn 👍 I went ahead and pushed a change to reflect your suggestion.

cmoussa1:
Thanks!

PR #15, opened 2020-05-13
accounting_cli.py: add command line interface

cmoussa1:
**Problem**: currently, the only way to access data within the accounting database file (**FluxAccounting.db**) is to open a sqlite3 shell from the same directory where the database file is located and run SQLite commands:

```
$ sqlite3

moussa1@trinity1 $ sqlite3
SQLite version 3.24.0 2018-06-04 14:10:15
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> .open FluxAccounting.db
sqlite> .headers on
sqlite> .mode column
sqlite> select * from association_table;
id_assoc    creation_time  mod_time    deleted     user_name   admin_level  account     parent_acct  shares      max_jobs    max_wall_pj
----------  -------------  ----------  ----------  ----------  -----------  ----------  -----------  ----------  ----------  -----------
1           1589225734     1589225734  0           fluxuser    1            acct        pacct        10          100         60  
``` 

**Proposal**: This PR looks to add a command line interface to flux-accounting so that instead of launching into a sqlite shell, we can use commands to:

- add a user to the accounting database: 

```
$ flux-accounting -a testuser 1 acct pacct 10 100 60
```

- view a user's information:

```
$ flux-accounting -u testuser
       3     1589384678  1589384678        0  testuser            1    acct       pacct      10       100           60
```

- edit a user's value:

```
$ flux-accounting -e testuser max_jobs 500 
```

- remove a user from the accounting database

```
$ flux-accounting -d testuser 
```

Instructions on how to install the command line interface and run the flux-accounting commands were added to the top level README as well. 

The unit tests for the added files `accounting_cli.py` and `accounting_cli_functions.py` can be run from the `flux-accounting` directory with: 

- `tox`,
- `python -m unittest discover`,
or
- `docker build -f accounting/test/docker/ubuntu/Dockerfile.ubuntu .`

I have this tagged as a [WIP] and I think it would be helpful for me to get some feedback on the labels I gave the command line arguments (e.g. is `-u` the best char to use for viewing a user's information?) and the way the command line syntax works:

1. Is the syntax for adding a user to the database perhaps too difficult to use? There are a lot of arguments (fields in the association_table) and the way the command works right now basically requires the admin to remember the order in which the fields should be entered. Would it be more convenient to add the fields one at a time, each labeled? 

An example:

```
enter a user_name: testuser
enter an admin_level: 1
.
.
.
```

2. Is `-u` a good flag for viewing a user's information? I first had thought of using `-v` but figured that would be confused with a `--verbose`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,117 @@ _______________________________summary _______________________________
   congratulations :)
 ```
 
-##### user account information
+To run the unit tests in a Docker container, you can use `docker build -f <path/to/Dockerfile> .` from the flux-accounting directory:
+
+```
+$ docker build -f accounting/test/docker/ubuntu/Dockerfile.ubuntu .
+Sending build context to Docker daemon  299.9MB
+Step 1/5 : FROM ubuntu:latest
+ ---> 1d622ef86b13
+Step 2/5 : LABEL maintainer="Christopher Moussa <moussa1@llnl.gov>"
+ ---> Using cache
+ ---> bdbc2ff632a5
+Step 3/5 : ADD . src/
+ ---> 7da92dbd8852
+ .
+ .
+ .
+ Step 5/5 : RUN cd src/   && pip3 install -r requirements.txt   && tox
+ .
+ .
+ .
+ python3.6 run-test: commands[0] | python -m unittest discover
+.........
+----------------------------------------------------------------------
+Ran 9 tests in 0.035s
+
+OK
+missing fields for account.
+
+FIELDS=user_name admin_level account parent_acct shares max_jobs max_wall_pj
+UNIQUE constraint failed: association_table.user_name
+no such column: bar
+___________________________________ summary ____________________________________
+  python3.6: commands succeeded
+  congratulations :)
+Removing intermediate container 45479512d947
+ ---> f29dd2ca5958
+Successfully built f29dd2ca5958
+```
+
+### User Account Information
 
 The accounting table in this database stores information like user name and ID, the account to submit jobs against, an optional parent account, the shares allocated to the user, as well as static limits, including max jobs submitted per user at a given time and max wall time per job per user.
+
+### Interacting With the Accounting DB
+
+There are two ways you can interact with the tables contained in the Accounting DB. Both of these require you to be in the same directory where the database file (**FluxAccounting.db**) is located. The first way is to launch into an interactive SQLite shell. From there, you can open the database file and interface with any of the tables using SQLite commands:
+
+```
+$ sqlite3
+SQLite version 3.24.0 2018-06-04 14:10:15
+Enter ".help" for usage hints.
+Connected to a transient in-memory database.
+Use ".open FILENAME" to reopen on a persistent database.
+
+sqlite> .open FluxAccounting.db
+sqlite> .tables
+association_table
+```
+
+To get nicely formatted output from queries (like headers for the tables and proper spacing), you can also set the following options in your shell:
+
+```
+sqlite> .mode columns
+sqlite> .headers on
+```
+
+This will output queries like the following:
+
+```
+sqlite> SELECT * FROM association_table;
+id_assoc    creation_time  mod_time    deleted     user_name   admin_level  account     parent_acct  shares      max_jobs    max_wall_pj
+----------  -------------  ----------  ----------  ----------  -----------  ----------  -----------  ----------  ----------  -----------
+1           1589225734     1589225734  0           fluxuser    1            acct        pacct        10          100         60  
+```
+
+The second way is to use flux-accounting's command line arguments. You can install them by running the following command from the `flux-accounting` directory:
+
+```
+$ pip3 install .
+
+Installing collected packages: flux-accounting
+  Running setup.py develop for flux-accounting
+Successfully installed flux-accounting
+```
+
+Then, from the same directory where the database file (**FluxAccounting.db**) is located, you can use flux-accounting's command line interface:
+
+```
+$ flux-accounting -h
+
+usage: flux-accounting [-h] {view-user,add-user,delete-user,edit-user} ...
+
+Description: Translate command line arguments into SQLite instructions for the
+Flux Accounting Database.
+
+positional arguments:
+  {view-user,add-user,delete-user,edit-user}
+                        sub-command help
+    view-user           view a user's information in the accounting database
+    add-user            add a user to the accounting database
+    delete-user         remove a user from the accounting database
+    edit-user           edit a user's value
+
+optional arguments:
+  -h, --help            show this help message and exit
+```
+
+With flux-accounting's command line tools, you can view a user's account information, add and remove users to the accounting database, and edit an existing user's account information. Just make sure you are in the same directory where the database file is located when you run the commands:
+
+```
+$ flux-account view-user fluxuser
+
+id_assoc  creation_time    mod_time  deleted user_name  admin_level account parent_acct  shares  max_jobs  max_wall_pj
+       1     1589225734  1589225734        0  fluxuser            1    acct       pacct      10       100           60
+```
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Run the unit tests with `tox` to ensure the correctness of this package on your 
 
 ```
 $ tox
-python3.6 run-test: commands[0] | python -m unittest discover
+python3.6 run-test: commands[0] | python -m unittest discover -b
 ....
 ----------------------------------------------------------------------
 Ran 4 tests in 0.008s
@@ -58,17 +58,12 @@ Step 3/5 : ADD . src/
  .
  .
  .
- python3.6 run-test: commands[0] | python -m unittest discover
+ python3.6 run-test: commands[0] | python -m unittest discover -b
 .........
 ----------------------------------------------------------------------
 Ran 9 tests in 0.035s
 
 OK
-missing fields for account.
-
-FIELDS=user_name admin_level account parent_acct shares max_jobs max_wall_pj
-UNIQUE constraint failed: association_table.user_name
-no such column: bar
 ___________________________________ summary ____________________________________
   python3.6: commands succeeded
   congratulations :)
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2020 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import sqlite3
+import argparse
+import time
+
+import pandas as pd
+
+from accounting import accounting_cli_functions as aclif
+
+
+def main():
+
+    parser = argparse.ArgumentParser(
+        description="""
+        Description: Translate command line arguments into
+        SQLite instructions for the Flux Accounting Database.
+        """
+    )
+    subparsers = parser.add_subparsers(help="sub-command help",)
+
+    subparser_view_user = subparsers.add_parser(
+        "view-user", help="view a user's information in the accounting database"
+    )
+    subparser_view_user.set_defaults(func="view_user")
+    subparser_view_user.add_argument("username", help="username", metavar=("USERNAME"))
+
+    subparser_add_user = subparsers.add_parser(
+        "add-user", help="add a user to the accounting database"
+    )
+    subparser_add_user.set_defaults(func="add_user")
+    subparser_add_user.add_argument(
+        "--username", help="username", metavar="USERNAME",
+    )
+    subparser_add_user.add_argument(
+        "--admin-level", help="admin level", metavar="ADMIN_LEVEL",
+    )
+    subparser_add_user.add_argument(
+        "--account", help="account to charge jobs against", metavar="ACCOUNT",
+    )
+    subparser_add_user.add_argument(
+        "--parent-acct", help="parent account", metavar="PARENT_ACCOUNT",
+    )
+    subparser_add_user.add_argument(
+        "--shares", help="shares", metavar="SHARES",
+    )
+    subparser_add_user.add_argument(
+        "--max-jobs", help="max jobs", metavar="MAX_JOBS",
+    )
+    subparser_add_user.add_argument(
+        "--max-wall-pj", help="max wall time per job", metavar="MAX_WALL_PJ",
+    )
+
+    subparser_delete_user = subparsers.add_parser(
+        "delete-user", help="remove a user from the accounting database"
+    )
+    subparser_delete_user.set_defaults(func="delete_user")
+    subparser_delete_user.add_argument(
+        "username", help="username", metavar=("USERNAME")
+    )
+
+    subparser_edit_user = subparsers.add_parser("edit-user", help="edit a user's value")
+    subparser_edit_user.set_defaults(func="edit_user")
+    subparser_edit_user.add_argument(
+        "--username", help="username", metavar="USERNAME",
+    )
+    subparser_edit_user.add_argument(
+        "--field", help="column name", metavar="FIELD",
+    )
+    subparser_edit_user.add_argument(
+        "--new-value", help="new value", metavar="VALUE",
+    )
+
+    args = parser.parse_args()
+
+    conn = sqlite3.connect("FluxAccounting.db")
+
+    try:
+        if args.func == "view_user":
+            aclif.view_user(conn, args.username)
+        elif args.func == "add_user":
+            aclif.add_user(
+                conn,
+                args.username,
+                args.admin_level,
+                args.account,
+                args.parent_acct,
+                args.shares,
+                args.max_jobs,
+                args.max_wall_pj,
+            )
+        elif args.func == "delete_user":
+            aclif.delete_user(conn, args.username)
+        elif args.func == "edit_user":
+            aclif.edit_user(conn, args.username, args.field, args.new_value)
+        else:
+            print(parser.print_usage())
+    finally:
+        conn.close()
+
+
+if __name__ == "__main__":
+    main()
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2020 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import sqlite3
+import argparse
+import time
+import sys
+
+import pandas as pd
+
+
+def view_user(conn, user):
+    try:
+        # get the information pertaining to a user in the Accounting DB
+        select_stmt = "SELECT * FROM association_table where user_name=?"
+        dataframe = pd.read_sql_query(select_stmt, conn, params=(user,))
+        # if the length of dataframe is 0, that means
+        # the user specified was not found in the table
+        if len(dataframe.index) == 0:
+            print("User not found in association_table")
+        else:
+            print(dataframe)
+    except pd.io.sql.DatabaseError as e_database_error:
+        print(e_database_error)
+
+
+def add_user(
+    conn, username, admin_level, account, parent_acct, shares, max_jobs, max_wall_pj
+):
+
+    # insert the user values into the database
+    try:
+        conn.execute(
+            """
+            INSERT INTO association_table (
+                creation_time,
+                mod_time,
+                deleted,
+                user_name,
+                admin_level,
+                account,
+                parent_acct,
+                shares,
+                max_jobs,
+                max_wall_pj
+            )
+            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+            """,
+            (
+                int(time.time()),
+                int(time.time()),
+                0,
+                username,
+                admin_level,
+                account,
+                parent_acct,
+                shares,
+                max_jobs,
+                max_wall_pj,
+            ),
+        )
+        # commit changes
+        conn.commit()
+    # make sure entry is unique
+    except sqlite3.IntegrityError as integrity_error:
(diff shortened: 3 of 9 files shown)


accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,108 @@
...
+                user_acct_info[6],
+            ),
+        )
+        # commit changes
+        conn.commit()
+    # make sure entry is unique
+    except sqlite3.IntegrityError as integrity_error:
+        print(integrity_error)
+    # make sure all fields are entered
+    except IndexError:
```
chu11, line 78: perhaps this could be checked earlier on as an user input error, so that we don't have to make a connection to the sql database to produce the error message.

accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,108 @@
...
+            "missing fields for account."
+            "\n\nFIELDS=user_name admin_level account parent_acct shares max_jobs max_wall_pj"
+        )
+
+def delete_user(user):
+    # open connection to database
+    conn = sqlite3.connect("FluxAccounting.db")
+    # delete user account from association_table
+    delete_stmt = (
+        "DELETE FROM association_table WHERE user_name='" + user + "'"
```
SteVwonder, line 89: Based on the [sqlite3 package documentation](https://docs.python.org/3.8/library/sqlite3.html), I recommend using the paramater substitution supported by the `execute` method for security reasons, similar to what you did in `add_user`:
```suggestion
        "DELETE FROM association_table WHERE user_name='?'"
```

[TIL there is also support for a named style](https://docs.python.org/3.8/library/sqlite3.html#sqlite3.Cursor.execute):
```suggestion
        "DELETE FROM association_table WHERE user_name=':name'"
```

accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,108 @@
...
+
+def edit_user(user_info):
+    # open connection to database
+    conn = sqlite3.connect("FluxAccounting.db")
+    try:
+        # edit value in accounting database
+        conn.execute(
+            "UPDATE association_table SET %s=%s WHERE user_name='%s'"
+            % (user_info[1], user_info[2], user_info[0])
+        )
```
SteVwonder, line 103: Same comment about parameter substitution:
```suggestion
        conn.execute(
            "UPDATE association_table SET ?=? WHERE user_name='?'",
            (user_info[1], user_info[2], user_info[0])
        )
```

accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,108 @@
...
+    # open connection to database
+    conn = sqlite3.connect("FluxAccounting.db")
+    # delete user account from association_table
+    delete_stmt = (
+        "DELETE FROM association_table WHERE user_name='" + user + "'"
+    )
+    conn.execute(delete_stmt)
+    # commit changes
+    conn.commit()
+
```
SteVwonder, line 94: missing a `conn.close()` here and in other functions, but see my note about making `conn` an argument.

accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,108 @@
...
+import argparse
+import time
+import sys
+
+import pandas as pd
+
+def view_user(user):
+    try:
+        # open connection to database
+        conn = sqlite3.connect("FluxAccounting.db")
```
SteVwonder, line 22: Rather than having a connection created at the start of every function (and having the database name/path hardcoded into each function), maybe it makes sense to add `conn` as an argument to these functions?  That would potentially enable future users of this code to perform multiple operations with a single connection, and it also means you only have to get the opening/closing logic correct once in `accounting_cli.py` rather than getting it correct multiple times in this file.

accounting/accounting_cli_functions.py
SteVwonder: Note, that my suggestion for this specific query may not work due to the first `?`.  It seems parameterization doesn't work for everything in a SQL statement, like table names [[1]](https://stackoverflow.com/a/597198).  Per that same SO thread:

> For situations where user input determines the table, I pull the name of the table out of a dict and raise an exception if the contents of the input are unexpected. Probably not the best way to go about it, but seems less liable to result in SQL injection.

accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,108 @@
...
+        # get the information pertaining to a user in the Accounting DB
+        select_stmt = (
+            "SELECT * FROM association_table where user_name='" + user + "'"
+        )
+        dataframe = pd.read_sql_query(select_stmt, conn)
```
SteVwonder, line 27: Looks like pandas also supports secure parameter substitution in a similar way [[1]](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_sql_query.html):

```suggestion
        select_stmt = (
            "SELECT * FROM association_table where user_name='?'"
        )
        dataframe = pd.read_sql_query(select_stmt, conn, params=(user,))
```

accounting/accounting_cli.py
```diff
@@ -0,0 +1,80 @@
...
+
+    -h, --help                              Display this message.
+    -u, --user USER                         View account information for a specific user.
+    -a, --add_user FIELDS                   Add a user to the accounting database.
+
+        FIELDS=user_name admin_level account parent_acct shares max_jobs max_wall_pj
+
+    -d, --delete_user USER                  Remove a user from the accounting database.
+    -e, --edit_value USER FIELD NEW_VALUE   Edit an existing user's value in the accounting database
+    """
```
SteVwonder, line 36: I think all of this should already be auto-generated by argparse.  Is there a particular part of the usage message that wasn't being properly generated?  If so, maybe we can fix that issue in a more generic way.

accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+
```
dongahn, line 2: The executable bit for this python file was off when I try this under Docker. 

`chmod u+x accounting_cli_functions.py`

dongahn:
@cmoussa1: Thank you for putting this together. I've added some minor requests inlined above. 

BTW, do you expect this command will ultimately be grown into a flux front end command so that scheduler administers will use `flux accounting ...` ? If so, good choice you have the "flux-" prefix.

But would it be a bit more succinct to rename the front end command name to `flux-account` as opposed to `flux-accounting` so that users can run `flux account -u ...` as opposed to `flux accounting -u" ?

accounting/accounting_cli_functions.py
cmoussa1: Sorry, I think I might need a little more clarification 😬 do you mean specifying the path to the .db file as an additional argument on the command line? as an example:

`flux-accounting accounting/FluxAccounting.db -u fluxuser`

accounting/create_db.py
```diff
@@ -28,17 +28,17 @@ def create_db(filepath):
     conn.execute(
         """
             CREATE TABLE IF NOT EXISTS association_table (
+                id_assoc      integer                           PRIMARY KEY,
```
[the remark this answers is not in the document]
cmoussa1: Haha, well, yes and no. Changing the type of `id_assoc` to the following:

`id_assoc integer PRIMARY KEY`

means that a value is created automatically without the need to specify an `AUTOINCREMENT` keyword. According to sqlite documentation, `AUTOINCREMENT` introduces a lot of overhead and should be avoided being used. [Here](https://www.sqlite.org/autoinc.html) is some documentation where I got this from.

accounting/accounting_cli.py
```diff
@@ -0,0 +1,80 @@
...
+    parser = argparse.ArgumentParser(usage=msg())
+
+    parser.add_argument(
+        "-u", "--user", dest="view_user", help="view a user's account information"
+    )
+    parser.add_argument(
+        "-a",
+        "--add-user",
+        dest="add_user",
+        nargs="*",
```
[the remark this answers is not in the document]
cmoussa1: I ended up going with `metavars`. The output help message seems pretty useful to me now, but what do you think? Does a sub parser still make more sense to implement? 

```
usage: flux-accounting [-h] [-u USER_NAME]
                       [-a USER_NAME ADMIN_LEVEL ACCOUNT PARENT_ACCT SHARES MAX_JOBS MAX_WALL_PJ]
                       [-d USER_NAME] [-e USER_NAME FIELD NEW VALUE]
```

accounting/accounting_cli_functions.py
cmoussa1: With some help from @chu11, I think I was able to make this change. All of the functions in `accounting_cli_functions.py` now take a **conn** parameter that's passed in from `accounting_cli.py`:

```
def main():

    conn = sqlite3.connect("FluxAccounting.db")

    ... parse args and stuff ...

    if args.view_user:
        aclif.view_user(conn, args.view_user)
    elif args.add_user:
        aclif.add_user(conn, args.add_user)
    elif args.delete_user:
        aclif.delete_user(conn, args.delete_user)
    elif args.edit_user:
        aclif.edit_user(conn, args.edit_user)

    conn.close()
```

This should only establish one connection while it runs, and then closes the connection when it exits. I hope I understood you correctly! Sorry in advance if it wasn't.

accounting/accounting_cli.py
cmoussa1: Yup, you're absolutely right. I wasn't paying close enough attention and had manually written my own help message. With the help of `metavars` I think I was able to produce a much cleaner help message with `flux-accounting -h`:

```
usage: flux-accounting [-h] [-u USER_NAME]
                       [-a USER_NAME ADMIN_LEVEL ACCOUNT PARENT_ACCT SHARES MAX_JOBS MAX_WALL_PJ]
                       [-d USER_NAME] [-e USER_NAME FIELD NEW VALUE]

Flux Accounting Database.

optional arguments:
  -u USER_NAME, --user USER_NAME
                        view a user's account information
  -a USER_NAME ADMIN_LEVEL ACCOUNT PARENT_ACCT SHARES MAX_JOBS MAX_WALL_PJ, --add-user USER_NAME ADMIN_LEVEL ACCOUNT PARENT_ACCT SHARES MAX_JOBS MAX_WALL_PJ
                        add a user to the accounting database
  -d USER_NAME, --delete-user USER_NAME
                        remove a user from the accounting database
  -e USER_NAME FIELD NEW VALUE, --edit-user USER_NAME FIELD NEW VALUE
                        edit a user's value
```

accounting/accounting_cli.py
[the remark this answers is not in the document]
chu11: I think someone commented below that a subcommand might be good b/c that'll match the rest of flux's command line usage.  Not a requirement, but something to consider.

accounting/create_db.py
```diff
@@ -14,7 +14,7 @@
 import logging
 
 
-LOGGER = logging.basicConfig(filename="accounting/db_creation.log", level=logging.INFO)
+LOGGER = logging.basicConfig(filename="db_creation.log", level=logging.INFO)
```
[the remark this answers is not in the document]
cmoussa1: Thanks for letting me know. I believe I made this change correctly and split these changes into two commits.

accounting/accounting_cli.py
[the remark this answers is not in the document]
cmoussa1: Oh oh! I'm sorry. I think I was misunderstanding. Instead of running something like `flux-accounting -u fluxuser`, the suggestion was to use subcommands (like `flux mini submit`):

```
$ flux-accounting view-user fluxuser

<output>
```

README.md
```diff
@@ -39,6 +39,116 @@ _______________________________summary _______________________________
...
+.........
+----------------------------------------------------------------------
+Ran 9 tests in 0.035s
+
+OK
+missing fields for account.
+
+FIELDS=user_name admin_level account parent_acct shares max_jobs max_wall_pj
+UNIQUE constraint failed: association_table.user_name
+no such column: bar
```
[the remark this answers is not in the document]
cmoussa1: Yes there is, you can pass a `-b` flag to suppress print output when running unit tests. I just went ahead and pushed this change; it should be reflected in the `tox` command and Docker builds.

cmoussa1:
OK, I pushed some changes to address the suggestions/comments above. Thanks for your feedback and advice so far. I will drop the [WIP] tag as I think this is getting closer to a merge-able PR.

accounting/accounting_cli.py
```diff
@@ -0,0 +1,148 @@
...
+    elif hasattr(args, "delete_user"):
+        aclif.delete_user(conn, args.delete_user)
+    elif hasattr(args, "edit_user"):
+        aclif.edit_user(conn, args.edit_user)
+
+    conn.close()
+
+
+if __name__ == "__main__":
+    main()
```
chu11, line 148: perhaps a stupid question, but why not just put all the code in `if __name__ ...`, why create a function called `main()`?

grondo: I have seen this a lot with Python code, it seems to be almost idiomatic.

See e.g. [this](https://stackoverflow.com/questions/22492162/understanding-the-main-method-of-python) SO post.

SteVwonder: I put most things into `main` b/c anything you put here gets put into global scope.  If I'm doing a quick and dirty script, I'll use that to my advantage and do arg parsing here.  Then I don't have to pass an `args` argument into every other function needlessly.  But it can bite you for the same reasons.

accounting/accounting_cli.py
```diff
@@ -0,0 +1,148 @@
...
+    args = parser.parse_args()
+
+    if hasattr(args, "view_user"):
+        aclif.view_user(conn, args.view_user)
+    elif hasattr(args, "add_user"):
+        aclif.add_user(conn, args.add_user)
+    elif hasattr(args, "delete_user"):
+        aclif.delete_user(conn, args.delete_user)
+    elif hasattr(args, "edit_user"):
+        aclif.edit_user(conn, args.edit_user)
```
SteVwonder, line 142: I think you probably want to wrap this whole block in a try-except, and then put the connection closing in a finally block.  That way the database connection is always closed, even if an uncaught exception bubbles up to the top-level and kills the program.

test/test_accounting_cli.py
```diff
@@ -0,0 +1,97 @@
...
+            aclif.edit_user(conn, ["fluxuser", "foo", "bar"])
+
+        self.assertEqual(cm.exception.code, 1)
+
+    # remove database and log file
+    @classmethod
+    def tearDownClass(self):
+        os.remove("FluxAccounting.db")
+        os.remove("db_creation.log")
+        conn.close()
```
SteVwonder, line 81: This ultimately probably doesn't matter too much, but the connection should probably be closed first and then the database file removed.

accounting/accounting_cli.py
```diff
@@ -0,0 +1,148 @@
...
+
```
SteVwonder, line 143: I'd also add an else condition for the situation where no subcommand is provided:

```suggestion
    else:
        parser.print_usage()
```

accounting/accounting_cli_functions.py
```diff
@@ -0,0 +1,110 @@
...
+                int(time.time()),
+                int(time.time()),
+                0,
+                user_acct_info[0],
+                user_acct_info[1],
+                user_acct_info[2],
+                user_acct_info[3],
+                user_acct_info[4],
+                user_acct_info[5],
+                user_acct_info[6],
```
SteVwonder, line 69: The ordering of the elements within the `add_user` list depends on the order that the user provides them in, but the `add_user` makes assumptions about their ordering.  If the user provides the args in the same order as the help message, all is good, but if you swap any arguments, the database will have the wrong values:

```
# sherbein at Stephens-MacBook-Pro.local in ~/Repositories/flux-framework/flux-accounting on git:front-end-tools ✖︎ [19:42:39]
→ PYTHONPATH=. ./accounting/accounting_cli.py add-user --user_name stephen --admin_level 1 --account foo --parent_acct NULL --shares 5 --max_jobs 6 --max_wall_pj 7

# sherbein at Stephens-MacBook-Pro.local in ~/Repositories/flux-framework/flux-accounting on git:front-end-tools ✖︎ [19:43:01]
→ PYTHONPATH=. ./accounting/accounting_cli.py view-user stephen
0         1     1590201763  1590201763        0   stephen            1     foo        NULL       5         6            7

# sherbein at Stephens-MacBook-Pro.local in ~/Repositories/flux-framework/flux-accounting on git:front-end-tools ✖︎ [19:43:34]
→ PYTHONPATH=. ./accounting/accounting_cli.py add-user --user_name mark --account bar --admin_level 2 --parent_acct NULL --shares 5 --max_jobs 6 --max_wall_pj 7

# sherbein at Stephens-MacBook-Pro.local in ~/Repositories/flux-framework/flux-accounting on git:front-end-tools ✖︎ [19:43:43]
→ PYTHONPATH=. ./accounting/accounting_cli.py view-user mark
   id_assoc  creation_time    mod_time  deleted user_name admin_level account parent_acct  shares  max_jobs  max_wall_pj
0         2     1590201823  1590201823        0      mark         bar       2        NULL       5         6            7
```

I recommend either making all of these arguments positional arguments to the function, or use `**kwargs`.  You will also need to not have every `dest=add_user` in the argparser.

accounting/accounting_cli.py
```diff
@@ -0,0 +1,148 @@
...
+    )
+    subparser_view_user.add_argument(
+        dest="view_user", type=str, help="username", metavar=("USER_NAME")
+    )
+
+    subparser_add_user = subparsers.add_parser(
+        "add-user", help="add a user to the accounting database"
+    )
+    subparser_add_user.add_argument(
+        "--user_name",
```
SteVwonder, line 44: For all of these long options with the double-dash, I recommend switching from underscores to hyphens.  It is the style recommended by the GNU style guide [[1]](https://stackoverflow.com/questions/1253679/should-command-line-options-in-posix-style-operating-systems-be-underscore-style)

```suggestion
        "--user-name",
```

Although in this specific case, isn't it normally a single word? `username` [[2]](https://english.stackexchange.com/questions/43436/username-user-name-or-user-name) (This is the definition of a nitpick so feel free to ignore this second half of the suggestion).

test/test_accounting_cli.py
cmoussa1: 🤦‍♂️ thanks for catching this. I will push a fix.

accounting/accounting_cli.py
chu11: additional food for thought, `--user` instead for consistency to other tools.  BUT the other tools sometimes take a userid too.

accounting/accounting_cli_functions.py
chu11: Additional idea, check inputs for validity?  Validity may not be perfect, but simple things like is a number, is a string with alphanumeric, etc.

accounting/accounting_cli.py
grondo: Yeah, using `--user` allows the option to take either user name or userid.

cmoussa1:
@SteVwonder thanks for your feedback. I really liked your suggestion about refactoring the commands that take multiple arguments, adding/editing users. I've refactored them so they do the following:

each subcommand now contains an argument like the following:

```
subparser_edit_user = subparsers.add_parser("edit-user", help="edit a user's value")
subparser_edit_user.set_defaults(func="edit_user")
```

which adds a `func` attribute to `Namespace`. Then, after parsing args, I check for the value of `func` in order to pass it to the right function:

```
elif args.func == "edit_user":
```

then I can pass in each parameter as its own variable instead of as a list:

```
aclif.edit_user(conn, args.username, args.field, args.new_value)
```

Now, the order of the parameters shouldn't matter (forgive me if you already knew this - I wanted to type it out so I can reinforce for myself what I just learned 😄).

accounting/accounting_cli.py
chu11: in this case wasn't clear if this tool could take a userid

grondo: Sorry, I commented without even looking at the code. Bad @grondo!
However, I meant that it leaves an opening to extend this option in the future to take either username or userid...

README.md
```diff
@@ -28,7 +28,7 @@ Run the unit tests with `tox` to ensure the correctness of this package on your
 
 ```
 $ tox
-python3.6 run-test: commands[0] | python -m unittest discover
+python3.6 run-test: commands[0] | python -m unittest discover -b
```
SteVwonder, line 31: Just out of curiosity, what does the `-b` flag do?  I couldn't find any info from googling.

accounting/accounting_cli.py
```diff
@@ -0,0 +1,112 @@
...
+import time
+
+import pandas as pd
+
+from accounting import accounting_cli_functions as aclif
+
+
+def main():
+
+    conn = sqlite3.connect("FluxAccounting.db")
```
SteVwonder, line 23: Nit pick, but I would put this connection establishment after the command line parsing, so that a connection is not made when a user provides bad input.

README.md
cmoussa1: This will suppress output from any `print()` statements while running the unit tests. Without the `-b` flag, running `python -m unittest discover` would output the following:

```
moussa1@trinity1 ~/src/flux-framework/flux-accounting$ python -m unittest discover
.UNIQUE constraint failed: association_table.user_name
..Field not found in association table
.....
Ran 8 tests in 0.032s

OK
```

I believe it was @dongahn who had mentioned that this output could be confusing to user's who were running the unit tests. Passing the `-b` flag, however, will output:

```
moussa1@trinity1 ~/src/flux-framework/flux-accounting$ python -m unittest discover -b
........
Ran 8 tests in 0.017s

OK
```

It took me a _long_ time to find a solution to this, but I stumbled across [this](https://stackoverflow.com/questions/24134343/suppress-print-output-in-unittests) Stack Overflow post which mentioned it 😄.

accounting/accounting_cli.py
cmoussa1: Thanks for pointing this out! I will make this change right now and push an updated branch.

cmoussa1:
@dongahn when you get the chance, can you take another look at this PR? I pushed some changes to address your suggestions, and I think it is ready for another review! 🙂

dongahn:
Will get to this soonish. Thanks!

dongahn:
OK. I spoke with @cmoussa1 by phone to go over some of the flux integration work items for the future. But I am putting the MWP label for this. Thanks @cmoussa1.

SteVwonder:
Merging manually since the mergify yaml is looking for check names that differ from the current check names.  @cmoussa1 is working on a fix in the meantime.

PR #17, opened 2020-06-08
Mergify: configuration update

cmoussa1:
**Problem**: mergify previously wasn't performing all of the checks when a PR was submitted; adding the `merge-when-passing` label didn't actually merge the PR.

This PR edits **.mergify.yml** so that all of the checks are performed when a new PR is submitted. Namely, the `validate commits (3.x)` checks are added to the file, since those checks already include the previous checks that were never run:

- `Lint with flake8`
- `Test with unittest`

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -2,9 +2,9 @@ pull_request_rules:
   - name: rebase and merge when passing all checks
     conditions:
       - base=master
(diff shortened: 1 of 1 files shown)


cmoussa1:
Thanks!

PR #19, opened 2020-06-17
make: add make install, update install instructions

### LINKED ISSUE #16: repo: integrate flux-accounting with Flux install
Speaking with @dongahn on Friday, we discussed that we need to integrate flux-accounting's install with the system instance install so that it gets picked up by the `flux(1)` command driver, so it can be called with the following:

```
$ flux account view-user fluxuser
```

What steps are required to accomplish this? Did I miss anything else @dongahn ?

cmoussa1:
This PR updates the install instructions for flux-accounting so that its commands can be picked up by Flux's command driver `flux(1)`. The repo can be pulled down and installed with `make install`, which will place **flux-account.py** in flux-core's `libexec/` directory. Then, like before, the SQLite database file can be interacted with using the command line arguments:

```
$ flux account view-user fluxuser
       1     1589225734  1589225734        0  fluxuser            1    acct       pacct      10       100           60
```

The instructions on how to install the package have been reflected in the **README** as well. 

Fixes #16

---

Since I have a conceptual question or two about the Makefile, I'll mark this as **[WIP]** for now: can the environment variables **PATH** and **PYTHONPATH** be set in the Makefile? Should they be? Or do we typically leave that to the user to set themselves?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,15 @@
-# minimal Makefile to run Python unit tests
-# for flux-accounting
+FLUX_INSTALL_PREFIX ?= /usr
+FLUX=${FLUX_INSTALL_PREFIX}/bin/flux
+FLUX_PYTHON_VERSION=$(shell ${FLUX} python --version | cut -f 2 -d ' ' | cut -f 1-2 -d.)
+
+dependencies:
+	${FLUX} python -m pip install -r requirements.txt --user
+
+install:
+	${FLUX_INSTALL_PREFIX}/bin/flux python setup.py install \
+	--prefix=${FLUX_INSTALL_PREFIX} \
+	--install-scripts=${FLUX_INSTALL_PREFIX}/libexec/flux/cmd/ \
+	--install-lib=${FLUX_INSTALL_PREFIX}/lib/flux/python${FLUX_PYTHON_VERSION}/
 
 check:
-	pip3 install -r requirements.txt
-	python3 -m unittest discover -b
+	${FLUX_INSTALL_PREFIX}/bin/flux python -m unittest discover -b
--- a/README.md
+++ b/README.md
@@ -16,15 +16,46 @@ flux-accounting requires the following packages to build:
 
 ### Install Instructions
 
-You can install the dependencies required by flux-accounting (located in **requirements.txt**) with the following command:
+You can install the dependencies required by flux-accounting as well as the package itself to be recognized by Flux's command driver `flux(1)` with `make`.
+
+1. Set your **FLUX_INSTALL_PREFIX** environment variable to point to flux-core's installation target:
 
 ```
-$ pip3 install -r requirements.txt
+$ export FLUX_INSTALL_PREFIX = ~/path/to/flux-core/install/
+```
+
+2. Pull down and install the flux-accounting repo:
+
+```
+$ git clone https://github.com/flux-framework/flux-accounting
+$ cd flux-accounting/
+$ make install
+```
+
+3. Run flux-accounting's commands:
+
+```
+$ flux account -h
+usage: flux-account.py [-h] {view-user,add-user,delete-user,edit-user} ...
+
+Description: Translate command line arguments into SQLite instructions for the
+Flux Accounting Database.
+
+positional arguments:
+  {view-user,add-user,delete-user,edit-user}
+                        sub-command help
+    view-user           view a user's information in the accounting database
+    add-user            add a user to the accounting database
+    delete-user         remove a user from the accounting database
+    edit-user           edit a user's value
+
+optional arguments:
+  -h, --help            show this help message and exit
 ```
 
 ### Test Instructions
 
-Run the unit tests with `tox` to ensure the correctness of this package on your platform::
+Run the unit tests with `tox` to ensure the correctness of this package on your platform:
 
 ```
 $ tox
@@ -108,22 +139,12 @@ id_assoc    creation_time  mod_time    deleted     user_name   admin_level  acco
 1           1589225734     1589225734  0           fluxuser    1            acct        pacct        10          100         60  
 ```
 
-The second way is to use flux-accounting's command line arguments. You can install them by running the following command from the `flux-accounting` directory:
+The second way is to use flux-accounting's command line arguments. Then, from the same directory where the database file (**FluxAccounting.db**) is located, you can use flux-accounting's command line interface:
 
 ```
-$ pip3 install .
+$ flux account -h
 
-Installing collected packages: flux-accounting
-  Running setup.py develop for flux-accounting
-Successfully installed flux-accounting
-```
-
-Then, from the same directory where the database file (**FluxAccounting.db**) is located, you can use flux-accounting's command line interface:
-
-```
-$ flux-accounting -h
-
-usage: flux-accounting [-h] {view-user,add-user,delete-user,edit-user} ...
+usage: flux-account.py [-h] {view-user,add-user,delete-user,edit-user} ...
 
 Description: Translate command line arguments into SQLite instructions for the
 Flux Accounting Database.
@@ -143,7 +164,7 @@ optional arguments:
 With flux-accounting's command line tools, you can view a user's account information, add and remove users to the accounting database, and edit an existing user's account information. Just make sure you are in the same directory where the database file is located when you run the commands:
 
 ```
-$ flux-account view-user fluxuser
+$ flux account view-user fluxuser
 
 id_assoc  creation_time    mod_time  deleted user_name  admin_level account parent_acct  shares  max_jobs  max_wall_pj
        1     1589225734  1589225734        0  fluxuser            1    acct       pacct      10       100           60
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ with open("README.md", "r") as fh:
 
 setuptools.setup(
     name="flux-accounting",
-    version="0.0.1",
+    version="0.1.0",
     author="Christopher Moussa",
     author_email="moussa1@llnl.gov",
     description="User accounting interface for Flux framework",
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setuptools.setup(
     packages=setuptools.find_packages(),
     entry_points={
         'console_scripts': [
-            'flux-account = accounting.accounting_cli:main',
+            'flux-account.py = accounting.accounting_cli:main',
         ]
     },
     classifiers=[


cmoussa1:
Just pushed a small update to add a default value for `FLUX_INSTALL_PREFIX` in case the user doesn't specify that environment variable on the command line or in their environment.

Makefile
```diff
@@ -1,6 +1,9 @@
-# minimal Makefile to run Python unit tests
-# for flux-accounting
+FLUX_INSTALL_PREFIX ?= /usr
+
+install:
+	${FLUX_INSTALL_PREFIX}bin/flux python -m pip install -r requirements.txt --user
+	${FLUX_INSTALL_PREFIX}bin/flux python -m pip install --install-option="--prefix=${FLUX_INSTALL_PREFIX}" \
+	--install-option="--install-scripts=${FLUX_INSTALL_PREFIX}libexec/flux/cmd/" . --upgrade
```
SteVwonder, line 6: Generally looks good.  Just need some extra slashes in case the user doesn't put a trailing slash at the end of `FLUX_INSTALL_PREFIX`.

```suggestion
	${FLUX_INSTALL_PREFIX}/bin/flux python -m pip install -r requirements.txt --user
	${FLUX_INSTALL_PREFIX}/bin/flux python -m pip install --install-option="--prefix=${FLUX_INSTALL_PREFIX}" \
	--install-option="--install-scripts=${FLUX_INSTALL_PREFIX}/libexec/flux/cmd/" . --upgrade
```

README.md
```diff
@@ -16,15 +16,53 @@ flux-accounting requires the following packages to build:
...
+1. Set your **FLUX_INSTALL_PREFIX** environment variable to point to flux-core's installation target:
+
+```
+$ export $FLUX_INSTALL_PREFIX = ~/path/to/flux-core/install/
+```
+
+2. After building flux-core, you'll need to set **PYTHONPATH** to point towards flux-core's Python site-packages as well as the **PATH** to point to Flux's executable:
+
+```
+$ export PYTHONPATH = ~/path/to/flux-core/install/lib/python3.7/site-packages/
```
SteVwonder, line 30: I don't think that exporting `PYTHONPATH` is necessary as long as the `flux` command is used rather than running `flux-account.py` directly.

README.md
```diff
@@ -16,15 +16,53 @@ flux-accounting requires the following packages to build:
...
+$ export PATH = ~/path/to/flux-core/install/bin/
```
SteVwonder, line 31: Looks good, we just want to make sure we don't wipe out existing paths in `PATH`.

```suggestion
$ export PATH = ~/path/to/flux-core/install/bin/:$PATH
```

SteVwonder: We can also leverage the previously exported prefix to reduce duplication.

```suggestion
$ export PATH = ${FLUX_INSTALL_PREFIX}/bin/:$PATH
```

README.md
```diff
@@ -16,15 +16,53 @@ flux-accounting requires the following packages to build:
...
-You can install the dependencies required by flux-accounting (located in **requirements.txt**) with the following command:
+You can install the dependencies required by flux-accounting as well as the package itself to be recognized by Flux's command driver `flux(1)` with `make`.
+
+1. Set your **FLUX_INSTALL_PREFIX** environment variable to point to flux-core's installation target:
+
+```
+$ export $FLUX_INSTALL_PREFIX = ~/path/to/flux-core/install/
+```
+
+2. After building flux-core, you'll need to set **PYTHONPATH** to point towards flux-core's Python site-packages as well as the **PATH** to point to Flux's executable:
```
SteVwonder, line 27: These two sections of `export`s can probably be combined into one step for simplicity.

Makefile
```diff
@@ -1,6 +1,9 @@
-# minimal Makefile to run Python unit tests
-# for flux-accounting
+FLUX_INSTALL_PREFIX ?= /usr
+
+install:
+	${FLUX_INSTALL_PREFIX}bin/flux python -m pip install -r requirements.txt --user
```
SteVwonder, line 4: It might be worth splitting this dependency installation step out into a separate target:

```make
dependencies:
	${FLUX_INSTALL_PREFIX}bin/flux python -m pip install -r requirements.txt --user
```

And then you can make that auto-run during a `make install` with:

```make
install: dependencies
      INSTALL_CMDS_GO_HERE
```

That way, if users want to run the testsuite without installing flux-accounting but don't have the dependencies installed, they can do:

```
make dependencies
make check
```

SteVwonder:
> can the environment variables PATH and PYTHONPATH be set in the Makefile? Should they be? Or do we typically leave that to the user to set themselves?

I would just leave those to be set by the user.  What you have in the Makefiles currently looks pretty good.

Makefile
cmoussa1: TIL! Thanks for explaining this - this is a neat feature.

cmoussa1:
Thanks for the great feedback @SteVwonder - I went ahead and pushed some changes to address the suggestions!

Makefile
```diff
@@ -1,6 +1,11 @@
-# minimal Makefile to run Python unit tests
-# for flux-accounting
+FLUX_INSTALL_PREFIX ?= /usr
+
+dependencies:
+        ${FLUX_INSTALL_PREFIX}/bin/flux python -m pip install -r requirements.txt --user
+
+install: dependencies
+        ${FLUX_INSTALL_PREFIX}/bin/flux python -m pip install --install-option="--prefix=${FLUX_INSTALL_PREFIX}" \
+        --install-option="--install-scripts=${FLUX_INSTALL_PREFIX}libexec/flux/cmd/" . --upgrade
```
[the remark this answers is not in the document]
cmoussa1: 🤦‍♂️I'm sorry for not catching this earlier. Just pushed an update.

Makefile
```diff
@@ -1,6 +1,11 @@
-# minimal Makefile to run Python unit tests
-# for flux-accounting
+FLUX_INSTALL_PREFIX ?= /usr
+
+dependencies:
+        ${FLUX_INSTALL_PREFIX}/bin/flux python -m pip install -r requirements.txt --user
```
SteVwonder, line 4: The beginning of these "recipe" lines (lines meant to be run in the shell) should start with a `<TAB>` rather than spaces [[1]](https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html).  It's an oddity of Makefiles that has also bitten me in the past:

```
sherbein ~/Repositories/flux-framework/flux-accounting (add-make-install)
❯ make install                                                                                                              16:25:21 ()
Makefile:4: *** missing separator (did you mean TAB instead of 8 spaces?).  Stop.
```

Makefile
```diff
@@ -1,6 +1,11 @@
...
+
+install: dependencies
+        ${FLUX_INSTALL_PREFIX}/bin/flux python -m pip install --install-option="--prefix=${FLUX_INSTALL_PREFIX}" \
```
SteVwonder, line 7: I ran into this error on pip v20, and this change seemed to fix it:
```
Installing collected packages: flux-accounting
    Running setup.py install for flux-accounting ... error
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-z41xl4kb/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-z41xl4kb/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'
"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-79duizhp/install-record.txt --single-version-externally-managed --user --prefix= --compile --install-headers /home/sherbein/.local/include/python3.8/flux-accountin
g --prefix=/home/sherbein/opt/flux/accounting-pr-19 --install-scripts=/home/sherbein/opt/flux/accounting-pr-19/libexec/flux/cmd/
         cwd: /tmp/pip-req-build-z41xl4kb/
    Complete output (2 lines):
    running install
    error: can't combine user with prefix, exec_prefix/home, or install_(plat)base
    ----------------------------------------
ERROR: Command errored out with exit status 1: /usr/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-z41xl4kb/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-z41xl4kb/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__)
;code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-79duizhp/install-record.txt --single-version-externally-managed --user --prefix= --compile --install-headers /home/sherbein/.local
/include/python3.8/flux-accounting --prefix=/home/sherbein/opt/flux/accounting-pr-19 --install-scripts=/home/sherbein/opt/flux/accounting-pr-19/libexec/flux/cmd/ Check the logs for full command output.
make: *** [Makefile:7: install] Error 1
```

```suggestion
        ${FLUX_INSTALL_PREFIX}/bin/flux python -m pip install --prefix=${FLUX_INSTALL_PREFIX} \
```

Also worth noting the deprecation warning which means that `--install-option="--install-scripts` might not work in the near future:
```
DEPRECATION: Location-changing options found in --install-option: ['--prefix', '--install-scripts'] from command line. This configuration may cause unexpected behavior and is unsupported. pip 20.2 will remove support for this functionality. A possible replacement is using
pip-level options like --user, --prefix, --root, and --target. You can find discussion regarding this at https://github.com/pypa/pip/issues/7309.
```

cmoussa1:
Thanks for your help iterating on this @SteVwonder. I'll go ahead and push the symlink "hack" that you had suggested, but I agree it might be better to look to see if we have a cleaner solution. I'll poke at the egg links and see if I can make any progress with it. In the meantime, I believe I addressed your suggestions and pushed those fixes.

cmoussa1:
Okay, just pushed an update to **Makefile**: 

> One thing we need to figure out is how to handle the workaround that flux-core uses to avoid "polluting" `PYTHONPATH` with system paths when users run flux: flux-framework/flux-core#1999. Ultimately the problem is that the directory that flux-accounting is installed into is correct, but it is not the directory that flux automatically adds to the `PYTHONPATH`.
I've been playing around with it some more, so I wanted to share where I got. I ended up using egg-links to just create a reference to where the `flux-accounting` source code lies (or at least, I think so 😅).

With my first iteration, after running `make install`, the source code for `flux-accounting` and its `.egg-info` directory were getting installed into `flux/accounting-target/lib/flux/python3.7/`. I changed the **Makefile** so that it instead contains the following `install` instruction:

```
${FLUX_INSTALL_PREFIX}/bin/flux python setup.py develop -xd ${FLUX_INSTALL_PREFIX}/lib/flux/python${FLUX_PYTHON_VERSION}/ install_scripts --install-dir=${FLUX_INSTALL_PREFIX}/libexec/flux/cmd/
```

which results in the following:

```
running develop
running egg_info
writing flux_accounting.egg-info/PKG-INFO
writing dependency_links to flux_accounting.egg-info/dependency_links.txt
writing entry points to flux_accounting.egg-info/entry_points.txt
writing top-level names to flux_accounting.egg-info/top_level.txt
writing manifest file 'flux_accounting.egg-info/SOURCES.txt'
running build_ext
Creating /g/g0/moussa1/flux/accounting-target/lib/flux/python3.7/site.py
Creating /g/g0/moussa1/flux/accounting-target/lib/flux/python3.7/flux-accounting.egg-link (link to .)
Adding flux-accounting 0.0.1 to easy-install.pth file

Installed /g/g0/moussa1/src/flux-framework/flux-accounting
Processing dependencies for flux-accounting==0.0.1
Finished processing dependencies for flux-accounting==0.0.1
running install_scripts
Installing flux-account.py script to /g/g0/moussa1/flux/accounting-target//libexec/flux/cmd/
```

So the `flux-account.py` executable still gets installed where we originally wanted, but the rest of `flux-accounting` remains in its original location, along with the newly-created `.egg-info` directory. And the contents of `~/flux/accounting-target/lib/flux/python3.7/` now contain:

```
total 16
drwx------ 2 moussa1 moussa1 4096 Jun 25 13:42 __pycache__
lrwxrwxrwx 1 moussa1 moussa1   70 Jun 25 13:29 _flux -> /g/g0/moussa1/flux/accounting-target/lib/python3.7/site-packages/_flux
-rw------- 1 moussa1 moussa1   49 Jun 25 13:42 easy-install.pth
lrwxrwxrwx 1 moussa1 moussa1   69 Jun 25 13:29 flux -> /g/g0/moussa1/flux/accounting-target/lib/python3.7/site-packages/flux
-rw------- 1 moussa1 moussa1   50 Jun 25 13:42 flux-accounting.egg-link
-rw------- 1 moussa1 moussa1 2302 Jun 25 13:42 site.py
```

I am still not too sure if we want **easy-install.pth** and **site.py** in this directory though. These get installed after running **setup.py**. The only contents of **easy-install.pth** is the path to the `flux-accounting` directory, and I _think_ **site.py** appends the `flux-accounting/` directory to `sys.path`. Thoughts @SteVwonder ?

SteVwonder:
Thanks @cmoussa1 for continuing to iterate on this!

> So the flux-account.py executable still gets installed where we originally wanted, but the rest of flux-accounting remains in its original location, along with the newly-created .egg-info directory. And the contents of ~/flux/accounting-target/lib/flux/python3.7/ now contain:

The layout and contents of `~/flux/accounting-target/lib/flux/python3.7/` looks correct.  The part that should probably change is that the links shouldn't point back to the repo directory in someone's home directory, but instead it should point to an installed directory (i.e., `${FLUX_INSTALL_PREFIX}/lib/python3.8/site-packages/`).  I think this is probably a function of using the `develop` subcommand to `setup.py`.  I wonder if switching to `setup.py install` will "just work" (not sure if that affects anything else).

> I am still not too sure if we want easy-install.pth and site.py in this directory though.

I'm not too familiar with these files, but after some searching on the web, I think they are probably fine to leave for now.  They may cause issues if we have other Flux projects that want to install python packages too, but I vote that we cross that bridge when we get to it (it also seems like easy-install.pth files support multiple paths, so maybe it would work just fine).

cmoussa1:
Thanks for your feedback @SteVwonder!

> The part that should probably change is that the links shouldn't point back to the repo directory in someone's home directory, but instead it should point to an installed directory (i.e., ${FLUX_INSTALL_PREFIX}/lib/python3.8/site-packages/).

Sorry, I think I might be a little confused here. Are you referring to these paths:

```
[moussa1@quartz2300:flux-accounting]$ ll ~/flux/accounting-target/lib/flux/python3.7/
total 8
lrwxrwxrwx 1 moussa1 moussa1 70 Jun 30 14:45 _flux -> /g/g0/moussa1/flux/accounting-target/lib/python3.7/site-packages/_flux
lrwxrwxrwx 1 moussa1 moussa1 69 Jun 30 14:45 flux -> /g/g0/moussa1/flux/accounting-target/lib/python3.7/site-packages/flux
```

isn't this an installed directory? For `$FLUX_INSTALL_PREFIX` refers to `/g/g0/moussa1/flux/accounting-target/`?

SteVwonder:
> Are you referring to these paths:

Sorry.  I forgot to paste in the links that I was referring to (too much multi-tasking):
```
❯ cat flux-accounting.egg-link                                                                                              15:04:22 ()
/data/hdd/Repositories/flux-framework/flux-accounting
```

That egg link points to the git repo in my home directory with the flux accounting source code.  Instead it should mirror the symlinks that you mentioned and point to a directory in the install prefix.

cmoussa1:
> That egg link points to the git repo in my home directory with the flux accounting source code. Instead it should mirror the symlinks that you mentioned and point to a directory in the install prefix.

Ah, okay. Thanks for explaining that!

So, I could be wrong, but my understanding is that we can only create an egg-link with `python setup.py develop`. This will create the egg, in-place (which I think is why the egg-link points to the source code in our home directory). The approved answer on [this](https://stackoverflow.com/questions/12901776/python-setup-py-develop-is-it-possible-to-create-egg-info-folder-not-in-so) Stack Overflow post is what I am referencing. 

I could instead use `python setup.py install` (like you mentioned above) while specifying some options:

```
${FLUX_INSTALL_PREFIX}/bin/flux python setup.py install 

--prefix=${FLUX_INSTALL_PREFIX}/lib/python${FLUX_PYTHON_VERSION}/site-packages/ \

--install-scripts=${FLUX_INSTALL_PREFIX}/libexec/flux/cmd/ \ 

--install-lib=${FLUX_INSTALL_PREFIX}/lib/flux/python${FLUX_PYTHON_VERSION}/
```

And the contents of `~/flux/accounting-target/lib/flux/python3.7/`:

```
total 36
drwx------ 2 moussa1 moussa1  4096 Jul  1 10:03 __pycache__
lrwxrwxrwx 1 moussa1 moussa1    70 Jul  1 10:02 _flux -> /g/g0/moussa1/flux/accounting-target/lib/python3.7/site-packages/_flux
-rw------- 1 moussa1 moussa1    34 Jul  1 10:03 easy-install.pth
lrwxrwxrwx 1 moussa1 moussa1    69 Jul  1 10:02 flux -> /g/g0/moussa1/flux/accounting-target/lib/python3.7/site-packages/flux
-rw------- 1 moussa1 moussa1 17937 Jul  1 10:03 flux_accounting-0.0.1-py3.7.egg
-rw------- 1 moussa1 moussa1  2302 Jul  1 10:03 site.py
```

So I don't need to create an egg-link, and **easy-install.pth** points to the .egg in the same directory.

SteVwonder:
> my understanding is that we can only create an egg-link with python setup.py develop

Gotcha!  That's kind of a bummer but I guess that makes sense.

> I could instead use python setup.py install (like you mentioned above) while specifying some options:

That looks good.  The more I think about it, it's probably more important to have the files in `lib/flux/python3.7/` than `lib/python3.7/site-packages` b/c the former is auto-added by the `flux` binary to `PYTHONPATH` while the latter would only be used if a user directly executes a script without using the `flux` binary.  And I suspect users will only ever do `flux account` and never do `${FLUX_INSTALL_PREFIX}/libexec/cmd/flux-account.py`.

> `--prefix=${FLUX_INSTALL_PREFIX}/lib/python${FLUX_PYTHON_VERSION}/site-packages/`

Out of curiosity, what does that directory end up looking like when that flag is used?

cmoussa1:
>> --prefix=${FLUX_INSTALL_PREFIX}/lib/python${FLUX_PYTHON_VERSION}/site-packages/

> Out of curiosity, what does that directory end up looking like when that flag is used?

Here are the contents of that directory: 

```
[moussa1@quartz2306:flux-accounting]$ ll ~/flux/accounting-target/lib/python3.7/site-packages/                    
total 8
drwx------ 3 moussa1 moussa1 4096 Jul  1 10:02 _flux
drwx------ 4 moussa1 moussa1 4096 Jul  1 10:02 flux
```

SteVwonder:
> Here are the contents of that directory:

Ok.  Unless I'm missing something, it looks like the prefix flag is kind of a noop since nothing new gets added there.  Is that right?

I guess that is because there are only libs and scripts to install, both of which are overridden with `--install-lib` and `--install-scripts`.  Maybe we just set `--prefix=${FLUX_INSTALL_PREFIX}` in case anything besides lib/scripts are added in the future?

Besides that, the new install command LGTM! :shipit:

cmoussa1:
> I guess that is because there are only libs and scripts to install, both of which are overridden with --install-lib and --install-scripts. Maybe we just set `--prefix=${FLUX_INSTALL_PREFIX}` in case anything besides lib/scripts are added in the future?

Agreed! I'll set `--prefix` accordingly. Just pushed an update that adds the new install command.

cmoussa1:
@dongahn when you get the chance, could you give this PR a quick review? 🙂 I think it is getting closer to being able to be merged.

Makefile
```diff
@@ -1,6 +1,15 @@
-# minimal Makefile to run Python unit tests
-# for flux-accounting
+FLUX_INSTALL_PREFIX ?= /usr
+FLUX=${FLUX_INSTALL_PREFIX}/bin/flux
+FLUX_PYTHON_VERSION=$(shell ${FLUX} python --version | cut -f 2 -d ' ' | cut -f 1-2 -d.)
+
+dependencies:
+	${FLUX} python -m pip install -r requirements.txt --user
+
+install: dependencies
```
SteVwonder, line 8: Sorry @cmoussa1, I know I suggested adding the `dependencies` target here so that they would be automatically installed on a `make install` call, but as I'm thinking more about system packaging, I'm not sure that is what we want.  If we package this up in an RPM, we won't want pip to install the dependencies in the user home directory.  To make packaging easier in the future, maybe we should drop this and require users to manually run `make dependencies` if they want the pip line to run.  Thoughts?

cmoussa1: No need to apologize @SteVwonder - your insight is really helpful! I think I'd agree with you. That change seems fine to me, so I'll go ahead and push a fix to remove `dependencies` from the `install:` target.

SteVwonder:
@mergifyio refresh

cmoussa1:
Woo! mergify is now working in this repo too 🙂 thanks @SteVwonder!

PR #21, opened 2020-06-29
create_db.py, accounting_cli.py: add configurable database path

### LINKED ISSUE #20: create_db.py: make file path to database configurable
**create_db.py**, the script responsible for creating the database file and the table inside the database, does not take any file path arguments. This means that when this script is called, the .db file (**FluxAccounting.db**) gets placed in the same directory:

```
.
|-- AUTHORS
|-- CONTRIBUTING.md
|-- COPYING
|-- LICENSE -> COPYING
|-- Makefile
|-- NOTICE -> NOTICE.LLNS
|-- NOTICE.LLNS
|-- README.md
|-- accounting
|   |-- FluxAccounting.db
|   |-- __init__.py
|   |-- accounting_cli.py
|   |-- accounting_cli_functions.py
|   |-- create_db.py
```

This is because of the `sqlite3.connect()` line in **create_db.py**, which takes a filepath string as an argument:

```python
conn = sqlite3.connect(filepath)
```

which, for now, is just set to the same directory as **create_db.py**:

```python
def main():
    create_db("FluxAccounting.db")
```

From talks with @SteVwonder, this functionality probably needs to undergo a couple of changes:

- the database file probably needs to created in a different location, one that can be specified by the user when they call this script. Probably something like the following:

```
$ ./create_db.py path/to/db
```

This change will _also_ need to be reflected in `accounting_cli.py`, because it establishes a connection to the .db file when it is called with any of the subcommands (`add-user`, `edit-user`, etc.). 

Maybe a strategy to accomplish this is to create a subcommand for _both_ scripts to set the .db file location? 

```
$ ./create_db.py set-db-location --filepath=/path/to/db
$ flux account set-db-location --filepath=/path/to/db
```

- the commands that make a connection to/interact with the database probably need a failsafe of some kind in case it does not establish a connection to the database. If it doesn't, the sqlite3 API will default to creating an empty .db file, which is no good. I think a workaround for this would be to interpret the database as a URI; this way, we can specify _only_ enable read/write mode for the subcommands [[1]](https://docs.python.org/3.7/library/sqlite3.html#sqlite3.connect). 

```python
conn = sqlite3.connect('file:path/to/db?mode=rw', uri=True)
```

cmoussa1:
**Problem**: Both **create_db.py** and **accounting_cli.py** make assumptions about the location of the database file **FluxAccounting.db**. When **create_db.py** gets run, the database file gets placed in the same directory as the source code:

```
|-- LICENSE -> COPYING
|-- Makefile
|-- NOTICE -> NOTICE.LLNS
|-- NOTICE.LLNS
|-- README.md
|-- accounting
|   |-- FluxAccounting.db
|   |-- accounting_cli.py
|   |-- accounting_cli_functions.py
|   |-- create_db.py
```

This forces all of the subcommands to be executed in the same directory as well.

---

This PR adds a new optional command line argument `-p`, `--path` that allows the user to specify the database location when both creating the database and executing any of the subcommands. If no `-p` is passed, both scripts will default to the current directory (which was the behavior beforehand). 

An example of creating the database with the new argument:

```
$ ./create_db.py -p some/other/location/FluxAccounting.db
```

And running the subcommands:

```
$ flux account -p ../FluxAccounting.db view-user fluxuser
0         1     1593451150  1593451150        0  fluxuser            1    acct       pacct      10       100         1440
```

The subcommands will only open **FluxAccounting.db** in read-write mode instead of read-write-create. This should prevent empty database files from being created in case it cannot find the file. 

Fixes #20

### DIFF
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -12,6 +12,7 @@
 import sqlite3
 import argparse
 import time
+import sys
 
 import pandas as pd
 
@@ -27,6 +28,10 @@ def main():
         """
     )
     subparsers = parser.add_subparsers(help="sub-command help",)
+
+    parser.add_argument(
+        "-p", "--path", dest="path", help="specify location of database file"
+    )
 
     subparser_view_user = subparsers.add_parser(
         "view-user", help="view a user's information in the accounting database"
@@ -82,7 +87,13 @@ def main():
 
     args = parser.parse_args()
 
-    conn = sqlite3.connect("FluxAccounting.db")
+    # try to open database file; will exit with -1 if database file not found
+    path = args.path if args.path else "FluxAccounting.db"
+    try:
+        conn = sqlite3.connect("file:" + path + "?mode=rw", uri=True)
+    except sqlite3.OperationalError:
+        print("Unable to open database file")
+        sys.exit(-1)
 
     try:
         if args.func == "view_user":
--- a/accounting/create_db.py
+++ b/accounting/create_db.py
@@ -12,6 +12,8 @@
 import sqlite3
 import pandas as pd
 import logging
+import argparse
+import sys
 
 
 LOGGER = logging.basicConfig(filename="db_creation.log", level=logging.INFO)
@@ -20,7 +22,7 @@ LOGGER = logging.basicConfig(filename="db_creation.log", level=logging.INFO)
 def create_db(filepath):
     # open connection to database
     logging.info("Creating Flux Accounting DB")
-    conn = sqlite3.connect(filepath)
+    conn = sqlite3.connect("file:" + filepath + "?mode:rwc", uri=True)
     logging.info("Created Flux Accounting DB sucessfully")
 
     # Association Table
@@ -47,7 +49,26 @@ def create_db(filepath):
 
 
 def main():
-    create_db("FluxAccounting.db")
+
+    parser = argparse.ArgumentParser(
+        description="""
+        Description: Create flux-accounting database to store
+        user account information.
+        """
+    )
+
+    parser.add_argument(
+        "-p", "--path", dest="path", help="specify location of database file"
+    )
+
+    args = parser.parse_args()
+
+    path = args.path if args.path else "FluxAccounting.db"
+    try:
+        create_db(path)
+    except sqlite3.OperationalError:
+        print("Unable to create database file")
+        sys.exit(-1)
 
 
 if __name__ == "__main__":


cmoussa1:
Okay, rebased after #19 and removed the **[WIP]** tag. I think this might be ready for some initial review

cmoussa1:
Just pushed a fix to address some incorrect syntax with opening the .db file in URI mode; instead of `mode:rw`, it should be `mode=rw`.

accounting/accounting_cli.py
```diff
@@ -82,7 +87,19 @@ def main():
...
+            conn = sqlite3.connect("file:" + args.path + "?mode=rw", uri=True)
+        except sqlite3.OperationalError:
+            print("Unable to open database file")
+            sys.exit(-1)
+    else:
+        try:
+            conn = sqlite3.connect("file:FluxAccounting.db?mode=rw", uri=True)
+        except sqlite3.OperationalError:
+            print("Unable to open database file")
+            sys.exit(-1)
```
SteVwonder, line 102: Nit, but you can eliminate some code duplication by leveraging a `path` variable. Something like:

```
        path = args.path if args.path else "FluxAccounting.db"
        try:
            conn = sqlite3.connect("file:" + path + "?mode=rw", uri=True)
        except sqlite3.OperationalError:
            sys.exit(-1)
```

accounting/create_db.py
```diff
@@ -47,7 +48,27 @@ def create_db(filepath):
...
+
+    args = parser.parse_args()
+
+    if args.path:
+        try:
+            create_db(args.path)
+        except sqlite3.OperationalError:
+            print("Unable to create database file")
+    else:
+        create_db("FluxAccounting.db")
```
SteVwonder, line 71: Does this create_db need to be wrapped in a try-catch as well?  Might want to leverage the suggestion above of leveraging a `path` variable.

accounting/create_db.py
```diff
@@ -47,7 +48,27 @@ def create_db(filepath):
...
+        "-p", "--path", dest="path", help="specify location of database file"
+    )
+
+    args = parser.parse_args()
+
+    if args.path:
+        try:
+            create_db(args.path)
+        except sqlite3.OperationalError:
+            print("Unable to create database file")
```
SteVwonder, line 69: If the database fails to be created, I think this script should exit with a non-zero exit code.

cmoussa1:
Thanks for taking the time to review this @SteVwonder! I believe I addressed your suggestions and pushed those up. Thanks for the suggestion about using a `path` variable; I hadn't seen that before.

cmoussa1:
Thanks! :-)

Issue #23, opened 2020-07-22
create_db.py: need to be able to create multiple accounts with the same username


cmoussa1:
The current structure of the `association_table` in flux-accounting has the restriction of having unique usernames. However, I recently learned that often times there is one user with multiple banks (accounts), so we’ll need the ability to create multiple users with the same username.

This should be a relatively simple fix, for the current table has the following parameters:

```
CREATE TABLE IF NOT EXISTS association_table (
                id_assoc      integer                           PRIMARY KEY,
                creation_time bigint(20)            NOT NULL,
                mod_time      bigint(20)  DEFAULT 0 NOT NULL,
                deleted       tinyint(4)  DEFAULT 0 NOT NULL,
                user_name     tinytext    UNIQUE    NOT NULL,
                admin_level   smallint(6) DEFAULT 1 NOT NULL,
                account       tinytext              NOT NULL,
                parent_acct   tinytext              NOT NULL,
                shares        int(11)     DEFAULT 1 NOT NULL,
                max_jobs      int(11)               NOT NULL,
                max_wall_pj   int(11)               NOT NULL
);
```

If we remove the UNIQUE tab from the `user_name` field, we should be able to create multiple accounts with the same username, but different accounts.

cmoussa1:
One of the things I've discovered is SQLite tables have a hidden `ROWID`. My original implementation of the primary key:

`id_assoc      integer                           PRIMARY KEY,`

was actually just an alias for the `ROWID`. On an INSERT, the `ROWID` will be filled automatically with an unused integer, usually one more than the largest `ROWID` currently in use [[1]](https://www.sqlite.org/autoinc.html). 

So, I should be able to just change the primary key for values in the association table to be a combination of `(username, account)`. This will ensure that multiple associations can be created with the same username, as long as they are under different accounts.

PR #25, opened 2020-07-22
create_db: change table structure to allow duplicate usernames

cmoussa1:
**Problem**: as discussed in #23, the current table structure of `association_table` has a **UNIQUE** constraint on the `username` field. However, there is a need for multiple entries to be created with the same username, but different accounts. 

This PR changes the structure of `association_table` to instead have a primary key of a combination of both `username` and `account`. It also removes the `assoc_id` field from the table, which will be replaced by SQLite's native `ROWID` field [[1]](https://www.sqlite.org/autoinc.html). This should allow multiple entries to be created with the same username, as long as they are under different accounts. 

### An example

Trying to add the same user under the same account should raise an `IntegrityError`:

```
$ flux account add-user --username=fluxuser --admin-level=1 --account=acct --parent-acct=pacct --shares=1 --max-jobs=100 --max-wall-pj=60
$ flux account add-user --username=fluxuser --admin-level=1 --account=acct --parent-acct=pacct --shares=1 --max-jobs=100 --max-wall-pj=60
UNIQUE constraint failed: association_table.user_name, association_table.account
```

But changing the account name results in a successful addition:

```
$ flux account -p accounting/FluxAccounting.db view-user fluxuser
   creation_time    mod_time  deleted user_name  admin_level     account parent_acct  shares  max_jobs  max_wall_pj
0     1595438356  1595438356        0  fluxuser            1        acct       pacct       1       100           60
1     1595438491  1595438491        0  fluxuser            1  other_acct       pacct       1       100           60
```

Some of the unit tests also needed to be changed to reflect the update in `association_table`, as well as the example output from the SQLite queries in the top level **README**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,9 +134,9 @@ This will output queries like the following:
 
 ```
 sqlite> SELECT * FROM association_table;
-id_assoc    creation_time  mod_time    deleted     user_name   admin_level  account     parent_acct  shares      max_jobs    max_wall_pj
-----------  -------------  ----------  ----------  ----------  -----------  ----------  -----------  ----------  ----------  -----------
-1           1589225734     1589225734  0           fluxuser    1            acct        pacct        10          100         60  
+creation_time  mod_time    deleted     user_name   admin_level  account     parent_acct  shares      max_jobs    max_wall_pj
+-------------  ----------  ----------  ----------  -----------  ----------  -----------  ----------  ----------  -----------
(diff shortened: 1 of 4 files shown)


cmoussa1:
Thanks @SteVwonder! Should I get another approval or am I free to set MWP? 🙂

SteVwonder:
I think it is probably fine.  Setting the label.  Thanks @cmoussa1 !

cmoussa1:
Thanks!

PR #33, opened 2020-08-18
job-archive: add front-end to job-archive

### LINKED ISSUE #30: job-archive: prefer a command line interface to accounting database to querying the database directly
On the question about whether it’s preferable for me to write my own scripts to directly query the database vs. having a front-end interface, I think it’s preferable to have a front-end interface. In general, I think more folks are more comfortable with a command line interface rather than directly interacting with the database and having a front-end interface can allow you more granularity in what types of queries different users are able to make. In Slurm, the separation of the database and the front-end interface also allows them to make changes to the database to add new functionality etc. without changing the user interface to the job data.

cmoussa1:
The power finally came back on at my place! I can continue where I left off before mother nature so rudely interrupted me 😅

**Problem**: flux-core's job-archive currently stores job records in a SQLite database that can be viewed and queried with SQLite commands: 

```
sqlite> select userid,id,t_submit,t_run,t_inactive,ranks,R from jobs;
userid      id            t_submit         t_run            t_inactive        ranks       R                                                                            
----------  ------------  ---------------  ---------------  ----------------  ----------  -----------------------------------------------------------------------------
2001        429932937216  1597722545.7235  1597722545.7498  1597722545.83819  0           {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-2"}}]}}
2001        408659427328  1597722544.4558  1597722544.4830  1597722544.58507  0           {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-2"}}]}}
2001        340007059456  1597722540.3639  1597722540.3879  1597722540.46995  0           {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-1"}}]}}
2001        319857623040  1597722539.1629  1597722539.1850  1597722539.27435  0           {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-1"}}]}}
2001        256154533888  1597722535.3659  1597722535.3874  1597722535.4645   0           {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}  
2001        241961009152  1597722534.5195  1597722534.5461  1597722534.63193  0           {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}} 
```

From talks with Ryan Day, it would be preferable to have some sort of front-end interface to this database to extract job records and group them by various options.  

---

This PR looks to add a front end to the job-archive so that job records can be queried without having to go into a SQLite shell and write SQL queries. The following subcommands have been added:

- **by-user**: allows a user to fetch all job records from a certain username
- **by-jobid**: allows user to fetch job information from a certain jobid
- **after-start-time**: allows a user to fetch job records after a certain start time
- **before-end-time**: allows a user to fetch job records before a certain end time

Each of the subcommands query the job-archive and add matching entries to a list of `JobRecord`s, which have the following structure:

```python
class JobRecord(object):
    """
    A record of an individual job.
    """

    def __init__(self, userid, username, jobid, t_submit, t_run, t_inactive, nnodes, R):
        self.userid = userid
        self.username = get_username(userid)
        self.jobid = jobid
        self.t_submit = t_submit
        self.t_run = t_run
        self.t_inactive = t_inactive
        self.nnodes = nnodes
        self.R = R
        return None

    @property
    def elapsed(self):
        return time.mktime(self.t_inactive) - time.mktime(self.t_run)

    @property
    def queued(self):
        return time.mktime(self.t_run) - time.mktime(self.t_submit)
```

These job records get written to a .csv file that can be specified on the command line; if no path is passed, the records get printed out to the screen in a `Dataframe` format:

```
$ flux account -p /var/lib/flux/jobs.db -o /var/lib/flux/job_records.csv by-user fluxuser
```

OR 

```
$ flux account -p /var/lib/flux/jobs.db by-user fluxuser
   UserID  Username         JobID         T_Submit            T_Run       T_Inactive  Nodes                                                                            R
0    2001  fluxuser  410739802112 1598469186.39630 1598469186.41970 1598469186.49578      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
1    2001  fluxuser  379114749952 1598469184.51090 1598469184.53045 1598469184.60123      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
2    2001  fluxuser  353646936064 1598469182.99277 1598469183.01287 1598469183.08738      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
3    2001  fluxuser  292040998912 1598469179.32122 1598469179.34539 1598469179.43041      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
```

Some notes:

The **nnodes** field is counted by reading the `ranks` field from the job-archive. It makes the assumption that each rank is associated with one node. We discussed in a coffee time (thanks @SteVwonder, @grondo, and @garlick for your input) that it is okay to make this assumption for now. 

Fixes #30

### DIFF
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -33,6 +33,13 @@ def main():
         "-p", "--path", dest="path", help="specify location of database file"
     )
 
+    parser.add_argument(
+        "-o",
+        "--output-file",
+        dest="output_file",
+        help="specify location of output file",
+    )
+
     subparser_view_user = subparsers.add_parser(
         "view-user", help="view a user's information in the accounting database"
     )
@@ -88,6 +95,38 @@ def main():
         "--new-value", help="new value", metavar="VALUE",
     )
 
+    subparser_view_jobs_by_username = subparsers.add_parser(
+        "by-user", help="show jobs run by username"
+    )
+    subparser_view_jobs_by_username.set_defaults(func="view_jobs_run_by_username")
+    subparser_view_jobs_by_username.add_argument(
+        "username", help="username", metavar="USERNAME",
+    )
+
+    subparser_view_job_by_jobid = subparsers.add_parser(
+        "by-jobid", help="show job info from jobid"
+    )
+    subparser_view_job_by_jobid.set_defaults(func="view_jobs_with_jobid")
+    subparser_view_job_by_jobid.add_argument(
+        "jobid", help="jobid", metavar="JOBID",
+    )
+
+    subparser_view_jobs_after_start_time = subparsers.add_parser(
+        "after-start-time", help="show jobs that completed after start time"
+    )
+    subparser_view_jobs_after_start_time.set_defaults(func="view_jobs_after_start_time")
+    subparser_view_jobs_after_start_time.add_argument(
+        "start_time", help="start time", metavar="START TIME",
+    )
+
+    subparser_view_jobs_before_end_time = subparsers.add_parser(
+        "before-end-time", help="show jobs that completed before end time"
+    )
+    subparser_view_jobs_before_end_time.set_defaults(func="view_jobs_before_end_time")
+    subparser_view_jobs_before_end_time.add_argument(
+        "end_time", help="end time", metavar="END TIME",
+    )
+
     args = parser.parse_args()
 
     # try to open database file; will exit with -1 if database file not found
@@ -97,6 +136,9 @@ def main():
     except sqlite3.OperationalError:
         print("Unable to open database file")
         sys.exit(-1)
+
+    # set path for output file
+    output_file = args.output_file if args.output_file else None
 
     try:
         if args.func == "view_user":
@@ -116,6 +158,14 @@ def main():
             aclif.delete_user(conn, args.username)
         elif args.func == "edit_user":
             aclif.edit_user(conn, args.username, args.field, args.new_value)
+        elif args.func == "view_jobs_run_by_username":
+            aclif.view_jobs_run_by_username(conn, args.username, args.output_file)
+        elif args.func == "view_jobs_with_jobid":
+            aclif.view_jobs_with_jobid(conn, args.jobid, args.output_file)
+        elif args.func == "view_jobs_after_start_time":
+            aclif.view_jobs_after_start_time(conn, args.start_time, args.output_file)
+        elif args.func == "view_jobs_before_end_time":
+            aclif.view_jobs_before_end_time(conn, args.end_time, args.output_file)
         else:
             print(parser.print_usage())
     finally:
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -13,8 +13,239 @@ import sqlite3
 import argparse
 import time
 import sys
+import pwd
+import csv
 
 import pandas as pd
+
+
+def count_ranks(ranks):
+    if "-" in ranks:
+        ranks_count = ranks.replace("-", ",").split(",")
+        return int(ranks_count[1]) - int(ranks_count[0]) + 1
+    else:
+        return int(ranks) + 1
+
+
+def get_username(userid):
+    try:
+        return pwd.getpwuid(userid).pw_name
+    except KeyError:
+        return str(userid)
+
+
+def get_uid(username):
+    try:
+        return pwd.getpwnam(username).pw_uid
+    except KeyError:
+        return str(username)
+
+
+def write_records_to_file(job_records, output_file):
+    with open(output_file, "w", newline="") as csvfile:
+        spamwriter = csv.writer(
+            csvfile, delimiter="|", quotechar="", escapechar="'", quoting=csv.QUOTE_NONE
+        )
+        spamwriter.writerow(
+            (
+                "UserID",
+                "Username",
+                "JobID",
+                "T_Submit",
+                "T_Run",
+                "T_Inactive",
+                "Nodes",
+                "R",
+            )
+        )
+        for record in job_records:
+            spamwriter.writerow(
+                (
+                    str(record.userid),
+                    str(record.username),
+                    str(record.jobid),
+                    str(record.t_submit),
+                    str(record.t_run),
+                    str(record.t_inactive),
+                    str(record.nnodes),
+                    str(record.R),
+                )
+            )
+
+
+def print_job_records(job_records):
+    records = {}
+    userid_arr = []
+    username_arr = []
+    jobid_arr = []
+    t_submit_arr = []
+    t_run_arr = []
+    t_inactive_arr = []
+    nnodes_arr = []
+    R_arr = []
+    hostname_arr = []
+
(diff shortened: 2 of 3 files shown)


accounting/accounting_cli_functions.py
```diff
@@ -13,10 +13,170 @@
...
+            + ","
+            + str(record.t_run)
+            + ","
+            + str(record.t_inactive)
+            + ","
+            + str(record.nnodes)
+            + ","
+            + str(record.R)
+            + ",\n"
+        )
```
SteVwonder, line 65: I'd recommend leveraging the `csv` module in the python standard library here.  It'll make it easier to optionally adjust the delimiter, and I think it'll cut the number of lines of code required in half (since you don't need to explicitly provide the delimiter after every element).

accounting/accounting_cli_functions.py
```diff
@@ -13,10 +13,170 @@
...
+
+def get_uid(username):
+    try:
+        return pwd.getpwnam(username).pw_uid
+    except KeyError:
+        return str(username)
+
+
+def write_records_to_file(job_records, output_file):
+    op = open(output_file, "a+")
```
SteVwonder, line 45: This file descriptor is not closed currently.  I'd recommend switching to the context manager usage:

```suggestion
    with open(output_file, "a+") as op:
```

accounting/accounting_cli.py
```diff
@@ -98,6 +137,9 @@ def main():
         print("Unable to open database file")
         sys.exit(-1)
 
+    # set path for output file
+    output_file = args.output_file if args.output_file else "job_records.csv"
```
SteVwonder, line 141: My initial instinct is that this command should write to stdout by default rather than a file.  Do the Slurm utilities write to a file by default?

cmoussa1: You're right - that was something I forgot to implement. I went ahead and changed this so that if no output file is specified, it should just write to stdout, in the same format as it would have if it were written to a file:

```
[fluxuser@137e364cd768 ~]$ flux account -p /var/lib/flux/jobs.db by-user fluxuser
UserID|Username|JobID|T_Submit|T_Run|T_Inactive|Nodes|R
2001|fluxuser|429932937216|1597722545.7235024|1597722545.7498028|1597722545.83819|1|{"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-2"}}]}}
2001|fluxuser|408659427328|1597722544.4558651|1597722544.4830627|1597722544.585075|1|{"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-2"}}]}}
2001|fluxuser|340007059456|1597722540.363928|1597722540.387967|1597722540.4699485|1|{"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-1"}}]}}
2001|fluxuser|319857623040|1597722539.162982|1597722539.1850276|1597722539.2743495|1|{"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0-1"}}]}}
2001|fluxuser|256154533888|1597722535.3659017|1597722535.3874228|1597722535.4645026|1|{"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
2001|fluxuser|241961009152|1597722534.5195136|1597722534.5461538|1597722534.6319313|1|{"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
```

cmoussa1:
Thanks for giving this a first pass so quickly @SteVwonder! I went ahead and addressed your suggestions for now. I will drop the **[WIP]** tag once I add a couple unit tests for the `by-user` subcommand - the function that it calls utilizes interacts with a passwd file and I haven't figured a way to create a sample one for the unit tests.

cmoussa1:
Okay, added a couple unit tests for the `jobs_run_by_username` function and pushed those changes up. I'll go ahead and drop the **[WIP]** tag - I think this is ready for an initial first review.

cmoussa1:
Also, @ryanday36, if you have the time and/or are willing, any feedback on this PR would be greatly appreciated! This PR is directly related to #30.

ryanday36:
Looks good @cmoussa1! My first feature requests would probably be to add a 'human readable' output option (or default) that prints things out in fixed-width columns and a by-bank job selection option. Also, is the host that the job was run on in the R field? is that what "R_lite" is in your example output? If not, that would also be useful information to include.

cmoussa1:
Thanks for your feedback @ryanday36 - just pushed a change so the format of output is slightly cleaner - I went with adding the job records to a `DataFrame` object:

```
 [fluxuser@33a301d49295 ~]$ flux account -p /var/lib/flux/jobs.db by-user fluxuser
   UserID  Username         JobID         T_Submit            T_Run       T_Inactive  Nodes                                                                            R
0    2001  fluxuser  410739802112 1598469186.39630 1598469186.41970 1598469186.49578      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
1    2001  fluxuser  379114749952 1598469184.51090 1598469184.53045 1598469184.60123      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
2    2001  fluxuser  353646936064 1598469182.99277 1598469183.01287 1598469183.08738      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
3    2001  fluxuser  292040998912 1598469179.32122 1598469179.34539 1598469179.43041      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
```

> and a by-bank job selection option. Also, is the host that the job was run on in the R field? is that what "R_lite" is in your example output? If not, that would also be useful information to include.

This is something I believe I would need to figure out with the flux-core team. I am in the process of writing up a ticket so we can get this integrated with job-info, where a _bank_ field could be extracted and added. From there, it should just be another filter option. 

I don't believe the hostname is currently included in `R`. I will reach out about this as well!

ryanday36:
> I don't believe the hostname is currently included in `R`. I will reach out about this as well!

Sounds good. The hostname / hostlist doesn't necessarily have to come from R, but it is a useful thing to have easily available for debugging user jobs.

chu11:
```
    by-user: allows a user to fetch all job records from a certain username
    by-jobid: allows user to fetch job information from a certain jobid
    after-start-time: allows a user to fetch job records after a certain start time
    before-end-time: allows a user to fetch job records before a certain end time
```

I know that this might be just round one of implementation so my input isn't necessarily for this PR but for a later one.

Isn't it possible a user would want to do multiple of these?  like fetch all job records of a certain username that occurred after a certain start time?

If so, perhaps these should be options instead of subcommands, in preparation for multiple at the same time support in a later PR.

PR #34, opened 2020-08-25
create_db.py: add bank table to Flux Accounting database 

### LINKED ISSUE #32: FluxAccounting.db: need a table to describe bank/account hierarchy
As discussed earlier this week, a new table describing bank and account information should be added to **FluxAccounting.db**, the database file storing user account information. 

Currently, the lone table in this database, `association_table`, has a column labeled `account` and an optional `parent_acct` for every user that gets added to this database. 

```
account       tinytext   NOT NULL
parent_acct   tinytext   NOT NULL
```

But the `account` column itself does not have any bank amounts or information itself. We could probably describe this by creating a new table, maybe called `bank_table`:

```
CREATE TABLE IF NOT EXISTS bank_table (
    account_name        text    NOT NULL,
    parent_acct_name    text,
    shares              int     NOT NULL,
    PRIMARY_KEY(account_name)
);
```

I believe creating a bank table would actually remove the need to have a `parent_acct` field in `association_table`, since that information would instead be stored in the bank table. This would also give us the ability to describe a bank/account hierarchy, similar to the usage files @ryanday36 has shared with us:

```
Account       | User   | 
root          |        |         
 root         | root   |  
 account_1    |        |  
  subacct_1   |        |  
  subacct_2   |        |  
   subacct_2_1| user1  |  
   subacct_2_2| user2  |  
  subacct_3   |        | 
```

_note that there is no usage data as of yet, as we still have to implement usage calculation per user and per account. This would probably be a next step after the creation of a bank/account table._

The three fields, `account_name, parent_acct_name, shares`, are the ones I can think of that could give some baseline functionality to creating a bank hierarchy. Can you think of any other fields that might be needed? @ryanday36

cmoussa1:
This is the other PR related to the [end-of-August milestone](https://github.com/flux-framework/flux-accounting/milestone/1).

**Problem**: Currently there is only one table in the flux-accounting database, `association table`, that stores user account information. One of the fields in every record in this table is an _account_ and optional _parent account_ field, which associates a user with an account they can charge jobs against. There is no information on the the account itself, however, such as information related to how many shares it gets on a machine.

---

The first part of this PR removes the _parent_acct_ field from `association_table`, since that information would now be captured as the _parent_bank_ field in `bank_table`. 

The second part looks to add a second table to the flux-accounting database, called `bank_table`. It stores bank names and information on how many shares are allocated to each bank. There are only three fields in the bank table:

```
bank            text    NOT NULL,
parent_bank     text,
shares          int     NOT NULL,
PRIMARY KEY     (bank)
```

With the addition of this new table, a new set of subcommands were added to `flux account` to be able to interact with it:

- `add-bank`: allows the addition and definition of banks, any sub-banks, and their allocated shares.

- `view-bank`: allows user to view bank information.

- `delete-bank`: allows the deletion of a bank. If the bank specified is a parent bank, it will delete all of its sub banks as well.

- `edit-bank`: allows editing of bank allocations.

- `print-hierarchy`: prints all banks, sub banks, and users associated with those banks via a `JOIN` operation between `association_table` and `bank_table` (where the _account_ field in `association_table` is equal to the _bank_ field in `bank_table`).   

Fixes #32

### DIFF
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -141,6 +141,12 @@ def main():
         "bank", help="bank name", metavar="BANK",
     )
 
+    subparser_delete_bank = subparsers.add_parser("delete-bank", help="remove a bank")
+    subparser_delete_bank.set_defaults(func="delete_bank")
+    subparser_delete_bank.add_argument(
+        "bank", help="bank name", metavar="BANK",
+    )
+
     args = parser.parse_args()
 
     # if we are creating the DB for the first time, we need
@@ -187,6 +193,8 @@ def main():
             aclif.view_jobs_before_end_time(conn, args.end_time, args.output_file)
         elif args.func == "view_bank":
             aclif.view_bank(conn, args.bank)
+        elif args.func == "delete_bank":
+            aclif.delete_bank(conn, args.bank)
         else:
             print(parser.print_usage())
     finally:
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -61,9 +61,6 @@ def main():
         "--account", help="account to charge jobs against", metavar="ACCOUNT",
     )
     subparser_add_user.add_argument(
-        "--parent-acct", help="parent account", default="", metavar="PARENT_ACCOUNT",
-    )
-    subparser_add_user.add_argument(
         "--shares", help="shares", default=1, metavar="SHARES",
     )
     subparser_add_user.add_argument(
@@ -164,7 +161,6 @@ def main():
                 args.username,
                 args.admin_level,
                 args.account,
-                args.parent_acct,
                 args.shares,
                 args.max_jobs,
                 args.max_wall_pj,
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -147,6 +147,17 @@ def main():
         "bank", help="bank name", metavar="BANK",
     )
 
+    subparser_edit_bank = subparsers.add_parser(
+        "edit-bank", help="edit a bank's allocation"
+    )
+    subparser_edit_bank.set_defaults(func="edit_bank")
+    subparser_edit_bank.add_argument(
+        "bank", help="bank", metavar="BANK",
+    )
+    subparser_edit_bank.add_argument(
+        "--shares", help="new shares value", metavar="SHARES",
+    )
+
     args = parser.parse_args()
 
     # if we are creating the DB for the first time, we need
@@ -195,6 +206,8 @@ def main():
             aclif.view_bank(conn, args.bank)
         elif args.func == "delete_bank":
             aclif.delete_bank(conn, args.bank)
+        elif args.func == "edit_bank":
+            aclif.edit_bank(conn, args.bank, args.shares)
         else:
             print(parser.print_usage())
     finally:
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -133,6 +133,14 @@ def main():
         "dbpath", help="specify location of database file", metavar=("DATABASE PATH")
     )
 
+    subparser_view_bank = subparsers.add_parser(
+        "view-bank", help="view bank information"
+    )
+    subparser_view_bank.set_defaults(func="view_bank")
+    subparser_view_bank.add_argument(
+        "bank", help="bank name", metavar="BANK",
+    )
+
     args = parser.parse_args()
 
     # if we are creating the DB for the first time, we need
@@ -177,6 +185,8 @@ def main():
             aclif.view_jobs_after_start_time(conn, args.start_time, args.output_file)
         elif args.func == "view_jobs_before_end_time":
             aclif.view_jobs_before_end_time(conn, args.end_time, args.output_file)
+        elif args.func == "view_bank":
+            aclif.view_bank(conn, args.bank)
         else:
             print(parser.print_usage())
     finally:
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -61,6 +61,9 @@ def main():
         "--account", help="account to charge jobs against", metavar="ACCOUNT",
     )
     subparser_add_user.add_argument(
+        "--parent-acct", help="parent account", default="", metavar="PARENT_ACCOUNT",
+    )
+    subparser_add_user.add_argument(
         "--shares", help="shares", default=1, metavar="SHARES",
     )
     subparser_add_user.add_argument(
@@ -131,6 +134,18 @@ def main():
     subparser_create_db.set_defaults(func="create_db")
     subparser_create_db.add_argument(
         "dbpath", help="specify location of database file", metavar=("DATABASE PATH")
+    )
+
+    subparser_add_bank = subparsers.add_parser("add-bank", help="add a new bank")
+    subparser_add_bank.set_defaults(func="add_bank")
+    subparser_add_bank.add_argument(
+        "bank", help="bank name", metavar="BANK",
+    )
+    subparser_add_bank.add_argument(
+        "--parent-bank", help="parent bank name", default="", metavar="PARENT BANK"
+    )
+    subparser_add_bank.add_argument(
+        "shares", help="number of shares to allocate to bank", metavar="SHARES"
     )
 
     subparser_view_bank = subparsers.add_parser(
@@ -207,6 +222,8 @@ def main():
             aclif.view_jobs_after_start_time(conn, args.start_time, args.output_file)
         elif args.func == "view_jobs_before_end_time":
             aclif.view_jobs_before_end_time(conn, args.end_time, args.output_file)
+        elif args.func == "add_bank":
+            aclif.add_bank(conn, args.bank, args.shares, args.parent_bank)
         elif args.func == "view_bank":
             aclif.view_bank(conn, args.bank)
         elif args.func == "delete_bank":
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -158,6 +158,11 @@ def main():
         "--shares", help="new shares value", metavar="SHARES",
     )
 
+    subparser_print_hierarchy = subparsers.add_parser(
+        "print-hierarchy", help="print accounting database"
+    )
+    subparser_print_hierarchy.set_defaults(func="print_hierarchy")
+
     args = parser.parse_args()
 
     # if we are creating the DB for the first time, we need
@@ -208,6 +213,8 @@ def main():
             aclif.delete_bank(conn, args.bank)
         elif args.func == "edit_bank":
             aclif.edit_bank(conn, args.bank, args.shares)
+        elif args.func == "print_hierarchy":
+            aclif.print_hierarchy(conn)
         else:
             print(parser.print_usage())
     finally:
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -263,9 +263,7 @@ def view_user(conn, user):
         print(e_database_error)
 
 
-def add_user(
-    conn, username, admin_level, account, parent_acct, shares, max_jobs, max_wall_pj
-):
+def add_user(conn, username, admin_level, account, shares, max_jobs, max_wall_pj):
 
     # insert the user values into the database
     try:
@@ -278,12 +276,11 @@ def add_user(
                 user_name,
                 admin_level,
                 account,
-                parent_acct,
                 shares,
                 max_jobs,
                 max_wall_pj
             )
-            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+            VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
             """,
             (
                 int(time.time()),
@@ -292,7 +289,6 @@ def add_user(
                 username,
                 admin_level,
                 account,
-                parent_acct,
                 shares,
                 max_jobs,
                 max_wall_pj,
@@ -319,7 +315,6 @@ def edit_user(conn, username, field, new_value):
         "user_name",
         "admin_level",
         "account",
-        "parent_acct",
         "shares",
         "max_jobs",
         "max_wall_pj",
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -275,6 +275,46 @@ def delete_bank(conn, bank):
     conn.commit()
 
 
+def edit_bank(conn, bank, shares):
+    total_sub_bank_shares = 0
+    parent_bank = ""
+    parent_bank_shares = 0
+    try:
+        # if bank is a parent bank, the new value
+        # should not be less than the total shares
+        # allocated to all sub banks
+        select_stmt = "SELECT shares FROM bank_table WHERE parent_bank=?"
+        dataframe = pd.read_sql_query(select_stmt, conn, params=(bank,))
+        for index, row in dataframe.iterrows():
+            total_sub_bank_shares += row["shares"]
+        if int(shares) < total_sub_bank_shares:
+            print(
+                "New shares amount would be less than total shares allocated to subaccounts"
+            )
+            sys.exit(-1)
+        # if bank is a sub bank, the new value
+        # should not be greater than its parent bank
+        select_stmt = "SELECT parent_bank FROM bank_table WHERE bank=?"
+        dataframe = pd.read_sql_query(select_stmt, conn, params=(bank,))
+        for index, row in dataframe.iterrows():
+            parent_bank = row["parent_bank"]
+        # if bank specified does not have a parent bank, just continue
+        if parent_bank != "":
+            check_parent_bank(conn, shares, parent_bank)
+        else:
+            pass
+
+    except pd.io.sql.DatabaseError as e_database_error:
+        print(e_database_error)
+
+    # edit value in bank_table
+    conn.execute(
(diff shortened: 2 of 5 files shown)


cmoussa1:
OK, rebased after #33 and fixed the conflicts. I think the one main thing I wanted to get a review on was the `edit-bank` subcommand. I only added the ability to edit the amount of shares that a bank was allocated, not the name of the bank itself. In my opinion I think it would be okay to leave off that functionality until later. Is that reasonable for now? @dongahn

accounting/accounting_cli.py
```diff
@@ -124,6 +127,18 @@ def main():
         "end_time", help="end time", metavar="END TIME",
     )
 
+    subparser_add_bank = subparsers.add_parser("add-bank", help="add a new bank")
+    subparser_add_bank.set_defaults(func="add_bank")
+    subparser_add_bank.add_argument(
+        "--bank", help="bank name", metavar="BANK",
```
dongahn, line 133: Same comment as before. Would it be more consistent to make this a mandatory argument? If you want to keep it this way, maybe you can change the way you specify `delete-bank` instead... As far as you make this consistent, this works for me.

accounting/accounting_cli_functions.py
```diff
@@ -248,6 +248,196 @@ def view_jobs_before_end_time(conn, time_before, output_file):
...
+            if int(total_sub_bank_shares) >= int(parent_bank_shares):
+                print("Total shares for parent account are already maxed")
+                sys.exit(-1)
+            # if the current total amount of shares in all the sub banks
+            # PLUS the number of shares trying to be added is greater than the
+            # parent's banks shares, the parent account WILL exceed its max
+            # allocation and thus the account won't be added
+            if int(total_sub_bank_shares) + int(shares) > int(parent_bank_shares):
+                print("Total shares will exceed parent account's allocation")
+                sys.exit(-1)
```
SteVwonder, line 294: I will admit that my familiarity with how we plan on implementing shares is very little, but it appears that the constraints enforced here (i.e., `parent shares >= sum(child shares)`) is not enforced by LLNL's existing banks: https://computing.llnl.gov/tutorials/moab/#Banks

It looks like the existing bank algorithms just apply a per-level normalization.  Do we instead plan to enforce a total share limit?  Is this a constraint that Ryan or Greg asked for?  If so, I think we may want to enable fractional shares to enable finer grained splitting (i.e., splitting 10 shares across 100 users).

accounting/accounting_cli_functions.py
```diff
@@ -248,6 +248,74 @@ def view_jobs_before_end_time(conn, time_before, output_file):
     return job_records
 
 
+def add_bank(conn, bank, shares, parent_bank=""):
+    parent_bank_shares = 0
+    total_sub_bank_shares = 0
+
+    # if the parent bank is not "", that means the account
+    # trying to be added wants to be placed under a parent bank
+    if parent_bank != "":
```
dongahn, line 257: Can the bank hierarchy share invariant/inequality checks be factored out so that you can use the same methods for either `add` or `edit`? If a bug is introduced, it is always to fix one place as opposed to multiple duplicate places.

accounting/accounting_cli_functions.py
```diff
@@ -248,6 +248,74 @@ def view_jobs_before_end_time(conn, time_before, output_file):
...
+            # if the current total amount of shares in all the sub banks
+            # PLUS the number of shares trying to be added is greater than the
+            # parent's banks shares, the parent account WILL exceed its max
+            # allocation and thus the account won't be added
+            if int(total_sub_bank_shares) + int(shares) > int(parent_bank_shares):
+                print("Total shares will exceed parent account's allocation")
+                sys.exit(-1)
+
+        except pd.io.sql.DatabaseError as e_database_error:
+            print(e_database_error)
```
dongahn, line 297: Same comments for error handling and exit code.

dongahn:
@cmoussa1: This is a significant addition based on the feedback you got from Ryan and others. Thank you for addressing them. 

The only thing is that there are quite a few changes that will need to be addressed. So I would say, have a few extra days before tagging your "Aug" release. My bad not being able to get to this sooner.

cmoussa1:
Thanks for your feedback @dongahn - I will work on addressing your suggestions and push an update as soon as I can.

accounting/accounting_cli_functions.py
```diff
@@ -248,6 +248,196 @@ def view_jobs_before_end_time(conn, time_before, output_file):
...
+    # if the parent bank is not "", that means the account
+    # trying to be added wants to be placed under a parent bank
+    if parent_bank != "":
+        try:
+            select_stmt = "SELECT shares FROM bank_table where bank=?"
+            dataframe = pd.read_sql_query(select_stmt, conn, params=(parent_bank,))
+            # if length of dataframe is 0, that means the parent bank wasn't found
+            if len(dataframe.index) == 0:
+                print("Parent account not found in bank table")
+                sys.exit(-1)
```
SteVwonder, line 264: Typically, you want fatal errors like this to get printed to stderr, and you don't want imported functions to fatally kill the interpreter with `sys.exit`.  My recommendation would be to raise an appropriately named exception (e.g., `RuntimeException`) with the string as the message within the exception.  Then have the top-level caller - `main` in this case - catch the exception, print the message to stderr and exit the interpreter with an exit code.

Same message applies elsewhere in this file that print and sys.exit are paired for an error.

accounting/accounting_cli_functions.py
```diff
@@ -248,6 +248,196 @@ def view_jobs_before_end_time(conn, time_before, output_file):
...
+    # if bank is a parent_bank, find all
+    # sub banks and remove them as well
+    while True:
+        select_stmt = "SELECT bank FROM bank_table WHERE parent_bank=?"
+        dataframe = pd.read_sql_query(
+            select_stmt, conn, params=(last_parent_bank_seen,)
+        )
+        for index, row in dataframe.iterrows():
+            cursor.execute(delete_stmt, (row["bank"],))
+            last_parent_bank_seen = row["bank"]
```
SteVwonder, line 351: If multiple rows are deleted, then some banks will be missed, right?  Take the following bank hierarchy for example:
```
      A
      |
   |-----|
   B     C
   |     |
 |---| |---|
 D   E F   G
```

If `A` is passed into `delete_bank`, then it will be deleted by the first delete execute.  The first `select_stmt` will return `B, C`.  `B` will be deleted and `last_parent_bank_seen` set to `B`.  `C` will be deleted and `last_parent_bank_seen` set to `C`.  The while loop will go to the next iteration, and the select statement will return `F` and `G` (which will then be deleted).   `D` and `E` will still be left in the database.

My recommendation would be to either make this function recursive or to make `last_parent_bank_seen` a queue that parent banks get pushed onto and popped from (rather than just a normal variable).

I also recommend adding a deep bank hierarchy example like the one in this comment to the test suite to ensure there aren't any orphaned banks after a delete.

accounting/create_db.py
```diff
@@ -44,6 +44,20 @@ def create_db(filepath):
...
 
+    # Bank Table
+    logging.info("Creating bank_table in DB...")
+    conn.execute(
+        """
+            CREATE TABLE IF NOT EXISTS bank_table (
+                bank            text    NOT NULL,
+                parent_bank     text,
+                shares          int     NOT NULL,
+                PRIMARY KEY     (bank)
```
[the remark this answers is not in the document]
cmoussa1: OK, I agree. I will open an issue to track this so I don't forget.

accounting/accounting_cli.py
cmoussa1: Agreed. I think both `bank` and `shares` should be mandatory arguments; `parent-bank` will be an optional argument. I'll make this change now.

accounting/accounting_cli.py
```diff
@@ -166,6 +207,8 @@ def main():
             aclif.view_jobs_after_start_time(conn, args.start_time, args.output_file)
         elif args.func == "view_jobs_before_end_time":
             aclif.view_jobs_before_end_time(conn, args.end_time, args.output_file)
+        elif args.func == "add_bank":
+            aclif.add_bank(conn, args.bank, args.shares, args.parent_bank)
```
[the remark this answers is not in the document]
cmoussa1: Ugh. I think things got mixed up trying to rebase after #33; I will try and fix this now.

accounting/accounting_cli_functions.py
```diff
@@ -248,6 +248,196 @@ def view_jobs_before_end_time(conn, time_before, output_file):
...
+    if parent_bank != "":
+        try:
+            select_stmt = "SELECT shares FROM bank_table where bank=?"
+            dataframe = pd.read_sql_query(select_stmt, conn, params=(parent_bank,))
+            # if length of dataframe is 0, that means the parent bank wasn't found
+            if len(dataframe.index) == 0:
+                print("Parent account not found in bank table")
+                sys.exit(-1)
+            # fetch parent account from account
+            for index, row in dataframe.iterrows():
```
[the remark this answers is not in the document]
cmoussa1: Sorry, I might be misunderstanding here - do you mean trying to add multiple matching banks should result in an error? If that's the case, that error is handled as an SQLite error:

```
$ flux account -p /var/lib/flux-ba/FluxAccounting.db add-bank root 100
$ flux account -p /var/lib/flux-ba/FluxAccounting.db add-bank root 100
UNIQUE constraint failed: bank_table.bank
```

accounting/accounting_cli_functions.py
cmoussa1: Thanks for this explanation and the reference - the `parent shares >= sum(child shares)` constraint should not be included here. My plan was just to emulate the current system. I'll remove that check now.

accounting/accounting_cli_functions.py
cmoussa1: Thanks for explaining this to me @SteVwonder! This is really helpful.

You're absolutely right. The `delete-bank` function will need some rework. Maybe we can take the same stance as the issue with the error propagation - I can create an issue to track this, and for now, the `delete-bank` subcommand will be refactored to just delete one bank at a time, regardless if it is a parent bank or not. Then once this lands, I can work on this in the immediate future. The reason being I think I might need some more time on figuring this out. 

Thoughts? Would that be okay with you?

accounting/accounting_cli_functions.py
```diff
@@ -288,6 +288,54 @@ def delete_bank(conn, bank):
...
+        # if bank is a parent bank, the new value
+        # should not be less than the total shares
+        # allocated to all sub banks
+        select_stmt = "SELECT shares FROM bank_table WHERE parent_bank=?"
+        dataframe = pd.read_sql_query(select_stmt, conn, params=(bank,))
+        for index, row in dataframe.iterrows():
+            total_sub_bank_shares += row["shares"]
+        if int(shares) < total_sub_bank_shares:
+            print(
+                "New shares amount would be less than total shares allocated to subaccounts"
```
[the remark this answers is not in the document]
cmoussa1: Yes, only inequality.

cmoussa1:
I'm sorry that took me a little while. Pushed a number of changes:

- split up the first two commits into 1) making `conn` a global variable in **test_create_db.py**, and 2) removing `parent_acct` from `association_table`.

- made the `bank` and `shares` args for the `add-bank` subcommand mandatory args.

- made the `bank` arg for the `edit-bank` subcommand a mandatory arg.

- added a common `check_parent_bank()` for both the `add-bank` and `edit-bank` subcommands instead of having similar code in both functions.

- removed the propagation feature in `delete-bank`. I think I am on planning on creating an issue ticket to track this and implement it instead in the near future, because I think I'll need more time to figure this out. The same will be done for the error handling. 

@dongahn and @SteVwonder thank you again for your feedback on this so far. I appreciate all the advice and suggestions.

accounting/accounting_cli_functions.py
SteVwonder: > Thoughts? Would that be okay with you?

Sounds good to me :)

cmoussa1:
OK, rebased (hopefully successfully this time 😅 ) to catch up after #36.

accounting/accounting_cli_functions.py
SteVwonder: > the parent shares >= sum(child shares) constraint should not be included here.

As far as I can tell, none of the share constraints exist in LLNL's existing banks.  The referenced example has a bank (`bank1`) with 3 shares that has children with 16, 48, and 32 shares each.

That being said, I ran `sshare` on a production cluster, and it looks like by convention that most every bank/account follows the constraints that you have guaranteed.  The only exceptions are the "expired" bank (which has all of the old accounts in it) and this one:
```
❯ sshare -A lprpp,lprpp-1,lprpp-2
             Account       User  RawShares  NormShares    RawUsage  EffectvUsage  FairShare
-------------------- ---------- ---------- ----------- ----------- ------------- ----------
lprpp                                 1945    0.026231  1375628740      0.133639
 lprpp-1                              1588    0.500000  1375628694      1.000000
 lprpp-2                              1588    0.500000          45      0.000000
```

In summary, I think the code either include all of the constraints that you initially had or none of them.  In the interest of getting this PR merged as soon as possible, I'm fine with having them all in here, and we can discuss this further in an issue.  Given that, do you mind adding the `parent shares >= sum(child shares)` constraint back in?

cmoussa1: This seems like a good plan. I don't mind at all - just pushed an update that adds that constraint back in.

cmoussa1:
Is there anything else that should be changed immediately before this gets approved? I believe the error propagation stuff and the rework for `delete-bank` was decided it would be considered future work after v0.2.0; I could be wrong though.

cmoussa1:
OK, just created issues to track the following based on feedback received in this PR:

#38 - error handling and propagation in the bank subcommands
#39 - using a fixed data type for the bank table in **FluxAccounting.db**
#40 - clarification on the constraints imposed on shares in the `bank_table`
#41 - implementing a recursive solution on the `delete-bank` subcommand for parent banks

@SteVwonder and @dongahn did I miss anything else?

SteVwonder (approved):
Thanks @cmoussa1!  LGTM!

test/test_accounting_cli.py
```diff
@@ -111,7 +111,7 @@ def setUpClass(self):
 
     # add a valid user to association_table
     def test_01_add_valid_user(self):
-        aclif.add_user(conn, "fluxuser", "1", "acct", "10", "100", "60")
+        aclif.add_user(acct_conn, "fluxuser", "1", "acct", "10", "100", "60")
```
dongahn, line 114: Hmmm. I am confused as to how this change was possible... What commit introduced the connection name change from `conn` to `acct_conn`? Doesn't this change belong that commit as opposed to this commit?

dongahn (changes requested):
Thank you for addressing all of my comments. The first two new reviews are just comments. But could you please clarify my question for the last commit? I couldn't keep track of some of the changes that this commit made. As it is, the revision before this commit seems incorrect as the `conn` variable doesn't exist?

test/test_accounting_cli.py
cmoussa1: I believe this change was made in db8d0282f516ecae4424625374c550f297ec6b2f, right? When all of the unit tests for the subcommands were added? Is there another commit that also makes this change?

dongahn: Yes. But I am curious about its previous commit. Was `conn` a valid variable in the previous commit? The reason I ask is, at the current commit db8d0282f516ecae4424625374c550f297ec6b2f, I don't see your change to define `acct_conn`. You just used it; that means that variable was probably defined previously. 

If that's the case, should the previous commit also not have the reference of `conn` but `acct_conn` instead? As is, I'm a bit concerned that the previous commit may have unintentionally broken.

cmoussa1: The first commit that introduced the `acct_conn` variable was 374242d4f6db01468d65bf376d04d6e888c7a146 - this is from the PR that added the front end to the job archive (#33), as two separate connections were needed (one to the jobs DB, and one to the accounting DB).

dongahn: Hmmm. From your 374242d, I see you made the necessary change:

```python
 aclif.add_user(acct_conn, "fluxuser", "1", "acct", "pacct", "10", "100", "60")
```

So why the diff says you made `conn` to `acct_conn` in the current commit?

dongahn: Ah I actually see the changes from 3c1951f75e46176d2f7b323fa2899960af1136cd. Perhaps the changes in that commit is incorrect?

cmoussa1: I'm not sure. I know I had to rebase this PR after #33, but I'm not sure why the change is recognized in this most recent commit...

cmoussa1: Ah; so I guess I made the change to `acct_conn` in 374242d, reverted it in 3c1951f, and then brought it back in the last commit, db8d028?

dongahn: My guess is this went wrong through rebasing and conflict resolution. Please drop that test script change in 3c1951f (you can use git rebase -i -- then edit) and see if that can clear up this churn.

cmoussa1: You are probably right - so sorry for the confusion. Just pushed an update.

dongahn:
OK. This looks good to me. I will add MWP so that you can proceed with your release.

dongahn (approved):

cmoussa1:
Thanks!

PR #36, opened 2020-08-28
accounting_cli.py: add create-db subcommand

### LINKED ISSUE #35: create_db.py: add ability to create database with subcommand instead of calling script directly
The current way to create the flux-accounting database is to call the Python script `create_db.py` directly:

```
$ ./create_db.py -p path/to/FluxAccounting.db
```

It might be a good idea to instead wrap this under a subcommand so that it doesn't have to be run directly from the same directory where the file lies:

```
$ flux account create-db path/to/FluxAccounting.db
```

This should be relatively straightforward change, since the script can already be imported in order to run some of the unit tests:

```python
from accounting import create_db as c

# create database and make sure it exists
def test_00_test_create_db(self):
    c.create_db("FluxAccounting.db")

    assert os.path.exists("FluxAccounting.db")
```

cmoussa1:
This PR would probably be nice to land before our end-of-August release deadline as well. :-)

**Problem**: As mentioned in #35, in order to initially create the flux-accounting database, you have to go into the `flux-accounting` directory and call the `create_db.py` script directly:

```
[flux-framework/flux-accounting/accounting] $ ./create_db.py -p path/to/FluxAccounting.db
```

---

This PR removes `main()` from **create_db.py** and instead adds a new subcommand to **accounting_cli.py**:  `create-db`, which creates **FluxAccounting.db**. It takes one argument, the path to the database file.

```
$ flux account create-db path/to/FluxAccounting.db
```

Keeping the same behavior as before, ownership of the directory that the database file is placed in is required for the database to be created.

Fixes #35

### DIFF
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -17,6 +17,7 @@ import sys
 import pandas as pd
 
 from accounting import accounting_cli_functions as aclif
+from accounting import create_db as c
 
 
 def main():
@@ -127,7 +128,21 @@ def main():
         "end_time", help="end time", metavar="END TIME",
     )
 
+    subparser_create_db = subparsers.add_parser(
+        "create-db", help="create the flux-accounting database"
+    )
+    subparser_create_db.set_defaults(func="create_db")
+    subparser_create_db.add_argument(
+        "dbpath", help="specify location of database file", metavar=("DATABASE PATH")
+    )
+
     args = parser.parse_args()
+
+    # if we are creating the DB for the first time, we need
+    # to ONLY create the DB and then exit out successfully
+    if args.func == "create_db":
+        c.create_db(args.dbpath)
+        sys.exit(0)
 
     # try to open database file; will exit with -1 if database file not found
     path = args.path if args.path else "FluxAccounting.db"
--- a/accounting/create_db.py
+++ b/accounting/create_db.py
@@ -46,30 +46,3 @@ def create_db(filepath):
     logging.info("Created association_table successfully")
 
     conn.close()
-
-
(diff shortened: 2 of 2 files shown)


dongahn:
@cmoussa1: this still has conflicting files. Did you want to resolve the conflicts and update the PR?

cmoussa1:
Oh! Sorry about that @dongahn - I resolved #34 but not this one. OK, conflicts should be resolved here.

dongahn:
@cmoussa1: 

The change looks good to me. Let me put MWP on this. 

However, I see that there isn't any error path for `create_db` -- in a form of neither return code nor python exception. 

So as is, even if `create_db` fails, `flux-account` would return 0 exit code, which can be 1) confusing and 2) making it difficult to write automatic tests around this. 

Could you open up an issue ticket so that you can address this later. 

If you can use that ticket as a way to look through the error paths for major methods, that'd be great.

Thanks.

cmoussa1:
Will do; thanks for pointing that out.

cmoussa1:
@Mergifyio refresh

cmoussa1:
Hm. Still hasn't been merged yet. Maybe I need to `refresh` once again?

cmoussa1:
@Mergifyio refresh

cmoussa1:
Oh, I think I know why this isn't being merged. I think it's because it hasn't officially been "approved" yet on GitHub. 

```
"#approved-reviews-by>0"
```

@dongahn if this looks good to you can you approve it when you get the chance? Then I think the MWP label will take effect.

dongahn:
Got it.

PR #42, opened 2020-09-09
bank_table: change primary key to a fixed data type

### LINKED ISSUE #39: create_db.py: primary key for bank_table should be a fixed data type
Peeled off from one of the comments from #34: 

> I think people typically use a fixed data type like integer as the primary key because this will have performance advantages: Please see, for example, the following -- https://stackoverflow.com/questions/15477005/what-are-the-pros-and-cons-for-choosing-a-character-varying-data-type-for-primar

We may want to consider changing the primary key for the `bank_table` to an integer, maybe in the form of a `bank_id` or something.

cmoussa1:
As mentioned in #39, the current primary key for the `bank_table` is the name of the bank, whose data type is `text`, a variable data type. Instead, it is considered best practice to use a fixed data type for a primary key (ref: [this StackOverflow post](https://stackoverflow.com/questions/15477005/what-are-the-pros-and-cons-for-choosing-a-character-varying-data-type-for-primar) that @dongahn pointed me to). 

---

This PR adds a new field to the `bank_table` called _bank_id_ which will serve as the new primary key. It is of type `int`. It will auto increment as more banks are added to this table. 

Fixes #39

### DIFF
--- a/accounting/create_db.py
+++ b/accounting/create_db.py
@@ -45,15 +45,15 @@ def create_db(filepath):
     logging.info("Created association_table successfully")
 
     # Bank Table
+    # bank_id gets auto-incremented with every new entry
     logging.info("Creating bank_table in DB...")
     conn.execute(
         """
             CREATE TABLE IF NOT EXISTS bank_table (
-                bank            text    NOT NULL,
-                parent_bank     text,
-                shares          int     NOT NULL,
-                PRIMARY KEY     (bank)
-
+                bank_id     integer PRIMARY KEY,
+                bank        text    NOT NULL,
+                parent_bank text,
+                shares      int     NOT NULL
         );"""
     )
     logging.info("Created bank_table successfully")
--- a/test/test_create_db.py
+++ b/test/test_create_db.py
@@ -90,6 +90,13 @@ class TestDB(unittest.TestCase):
         dataframe = pd.read_sql_query(select_stmt, conn)
         self.assertEqual(len(dataframe.index), 2)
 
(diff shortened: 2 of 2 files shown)


chu11:
LGTM, although I'm curious how its autoincrementing.  I thought you had to specify the `AUTO_INCREMENT` keyword?  If mysql defaults to turning it on, perhaps just add a comment to that effect.

chu11 (approved):

cmoussa1:
Thanks @chu11! 

[Here](https://sqlite.org/autoinc.html) is the reference to how I implemented this, but essentially just using an `INTEGER PRIMARY KEY` serves as an alias for a `ROWID` (a 64-bit signed integer), which gets auto-incremented with a new entry.

> On an INSERT, if the ROWID or INTEGER PRIMARY KEY column is not explicitly given a value, then it will be filled automatically with an unused integer, usually one more than the largest ROWID currently in use. This is true regardless of whether or not the AUTOINCREMENT keyword is used.

The `AUTOINCREMENT` apparently introduces a lot of overhead and is recommended to be avoided if possible.

chu11:
@cmoussa1  ok, how about adding a comment in `create_db.py` just so its documented somewhere.

cmoussa1:
Will do. Just pushed an update that adds a comment about `bank_id`'s auto-increment feature.

cmoussa1:
@chu11 am I okay to add MWP? :-)

chu11:
@cmoussa1 yup, go for it!

PR #44, opened 2020-09-14
bank_table: remove bank constraints from add_bank(), edit_bank()

### LINKED ISSUE #40: bank_table: need clarification on constraints with shares
Noted in the comments from #36 (thanks @SteVwonder for bringing this up!):

> And with the shares, it doesn't look like the existing bank systems enforce that the sum of the child shares must be less than or equal to the parents shares. There is just a per-level normalization step that needs to happen (see my inline comment for more details). Maybe this is something we need to discuss during a coffee hour, but it could also just be a design choice that I wasn't aware of, so feel free to ignore if so.

Some of the constraints I included in the `add-bank` subcommands turned out to be not technically enforced by LLNL's existing banks, notably that `parent shares >= sum(child shares)`. In fact, none of the share constraints exist in LLNL's existing banks. The referenced example has a bank (bank1) with 3 shares that has children with 16, 48, and 32 shares each.

If that is in fact the case, then I need to remove those constraints from `add-bank` and `edit-bank`. 

I should confirm with @ryanday36 on this - are constraints on shares required? Should we start to enforce those constraints in flux-accounting?

cmoussa1:
**Problem**: Currently, `add_bank()` and `edit_bank()` impose constraints on the amount of shares sub banks can have compared to their parent bank; notably, `sum(sub_bank_shares) <= parent_bank_shares`. Thanks to @SteVwonder for first pointing this out (and to @ryanday36 for clarifying in #40), but these constraints are actually not required.

--- 

This PR removes those checks from `add_bank()` and `edit_bank()` and removes the unit tests that interacts with those constraints. 

Fixes #40

### DIFF
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -248,56 +248,16 @@ def view_jobs_before_end_time(conn, time_before, output_file):
     return job_records
 
 
-def check_parent_bank(conn, shares, parent_bank):
-    select_stmt = "SELECT shares FROM bank_table where bank=?"
-    dataframe = pd.read_sql_query(select_stmt, conn, params=(parent_bank,))
-    # if length of dataframe is 0, that means the parent bank wasn't found
-    if len(dataframe.index) == 0:
-        print("Parent account not found in bank table")
-        sys.exit(-1)
-    # fetch parent account from account
-    for index, row in dataframe.iterrows():
-        parent_bank_shares = row["shares"]
-    if int(shares) > int(parent_bank_shares):
-        print("Shares must not be greater than parent account's shares")
-        sys.exit(-1)
-
-    return parent_bank_shares
-
-
 def add_bank(conn, bank, shares, parent_bank=""):
-    parent_bank_shares = 0
-    total_sub_bank_shares = 0
-
     # if the parent bank is not "", that means the account
     # trying to be added wants to be placed under a parent bank
     if parent_bank != "":
         try:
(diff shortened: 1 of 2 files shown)


cmoussa1:
Rebased to catch up after #42.

accounting/accounting_cli_functions.py
```diff
@@ -349,44 +309,16 @@ def delete_bank(conn, bank):
...
-        dataframe = pd.read_sql_query(select_stmt, conn, params=(bank,))
-        for index, row in dataframe.iterrows():
-            parent_bank = row["parent_bank"]
-        # if bank specified does not have a parent bank, just continue
-        if parent_bank != "":
-            check_parent_bank(conn, shares, parent_bank)
-        else:
-            pass
-
+        # edit value in bank_table
```
SteVwonder, line 313: Does this function need to check for `shares <= 0` and raise an exception if so?

SteVwonder (commented):
Generally LGTM! One error checking question below:

accounting/accounting_cli_functions.py
cmoussa1: That's a good idea. Thanks for pointing that out! Just pushed an update that adds an exception if `shares <= 0`, along with a unit test that interacts with it.

SteVwonder (approved):
Thanks @cmoussa1! LGTM!

cmoussa1:
Awesome! I'll go ahead and set MWP then. Thanks!

PR #45, opened 2020-09-14
repo: add Travis CI, Codecov to flux-accounting

cmoussa1:
This PR adds Travis CI and Codecov to `flux-accounting`, as well as the badges indicating successful builds and the coverage percentage to the top of the **README**.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+python:
+  - "3.6"
+  - "3.7"
+  - "3.8"
+install:
+  - pip install pandas
+  - pip install codecov
+# command to run tests
+script:
+  - coverage run -m unittest discover
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+[![Build Status](https://travis-ci.org/flux-framework/flux-accounting.svg?branch=master)](https://travis-ci.org/flux-framework/flux-accounting)
+[![codecov](https://codecov.io/gh/flux-framework/flux-accounting/branch/master/graph/badge.svg)](https://codecov.io/gh/flux-framework/flux-accounting)
+
 _NOTE: The interfaces of flux-accounting are being actively developed and are not yet stable. The Github issue tracker is the primary way to communicate with the developers._
 
 ## flux-accounting
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,13 @@
+coverage:
+ round: down
+ range: "65...100"
+
+ # Allow coverage to drop up to 0.5%
+ status:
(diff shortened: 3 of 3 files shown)


cmoussa1:
Okay, now that I've added the **codecov.yml** file, I'll drop the [WIP] tag.

cmoussa1:
Rebased to catch up after #42.

dongahn:
@cmoussa1: this LGTM. @grondo?

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+  - "3.7"
+  - "3.8"
+install:
+  - pip install pandas
+  - pip install codecov
+# command to run tests
+script:
+  - coverage run -m unittest discover
+after_success:
+  - bash <(curl -s https://codecov.io/bash) -t b5b039a8-7446-4655-b25e-bda6f3168fd7
```
grondo, line 13: The private codecov token for flux-accounting should not be checked in to your repo in the clear.

Instead set `CODECOV_TOKEN` as a "secret" environment variable in the Travis-CI settings page for flux-accounting:

https://travis-ci.com/github/flux-framework/flux-accounting/settings

Then that variable and your private codecov token will not be exposed to anyone except maintainers.

grondo (changes requested):
Just onec comment about the private codecov token. This should be fixed before merging the PR.

.travis.yml
cmoussa1: Thanks Mark. I believe I was able to set that environment variable correctly and pushed those changes up. Let me know if this is what you were thinking of. 

Looks like Codecov said that no coverage was uploaded, but looking at the [Error Reference](https://docs.codecov.io/docs/error-reference#section-missing-base-commit), I am wondering if it is because these changes haven't landed on the `master` branch yet?

grondo: Yeah, since there is no coverage report for the main branch, codecov cannot report changes. That is expected.

grondo (approved):

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+  - "3.7"
+  - "3.8"
+install:
+  - pip install pandas
+  - pip install codecov
+# command to run tests
+script:
+  - coverage run -m unittest discover
+after_success:
+  - bash <(curl -s https://codecov.io/bash) -t $CODECOV_TOKEN
```
grondo, line 13: You don't actually need to use `-t $CODECOV_TOKEN` here, just leave this off, the script will look for CODECOV_TOKEN.

Otherwise, LGTM!

cmoussa1: Oh! My apologies. Just pushed a fix to remove that.

grondo:
LGTM! Feel free to set MWP!

PR #47, opened 2020-09-21
job-archive interface: turn current subcommands into options

### LINKED ISSUE #46: job-archive interface: filtering subcommands should instead be options
The job-archive interface currently has four subcommands that interact with the database and fetch job records:

- **by-user**
- **by-jobid**
- **after-start-time**
- **before-end-time**

These subcommands each query the database to fetch job records, but they cannot be combined with one another. For example, if you wanted to filter jobs by a specific user _also_ before some certain time _t_, you would have to run two separate subcommands:

```
$ flux account -p path/to/db-file by-user fluxuser
...
$ flux account -p path/to/db-file before-end-time 1234
...
```

These should instead function like optional arguments, where combinations can be passed in to further refine queries:

```
$ flux account -p path/to/db-file view-job-records -u fluxuser --after-start-time 1234
```

cmoussa1:
**Problem**: As described in #46, the current way to interact with the job-archive interface is through a series of subcommands. These, however, limit the user to only being able to run one of these filters at a time.

---

This PR turns the existing subcommands into optional arguments under a new subcommand, `view-job-records`. The filter options are the same:

```
optional arguments:
  -u USERNAME, --user USERNAME
                        username
  -j JOBID, --jobid JOBID
                        jobid
  -a START TIME, --after-start-time START TIME
                        start time
  -b END TIME, --before-end-time END TIME
                        end time
```

except now they can be combined (e.g. I want to see all jobs run by a user _u_ before some timestamp _t_).

The current implementation grabs the optional args passed in and performs a series of `if-else` checks to determine which `SELECT` statement to use on the job-archive DB. If a combination of parameters are passed in that are not one of the combinations checked by the `if-else` branches, the `SELECT` statement should default to the first parameter that was passed in (e.g. I pass in a jobid _j_ but also some timestamp _t_ - this will just use the jobid _j_). 

With the new `view_job_records()` function in **accounting_cli_functions.py**, I went ahead and removed the individual job-archive interface subcommands and functions, and changed their unit tests to instead interact with the new function, `view_job_records()`.  

Fixes #46

### DIFF
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -168,76 +168,50 @@ def add_job_records(dataframe):
     return job_records
 
 
-def view_jobs_run_by_username(conn, username, output_file):
-    # look up userid with username
-    userid = get_uid(username)
-    # get the information pertaining to a user in the jobs DB
-    select_stmt = (
-        "SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs where userid=?"
-    )
-    dataframe = pd.read_sql_query(select_stmt, conn, params=(userid,))
-    # if the length of dataframe is 0, that means
-    # the user specified was not found in the table
-    if len(dataframe.index) == 0:
-        return "User not found in jobs table"
-    else:
-        job_records = add_job_records(dataframe)
-        if output_file is None:
-            print_job_records(job_records)
+def view_job_records(conn, output_file, **kwargs):
+    job_records = []
+
+    # find out which args were passed and place them in a dict
+    valid_params = ("user", "after_start_time", "before_end_time", "jobid")
+    params = {}
+    params_list = []
+
+    params = {
+        key: val for (key, val) in kwargs.items() if val != None and key in valid_params
+    }
+
+    select_stmt = "SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs "
+    where_stmt = ""
+
+    def append_to_where(where_stmt, conditional):
+        if where_stmt != "":
+            return "{} AND {} ".format(where_stmt, conditional)
         else:
-            write_records_to_file(job_records, output_file)
+            return "WHERE {}".format(conditional)
 
-    return job_records
+    # generate the SELECT statement based on the parameters passed in
+    if "user" in params:
+        params["user"] = get_uid(params["user"])
+        params_list.append(params["user"])
+        where_stmt = append_to_where(where_stmt, "userid=? ")
+    if "after_start_time" in params:
+        params_list.append(params["after_start_time"])
+        where_stmt = append_to_where(where_stmt, "t_run > ? ")
+    if "before_end_time" in params:
+        params_list.append(params["before_end_time"])
+        where_stmt = append_to_where(where_stmt, "t_inactive < ? ")
+    if "jobid" in params:
+        params_list.append(params["jobid"])
+        where_stmt = append_to_where(where_stmt, "id=? ")
 
+    select_stmt += where_stmt
 
-def view_jobs_with_jobid(conn, jobid, output_file):
-    # get the information pertaining to a job in the jobs DB
-    select_stmt = (
-        "SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs where id=?"
-    )
-    dataframe = pd.read_sql_query(select_stmt, conn, params=(jobid,))
+    dataframe = pd.read_sql_query(select_stmt, conn, params=((*tuple(params_list),)))
     # if the length of dataframe is 0, that means
-    # the jobid specified was not found in the table
+    # no job records were found in the jobs table,
+    # so just return an empty list
     if len(dataframe.index) == 0:
-        return "Job not found in jobs table"
-    else:
-        job_records = add_job_records(dataframe)
-        if output_file is None:
-            print_job_records(job_records)
-        else:
-            write_records_to_file(job_records, output_file)
-
-    return job_records
-
-
-def view_jobs_after_start_time(conn, time_after, output_file):
-    # get jobs that have completed after a certain time
-    select_stmt = (
-        "SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs WHERE t_run > ?"
-    )
-    dataframe = pd.read_sql_query(select_stmt, conn, params=(time_after,))
-    # if the length of dataframe is 0, that means
-    # the time specified resulted in no jobs found
-    if len(dataframe.index) == 0:
-        return "No jobs found after time specified"
-    else:
-        job_records = add_job_records(dataframe)
-        if output_file is None:
-            print_job_records(job_records)
-        else:
-            write_records_to_file(job_records, output_file)
-
-    return job_records
-
-
-def view_jobs_before_end_time(conn, time_before, output_file):
-    # get jobs that have completed before a certain time
-    select_stmt = "SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs WHERE t_inactive < ?"
-    dataframe = pd.read_sql_query(select_stmt, conn, params=(time_before,))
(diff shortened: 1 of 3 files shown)


cmoussa1:
I think feedback/review on the structure of the `if-else` branches (i.e. feedback on how the right `SELECT` statement is picked) would be especially helpful. I'm not sure if there might be another/easier way to do this. 

@SteVwonder do you think when you get the chance you could give this a look?

chu11:
> I think feedback/review on the structure of the if-else branches (i.e. feedback on how the right SELECT statement is picked) would be especially helpful. I'm not sure if there might be another/easier way to do this.

Looking through the select statements would logic like the following work?

```
str=""
if --user:
   append "--user=?"
if --after-start-time:
  if str != "":
    append " AND "
  append "t_run > ?"
if --before-end-time:
  if str != "":
    append " AND "
  append "t_inactive < ?"
if --jobid:
  if str != "":
    append " AND "
  append "job-id=?"
```

Some Python nice-ness can make this cleaner, but hopefully you get the logic of what I'm saying.

cmoussa1:
Thanks @chu11! I just pushed a change using your suggestion. What I like about it is that it keeps its current functionality, but I think it also makes it easier to add more filters later in the future, like `hostname`, `nnodes`, things like that.

accounting/accounting_cli_functions.py
```diff
@@ -168,76 +168,45 @@ def add_job_records(dataframe):
...
+    for key, val in vars(args).items():
+        if val != None and key != "path" and key != "output_file" and key != "func":
+            params[key] = val
+
+    select_stmt = ""
+    # generate the SELECT statement based on the parameters passed in
+    if "user" in params:
+        params["user"] = get_uid(params["user"])
+        params_list.append(params["user"])
+        select_stmt += "SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs WHERE userid=? "
```
chu11, line 185: is the user allowed to not specify any of the options?  so they can output everything?  Perhaps could do:

```
select_stmt = "SELECT ....".
where_stmt = ""
```

and at the very end

`select_stmt += where_stmt`

and if where_stmt is empty string, no biggie?

cmoussa1: Oh, I see what you mean. That's probably a good idea. I didn't include returning all results as an option at first, but that update should be pushed now:

```
$ flux account -p path/to/db-file view-job-records
   UserID  Username         JobID         T_Submit            T_Run       T_Inactive  Nodes                                                                            R
0    2201     user1  530059362304 1600808153.99083 1600808154.01275 1600808154.09784      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
1    2201     user1  518164316160 1600808153.28221 1600808153.30560 1600808153.37464      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
2    2201     user1  504692211712 1600808152.47932 1600808152.50098 1600808152.57389      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
3    2201     user1  492377735168 1600808151.74527 1600808151.76970 1600808151.84127      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
4    2001  fluxuser  291655122944 1600808139.78101 1600808139.80519 1600808139.88820      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
5    2001  fluxuser  281706233856 1600808139.18806 1600808139.21133 1600808139.28879      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
6    2001  fluxuser  269895073792 1600808138.48468 1600808138.50479 1600808138.57930      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
7    2001  fluxuser  258989883392 1600808137.83465 1600808137.85604 1600808137.93684      1  {"version":1,"execution":{"R_lite":[{"rank":"0","children":{"core":"0"}}]}}
```

accounting/accounting_cli_functions.py
```diff
@@ -168,76 +168,52 @@ def add_job_records(dataframe):
...
-    # get jobs that have completed before a certain time
-    select_stmt = "SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs WHERE t_inactive < ?"
-    dataframe = pd.read_sql_query(select_stmt, conn, params=(time_before,))
+    dataframe = pd.read_sql_query(select_stmt, conn, params=((*tuple(params_list),)))
     # if the length of dataframe is 0, that means
-    # the time specified resulted in no jobs found
+    # no job records were found in the jobs table
     if len(dataframe.index) == 0:
-        return "No jobs found before time specified"
+        return "Record(s) not found in jobs table"
```
chu11, line 216: although you should defer to any users you are talking to, my preference would be to have no output if no records match.  But that's me.

accounting/accounting_cli_functions.py
```diff
@@ -168,76 +168,52 @@ def add_job_records(dataframe):
...
-        job_records = add_job_records(dataframe)
-        if output_file is None:
-            print_job_records(job_records)
+def view_job_records(conn, output_file, args):
+    # find out which args were passed and place them in a dict
+    params = {}
+    params_list = []
+
+    for key, val in vars(args).items():
+        if val != None and key != "path" and key != "output_file" and key != "func":
```
chu11, line 177: if you add a new general option to the tool, you'll have to update this loop check, which people will likely forget to do and could lead to errors.  Instead, how about only matching options this function cares about.

```
params_i_care_about = ("user", "after_start_time", "before_start_time", "jobdid)
for key, val in vars(args).items(): 
    if val != None and key in params_i_care_about:
       ...
```

accounting/accounting_cli_functions.py
cmoussa1: I like this suggestion - thanks. No output would actually be more in line with SQLite's behavior, which I believe just returns an empty `Cursor` object:

```
sqlite> select * from jobs where user='bad_username'
sqlite> 
``` 

In my specific case, I could probably just return an empty `job_records` list.

cmoussa1:
Just pushed a fix per @chu11's suggestions - instead of returning a _Record(s) not found in jobs table_ string on an empty result, just return an empty `job_records` list, as well as adding valid parameters to a set which is checked instead of listing them one-by-one in a long `if` check.

chu11 (approved):
LGTM, approving.  Although you may want a skim from the more python centric folks.

test/test_job_archive_interface.py
```diff
@@ -108,64 +109,106 @@ def setUpClass(self):
     # passing a valid jobid should return
     # its job information
     def test_01_with_jobid_valid(self):
-        job_records = aclif.view_jobs_with_jobid(jobs_conn, 102, op)
+        args = Namespace(jobid=102)
+        job_records = aclif.view_job_records(jobs_conn, op, args)
```
SteVwonder, line 113: Rather than populating and passing a `Namespace` down into a function, I think the more re-usable method would be to have `view_job_records` have a `**kwargs` argument, and then you could call the function in a few different ways:
```
aclif.view_job_records(jobs_conn, op, **my_dict)
aclif.view_job_records(jobs_conn, op, user="102", after_start_time=5)
```

Which is nice from the testing perspective but also if anyone wants to call `view_job_records` directly in the future.

SteVwonder (commented):
LGTM! Couple of suggestions below.  Feel free to take them or leave them.

cmoussa1:
Thanks for giving this a pass @SteVwonder - I especially liked the suggestion about using a function inside a function - had no idea that was possible. I just pushed an update including your suggestions.

SteVwonder (approved):
LGTM! Thanks @cmoussa1!  Feel free to add MWP when you are ready.

Issue #38, opened 2020-08-31
create_db.py: add error handling for failed DB creation


cmoussa1:
As mentioned in the comments in #36, there is no error-handling the `create_db` function. Even if the database fails to be created, there will be no error message or return code. We'll need to add some output in case the database fails to get created.

dongahn:
> We'll need to add some output in case the database fails to get created.

Yes. And you would also need to be able to propagate any failure as the exit code from flux account create-db. Normally, exit with the return code of 1 with some reasonable logging of error messages will do.

cmoussa1:
I'll go ahead and add here that error propagation should be added for multiple subcommands (as noted in #34, some of the bank subcommands will need to undergo this change as well). This will be future work following this most recent milestone.

PR #49, opened 2020-09-30
edit error handling in create-db(), edit-user()

cmoussa1:
**Problem**: Noted in #38, there was no sort of error handling if `create-db()` failed for whatever reason while creating the database. There was also an instance in `edit-user()` where if somebody tried to edit a user's account information and specified a non-existent field, the program would call `sys.exit(1)`.

---

This PR wraps `create-db()` in a `try-except` block, and will print any SQLite operational error to the screen. It also changes `edit-user()` to raise an `Exception` instead of calling `sys.exit()` if a non-existent field is specified.

I also noticed that `accounting_cli_functions.py` had a couple of unused `import` statements, so I figured I would remove those as well in this PR (if anything, I can always open another tiny PR that removes it instead of including it in this one).

I believe now there is only one instance where `sys.exit()` is called - in `accounting_cli.py`:

```python
# try to open database file; will exit with -1 if database file not found
path = args.path if args.path else "FluxAccounting.db"
try:
    conn = sqlite3.connect("file:" + path + "?mode=rw", uri=True)
except sqlite3.OperationalError:
    sys.exit(-1)
```

I figured not being able to connect to the database file would warrant a `sys.exit()` since essentially none of the subcommands would work if you can't connect to the database file... is this the right intuition to have? Should an `Exception` instead be raised?

### DIFF
--- a/accounting/create_db.py
+++ b/accounting/create_db.py
@@ -20,10 +20,14 @@ LOGGER = logging.basicConfig(filename="db_creation.log", level=logging.INFO)
 
 
 def create_db(filepath):
-    # open connection to database
-    logging.info("Creating Flux Accounting DB")
-    conn = sqlite3.connect("file:" + filepath + "?mode:rwc", uri=True)
-    logging.info("Created Flux Accounting DB sucessfully")
+    try:
+        # open connection to database
+        logging.info("Creating Flux Accounting DB")
+        conn = sqlite3.connect("file:" + filepath + "?mode:rwc", uri=True)
+        logging.info("Created Flux Accounting DB successfully")
+    except sqlite3.OperationalError as e:
+        logging.error(e)
+        sys.exit(1)
 
     # Association Table
     logging.info("Creating association_table in DB...")
--- a/accounting/print_hierarchy.py
+++ b/accounting/print_hierarchy.py
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: LGPL-3.0
 ###############################################################
 import sqlite3
+import sys
 import pandas as pd
 
 # this will print the full hierarchy of banks
@@ -27,10 +28,14 @@ def print_full_hierarchy(conn):
         """
     dataframe = pd.read_sql_query(select_stmt, conn)
 
-    if len(dataframe) == 0:
-        raise Exception("No root bank found")
-    elif len(dataframe) > 1:
-        raise Exception("More than one root bank found")
+    try:
+        if len(dataframe) == 0:
+            raise IndexError("No root bank found")
+        elif len(dataframe) > 1:
+            raise IndexError("More than one root bank found")
+    except IndexError as e:
+        print(e)
+        sys.exit(1)
 
     root = dataframe.iloc[0]
 
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -179,7 +179,7 @@ def main():
         conn = sqlite3.connect("file:" + path + "?mode=rw", uri=True)
     except sqlite3.OperationalError:
         print("Unable to open database file")
-        sys.exit(-1)
+        sys.exit(1)
 
     # set path for output file
     output_file = args.output_file if args.output_file else None
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -167,16 +167,18 @@ def edit_user(conn, username, field, new_value):
         "max_jobs",
         "max_wall_pj",
     ]
-    if field in fields:
-        the_field = field
-    else:
-        print("Field not found in association table")
-        sys.exit(1)
+    try:
+        if field in fields:
+            the_field = field
 
-    # edit value in accounting database
-    conn.execute(
-        "UPDATE association_table SET " + the_field + "=? WHERE user_name=?",
-        (new_value, username,),
-    )
-    # commit changes
-    conn.commit()
+            # edit value in accounting database
+            conn.execute(
+                "UPDATE association_table SET " + the_field + "=? WHERE user_name=?",
+                (new_value, username,),
+            )
+            # commit changes
+            conn.commit()
+        else:
+            raise ValueError("Field not found in association table")
+    except Exception as e:
+        print(e)
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -10,9 +10,7 @@
 # SPDX-License-Identifier: LGPL-3.0
 ###############################################################
 import sqlite3
-import argparse
 import time
-import sys
 import pwd
 import csv
 
--- a/test/test_bank_subcommands.py
+++ b/test/test_bank_subcommands.py
@@ -165,12 +165,39 @@ A||1
 
         self.assertEqual(test, expected)
 
+    # having more than one root bank should result
+    # in an exception being thrown
+    def test_09_print_hierarchy_failure_1(self):
+        c.create_db("flux_accounting_failure_1.db")
+        acct_conn = sqlite3.connect("flux_accounting_failure_1.db")
+
+        aclif.add_bank(acct_conn, bank="A", shares=1)
+        aclif.add_bank(acct_conn, bank="B", shares=1)
+
+        with self.assertRaises(SystemExit) as cm:
+            p.print_full_hierarchy(acct_conn)
+
+        self.assertTrue(cm.exception.code, 1)
+
+    # having no root bank should also result in an
+    # exception being thrown
+    def test_10_print_hierarchy_failure_2(self):
+        c.create_db("flux_accounting_failure_2.db")
+        acct_conn = sqlite3.connect("flux_accounting_failure_2.db")
+
+        with self.assertRaises(SystemExit) as cm:
+            p.print_full_hierarchy(acct_conn)
+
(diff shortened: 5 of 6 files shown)


accounting/create_db.py
```diff
@@ -20,42 +20,45 @@
...
+                CREATE TABLE IF NOT EXISTS bank_table (
+                    bank_id     integer PRIMARY KEY,
+                    bank        text    NOT NULL,
+                    parent_bank text,
+                    shares      int     NOT NULL
+            );"""
+        )
+        logging.info("Created bank_table successfully")
+
+        conn.close()
```
chu11, line 62: if an exception happens before the `conn.close()` this won't be cleaned up properly.  perhaps should use a `finally` clause here?

accounting/create_db.py
```diff
@@ -20,42 +20,45 @@
...
+    except sqlite3.OperationalError as e:
+        print(e)
```
chu11, line 64: IMO you shouldn't `print(e)`.  Raise the exception and let the caller decide if they want to print or not.

chu11 (commented):
> I figured not being able to connect to the database file would warrant a sys.exit() since essentially none of the subcommands would work if you can't connect to the database file... is this the right intuition to have? Should an Exception instead be raised?

I think this is fine.

accounting/create_db.py
cmoussa1: Oh yeah, I forgot about that. Just pushed a fix that adds a `finally` clause that closes the SQLite connection.

cmoussa1:
Thanks @chu11! I just pushed an update with your suggestions. Thanks for providing your input on the use of `sys.exit()`.

chu11:
Do you have to try-catch the call to `create_db`?

```
    if args.func == "create_db":
        c.create_db(args.dbpath)
        sys.exit(0)

```

cmoussa1:
Hm. I'm not sure. 

Well the `create_db()` function that gets called is itself wrapped in a `try-except`. Does the call to `create_db()` also have to be wrapped in `try-except`?

chu11:
> Well, the create_db() function that gets called is itself wrapped in a try-except. Does the call to create_db() also have to be wrapped in try-except?

Before you `print(e)`'-ed in `create_db()` so I thought you'd want to capture and print out the `e` now.  But you don't necessarily have to.

cmoussa1:
Yeah, `edit-user()`  does the same thing, it just raises the exception, so I figured I might make it consistent with that. Do you think it would be better to instead print out `e`?

chu11:
I'm not sure what is preferred.  Perhaps @SteVwonder has an opinion.  I certainly have met my share of programs that capture exceptions and print something pretty, and others that just happily let the exception & stack trace be output to the user directly.  But unsure if there's a more common approach in the Python world.

SteVwonder:
> I'm not sure what is preferred. Perhaps @SteVwonder has an opinion. I certainly have met my share of programs that capture exceptions and print something pretty, and others that just happily let the exception & stack trace be output to the user directly.

Personally, I don't have a strong preference one way or the other.  Over in flux-core at least, I think we have tried to keep the default to be "pretty" and then provide env vars that turn up the verbosity.  With these tools being more sys admin facing, I could see an argument for being more verbose by default.

FWIW, if you decorate your top-level function (i.e., main) with the `CLIMain` decorator, then the exception handling gets handled automatically for you.  Any uncaught exceptions are "pretty printed" (no stack trace) by default, and the full stack trace is included if `FLUX_PYCLI_LOGLEVEL` is set.  The decorator also sets up the `logger` for you if you want to use that rather than printing directly.  Sorry if you are already aware of all of that and using it.

cmoussa1:
> Personally, I don't have a strong preference one way or the other. Over in flux-core at least, I think we have tried to keep the default to be "pretty" and then provide env vars that turn up the verbosity. With these tools being more sys admin facing, I could see an argument for being more verbose by default.

That makes sense. To me, it seems reasonable to pretty print an error in `edit_user()`'s case (I went ahead and changed it to raise a `ValueError` and print that to the screen). 

In `create_db()`'s case, I wonder if it's reasonable to use `sys.exit()` if the database file can't be created:

```python
try:
    # open connection to database
    logging.info("Creating Flux Accounting DB")
    conn = sqlite3.connect("file:" + filepath + "?mode:rwc", uri=True)
    logging.info("Created Flux Accounting DB successfully")
    print("Error creating database:", e)
    logging.info(e)
    sys.exit(-1)
```

If there is an error trying to create the database here (for example, if the user doesn't have permission to the directory they are trying to place the database file in), the error will be pretty printed to the screen and written to the log file, but the program will exit before it moves on and tries to create the two tables. I think this will also prevent the `UnboundLocalError` exception that would get raised when it tries to close the connection in the `finally` block. 

Thoughts on this? Would this be an appropriate use case of `sys.exit()`?

SteVwonder:
> In create_db()'s case, I wonder if it's reasonable to use sys.exit() if the database file can't be created:

I think the code snippet you have looks fine.  The print immediately followed by a log seems weird to me.  If you change `logging.info` to `logging.error`, does it print out to the console?  If so, I'd drop the print.  If you need to keep the print, I'd make sure to print to stderr `print("error message", file=sys.stderr)`.

cmoussa1:
> If you change logging.info to logging.error, does it print out to the console?

I don't believe it does, it just gets written to the `.log` file with level `ERROR`:

```
[fluxuser@8e373ae13a64 flux-ba]$ cat db_creation.log 
INFO:root:Creating Flux Accounting DB
ERROR:root:unable to open database file
```

But I don't necessarily need to print anything to the screen, so I can drop the `print()`.

cmoussa1:
Rebased to catch up after #55.

dongahn:
@cmoussa1 got reviews from two people already. Is this good to go? 

I would like to start my current fair-tree effort from a clean slate as much as possible.

cmoussa1:
I just rebased to catch up after #57; this should be good to go on my end. 👍

cmoussa1:
@dongahn pointed out that the output format of the `Exception`s that get raised from `print_hierarchy()` could be improved, so I attached it here (fc7319a) since it is similar to the other fixes attached in this PR.

accounting/create_db.py
```diff
@@ -20,10 +20,14 @@
...
-    conn = sqlite3.connect("file:" + filepath + "?mode:rwc", uri=True)
-    logging.info("Created Flux Accounting DB sucessfully")
+    try:
+        # open connection to database
+        logging.info("Creating Flux Accounting DB")
+        conn = sqlite3.connect("file:" + filepath + "?mode:rwc", uri=True)
+        logging.info("Created Flux Accounting DB successfully")
+    except sqlite3.OperationalError as e:
+        logging.error(e)
+        sys.exit(-1)
```
chu11, line 30: is there a reason for exiting -1 instead of 1?  is it a python thing?

accounting/print_hierarchy.py
```diff
@@ -28,9 +29,17 @@ def print_full_hierarchy(conn):
     dataframe = pd.read_sql_query(select_stmt, conn)
 
     if len(dataframe) == 0:
-        raise Exception("No root bank found")
+        try:
```
chu11, line 32: i'm a little confused by this design pattern.  why capture the exception here.  why not capture the exception when `print_full_hierarchy()` is called?

accounting/create_db.py
cmoussa1: Yeah, I also exited -1 here in **accounting_cli.py**:

```python
# try to open database file; will exit with -1 if database file not found
path = args.path if args.path else "FluxAccounting.db"
try:
    conn = sqlite3.connect("file:" + path + "?mode=rw", uri=True)
except sqlite3.OperationalError:
    sys.exit(-1)
```

accounting/print_hierarchy.py
cmoussa1: Are you thinking to instead capture the exception in **accounting_cli.py**?

```python
elif args.func == "print_hierarchy":
    print(ph.print_full_hierarchy(conn))
```

I guess my reasoning was I was capturing exceptions from inside the subcommands themselves, not the program (**accounting_cli.py**) that calls those functions, so I tried to make it consistent. 

Does it make more sense to capture the exception in **accounting_cli.py**?

accounting/create_db.py
chu11: i would say just do `sys.exit(1)` for errors.  I think it's the standard convention unless there's a special python convention I'm unaware of: https://tldp.org/LDP/abs/html/exitcodes.html

accounting/print_hierarchy.py
chu11: ok.  In that case, perhaps I'd suggest;

```
try:
  if ...:
     raise ...
  elif ...:
     raise ...
except:
   print(e)
   sys.exit(1)
```

so no need for the double try/exception.  I think that double try/except is what confused me the most.

cmoussa1: Ah, okay. Thanks. Sorry for the confusion. That change should be pushed up now.

cmoussa1:
pushed a small change that adds a couple of unit tests that interact with the `Exception`s that get thrown in `print_full_hierarchy()`.

chu11 (approved):
LGTM, set MWP when reading

PR #51, opened 2020-10-01
print-hierarchy: improve output format

cmoussa1:
**Problem**: The current output of `print-hierarchy()` is pretty clunky and hard to read if there are lots and lots of banks and sub banks, each which might hold lots of user accounts.

In reality, with lots of banks and accounts, we need output that is cleaner and has more of a directory tree layout. 

---

This PR moves `print_hierarchy()` to its own file, **print-hierarchy.py**, and changes its output to more closely resemble the `sshare` command. For example, take the following bank structure

```
      A
      |
   |-----|
   B     C
   |     |
 |---| |---|
 D   E F   G
```

With a call to `print-hierarchy()`, this should result in the following output:

```
Bank|User|RawShares
A||1
 B||1
  D||1
   D|user1|1
  E||1
 C||1
  F||1
   F|user2|1
   F|user3|1
  G||1
   G|user4|1
```

This PR is the first step to gathering bank/user data needed to perform usage calculations. Eventually, the output from this function would probably contain more columns like the `sshare` command, like `RawUsage` and `FairShare` values. But for now, this PR just fetches static information about banks and their associations from `flux-accounting`'s database file.

### DIFF
--- a/accounting/accounting_cli.py
+++ b/accounting/accounting_cli.py
@@ -18,6 +18,7 @@ import pandas as pd
 
 from accounting import accounting_cli_functions as aclif
 from accounting import create_db as c
+from accounting import print_hierarchy as ph
 
 
 def main():
@@ -231,7 +232,7 @@ def main():
         elif args.func == "edit_bank":
             aclif.edit_bank(conn, args.bank, args.shares)
         elif args.func == "print_hierarchy":
-            aclif.print_hierarchy(conn)
+            print(ph.print_full_hierarchy(conn))
         else:
             print(parser.print_usage())
     finally:
--- a/accounting/accounting_cli_functions.py
+++ b/accounting/accounting_cli_functions.py
@@ -325,40 +325,6 @@ def edit_bank(conn, bank, shares):
         print(e_database_error)
 
 
-def print_hierarchy(conn):
-    print("Bank = RawShares\n  | User = RawShares\n")
-    select_stmt = """
-        SELECT bank_table.bank,
-        bank_table.shares,
-        bank_table.parent_bank
-        FROM bank_table;
-        """
-    dataframe = pd.read_sql_query(select_stmt, conn)
-    for index, row in dataframe.iterrows():
-        print("-" * 50)
-        print(
-            "Bank:",
-            row["bank"],
-            "| Shares =",
-            row["shares"],
-            "| Parent Bank:",
-            row["parent_bank"],
-        )
-        print("-" * 50)
-        select_stmt = """
-            SELECT association_table.user_name,
-            association_table.shares AS user_shares
-            FROM association_table
-            INNER JOIN bank_table ON association_table.account = ?;
-            """
-        dataframe = pd.read_sql_query(select_stmt, conn, params=(row["bank"],))
-        users_seen = []
-        for index, row in dataframe.iterrows():
-            if row["user_name"] not in users_seen:
-                print("  |", row["user_name"], "| Shares =", row["user_shares"])
-                users_seen.append(row["user_name"])
-
-
 def view_user(conn, user):
     try:
         # get the information pertaining to a user in the Accounting DB
--- a/accounting/print_hierarchy.py
+++ b/accounting/print_hierarchy.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2020 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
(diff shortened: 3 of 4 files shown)


SteVwonder (commented):
Thanks @cmoussa1 for putting this together!  I think this is pretty close.  The one thing that I noticed is that the `print_full_hierarchy` function output changes depending on the insertion order into the database.  For example, if you apply this patch to the tests:

```diff
diff --git a/test/test_bank_subcommands.py b/test/test_bank_subcommands.py
index 109f7b5..04ae053 100644
--- a/test/test_bank_subcommands.py
+++ b/test/test_bank_subcommands.py
@@ -101,9 +101,9 @@ class TestAccountingCLI(unittest.TestCase):

         aclif.add_bank(acct_conn, bank="A", shares=1)
         aclif.add_bank(acct_conn, bank="B", parent_bank="A", shares=1)
+        aclif.add_bank(acct_conn, bank="C", parent_bank="A", shares=1)
         aclif.add_bank(acct_conn, bank="D", parent_bank="B", shares=1)
         aclif.add_bank(acct_conn, bank="E", parent_bank="B", shares=1)
-        aclif.add_bank(acct_conn, bank="C", parent_bank="A", shares=1)
         aclif.add_bank(acct_conn, bank="F", parent_bank="C", shares=1)
         aclif.add_bank(acct_conn, bank="G", parent_bank="C", shares=1)
```

The database is still semantically correct, and the insertions all work correctly. But the order of the rows within the database is different, so the output from the print subcommand is incorrect:

```
======================================================================
FAIL: test_08_print_hierarchy (test.test_bank_subcommands.TestAccountingCLI)
  File "/Users/herbein1/Repositories/flux-framework/flux-accounting/test/test_bank_subcommands.py", line 166, in test_08_print_hierarchy
    self.assertEqual(test, expected)
AssertionError: 'Bank[26 chars]|1\n C||1\n  D||1\n   D|user1|1\n  E||1\n  F||[49 chars]|1\n' != 'Bank[26 chars]|1\n  D||1\n   D|user1|1\n  E||1\n C||1\n  F||[49 chars]|1\n'
  Bank|User|RawShares
  A||1
   B||1
-  C||1
    D||1
     D|user1|1
    E||1
+  C||1
    F||1
     F|user2|1
     F|user3|1
    G||1
     G|user4|1
```

This is because the current `print_full_hierarchy` function does a single read from the database and then prints the banks in the order they are returned by the query.  This works fine when the database ordering is the same as a [pre-order traversal](https://en.wikipedia.org/wiki/Tree_traversal#Pre-order_(NLR)) of the bank hierarchy, but it breaks when the ordering is different (side-note: I always get the names for pre-order and in-order mixed up.....one of these days it'll actually stick 😆 ).

I think the only way around this is to perform a pre-order traversal of the bank hierarchy by issuing N select queries (1 per bank); starting with the hierarchy root and working your way down through the entire hierarchy.  If you go this route, one benefit will be that it will be relatively easy to keep track of your depth within the hierarchy, so the `how_many_parent_banks` function can be eliminated, saving you quite a few DB queries.  I think after a refactor, you will actually end up with fewer DB queries being issued overall.

cmoussa1:
Thanks for your feedback @SteVwonder and for the suggestion! (With your help) I pushed an update with a refactor to `print_full_hierarchy()`:

The function now traverses the bank hierarchy by starting at the root bank and traversing its sub banks until it reaches a "leaf bank" (one with no sub banks). Once it arrives there, it will print out all associations in that sub bank. This should resolve the problem of depending on the order banks are inserted into the `flux-accounting` DB. 

```
insert order: [A B D E C F G]

Bank|User|RawShares
A||1
 B||1
  D||1
   D|user1|1
  E||1
 C||1
  F||1
   F|user2|1
   F|user3|1
  G||1
   G|user4|1
```

```
insert order: [A B C D E F G]

Bank|User|RawShares
A||1
 B||1
  D||1
   D|user1|1
  E||1
 C||1
  F||1
   F|user2|1
   F|user3|1
  G||1
   G|user4|1
```

I'll go ahead and remove the **[WIP]** tag and post this as an official PR.

SteVwonder:
Thanks @cmoussa1! I'll leave this to another flux team member to review since we pair programmed on part of it.  Btw, if you want, you can use Github's [co-authorship](https://github.blog/2018-01-29-commit-together-with-co-authors/) feature to ascribe a single commit to multiple people.

cmoussa1:
> Btw, if you want, you can use Github's [co-authorship](https://github.blog/2018-01-29-commit-together-with-co-authors/) feature to ascribe a single commit to multiple people.

TIL! Thanks for showing me this. I believe I was able to successfully add you to the commit.

chu11 (approved):
LGTM!

cmoussa1:
Thanks @chu11! Am I good to set MWP?

chu11:
go for it!

PR #53, opened 2020-10-06
add-user, association_table: add default values for function, rename 'account' field to 'bank'

cmoussa1:
**Problems**

- If you want to use `add_user()` in a function call, you have to pass in every field when adding a new user. However, in #26, we added default argument values for the `add-user` subcommand so that we could just set a username and bank when creating a new user.

- The field where a 'bank' should be specified in `add_user()` is named 'account', and can be confusing. 

---

This is a small PR that adds those same default arg values from the subcommand into the function definition of `add-user()`. This required a small rework of some of the `add-user()` unit tests since the ordering of parameters had to be changed when defining the function:

```diff
- def add_user(conn, username, admin_level, account, shares, max_jobs, max_wall_pj):
+ def add_user(conn, username, account, admin_level=1, shares=1, max_jobs=1, max_wall_pj=60):
```

It then renames the 'account' field to 'bank' in the `association_table` in the flux-accounting database. The name change should be reflected throughout its use cases in flux-accounting (function calls, exception messages, and unit tests).


SteVwonder (approved):
LGTM! Thanks @cmoussa1!

PR #64, opened 2020-10-29
add print_hierarchy() C++ implementation to weighted tree lib

cmoussa1:
This experimental PR is my C++ implementation of the `print_hierarchy()` function, which will be used to display output from the flux-accounting database.

### current implementation

The program is compiled with the following:

```console
g++ print_hierarchy.cpp -l sqlite3 --std=c++11 -o print_hierarchy 
```

And is called with one command line argument, a path to the flux-accounting database file:

```console
./print_hierarchy path/to/FluxAccounting.db
```

In `main()`, the program attempts to open the database file in read-write mode. Once it opens the file, it will look for a root bank (one that sits on top of the hierarchical 'tree'). The program should exit if the following cases are met:

- the `bank_table` does not have a root bank (i.e. there is no bank in the `bank_table` that does not have a `parent_bank`)

- the `bank_table` has _more than one_ root bank (i.e there is more than one bank in the `bank_table` that does not have a `parent_bank`). 

After it finds a root bank, it calls `get_sub_banks()`, passing in the database handle and the name of the root bank. `get_sub_banks()` performs a depth-first-search of the `bank_table`, looking for sub banks. Upon coming across a bank with no sub banks (i.e. a 'leaf' bank), it will append all user accounts and their respective shares under that leaf bank to a global string, `hierarchy`. After completing a DFS traversal of the `bank_table`, it will print out the `hierarchy` string and close its connection to the DB.

```
Bank|User|RawShares
A||1|
 B||1|
  D||1|
   D|user1|1
  E||1|
 C||1|
  F||1|
   F|user2|1
   F|user3|1
  G||1|
   G|user4|1
```

### TODO

This program needs tests to be added. As of now, I've just been testing manually by passing in four different types of `bank_table`s from the flux-accounting database:

1. a standard 'successful' `bank_table`, like the one shown above

2. a `bank_table` with no root bank

3. a `bank_table` with more than one root bank

4. a successful `bank_table` where the banks are ordered in a different order

We can probably add sharness tests to confirm the correct output of running this program to automate this process. 

---

I'm not sure if we'll want to make this program a library instead with just functions that get called instead of one with a `main()` function. I'll leave that to @dongahn to give his feedback and suggestions on what he might need.

All of the SQLite database interactions in this program used the `sqlite3` C++ interface, which was actually pretty intuitive. The big difference I noticed between this and the Python interface was that the C++ version (to my knowledge) requires manually "stepping through" each result of a SQLite query, whereas Python can store its result in one `DataFrame` object, which can be indexed and iterated through.  

---

C++ styling: I'm not sure if there is a certain styling guide we use for C++, so I am sorry if there are formatting or other code conventions that I missed!

### DIFF
--- a/src/fairness/print_hierarchy/print_hierarchy.cpp
+++ b/src/fairness/print_hierarchy/print_hierarchy.cpp
@@ -1,0 +1,202 @@
+/*
+Copyright 2020 Lawrence Livermore National Security, LLC
+(c.f. AUTHORS, NOTICE.LLNS, COPYING)
+
+This file is part of the Flux resource manager framework.
+For details, see https://github.com/flux-framework.
+
+SPDX-License-Identifier: LGPL-3.0
+*/
+#include <iostream>
+#include <sqlite3.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <vector>
+#include <tuple>
+
+std::string hierarchy = "Bank|User|RawShares\n";
+char* messageError;
+
+sqlite3_stmt *b_select_root_bank_stmt;
+sqlite3_stmt *b_select_shares_stmt;
+sqlite3_stmt *b_select_sub_banks_stmt;
+sqlite3_stmt *b_select_associations_stmt;
+
+int get_sub_banks(sqlite3* DB, const std::string& bank_name, const std::string& indent) {
+  int exit = 0;
+
+  // bind parameter to prepared SQL statement
+  exit = sqlite3_bind_text(b_select_shares_stmt, 1, bank_name.c_str(), -1, NULL);
+  if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
+  }
+  // execute SQL statement and store result
+  exit = sqlite3_step(b_select_shares_stmt);
+  std::string bank_shares = reinterpret_cast<char const*> (sqlite3_column_text(b_select_shares_stmt, 0));
+
+  hierarchy.append(indent + bank_name + "||" + bank_shares + "\n");
+
+  exit = sqlite3_bind_text(b_select_sub_banks_stmt, 1, bank_name.c_str(), -1, NULL);
+  if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
+  }
+  exit = sqlite3_step(b_select_sub_banks_stmt);
+
+  // vector of tuples to store usernames and their respective shares
+  std::vector<std::tuple<std::string, std::string>> users;
+
+  // vector of strings to hold sub banks
+  std::vector<std::string> banks;
+
+  // we've reached a bank with no sub banks, so print
+  // out all associations under this sub bank
+  if (exit != SQLITE_ROW) {
+    exit = sqlite3_bind_text(b_select_associations_stmt, 1, bank_name.c_str(), -1, NULL);
+    if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
+    }
+
+    // execute SQL statement
+    exit = sqlite3_step(b_select_associations_stmt);
+    while (exit == SQLITE_ROW) {
+      std::string user = reinterpret_cast<char const*> (sqlite3_column_text(b_select_associations_stmt, 0));
+      std::string single_user_shares = reinterpret_cast<char const*> (sqlite3_column_text(b_select_associations_stmt, 1));
+      // place the user and their allocated shares into vector
+      users.emplace_back(user, single_user_shares);
+      exit = sqlite3_step(b_select_associations_stmt);
+    }
+    // iterate through associations and append them to the hierarchy string
+    for (unsigned int i=0; i < users.size(); i++) {
+      std::string username = std::get<0>(users[i]);
+      std::string user_shares = std::get<1>(users[i]);
+      hierarchy.append(indent + " " + bank_name + "|" + username + "|" + user_shares + "\n");
+    }
+  }
+  // otherwise, this bank has sub banks, so call this helper
+  // function again with the first sub bank it found
+  else {
+    while (exit == SQLITE_ROW) {
+      // execute SQL statement
+      std::string bank = reinterpret_cast<char const*> (sqlite3_column_text(b_select_sub_banks_stmt, 0));
+      banks.push_back(bank);
+      exit = sqlite3_step(b_select_sub_banks_stmt);
+    }
+    for (std::string b : banks) {
+      // reset the prepared statements back to their initial state and
+      // clear their bindings
+      sqlite3_clear_bindings(b_select_associations_stmt);
+      sqlite3_reset(b_select_associations_stmt);
+      sqlite3_clear_bindings(b_select_sub_banks_stmt);
+      sqlite3_reset(b_select_sub_banks_stmt);
+      sqlite3_clear_bindings(b_select_shares_stmt);
+      sqlite3_reset(b_select_shares_stmt);
+      get_sub_banks(DB, b, indent + " ");
+    }
+  }
+
+  return 0;
+}
+
+
+/*
+print_hierarchy.cpp takes one argument, the path to a flux-accounting DB file.
+
+It will look for a parent bank in the bank table. It will exit if it does not
+find a root bank, or if it finds more than one root bank. Once it finds a root
+bank, it will call get_sub_banks, which is a recursive function that traverses
+the bank table with a depth-first search. Once it finds a bank with no sub
+banks, it will print any associations and their corresponding shares under that
+bank.
+*/
+int main(int argc, char** argv) {
+  std::string indent = "";
+
+  // only one argument should be passed in; a filepath to the flux-accounting DB
+  if (argc != 2) {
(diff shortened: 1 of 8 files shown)


dongahn:
Great. Let's have a deeper dive into this PR  and my WIP at 3PM to see how two things can be combined.

libweighted_tree/print_hierarchy.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+int get_sub_banks(sqlite3* DB, const std::string& bank_name, const std::string& indent) {
+  int exit = 0;
+
+  // get shares of the current bank
+  std::string select_shares_stmt = "SELECT bank_table.shares "
+          "FROM bank_table "
+          "WHERE bank='";
+  select_shares_stmt.append(bank_name + "'");
+  exit = sqlite3_prepare_v2(DB, select_shares_stmt.c_str(), -1, &res, 0);
```
SteVwonder, line 31: I'm not too familiar with the sqlite3 API, but it looks like from the documentation that you can actually `prepare`/compile the statement with `?`s (just like with pandas) and then use the `bind` function to apply specific parameters to the prepared statement.  Prepared Statement documentation w/ the lifecycle: https://sqlite.org/c3ref/stmt.html

I think this has two implications for the code:

1. You will only need to prepare the statement one time, but it is bound and reset multiple times.
2. You will need to replace the statment append with a call to a `sqlite3_bind_*()` function

For example:
```C++
  // put this in the main function
  std::string select_shares_stmt = "SELECT bank_table.shares "
          "FROM bank_table "
          "WHERE bank='?'";
  sqlite3_stmt select_bank_stmt;
  exit = sqlite3_prepare_v2(DB, select_shares_stmt.c_str(), -1, &select_bank_stmt, 0);
  
  // keep this in the get_sub_banks function
  exit = sqlite3_bind_text(&select_bank_stmt, 0, bank_name.c_str(), -1, NULL);
  exit = sqlite3_step(select_bank_stmt);
  std::string bank_shares = ... sqlite3_column_text(...);
  sqlite3_reset(&select_shares_stmt);
```

cmoussa1: Thanks @SteVwonder! This was super useful feedback. I believe I was able to get this working, and I pushed those changes up. I noticed a couple of tiny subtleties but I was able to get the same results as before:

- I think parameters are automatically quoted and escaped, so we don't actually need to surround the parameter with quotes. I referred to this GitHub thread https://github.com/mapbox/node-sqlite3/issues/999:

```c
std::string select_shares_stmt = "SELECT bank_table.shares FROM bank_table WHERE bank=?";
```

- The second parameter to `sqlite3_bind_text()` is the index of the SQL parameter to be set, and I guess the leftmost index is 1 instead of 0. Reference: https://sqlite.org/c3ref/bind_blob.html

Let me know what you think of these changes. 

I think ultimately this PR should wait until @dongahn's weighted tree-based fairness PR gets landed to merge this one, so I'll keep it as a draft for now. 🙂

libweighted_tree/print_hierarchy.cpp
```diff
@@ -0,0 +1,189 @@
...
+
+int get_sub_banks(sqlite3* DB, const std::string& bank_name, const std::string& indent) {
+  int exit = 0;
+
+  // get shares of the current bank
+  std::string select_shares_stmt = "SELECT bank_table.shares "
+          "FROM bank_table "
+          "WHERE bank=?";
+  // compile SQL statement with bank name
+  exit = sqlite3_prepare_v2(DB, select_shares_stmt.c_str(), -1, &select_stmt, 0);
```
SteVwonder, line 32: Looking much better!  One additional change that could make things better:

One of the major benefits of using the `sqlite3_prepare_v2` function is that you only need to prepare a statement 1 time for the program's entire execution.  Right now, this statement gets prepared for every call of `get_sub_bank` (which is called many times recursively).  I would move the prepare statements out of this function and into `main`.  That will require a unique `sqlite3_stmt *` for each statement that is prepared.  The `sqlite3_bind_*` functions will be the ones that stay within this recursive function.

Also, for every corresponding call to `sqlite3_prepare_v2` there should be a call to the destructor [`sqlite3_finalize`](https://sqlite.org/c3ref/finalize.html).  Similar to how malloc/free and new/delete should be balanced.

cmoussa1: Ah, I see what you are saying. I think that makes sense.

From trying to convert it this morning, I found that I was able to `prepare` the first SQL statement (the one that gets the shares of the bank passed in), but not the other two (the ones that get all potential sub banks and associations). 

My guess is this has something to do with the value returned when `sqlite3_step()` returns something other than `SQLITE_OK` (according to their documentation, the resulting value returned is undefined), which results in a `Segmentation fault`. I will try and work with it some more to see if I can get it working.

cmoussa1: Well, I guess after enough `cout` statements, I have found the culprit. 😆

Before each recursive call, I think the SQL statements must have their bindings cleared and reset so that the new bindings can be placed in:

```c++
sqlite3_clear_bindings(stmt);
sqlite3_reset(stmt);
```

Now I can `prepare` each SQL statement just once in `main()`, and `bind` the parameters when `get_sub_banks()` gets called. I will squash and push those changes up now.

cmoussa1:
Just pushed a change to remove an unused `sqlite3_stmt*` I accidentally left in **print_hierarchy.cpp**.

cmoussa1:
Just rebased to catch up after #65. For now, I just created a separate directory under `src/fairness/` called `print_hierarchy/` that stores my C++ program. But, I assume we will have some sort of discussion today to talk about what needs to be changed so that my changes can be integrated with @dongahn's most recent PR. 🙂

src/fairness/print_hierarchy/print_hierarchy.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include <iostream>
+#include <sqlite3.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <vector>
+#include <tuple>
+
+std::string hierarchy = "Bank|User|RawShares\n";
+std::string indent = "";
```
SteVwonder, line 19: I think this should be moved into `main` since it is passed down into `get_sub_banks` as an argument.  I don't think there is a need currently for it to be a global.

src/fairness/print_hierarchy/print_hierarchy.cpp
```diff
@@ -0,0 +1,201 @@
...
+char* messageError;
+
+sqlite3_stmt *b_select_root_bank_stmt;
+sqlite3_stmt *b_select_shares_stmt;
+sqlite3_stmt *b_select_sub_banks_stmt;
+sqlite3_stmt *b_select_associations_stmt;
```
SteVwonder, line 25: For a self-contained file, I think having these as globals is fine. If we want to expose `get_sub_banks` as a function that other parts of the codebase can call, then I think these should be arguments that are declared in `main` (or a public-facing API function, see note below on `main`) and then passed down into `get_sub_banks`.

src/fairness/print_hierarchy/print_hierarchy.cpp
```diff
@@ -0,0 +1,201 @@
...
+print_hierarchy.cpp takes one argument, the path to a flux-accounting DB file.
+
+It will look for a parent bank in the bank table. It will exit if it does not
+find a root bank, or if it finds more than one root bank. Once it finds a root
+bank, it will call get_sub_banks, which is a recursive function that traverses
+the bank table with a depth-first search. Once it finds a bank with no sub
+banks, it will print any associations and their corresponding shares under that
+bank.
+*/
+int main(int argc, char** argv) {
```
SteVwonder, line 122: Depending on how we plan on using this code, I think it might make sense to have `main` just do argument parsing, and then put the sqlite opening, prepare, finalize, etc code into a separate function that is public-facing and easily called by other parts of the code base.

As a strawman, the function declarations in this file could be something like:
```
int main(int argc, char** argv);
static int get_sub_banks(sqlite3* DB, const std::string& bank_name, const std::string& indent, sqlite3_stmt *b_select_shares_stmt, sqlite3_stmt *b_select_sub_banks_stmt, sqlite3_stmt *b_select_associations_stmt);
int print_banks(char *db_path);
```

Where `print_banks` would be the public-facing API and `get_sub_banks`/`main` are "internal" functions.

That being said, I think our plan is to expose a public API that populates @dongahn's fair tree data structure, and then the printing is just a walker on that populated tree.  So maybe the public API should be something like `weighted_tree_node_t* populate_banks(char *db_path)`.  Something to discuss in more detail in our 3pm call.

SteVwonder (commented):
This looks really great @cmoussa1!  The only thing I see left are the global variables.  The other comment has to do with the structuring of the "public API", but we should discuss that in more detail with @dongahn at the 3pm call today.

SteVwonder:
Looks like the `print_hierarchy` binary was committed in the most recent rebase.  You'll want to drop that out on your next pass.

cmoussa1:
@SteVwonder: thanks again for your help with automake yesterday! I just pushed those changes; let me know what you think. I hope I didn't miss anything. I basically confirmed this by `rsync`-ing my directory and going through the build stages - `./autogen.sh`, `./configure`, `make -j`, `make check` - everything seems to be working okay I think.

cmoussa1:
Just rebased to catch up after #49. I'll go ahead and drop the [WIP] tag and experimental label as well.

SteVwonder (approved):
LGTM! Thanks @cmoussa1 for putting this together! We should handle the global variables in the refactor that takes place after this release.

SteVwonder:
Feel free to set MWP when you are ready.

cmoussa1:
Thank you @SteVwonder! I will make an issue on the global variables right now for the next release.

PR #72, opened 2020-11-11
print_hierarchy.cpp: move global variables into main(), add another test

### LINKED ISSUE #71: print_hierarchy.cpp: move variables from 'global' scope
The following variables are defined in a `global` scope:

```
sqlite3_stmt *b_select_root_bank_stmt;
sqlite3_stmt *b_select_shares_stmt;
sqlite3_stmt *b_select_sub_banks_stmt;
sqlite3_stmt *b_select_associations_stmt;
```

As @SteVwonder noted in #64, if we want to expose `get_sub_banks()` as a function that other parts of the codebase can call, then we should instead move these definitions to inside `main()` and passed down to `get_sub_banks()`.

cmoussa1:
**Problem**: As noted in #71, a number of the `sqlite3_stmt` variables where defined in global scope. If we wanted to call `get_sub_banks()` from other parts of the codebase, we should instead move those variables inside `main()` and pass them into `get_sub_banks()`. 

---

This small PR does just that; moves the `sqlite3_stmt` variables into `main()` and reformats the function header and calls for `get_sub_banks()`. 

I also added another test to **test-print-hierarchy.sh**: when a flux-accounting database is created with no root bank, the program should detect that and exit out. 

Fixes #71

### DIFF
--- a/src/fairness/print_hierarchy/print_hierarchy.cpp
+++ b/src/fairness/print_hierarchy/print_hierarchy.cpp
@@ -16,14 +16,15 @@ SPDX-License-Identifier: LGPL-3.0
 #include <tuple>
 
 std::string hierarchy = "Bank|User|RawShares\n";
-char* messageError;
 
-sqlite3_stmt *b_select_root_bank_stmt;
-sqlite3_stmt *b_select_shares_stmt;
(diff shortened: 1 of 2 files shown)


SteVwonder (approved):
LGTM! Thanks @cmoussa1!

Issue #59, opened 2020-10-21
add docker-run-checks.sh


cmoussa1:
This repo will need a `docker-run-checks.sh` file, similar to the ones in flux-core and flux-sched to build the flux-accounting docker image and run tests (so, in a full directory layout, probably under `src/test/docker`). 

They will be part of the strategy for CI testing of the flux-accounting project. This will help speed up deployment of an environment with correct build dependencies and to keep a docker image with the latest master build.

PR #73, opened 2020-11-12
build: convert python code and tests over to autotools, TAP, and sharness

### LINKED ISSUE #52: build system: convert to autotools
In a meeting today, we discussed converting over from setuptools to autotools.  Benefits:

- More consistent across the various flux-framework repos
- Allows C++ within this repo
- Flux-core already has the automake/TAP magic to get python unittests working with the build system and `make check`
- Flux-sched already has the automake/configure magic to easily install a project within the same prefix as flux-core

Plan of attack:
- @dongahn will do an initial copy over from flux-sched to support C++
- @SteVwonder will do a copy over from flux-core of the python/TAP magic

SteVwonder:
Make flux-accounting mirror flux-core and flux-sched with its build system and file structure.  Drop tox and setuptools in favor of autotools, TAP, and sharness.  All of the tests are driven my autotools and our custom TAP drivers (one for shell tests and another for python tests).  The python unit tests also leverage the vendored pycotap module.  There is some code copied from flux-sched to make sure flux-accounting uses the same python that flux-core was configured against (if the install prefixes match).  Finally, there are some fixes to make sure everything works under `make distcheck`.

If others are OK with it, I'd like to leave hooking everything up to the CI for another PR.  This PR already rearranges and touches a lot of files, so I want to get it in sooner rather than later to minimize conflicts down the road.

I'm also going to open an issue about how we want to install the flux accounting python modules down the road.  Currently they are installed in their own module called `accounting`, rather than in the `flux` python module.

Closes #52


configure.ac
```diff
@@ -45,6 +45,19 @@ fi
...
+
+AM_PATH_PYTHON([$ac_python_version])
+if test "X$PYTHON" = "X"; then
+  AC_MSG_ERROR([could not find python])
+fi
+
+AC_SUBST(PYTHON_LIBRARY, lib${ac_python_library}.so)
+AC_DEFINE_UNQUOTED([PYTHON_INTERPRETER], ["$PYTHON"], [The python interpreter flux is configured with])
+AC_SUBST(PYTHON)
+
```
dongahn, line 60: Can we add a Panda dependency check here or around here?

dongahn:
From our 3PM flux-accounting meeting, we decided to move the accounting module under `flux` subdirectory (exactly same as flux-core) so that we can do `make check` for in tree and out tree with ease.

dongahn:
Not as part of this PR, but we will add `src/test/docker` support to add more formal docker support which then will allow us to do travis CI testing with more rigor. @dongahn can take a look at it. (@cmoussa1: I will appreciate if you can create an issue ticket for this.)

@grondo: also it will be great if we can pick your brain a little as to how fluxorama is used for flux-core CI tests if you indeed do that. This would be useful for flux-accounting to test access control if nothing else.

cmoussa1:
@dongahn: I created an issue that I think encapsulates what you mentioned in #59; let me know if this is what you're looking for!

cmoussa1:
Another thing we discussed yesterday during our flux-accounting meeting: we will need to update our docs for flux-accounting to reflect the installation instructions that follow as a result of this PR. I will open up an issue ticket on this now to track.

dongahn:
> @dongahn: I created an issue that I think encapsulates what you mentioned in #59; let me know if this is what you're looking for!

Yup. this sounds good. Thanks.

SteVwonder:
@dongahn and @cmoussa1 : this PR is ready for another review.  Since the last review, I have:

- added a check for pandas to `configure`
- made sure that `flux-account.py` is installed with the other flux commands so that you can run it with `flux account`
- restructured the python module so that `accounting` is now installed as a subpackage of `flux`. You import it like `import flux.accounting`.  This required some moving around within the source tree so that running with the in-tree version looks the same as running with the installed version.
- created a `__init__.py.in` that gets populated with the "compiled-in" default location for the Flux accounting DB; then tweaked `flux-account.py` to use that default unless overridden by the user at the command line
- a few minor tweaks and fixups to flux-account.py, a unittest, and some of my previous commits (the latter to ensure that the python flux-core was configured against is used to run the tests, not just whatever `python3` is first in your `PATH`)

cmoussa1:
Thanks for your work on this @SteVwonder - this LGTM!

cmoussa1 (approved):

SteVwonder:
Thanks @cmoussa1!

PR #77, opened 2020-11-16
README.md: update build/install instructions

### LINKED ISSUE #75: docs: update build, install instructions for flux-accounting
After #73 gets merged, we'll want to update our build and install instruction docs for flux-accounting to keep them up to date since they will be changed. This will include updating the top-level **README.md**.

cmoussa1:
This PR updates **README.md** to reflect the install changes after #73 gets merged. It removes the `tox` and `python3-pip` package requirements from the 'Requirements' section, and updates the example output of the `help` and  `view-user` subcommands.

Fixes #75


README.md
```diff
@@ -9,68 +9,51 @@ Development for a bank/accounting interface for the Flux resource manager. Write
...
+flux-accounting requires the following Python packages to build:
 
-| centos8       | ubuntu      | version |
-| ------        | --------    | ------- |
-| python3-devel | python3-dev | >= 3.6  |
-| python3-pip   | python3-pip | 20.0.2  |
-| tox           | tox         | 3.15.0  |
+| Package Name | Version |
+| ------------ | ------- |
+| pandas       | 0.25.3  |
```
SteVwonder, line 16: It turns out that python/3.7.2 on LC systems only have pandas version 0.24.1. So I think we should probably loosen this up a bit to >= 0.24.1 (I also need to update the requirements.txt in my PR...whoops).

(Sorry that the other comment wasn't a part of a formal review).

README.md
```diff
@@ -9,68 +9,51 @@ Development for a bank/accounting interface for the Flux resource manager. Write
...
-
-2. Pull down and install the flux-accounting repo:
-
-```
-$ git clone https://github.com/flux-framework/flux-accounting
-$ cd flux-accounting/
-$ make install
+```console
+./autogen.sh
+PYTHON_VERSION=3.7 ./configure
```
SteVwonder, line 24: Maybe we should make a note of the `PYTHON_VERSION` variable below this code block, but in the canonical example not include it since it isn't required?

README.md
```diff
@@ -108,7 +91,7 @@ Successfully built f29dd2ca5958
```
SteVwonder, line 91: Nit: for the docker output, I think that will have changed now that we no longer use tox.

README.md
```diff
@@ -136,38 +119,18 @@ This will output queries like the following:
...
-                        sub-command help
-    view-user           view a user's information in the accounting database
-    add-user            add a user to the accounting database
-    delete-user         remove a user from the accounting database
-    edit-user           edit a user's value
-
-optional arguments:
-  -h, --help            show this help message and exit
-```
+The second way is to use flux-accounting's command line arguments. You can pass in a path to the database file, or be in the same directory where the database file (**FluxAccounting.db**) is located.
```
SteVwonder, line 127: The "be in the same directory" will not be true after #73 is merged.  If you don't provide an explicit path with `-p`, it will default to the "compiled-in" path of `${prefix}/var/FluxAccounting.db`.

SteVwonder (changes requested):
Thanks @cmoussa1! A couple of small suggested tweaks based on some changes in #73

SteVwonder:
Per the call today, can you also make a note that flux-accounting needs to be configured against the same version of python as flux-core that it is configured against (and that this is the default behavior of configure if you choose the same prefix for flux-core and flux-accounting)?

cmoussa1:
Thanks for the feedback @SteVwonder! I just pushed up some changes per your suggestions.

Thanks for pointing out the use of `tox` in the Dockerfile console output. I ended up adding a commit that removes the `tox` installation and use in both of the Dockerfiles. I don't mind splitting that off into a different PR if that is more appropriate.

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

SteVwonder:
@cmoussa1: it looks like now that #73 is merged  there is a conflict with this PR.  Can you rebase and resolve any conflicts?  This should get merged automatically after that.

cmoussa1:
Thanks @SteVwonder! Looks like there was a conflict with the location of the `Dockerfile`s. Think I was able to get it resolved. 👍

PR #78, opened 2020-11-16
delete_bank(): make function recursive

### LINKED ISSUE #41: bank_table: delete-bank should be recursive for parent banks
In `delete-bank`, my current implementation fails on the following case:

```
      A
      |
   |-----|
   B     C
   |     |
 |---| |---|
 D   E F   G
```

If `A` is passed into `delete_bank`, then it will be deleted by the first delete execute.  The first `select_stmt` will return `B, C`.  `B` will be deleted and `last_parent_bank_seen` set to `B`.  `C` will be deleted and `last_parent_bank_seen` set to `C`.  The while loop will go to the next iteration, and the select statement will return `F` and `G` (which will then be deleted).   `D` and `E` will still be left in the database.

We should probably make `delete-bank` recursive so that we can cover a case like this. This should also include adding a deep hierarchy example in my tests that confirms it works.

cmoussa1:
_note: this PR should probably be reviewed after #73 gets approved and landed since that is a higher priority at the moment, so I'll leave it as a Draft PR for now._ 

**Problem**: As noted in #41, the `delete-bank` subcommand only deletes the bank that is passed in. If that bank happens to be a parent bank, then it will still leave its sub banks as "parentless" banks after it gets deleted. 

---

This PR turns `delete_bank()` into a recursive function where all sub banks (if any) also get deleted. It contains a similar `get_sub_banks()` helper function like in **print_hierarchy.py**. I also added in another test to **test_bank_subcommands.py** to ensure that this works, where I add a bank hierarchy tree like the following:

```
      A
      |
   |-----|
   B     C
   |     |
 |---| |---|
 D   E F   G
```

and delete bank "A", which should result in an empty hierarchy after `delete_bank(A)` is called. 

Fixes #41

### DIFF
--- a/src/bindings/python/flux/accounting/accounting_cli_functions.py
+++ b/src/bindings/python/flux/accounting/accounting_cli_functions.py
@@ -66,15 +66,54 @@ def view_bank(conn, bank):
 
 
 def delete_bank(conn, bank):
-    last_parent_bank_seen = bank
+    try:
+        # delete the first bank passed in
+        delete_stmt = """
+            DELETE FROM bank_table
+            WHERE bank=?
+            """
+        cursor = conn.cursor()
+        cursor.execute(delete_stmt, (bank,))
 
-    # delete bank from bank_table
-    delete_stmt = "DELETE FROM bank_table WHERE bank=?"
-    cursor = conn.cursor()
-    cursor.execute(delete_stmt, (bank,))
+        # construct a DataFrame object out of the
+        # bank passed in
+        df = pd.DataFrame([bank], columns=["bank"])
+        bank = df.iloc[0]
 
-    # commit changes
-    conn.commit()
+        # helper function to traverse the bank table
+        # and delete all of its sub banks
+        def get_sub_banks(row):
+            select_stmt = """
+                SELECT bank
+                FROM bank_table
+                WHERE parent_bank=?
+                """
+            cursor = conn.cursor()
+            dataframe = pd.read_sql_query(select_stmt, conn, params=(row["bank"],))
+
+            # we've reached a bank with no sub banks
+            if len(dataframe) == 0:
+                select_associations_stmt = """
+                    SELECT user_name, bank
+                    FROM association_table
+                    WHERE bank=?
+                    """
+                for row in cursor.execute(select_associations_stmt, (row["bank"],)):
+                    delete_user(conn, user=row[0], bank=row[1])
+            # else, delete all of its sub banks and continue
+            # traversing
+            else:
+                for index, sub_bank_row in dataframe.iterrows():
+                    cursor.execute(delete_stmt, (sub_bank_row["bank"],))
+                    get_sub_banks(sub_bank_row)
+
+        get_sub_banks(bank)
+    # if an exception occcurs while recursively deleting
+    # the parent banks, then throw the exception and roll
+    # back the changes made to the DB
+    except sqlite3.OperationalError as e:
+        print(e)
+        conn.rollback()
 
 
 def edit_bank(conn, bank, shares):
@@ -147,13 +186,11 @@ def add_user(conn, username, bank, admin_level=1, shares=1, max_jobs=1, max_wall
         print(integrity_error)
 
 
-def delete_user(conn, user):
+def delete_user(conn, user, bank):
     # delete user account from association_table
-    delete_stmt = "DELETE FROM association_table WHERE user_name=?"
+    delete_stmt = "DELETE FROM association_table WHERE user_name=? AND bank=?"
     cursor = conn.cursor()
-    cursor.execute(delete_stmt, (user,))
-    # commit changes
-    conn.commit()
+    cursor.execute(delete_stmt, (user, bank,))
 
 
 def edit_user(conn, username, field, new_value):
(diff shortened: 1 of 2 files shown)


SteVwonder (commented):
Generally LGTM!

Two high-level questions:

1. Should this also delete the banks from the association table?
2. Can `cursor.execute` or any of the other DB functions throw an exception? If so, this function could leave the DB in a "broken" state, where some banks are orphaned.  This can be implemented in a different PR, but if it is the case, we should probably open an issue.

EDIT: It looks like you can tell SQLite to not allow transactions that would delete a foreign key without also deleting the child row: the " ON DELETE and ON UPDATE Actions" section of https://sqlite.org/foreignkeys.html

cmoussa1:
> 1. Should this also delete the banks from the association table?

This is a great question. As the current implementation stands, only banks (and any sub banks) will get deleted with the `delete_bank()` function. That leaves the possibility of leaving "orphan" associations if their bank gets deleted, though, so I could definitely see the argument for also deleting associations that belong to the bank that is being deleted. Should we adopt this behavior?

2. Can cursor.execute or any of the other DB functions throw an exception? If so, this function could leave the DB in a "broken" state, where some banks are orphaned. This can be implemented in a different PR, but if it is the case, we should probably open an issue.

Hm, I think I might not understand the case fully here, sorry about that. Are you asking if `cursor.execute()` throws an Exception if it tries to delete a bank or an association that does not exist? If so, I think the behavior of SQLite is to accept the execution of a statement, even if nothing happens. For example, if I have 5 banks in the `bank_table`: `[A, B, C, D, E]` and I try to `DELETE` bank `F`, the SQLite statement shouldn't throw an exception even though `F` does not exist in the table. See [this](https://stackoverflow.com/questions/55075297/how-can-i-handle-the-failure-of-a-delete-statement-to-delete-any-rows) SO post that I referenced. 

Sorry if I completely misunderstood what you were asking.

cmoussa1:
@SteVwonder thanks again for raising the point about raising an exception during the recursive `delete_bank`. One possible solution that came to mind this morning was to raise an `Exception` if an `sqlite3.OperationalError` occurs while in the `delete_bank` function. If the `Exception` is raised, we can issue a `"ROLLBACK"` [statement](), which doesn't save any changes while in the function, which should in turn leave the database in the state it was before `delete_bank()` was called:

```python
# if an exception occcurs while recursively deleting
# the parent banks, then throw the exception and roll
# back the changes made to the DB
    print(e)
    conn.execute('ROLLBACK')
```

cmoussa1:
Just pushed another couple commits that changes `delete_bank()`:

Add `Exception` handling that gets thrown if an `sqlite3.OperationalError` occurs while deleting a bank. If the `Exception` is thrown, issue a SQL `ROLLBACK` statement that will roll back any changes to the DB before `delete_bank()` was called.

Remove any associations under the bank(s) that were deleted.

src/bindings/python/flux/accounting/accounting_cli_functions.py
```diff
@@ -66,15 +66,56 @@ def view_bank(conn, bank):
...
 
-    # commit changes
-    conn.commit()
+        conn.commit()
+    # if an exception occcurs while recursively deleting
+    # the parent banks, then throw the exception and roll
+    # back the changes made to the DB
+    except sqlite3.OperationalError as e:
+        print(e)
+        conn.execute("ROLLBACK")
```
SteVwonder, line 118: It looks like python's `sqlite3` module has built-in support for rollback: https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.rollback

```suggestion
        conn.rollback()
```

src/bindings/python/flux/accounting/accounting_cli_functions.py
```diff
@@ -66,15 +66,56 @@ def view_bank(conn, bank):
...
+
+            # we've reached a bank with no sub banks
+            if len(dataframe) == 0:
+                select_associations_stmt = """
+                    SELECT user_name, bank
+                    FROM association_table
+                    WHERE bank=?
+                    """
+                for row in cursor.execute(select_associations_stmt, (row["bank"],)):
+                    delete_user(conn, user=row[0], bank=row[1])
```
SteVwonder, line 102: `delete_user` calls `conn.commit`, which will prevent an effective rollback in the case of an exception.  If `delete_user` is only called by this function, then I'd remove the `conn.commit()` call.  If it is called by other functions, maybe add an optional argument `commit` to allow committing to be turned off for this particular call?

SteVwonder (changes requested):
Thanks @cmoussa1!  The rollback looks like the right change.  One comment about commits being done in `delete_user` and then I think this PR is good to go!

src/bindings/python/flux/accounting/accounting_cli_functions.py
cmoussa1: Ah, good point. I don't believe `delete_user` is called by anything else, so I'll go ahead and remove that `commit()` call. Thanks for pointing that out.

cmoussa1:
Thanks @SteVwonder for your feedback! Just pushed a couple of changes per your suggestions.

src/bindings/python/flux/accounting/accounting_cli_functions.py
SteVwonder: Hmmm. I'm still seeing `conn.commit` in `delete_user` in the diff.  Maybe just a new push is needed?

cmoussa1: That's a fail on my part. Made the change on my machine but didn't push it. Sorry about that!

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

Issue #70, opened 2020-11-09
accounting_cli.py: change 'account' arg to 'bank' 


cmoussa1:
This will keep consistency between what the 'account' field is really supposed to represent - the bank that the user belongs to:

```diff
- subparser_add_user.add_argument(
-    "--account", help="account to charge jobs against", metavar="ACCOUNT",
- )
+ subparser_add_user.add_argument(
+    "--bank", help="bank to charge jobs against", metavar="BANK",
+ )
```
...
```diff
- elif args.func == "add_user":
-    aclif.add_user(
-        ...
-        args.account,
-        ...
-    )
+ elif args.func == "add_user":
+     aclif.add_user(
+         ...
+        args.bank,
+        ....
+     )
```

Issue #67, opened 2020-11-06
create_db.py: change 'user_name' field to 'username'


cmoussa1:
This will require changes in the following files:

**create_db.py**
`add_user()` in **accounting_cli_functions.py**
**print_hierarchy.py**
**print_hierarchy.cpp**

PR #79, opened 2020-11-18
fairshare implementation part 1: add effective job usage factor calculation  

### LINKED ISSUE #58: job_archive_interface: implement job usage calculation
Okay, now that #57 has landed and the job-archive interface functions are their own module, we can move on and start to extend this module's functionality beyond just job record listing. 

We discussed last week about this module playing a part in some of the "internal" work for fair share calculation. Although the static information comes directly from the flux-accounting database (user account information and bank information), this specific module (which interacts with flux-core's job-archive module) can fetch information pertaining to job history and calculate the usage factor in fair share. 

#### Slurm's definition of usage

If we wanted to calculate job usage according to Slurm's definition, it would be the following:

```
U_normalized = U_user / U_total
```

where: 

> `U_user` is the processor * seconds consumed by all of a user's jobs in a given account for over a fixed time period and `U_total` is the total number of processor * seconds utilized across the cluster during that same time period. 

So, I _think_ this would entail fetching job records for some user _U_, summing up the number of cores and multiplying it by `t_elapsed` across those same jobs, and dividing it by the number of cores * `t_elapsed` across all job records in the job-archive.  To only take the jobs that have completed in the past _x_ minutes, we can add another filter to fetching job records to only fetch jobs that have completed after some certain time _t_ (this could probably help represent the _decay factor_  in calculating job usage, where more recent jobs weigh heavier than older jobs). 

One note I thought of that I think we should keep in mind: in another issue thread, @ryanday36 made this good comment about charging usage:

> Accounting by total number of cores works fine as long as users get charged for all of the cores when we give them a whole node. i.e., in slurm, a user can run something like 'srun -n1 ...' and only ask for 1 task on one core, but, on most of our clusters, we're giving them the whole node anyway, so we need to charge them for all 36 (or whatever) cores.

So, if I am understanding this correctly, that means even if a user submits a job with the following job spec:

```bash
> flux mini submit -n 1 hostname
```

this would mean we would need to charge for the whole node when performing the usage calculation, right? If that's the case, would our usage calculation instead become a product of the total `num_nodes` and `t_elapsed` across a user's jobs? 

FWIW, I submitted two jobs in the `fluxorama` Docker container:

```bash
> [fluxuser@d215a637d63f flux-ba]$ flux mini submit -n 1 hostname
ƒAXNoQrj
> [fluxuser@d215a637d63f flux-ba]$ flux mini submit -N 1 hostname
ƒD3dvZMq
```

and the result of their two job specs are fairly similar:

```json
ƒAXNoQrj

[
  {
    "type": "slot",
    "count": 1,
    "with": [
      {
        "type": "core",
        "count": 1
      }
    ],
    "label": "task"
  }
]
```

```json
ƒD3dvZMq

[
  {
    "type": "node",
    "count": 1,
    "with": [
      {
        "type": "slot",
        "count": 1,
        "with": [
          {
            "type": "core",
            "count": 1
          }
        ],
        "label": "task"
      }
    ]
  }
]
```

So I could probably re-use the  `count_ranks()` helper function to count the number of nodes for a users' jobs and the `t_elapsed` property for all of the job records. And, from speaking with @chu11, to get a core count from a node, we'd have to look up node hardware info, which isn't stored in a job's `R` as of now.

So, maybe to start, I can calculate a usage factor based on `nnodes` * `t_elapsed`, and then adjust if we need to instead calculate a precise number of cores.

cmoussa1:
_note: this PR should probably be reviewed after #73 gets approved and landed since that is a higher priority at the moment, so I'll leave it as a Draft PR for now._ 

_note 2: there are a couple of smaller fixes that I included as a part of this PR as well - more specifically issues #67 and #70. If it is more appropriate to cut those out and post as separate PRs, I would be happy to do so._

This PR is part 1 of implementing a [Level Fairshare](https://slurm.schedmd.com/fair_tree.html) calculation into flux-accounting - calculating an effective usage factor for a given user in the flux-accounting database. It looks at a user's given jobs in the job-archive, calculates a raw usage factor (while also applying a half-life decay to that user's past job usage factors), and then divides that by a job usage factor calculated by the user's siblings' jobs. This returns an effective usage factor. Users who run many jobs will have a larger usage factor than those who do not run as many jobs. 

More specific details for how the FluxAccounting database stores past job usage factors, how the job usage factor calculation is performed, and a walkthrough of an example can be found on [this](https://github.com/flux-framework/flux-accounting/wiki/Job-Usage-Factor-Calculation-Notes) Wiki page. 

Essentially, the table that stores past job usage factors (the ones that have the half-life factor applied to them) is dynamically sized when the database is first created via two parameters, **PriorityDecayHalfLife** (how long we want one period of time to represent one job usage factor) and **PriorityUsageResetPeriod** (the amount of time before jobs no longer play a role in calculating a job usage factor). For now, both of these parameters represent a time period in weeks - however, I could see how we might want to make these bins shorter (in days, for example). 

I see the main function `calc_usage_factor()` as a function that gets called via a `cron` job, one that runs every _x_ period of time or so and updates all of the job usage factors for users in the `association_table`.

Fixes #58

### DIFF
--- a/src/bindings/python/flux/accounting/create_db.py
+++ b/src/bindings/python/flux/accounting/create_db.py
@@ -15,8 +15,64 @@ import logging
 import argparse
 import sys
 import pathlib
+import math
+import time
 
-def create_db(filepath):
+
+def add_usage_columns_to_table(
+    conn, table_name, priority_usage_reset_period=None, priority_decay_half_life=None
+):
+    # the number of columns (or 'bins') holding usage factors is determined by
+    # the following:
+    #
+    # PriorityUsageResetPeriod / PriorityDecayHalfLife
+    #
+    # each parameter represents a number of weeks by which to hold usage
+    # factors up to the time period where jobs no longer play a factor in
+    # calculating a usage factor
+    column_name = "usage_factor_period_0"
+    if priority_decay_half_life != None and priority_usage_reset_period != None:
+        num_columns = math.ceil(
+            int(priority_usage_reset_period) / int(priority_decay_half_life)
+        )
+    else:
+        num_columns = 4
+    for i in range(num_columns):
+        alter_stmt = (
+            "ALTER TABLE "
+            + table_name
+            + " ADD COLUMN "
+            + column_name
+            + " REAL DEFAULT 0.0"
+        )
+        conn.execute(alter_stmt)
+        conn.commit()
+        column_name = "usage_factor_period_" + str(i + 1)
+
+
+def set_half_life_period_end(conn, priority_decay_half_life=None):
+    if priority_decay_half_life != None:
+        # convert number of weeks to seconds; this will be appended
+        # to the current time to represent one 'half-life' period
+        # for the first usage bin
+        half_life_period = int(priority_decay_half_life) * 604800
+        half_life_period_end = time.time() + half_life_period
+    else:
+        half_life_period = 604800
+        half_life_period_end = time.time() + half_life_period
+
+    update_stmt = """
+        UPDATE t_half_life_period_table
+        SET end_half_life_period=?
+        WHERE cluster='cluster'
+        """
+    conn.execute(update_stmt, (str(half_life_period_end),))
+    conn.commit()
+
+
+def create_db(
+    filepath, priority_usage_reset_period=None, priority_decay_half_life=None
+):
     db_dir = pathlib.PosixPath(filepath).parent
     db_dir.mkdir(parents=True, exist_ok=True)
     try:
@@ -60,5 +116,49 @@ def create_db(filepath):
         );"""
     )
     logging.info("Created bank_table successfully")
+
+    # Job Usage Factor Table
+    # stores past job usage factors for users
+    logging.info("Creating job_usage_factor table in DB...")
+    conn.execute(
+        """
+            CREATE TABLE IF NOT EXISTS job_usage_factor_table (
+                username            tinytext                    NOT NULL,
+                bank                tinytext                    NOT NULL,
+                last_job_timestamp  real        DEFAULT 0.0,
+                PRIMARY KEY (username, bank),
+                FOREIGN KEY (username, bank)
+                    REFERENCES association_table (username, bank)
+                        ON UPDATE CASCADE
+                        ON DELETE CASCADE
+        );"""
+    )
+    add_usage_columns_to_table(
+        conn,
+        "job_usage_factor_table",
+        priority_usage_reset_period,
+        priority_decay_half_life,
+    )
+    logging.info("Created job_usage_factor_table successfully")
+
+    # Half Life Timestamp Table
+    # keeps track of current half-life period
+    logging.info("Creating t_half_life_period_table in DB...")
+    conn.execute(
+        """
+            CREATE TABLE IF NOT EXISTS t_half_life_period_table (
+                cluster               tinytext DEFAULT 'cluster',
+                end_half_life_period  real     DEFAULT 0.0
+
+        );"""
+    )
+    conn.execute(
+        """
+            INSERT INTO t_half_life_period_table (cluster, end_half_life_period)
+            VALUES ('cluster', 0.0);
+        """
+    )
+    set_half_life_period_end(conn, priority_decay_half_life)
+    logging.info("Created t_half_life_period_table successfully")
 
     conn.close()
--- a/src/bindings/python/flux/accounting/create_db.py
+++ b/src/bindings/python/flux/accounting/create_db.py
@@ -36,13 +36,13 @@ def create_db(filepath):
                 creation_time bigint(20)            NOT NULL,
                 mod_time      bigint(20)  DEFAULT 0 NOT NULL,
                 deleted       tinyint(4)  DEFAULT 0 NOT NULL,
-                user_name     tinytext              NOT NULL,
+                username      tinytext              NOT NULL,
                 admin_level   smallint(6) DEFAULT 1 NOT NULL,
                 bank          tinytext              NOT NULL,
                 shares        int(11)     DEFAULT 1 NOT NULL,
                 max_jobs      int(11)               NOT NULL,
                 max_wall_pj   int(11)               NOT NULL,
-                PRIMARY KEY   (user_name, bank)
+                PRIMARY KEY   (username, bank)
         );"""
     )
     logging.info("Created association_table successfully")
--- a/src/bindings/python/flux/accounting/flux-account.py
+++ b/src/bindings/python/flux/accounting/flux-account.py
@@ -11,11 +11,8 @@
 ###############################################################
 import sqlite3
 import argparse
-import time
 import sys
 import os
-
-import pandas as pd
 
 import flux.accounting
 from flux.accounting import accounting_cli_functions as aclif
@@ -32,8 +29,7 @@ def main():
         SQLite instructions for the Flux Accounting Database.
         """
     )
-    subparsers = parser.add_subparsers(help="sub-command help",
-                                       dest="subcommand")
+    subparsers = parser.add_subparsers(help="sub-command help", dest="subcommand")
     subparsers.required = True
 
     parser.add_argument(
@@ -123,6 +119,19 @@ def main():
         "create-db", help="create the flux-accounting database"
     )
     subparser_create_db.set_defaults(func="create_db")
+    subparser_create_db.add_argument(
+        "dbpath", help="specify location of database file", metavar=("DATABASE PATH")
+    )
+    subparser_create_db.add_argument(
+        "--priority-usage-reset-period",
+        help="the number of weeks at which usage information gets reset to 0",
+        metavar=("PRIORITY USAGE RESET PERIOD"),
+    )
+    subparser_create_db.add_argument(
+        "--priority-decay-half-life",
+        help="the contribution of historical usage in weeks on the composite usage value",
+        metavar=("PRIORITY DECAY HALF LIFE"),
+    )
 
     subparser_add_bank = subparsers.add_parser("add-bank", help="add a new bank")
     subparser_add_bank.set_defaults(func="add_bank")
@@ -172,7 +181,9 @@ def main():
     # if we are creating the DB for the first time, we need
     # to ONLY create the DB and then exit out successfully
     if args.func == "create_db":
-        c.create_db(path)
+        c.create_db(
+            args.dbpath, args.priority_usage_reset_period, args.priority_decay_half_life
+        )
         sys.exit(0)
 
     # try to open database file; will exit with -1 if database file not found
@@ -183,6 +194,8 @@ def main():
     db_uri = "file:" + path + "?mode=rw"
     try:
         conn = sqlite3.connect(db_uri, uri=True)
+        # set foreign keys constraint
+        conn.execute("PRAGMA foreign_keys = 1")
     except sqlite3.OperationalError:
         print(f"Unable to open database file: {db_uri}", file=sys.stderr)
         sys.exit(1)
--- a/src/bindings/python/flux/accounting/flux-account.py
+++ b/src/bindings/python/flux/accounting/flux-account.py
@@ -64,7 +64,7 @@ def main():
         "--admin-level", help="admin level", default=1, metavar="ADMIN_LEVEL",
     )
     subparser_add_user.add_argument(
-        "--account", help="account to charge jobs against", metavar="ACCOUNT",
+        "--bank", help="bank to charge jobs against", metavar="BANK",
     )
     subparser_add_user.add_argument(
         "--parent-acct", help="parent account", default="", metavar="PARENT_ACCOUNT",
@@ -197,7 +197,7 @@ def main():
             aclif.add_user(
                 conn,
                 args.username,
-                args.account,
+                args.bank,
                 args.admin_level,
                 args.shares,
                 args.max_jobs,
--- a/src/bindings/python/flux/accounting/job_archive_interface.py
+++ b/src/bindings/python/flux/accounting/job_archive_interface.py
@@ -13,6 +13,8 @@ import sqlite3
 import time
 import pwd
 import csv
+import os
+import math
 
 import pandas as pd
 
@@ -218,3 +220,202 @@ def view_job_records(conn, output_file, **kwargs):
             write_records_to_file(job_records, output_file)
 
     return job_records
+
+
+def fetch_old_usage_factors(acct_conn, user=None, bank=None):
+    past_usage_factors = []
+
+    select_stmt = "SELECT * from job_usage_factor_table WHERE username=? AND bank=?"
+    dataframe = pd.read_sql_query(select_stmt, acct_conn, params=(user, bank,))
+
+    for val in dataframe.iloc[0].values[3:]:
+        if isinstance(val, float):
+            past_usage_factors.append(val)
+
+    return past_usage_factors
+
+
+def apply_decay_factor(decay_factor, acct_conn, user=None, bank=None):
+    past_usage_factors = []
+    past_usage_factors_w_decay = []
+
+    past_usage_factors = fetch_old_usage_factors(acct_conn, user, bank)
+
+    # apply decay factor to past usage periods of a user's jobs
+    for power, usage_factor in enumerate(past_usage_factors, start=1):
+        past_usage_factors_w_decay.append(usage_factor * math.pow(decay_factor, power))
+
+    # update job_usage_factor_table with new values, starting with period-2;
+    # the last usage factor in the table will get discarded after the update
+    period = 1
+    for usage_factor in past_usage_factors_w_decay[
+        1 : len(past_usage_factors_w_decay) - 1
+    ]:
+        update_stmt = (
+            "UPDATE job_usage_factor_table SET usage_factor_period_"
+            + str(period)
+            + "=? WHERE username=? AND bank=?"
+        )
+        acct_conn.execute(update_stmt, (str(usage_factor), user, bank,))
+        acct_conn.commit()
+        period += 1
+
+    return sum(past_usage_factors_w_decay)
+
+
+def calc_usage_factor(
+    jobs_conn, acct_conn, priority_decay_half_life, user, bank,
+):
+
+    # half_life_period represents the number of weeks (converted to
+    # seconds) that represents one usage bin
+    half_life_period = priority_decay_half_life * 604800
+
+    # fetch timestamp of last seen job (will fetch jobs that
+    # have run after this time)
+    fetch_timestamp_query = """
+        SELECT last_job_timestamp
+        FROM job_usage_factor_table
+        WHERE username=? AND bank=?
+        """
+    dataframe = pd.read_sql_query(
+        fetch_timestamp_query, acct_conn, params=(user, bank,)
+    )
+    last_job_timestamp = dataframe.iloc[0]
+
+    # fetch timestamp of the end of the current half-life period
+    fetch_half_life_timestamp_query = """
+        SELECT end_half_life_period
+        FROM t_half_life_period_table
+        """
+    dataframe = pd.read_sql_query(fetch_half_life_timestamp_query, acct_conn)
+    t_end_half_life_period = dataframe.iloc[0]
+
+    # get the total number of nodes and time elapsed across
+    # all of a user's jobs that completed since the last completed
+    # job that was retrieved
+    job_totals_user = view_job_records(
+        jobs_conn,
+        output_file=None,
+        user=user,
+        after_start_time=float(last_job_timestamp),
+    )
+
+    last_t_inactive = 0.0
+    usage_user_current = 0.0
+
+    if len(job_totals_user) > 0:
+        # sort jobs by job.t_inactive
+        job_totals_user.sort(key=lambda job: job.t_inactive)
+
+        # one per job factor = nnodes * t_elapsed
+        # usage factors = total(nnodes * t_elapsed)
+        per_job_factors = []
+        for job in job_totals_user:
+            per_job_factors.append(round((job.nnodes * job.elapsed), 5))
+
+        last_t_inactive = job_totals_user[-1].t_inactive
+        usage_user_current = sum(per_job_factors)
+
+    # if no new jobs were found and we are still in the same half-life
+    # period, then the current usage remains the same
+    if len(job_totals_user) == 0 and (
+        float(t_end_half_life_period) > (time.time() - half_life_period)
+    ):
+        # fetch past usage factors
+        past_usage_factors = fetch_old_usage_factors(acct_conn, user, bank)
+
+        usage_user_historical = sum(past_usage_factors)
+    # if no new jobs were found but we are past the most recent half-life
+    # period, then the current usage needs to have a decay factor applied
+    # to it
+    elif len(job_totals_user) == 0 and (
+        float(t_end_half_life_period) < (time.time() - half_life_period)
+    ):
+        # fetch past usage factors
+        past_usage_factors = fetch_old_usage_factors(acct_conn, user, bank)
+
+        usage_user_historical = apply_decay_factor(0.5, acct_conn, user, bank)
+    # if last_t_inactive - t_end_half_life_period < half_life_period,
+    # append newly found jobs to the most recent usage factor
+    elif (last_t_inactive - float(t_end_half_life_period)) < half_life_period:
+        # append current usage to first usage factor bin
+        fetch_current_usage_factor = """
+            SELECT usage_factor_period_0
+            FROM job_usage_factor_table
+            WHERE username=?
+            AND bank=?
+            """
+        dataframe = pd.read_sql_query(
+            fetch_current_usage_factor, acct_conn, params=(user, bank,)
+        )
+        usage_factor_period_0 = dataframe.iloc[0]
+
+        usage_user_current += float(usage_factor_period_0)
+
+        # usage_user_past will just be the sum of the older factors since
+        # they will already had their decay factor applied to them in this
+        # current half-life period
+        usage_user_past = fetch_old_usage_factors(acct_conn, user, bank)
+
+        # calculate historical usage factor for user
+        usage_user_historical = usage_user_current + sum(usage_user_past[1:])
+    # else, create a new bin, move the older factors (and throw out the oldest
+    # one), and set a new end timestamp for the next half-life period
+    else:
+        # apply decay factor to past usage periods of a user's jobs
+        usage_user_past = apply_decay_factor(0.5, acct_conn, user, bank)
+
+        # calculate historical usage factor for user
+        usage_user_historical = usage_user_current + usage_user_past
+
+    # write last t_inactive to last_job_timestamp for user
+    update_timestamp_stmt = """
+        UPDATE job_usage_factor_table SET last_job_timestamp=?
+        WHERE username=?
+        AND bank=?
+        """
+    acct_conn.execute(update_timestamp_stmt, (last_t_inactive, user, bank,))
+    acct_conn.commit()
+
+    # write historical usage to first column in job_usage_factor_table
+    update_stmt = """
+        UPDATE job_usage_factor_table
+        SET usage_factor_period_0=?
+        WHERE username=?
+        AND bank=?
+        """
+    acct_conn.execute(update_stmt, (usage_user_historical, user, bank,))
+    acct_conn.commit()
+
+    return usage_user_historical
+
+
+def update_end_half_life_period(acct_conn, priority_decay_half_life):
+    # half_life_period represents the number of weeks (converted to
+    # seconds) that represents one usage bin
+    half_life_period = priority_decay_half_life * 604800
+
+    # fetch timestamp of the end of the current half-life period
+    fetch_half_life_timestamp_query = """
+        SELECT end_half_life_period
+        FROM t_half_life_period_table
+        WHERE cluster='cluster'
+        """
+    dataframe = pd.read_sql_query(fetch_half_life_timestamp_query, acct_conn)
+    t_end_half_life_period = dataframe.iloc[0]
+
+    # check to see if we are still in the same half-life period;
+    # if not, we need to update the end_half_life_period timestamp
+    # with a new timestamp value
+    if float(t_end_half_life_period) < (time.time() - half_life_period):
+        # update new end of half-life period timestamp
+        update_timestamp_stmt = """
+            UPDATE t_half_life_period_table
+            SET end_half_life_period=?
+            WHERE cluster='cluster'
+            """
+        acct_conn.execute(
+            update_timestamp_stmt, ((float(t_end_half_life_period) + half_life_period),)
+        )
+        acct_conn.commit()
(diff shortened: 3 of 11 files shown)


cmoussa1:
Based on our meeting today, the following changes/additions should make their way into this PR (thanks @SteVwonder for the feedback; let me know if I missed anything else):

My current implementation for calculating a user's **RawUsage** factor is by summing up a user's nodes, summing up a user's time elapsed across all of their jobs, and multiplying the two values:

```
RawUsage = sum(nnodes) * sum(t_elapsed)
```

This should instead calculate a **RawUsage** value for each `nnodes * t_elapsed` and sum the **RawUsage** values. 

```
RawUsage = sum(nodes * t_elapsed)
```

---

The Wiki page I created to document my process for calculating the usage factors contains an example of the past job usage factors with a decay factor applied to them:

```
  username bank  usage_factor_period_0  usage_factor_period_1  usage_factor_period_2  usage_factor_period_3
0     1002    C               16.00000               16.00000               16.00000               16.00000
```

which results in the following list of decay factors: `[8.0, 4.0, 2.0, 1.0]`

This is pretty confusing; I should change the past usage factors to not be the same values.

---

I should add this Wiki page as a **README.md** to the repo itself. 

---

The **PriorityDecayHalfLife** and **PriorityUsageResetPeriod** as of now are hard coded into `calc_usage_factor()`. They should probably be stored somewhere permanent in the database, where they can be read from to calculate job usage factors.

test/test_job_archive_interface.py
```diff
@@ -47,63 +57,72 @@ def setUpClass(self):
...
-                        ranks,
-                        t_submit,
-                        t_sched,
-                        t_run,
-                        t_cleanup,
-                        t_inactive,
-                        eventlog,
-                        jobspec,
-                        R
+        def populate_job_archive_db(jobs_conn, userid, username, ranks, num_entries):
```
SteVwonder, line 77: Might be good to add `runtime`/`elapsed` as an argument to this function so that you can test with jobs that have different runtimes.

accounting/job_archive_interface.py
```diff
@@ -218,3 +220,129 @@ def append_to_where(where_stmt, conditional):
...
+
+    for val in dataframe.iloc[0].values:
+        if isinstance(val, float):
+            past_usage_factors.append(val)
+
+    # apply decay factor to past usage periods of a user's jobs
+    power = 1
+    for usage_factor in past_usage_factors:
+        past_usage_factors_w_decay.append(usage_factor * math.pow(decay_factor, power))
+        power += 1
```
SteVwonder, line 240: Totally fine to leave this code as is. Just wanted to share a neat little trick I learned a while back:

```python
for power, usage_factor in enumerate(past_usage_factors, start=1):
	past_usage_factors_w_decay.append(usage_factor * math.pow(decay_factor, power))
```

https://docs.python.org/3/library/functions.html#enumerate

accounting/job_archive_interface.py
```diff
@@ -218,3 +220,129 @@ def append_to_where(where_stmt, conditional):
...
+
+    # convert PriorityDecayHalfLife to a number of seconds
+    # that determine how far back to go in the job-archive
+    after_start_time = priority_decay_half_life * 604800
+
+    # get the total number of nodes and time elapsed across
+    # all of a user's jobs that completed before an end time;
+    # just store job records in a temporary .csv file that gets
+    # deleted when the calculation are done.
+    output_file = "tmp.csv"
```
SteVwonder, line 283: Can we avoid writing these files to a CSV?  Are the values too large to fit into memory?

SteVwonder (commented):
Thanks @cmoussa1 for putting this together!  I left some inline comments to complement the high-level stuff that we discussed on the call today.

test/test_create_db.py
```diff
@@ -97,10 +97,39 @@ def test_05_check_bank_ids(self):
...
+    # if PriorityDecayHalfLife and PriorityUsageResetPeriod
+    # are not specified, the job_usage_factor_table will have
+    # 4 bins, each representing 1 week's worth of jobs
+    def test_06_job_usage_factor_table_default(self):
+        c.create_db("flux_accounting_test_1.db")
+        columns_query = "PRAGMA table_info(job_usage_factor_table)"
+        test_conn = sqlite3.connect("flux_accounting_test_1.db")
+        cursor = test_conn.cursor()
+        cursor.execute(columns_query)
+        num_columns = len(cursor.fetchall())
```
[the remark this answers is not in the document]
cmoussa1: I don't know if it's possible to only query for usage bins, but maybe instead I could do a list comparison - in other words, check to see that `['usage_period_0', 'usage_period_1', ..., 'usage_period_n']` is in this query? That change could be made to both `test_06` and `test_07`.

accounting/job_archive_interface.py
cmoussa1: Ah, so this is because of the way `view_job_records()` works. If no `output_file` is specified, then the function defaults to printing out the job records to the screen. Do you think this is OK behavior to have when running `calc_usage_factor()`? Maybe we should change `view_job_records()` to only print out records if a certain flag is passed?

cmoussa1:
Thanks for your detailed feedback @SteVwonder! I just pushed up some changes per your suggestions - I've just appended new commits for now, and will squash and rebase later - they are commits 67ced37 through 215c078. I've changed the following:

- added documentation for `calc_usage_factor()` in the form of a **README.md** (which, as I am typing this, am realizing that the docs will have to be updated with the changes made in these new commits 😅)

- added a timestamp field to `job_usage_factor_table`. This will store the last `t_inactive` value found for a user's jobs, so that when the next query for job records is run, it will look for job records that have completed after this timestamp.

- refined the query to fetch job records to instead look for jobs that have completed since the last recorded `t_inactive`.

- refactored usage calculation to instead calculate `sum(nnodes * t_elapsed)` rather than `sum(nnodes) * sum(t_elapsed)` (the description for this change is also incorrect in the `fixup!` commit 9289f48 message, sorry about that)

- removed storing job records into a `.csv` file that gets removed once the function returns the usage value (please see my response to your comment above about how we want to handle this).

- removed siblings' usage calculation from `calc_usage_factor()`.

cmoussa1:
Just pushed a couple more `fixup!` commits:

- updated the `calc_usage_factor()` **README.md**

- removed the `priority_decay_half_life` and `priority_usage_reset_period` arguments from `calc_usage_factor()`

accounting/job_archive_interface.py
```diff
@@ -274,26 +274,47 @@ def calc_usage_factor(
...
+    update_timestamp_stmt = (
+        "UPDATE job_usage_factor_table SET last_job_timestamp="
+        + str(last_t_inactive)
+        + " WHERE username='"
+        + user
+        + "' AND bank='"
+        + bank
+        + "'"
+    )
+    acct_conn.execute(update_timestamp_stmt)
```
SteVwonder, line 307: Rather than using string formatting or string concatenation, I'd recommend this use the parameter substitution supported by the `execute` funcion:

```python
    update_timestamp_stmt = (
        "UPDATE job_usage_factor_table SET last_job_timestamp=?" \
        " WHERE username='?' AND bank='?'"
    )
    acct_conn.execute(update_timestamp_stmt, (user, bank))
```

there is also a named variant:

```python
    update_timestamp_stmt = (
        "UPDATE job_usage_factor_table SET last_job_timestamp=?" \
        " WHERE username=':name' AND bank=':bank'"
    )
    acct_conn.execute(update_timestamp_stmt, {"user": user, "bank": bank})
```

SteVwonder: Same comment for the `update_stmt` below

SteVwonder (commented):
Thanks @cmoussa1 for putting this together!

Beyond the comments listed below, there is the comment I made on the concall: we need to track the end time of the previous "half-life period" so that we can tell when it is time to create a new half-life period entry in the database (and throw out the oldest one).

cmoussa1:
Thanks for your feedback @SteVwonder!

> we need to track the end time of the previous "half-life period" so that we can tell when it is time to create a new half-life period entry in the database (and throw out the oldest one).

Just so I am clear: by tracking the end time of the previous "half-life period", when we re-call `calc_usage_factor()`, would we do a comparison of the first new timestamp we come across vs. the last job's timestamp written? And if the difference is greater than one "half-life" period, we would then write it to a new usage bin?

SteVwonder:
> when we re-call calc_usage_factor(), would we do a comparison of the first new timestamp we come across vs. the last job's timestamp written (last_job_timestamp)? And if the difference is greater than one "half-life" period, we would then write it to a new usage bin?

Pretty much. I think the one difference is that you would compare the new `last_job_Timestamp` versus the end of the last decay period. And if the difference of that is greater than a half-life period, it signals the start of a new half-life period (and that some book keeping needs to be done to account for that new period).

cmoussa1:
Thanks for iterating on this with me @SteVwonder - I've pushed some more changes that consist of the following:

I added a new helper function, `fetch_old_usage_factors()` to retrieve old usage factors.

I added a couple of checks in `calc_usage_factor()` to see if the new last_job_timestamp occurs after the current half-life period. 

The workflow goes like this:

- jobs are fetched from the job-archive with a filter on jobs that have started after the `last_job_timestamp`
- a half-life period is calculated (in seconds) by multiplying the **PriorityDecayHalfLife** parameter by 604800 (the number of seconds in a week)

A series of checks are then performed:

- if no new jobs were found _and_ we are still in the same half-life period, then the current usage should remain the same
- if no new jobs were found but we are no longer in the same half-life period, then the current usage needs to have a decay factor applied to it
- if the difference between the last `t_inactive` of the newly found jobs and the `last_job_timestamp` is less than one half-life period, then I apply the newly found jobs to _just_ the most recent usage factor
- otherwise, I need to create a new bin and move the older factors (and throw out the oldest bin)

I hope I documented these changes okay and that the additions make sense.

cmoussa1:
It also looks like GitHub actions is now failing with some `flake8` formatting errors:

```
Run pip install flake8
Collecting flake8
  Downloading flake8-3.8.4-py2.py3-none-any.whl (72 kB)
.
.
.
Building wheels for collected packages: black
  Building wheel for black (PEP 517): started
  Building wheel for black (PEP 517): finished with status 'done'
  Created wheel for black: filename=black-20.8b1-py3-none-any.whl size=124185 sha256=eb33110fbf50ad81d4e844719fdf5f9cfc3a7294846c162429f8a2fba84275e3
  Stored in directory: /home/runner/.cache/pip/wheels/dd/c8/94/e774504958e4667180d81202d1d45cdc35d6222ae39229871b
Successfully built black
Installing collected packages: typing-extensions, typed-ast, toml, regex, pathspec, mypy-extensions, dataclasses, click, appdirs, black
Successfully installed appdirs-1.4.4 black-20.8b1 click-7.1.2 dataclasses-0.8 mypy-extensions-0.4.3 pathspec-0.8.1 regex-2020.11.13 toml-0.10.2 typed-ast-1.4.1 typing-extensions-3.7.4.3
./src/bindings/python/flux/accounting/flux-account.py:1:1: E902 IndentationError: unindent does not match any outer indentation level
1     E902 IndentationError: unindent does not match any outer indentation level
1
Error: Process completed with exit code 1.
```

I haven't seen this before.

cmoussa1:
Looks like there was some junk that snuck into `flux-account.py` while I was trying to rebase, so I just pushed a small commit that I think fixes it.

SteVwonder:
> if the difference between the last t_inactive of the newly found jobs and the last_job_timestamp is less than one half-life period, then I apply the newly found jobs to just the most recent usage factor

I think what you have is mostly right, but I think `last_job_timestamp` in the above quote should actually be the time when the last half-life period ended, not when the last job finished.  With how it is written currently, as long as a job completes once every few days, no new usage bins will be created.

cmoussa1:
Sorry for misunderstanding you @SteVwonder, hopefully I understood what you were saying correctly this time.

> I think what you have is mostly right, but I think last_job_timestamp in the above quote should actually be the time when the last half-life period ended, not when the last job finished. With how it is written currently, as long as a job completes once every few days, no new usage bins will be created.

I tried to think of how to store the time when the last half-life period ended; for now I created another table in the FluxAccounting database, `t_half_life_period_table`, that just contains one entry for the cluster, **end_half_life_period**. It stores the end time of the current half life period. When a new half-life period begins, the table gets updated with the next half-life period (by adding one half-life period to the current timestamp in `t_half_life_period_table`. It should get updated if either:

- `calc_usage_factor()` gets run, there **are no** newly found jobs and we are in a new half-life period, or

- `calc_usage_factor()` gets run, there **are** newly found jobs and  are in a new half-life period

The last unit test in `tests/test_job_archive_interface.py` simulates adding a small job that is run after the current half-life period, which should result in a smaller usage value since the previous usage should undergo a decay. 

The changes I just made should be reflected in commits edd75dc and 276f55e.

SteVwonder:
Awesome! That sounds right.  Is this ready for another review, or do you still have changes in flight?

cmoussa1:
I think this should be ready for another round of review! All of my pushes from today should be up.

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -288,6 +311,77 @@ def calc_usage_factor(
 
     usage_user_current = sum(per_job_factors)
 
+    # if no new jobs were found and we are still in the same half-life
+    # period, then the current usage remains the same
+    if len(job_totals_user) == 0 and (
+        float(t_end_half_life_period) > (time.time() - half_life_period)
```
SteVwonder, line 317: I think this logic is perfectly fine.  Typically, using `time.time()` directly can cause issues with testing.  Sidenote (just learned this today), we can use `unittest`s `mock` to override `time.time` within our tests: https://stackoverflow.com/questions/35783760/faking-a-time-stamp-from-time-time-in-nose

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -288,6 +311,77 @@ def calc_usage_factor(
...
+
+        # update new end of half-life period timestamp
+        update_timestamp_stmt = """
+            UPDATE t_half_life_period_table
+            SET end_half_life_period=?
+            WHERE cluster='cluster'
+            """
+        acct_conn.execute(
+            update_timestamp_stmt, ((float(t_end_half_life_period) + half_life_period),)
+        )
```
SteVwonder, line 342: Because this value applies to the entire table, this update should only happen after every user/bank has it's new period and decay factors applied.

It would work here if it only applied to an individual bank/user row.

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -288,6 +311,77 @@ def calc_usage_factor(
...
+        usage_user_historical = usage_user_current + usage_user_past
+
+        update_timestamp_stmt = """
+            UPDATE t_half_life_period_table
+            SET end_half_life_period=?
+            WHERE cluster='cluster'
+            """
+        acct_conn.execute(
+            update_timestamp_stmt, ((float(t_end_half_life_period) + half_life_period),)
+        )
```
SteVwonder, line 383: Same comment as above about this being global vs per row.

SteVwonder (changes requested):
Thanks @cmoussa1!  I think we are almost there.  At a high-level, I think this is on target now.  The only major thing is handling the fact that the period end is global versus the update being per bank/user pair.  Other than that, I just have minor comments below:

SteVwonder:
PS - For any changes you make in the next round of edits, can you leave the current set of commits unedited and only append commits (even if they are fixup commits added with `git commit --fixup`)?  That should make reviewing easier.  Happy to explain further if that doesn't make sense.

src/bindings/python/flux/accounting/create_db.py
```diff
@@ -27,6 +27,84 @@ def create_db(filepath):
     except sqlite3.OperationalError as e:
         logging.error(e)
         sys.exit(1)
+import math
+import time
```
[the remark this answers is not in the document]
cmoussa1: Oops. I believe I fixed that error (moving the `import` statements to the top of the file) in commit 430cb5d - that was a result of me probably messing up somewhere during a rebase after #73. I'm sorry about that - my fault for not being clearer that 430cb5d should probably be a `--fixup` commit (now I understand why it's easier to review just appended commits 😅 ).

src/bindings/python/flux/accounting/create_db.py
```diff
@@ -27,6 +27,84 @@ def create_db(filepath):
...
+        half_life_period = int(priority_decay_half_life) * 604800
+        half_life_period_end = time.time() + half_life_period
+    else:
+        half_life_period = 604800
+        half_life_period_end = time.time() + half_life_period
+
+    update_stmt = """
+        UPDATE t_half_life_period_table
+        SET end_half_life_period=?
+        WHERE cluster='cluster'
```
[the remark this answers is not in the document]
cmoussa1: So my thinking was to have one DB per cluster in a facility. Each cluster would have their own unique `half_life_period`, depending on how the DB is configured when it is first created. 

My reasoning behind the `cluster` column was just to provide some additional context on what the `end_half_life_period` timestamp represented - the end of the current half-life period for a given cluster. Although, I could see why it might be clearer to just have a one-column, one-row table that has one timestamp value:

**t_half_life_period_table**

| end_half_life_period |
| -------------------- |
| 123456789.0            |  

I hope this makes sense.

src/bindings/python/flux/accounting/create_db.py
```diff
@@ -61,4 +139,48 @@ def create_db(filepath):
...
+        """
+            CREATE TABLE IF NOT EXISTS job_usage_factor_table (
+                username            tinytext                    NOT NULL,
+                bank                tinytext                    NOT NULL,
+                last_job_timestamp  real        DEFAULT 0.0,
+                PRIMARY KEY (username, bank),
+                FOREIGN KEY (username, bank)
+                    REFERENCES association_table (username, bank)
+                        ON UPDATE CASCADE
+                        ON DELETE CASCADE
```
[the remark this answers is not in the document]
cmoussa1: I like this idea - thanks for explaining that. I think opening an issue on this is the right way to go. I will go ahead and open one up.

src/bindings/python/flux/accounting/flux-account.py
```diff
@@ -172,7 +185,13 @@ def main():
     # if we are creating the DB for the first time, we need
     # to ONLY create the DB and then exit out successfully
     if args.func == "create_db":
+<<<<<<< HEAD:src/bindings/python/flux/accounting/flux-account.py
         c.create_db(path)
+=======
+        c.create_db(
+            args.dbpath, args.priority_usage_reset_period, args.priority_decay_half_life
+        )
+>>>>>>> create-db: add job usage factor table:accounting/accounting_cli.py
```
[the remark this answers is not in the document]
cmoussa1: Yeah, so sorry about that - please see my comment above responding to the `import` statements - they should be fixed now by 430cb5d. I should squash that commit when it comes time to rebase the commits in this current PR.

src/bindings/python/flux/accounting/test/test_create_db.py
```diff
@@ -40,8 +40,13 @@ def test_01_tables_exist(self):
...
 
-        expected = ["association_table", "bank_table"]
-        test = list_of_tables[:2]
+        expected = [
+            "association_table",
+            "bank_table",
+            "job_usage_factor_table",
+            "t_half_life_period_table",
+        ]
+        test = list_of_tables[:4]
```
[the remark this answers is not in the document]
cmoussa1: Ah, no there aren't. I'll append a fixup that just does a comparison between `list_of_tables` and `expected` - no need for `test`.

src/bindings/python/flux/accounting/test/test_job_archive_interface.py
```diff
@@ -47,63 +57,72 @@ def setUpClass(self):
...
+                        (
+                            id,
+                            userid,
+                            username,
+                            ranks,
+                            time.time() - 2000,
+                            time.time() - 1000,
+                            time.time(),
+                            time.time() + 1000,
+                            time.time() + 2000,
```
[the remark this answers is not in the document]
cmoussa1: Oops. `t_submit` and the other `t_*` variables were accidentally left over from when I converted to using `time.time()`. I will append a `fixup` commit that removes those variables since they're not used.

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -218,3 +220,100 @@ def append_to_where(where_stmt, conditional):
...
+        jobs_conn, output_file=None, user=user, after_start_time=float(timestamp),
+    )
+
+    # one per job factor = nnodes * t_elapsed
+    # usage factors = total(nnodes * t_elapsed)
+    per_job_factors = []
+    last_t_inactive = 0.0
+    for job in job_totals_user:
+        per_job_factors.append(round((job.nnodes * job.elapsed), 5))
+        last_t_inactive = job.t_inactive
```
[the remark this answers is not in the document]
cmoussa1: No it is not - thanks for pointing that out. I'll go ahead and append a commit to add an in-place sort by `t_inactive`:

```python
# sort jobs by job.t_inactive
job_totals_user.sort(key=lambda job: job.t_inactive)
```

src/bindings/python/flux/accounting/create_db.py
```diff
@@ -27,6 +27,84 @@ def create_db(filepath):
...
+        for i in range(num_columns):
+            alter_stmt = (
+                "ALTER TABLE "
+                + table_name
+                + " ADD COLUMN "
+                + column_name
+                + " REAL DEFAULT 0.0"
+            )
+            conn.execute(alter_stmt)
+            conn.commit()
```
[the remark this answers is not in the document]
cmoussa1: Unfortunately, I don't think parameter substitution can be used with table/column names. I believe it is a part of the SQL standard. I referenced [this](https://stackoverflow.com/questions/48939214/why-can-parameter-substitution-not-be-used-with-tables-in-sqlite3) Stack Overflow post when I was trying to use parameter substitution. 

I don't mind adding a comment explaining the reasoning behind this in **create_db.py** if you prefer.

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -218,3 +220,100 @@ def append_to_where(where_stmt, conditional):
...
+
+    usage_user_current = sum(per_job_factors)
+
+    # write last t_inactive to last_job_timestamp for user
+    update_timestamp_stmt = """
+        UPDATE job_usage_factor_table SET last_job_timestamp=?
+        WHERE username=?
+        AND bank=?
+        """
+    acct_conn.execute(update_timestamp_stmt, (str(last_t_inactive), user, bank,))
```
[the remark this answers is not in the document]
cmoussa1: Oops. That accidentally made its way in there. Thanks for pointing that out. I'll append a `fixup` commit to remove that cast to `str`.

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -218,3 +220,100 @@ def append_to_where(where_stmt, conditional):
...
+    update_stmt = (
+        "UPDATE job_usage_factor_table SET usage_factor_period_0="
+        + str(usage_user_historical)
+        + " WHERE username='"
+        + user
+        + "' AND bank='"
+        + bank
+        + "'"
+    )
+    acct_conn.execute(update_stmt)
```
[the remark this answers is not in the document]
cmoussa1: Yes, sorry for not having switched that earlier! I will append a `fixup` commit to switch this.

src/bindings/python/flux/accounting/job_archive_interface.py
cmoussa1: Thanks for pointing this out. Maybe then the update _should_ occur per user/bank combination, instead of globally? Maybe then, the `end_half_life_period` column could instead just be added to the `job_usage_factor_table` for every user/bank combination, instead of in its own table?

cmoussa1: Or, maybe this update can happen after every user/bank has its new period and decay factors applied. Maybe the cron job that calls `calc_usage_factor()` can have a workflow like the following, where:

- it iterates through every user/bank combination and calls `calc_usage_factor()` for each one

- once it finishes and every user/bank combo is updated, it can check to see if we are in still in the same half-period, and if not, update `t_half_life_period_table` with the new `end_half_life_period` timestamp. I think that could remove the `update_timestamp_stmt` from this `if-else` branch. 

Does that make sense?

src/bindings/python/flux/accounting/create_db.py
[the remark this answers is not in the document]
SteVwonder: > So my thinking was to have one DB per cluster in a facility. Each cluster would have their own unique half_life_period, depending on how the DB is configured when it is first created.

Ok. Makes sense.  Is that how Slurm is configured on LC systems currently?  One usage/accounting database per cluster?  If so, that works for me.  If not, maybe we need to ping Ryan and others to see if one DB per cluster will work for them.

src/bindings/python/flux/accounting/create_db.py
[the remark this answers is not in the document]
SteVwonder: Ah! I see.  Thanks for explaining that!  No need for a comment, then.  Sorry for the noise.

src/bindings/python/flux/accounting/job_archive_interface.py
SteVwonder: Yeah. I think either of those would work.  I have a slight preference for the latter, since the period end time conceptually seems like it should be independent of the user/bank, and it might be a little weird for every user/bank to have their own end periods.

cmoussa1:
Thanks for giving such great detailed feedback @SteVwonder - this has been super helpful so far! I pushed up a series of `fixup` commits that address your suggestions (except for the unittest `mock.mock()` suggestion - maybe I can pick your brain a little bit to see how best to implement that into the unit tests - maybe I should open an issue on it?). Let me know if I can clarify anything. The last two `fixup` commits moves the update to the global `end_half_life_period` timestamp into its own function, `update_end_half_life_period`. 

Instead of this global update happening while iterating through user/bank combinations, this new `update_end_half_life_period` function would get called after the iterating is complete. If we are in a new half-life period, the table should be updated with a new timestamp value.

src/bindings/python/flux/accounting/job_archive_interface.py
cmoussa1: I also have a preference for the latter - I think it makes sense to keep it separate from updating individual user/bank values. `calc_usage_factor()` was getting pretty long as it is and it would probably be cleaner to move it into its own function. That implementation should be pushed up with the new `fixup` commits.

src/bindings/python/flux/accounting/create_db.py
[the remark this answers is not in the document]
cmoussa1: Just pinged Ryan to double check. If I remember correctly, I believe it was okay to to have one usage/accounting database per cluster, but I will confirm and get back to you as soon as I can.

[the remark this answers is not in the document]
cmoussa1: Wanted to follow up on this. One usage/accounting DB per cluster is okay to start - long term, I think the plan would be to have one DB per network.

cmoussa1:
Just pushed a couple more `fixup` commits:

- realized `last_t_inactive` and `usage_user_current` should be initialized to some default values (I chose `0.0` for both)

- added a couple of tests that assure the following:

1) a user's usage value should decrease if we are in a new half-life period and the user has not run any jobs during that time

2) if we are in a new half-life period, `update_end_half_life_period()` should update the timestamp to reflect the end of the new half-life period.

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -394,16 +394,13 @@ def calc_usage_factor(
...
-        + "'"
-    )
-    acct_conn.execute(update_stmt)
+    update_stmt = """
+        UPDATE job_usage_factor_table
+        SET usage_factor_period_0=?
+        WHERE username=?
+        AND bank=?
+        """
+    acct_conn.execute(update_stmt, (str(usage_user_historical), user, bank,))
```
SteVwonder, line 388: Does this require the `str` call? Or can it be dropped?

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -218,3 +220,202 @@ def append_to_where(where_stmt, conditional):
...
+    acct_conn.commit()
+
+    return usage_user_historical
+
+
+def update_end_half_life_period(acct_conn, priority_decay_half_life):
```
SteVwonder, line 394: Where does this function get called from outside of the tests?  I'm not seeing it in any of the diffs.

SteVwonder (commented):
Thanks @cmoussa1 for all the fixups.  Structuring it that way made it much easier to review.  You can go ahead and squash those down now :)

Thanks also for adding those mock tests in the last commit.  Those are great!

Just two questions (one is super minor) and then I think this is ready to go in.

src/bindings/python/flux/accounting/job_archive_interface.py
cmoussa1: This does not - thanks for pointing that out. I will append a `fixup` commit to address this.

src/bindings/python/flux/accounting/job_archive_interface.py
cmoussa1: What I had in mind was that this function would get called by the cron job script (that isn't written yet) that runs every **PriorityCalcPeriod** that iterates through the user/bank combinations and calls `calc_usage_factor()` on each of those combinations. After it finishes iterating through those, it would call `update_end_half_life_period()` to see if we are in a new half-life period and update the appropriate tables.

I was thinking that in the next couple stages and PR's where we integrate `calc_usage_factor()` with `print-hierarchy`, Dong's fair tree library, write the `cron` job script, we would actually get to use and call `update_end_half_life_period()`. I hope that makes sense.

P.S Maybe I should document this in the **README.md** that talks about `calc_usage_factor()` and describe this workflow for that `cron` job script.

SteVwonder: >  I hope that makes sense.

Yeah. That does! Thanks for the explanation!

> P.S Maybe I should document this in the README.md that talks about calc_usage_factor() and describe this workflow for that cron job script.

That would be a great addition if you have the time.

src/bindings/python/flux/accounting/job_archive_interface.py
SteVwonder: LGTM! Feel free to squash that fixup commit.

SteVwonder (approved):
Thanks @cmoussa1 for all the hard work on this!  I think this is ready to go.  I'm approving, and feel free to set MWP once you are happy with the PR.

src/bindings/python/flux/accounting/job_archive_interface.py
cmoussa1: Not a problem - I went ahead and appended the above (hopefully in nicer wording 😆) to the **README.md** that outlines `calc_usage_factor()` in a `fixup` commit.

cmoussa1:
Woo! Thanks @SteVwonder! I went ahead and squashed those `fixup` commits - I will set MWP.

PR #90, opened 2020-12-22
repo: remove unused imports, variable in accounting/, change license in fairness/ to LGPL

cmoussa1:
**Problem(s)**: 

- `flake8` revealed a number of unused `import` statements throughout `src/bindings/python/flux/accounting/`.

- The headers in `fairness/` are marked as GPL, but the repo is licensed LGPL.

---

This PR removes the unused `import` statements (as well as an unused variable in **job_archive_interface.py**) as well as changes the headers in `fairness/` to LGPL. I simply copied the license header from `src/fairness/print_hierarchy/print_hierarchy.cpp`.

I'm not sure if we want to squash down the four commits that remove the `import` statements into one larger commit, but for now I just left them as separate commits. 

Fixes #83, #66

Issue #63, opened 2020-10-23
Weighted tree library


dongahn:
@cmoussa1, @SteVwonder, @ryanday36 and @garlick chatted about this at yesterday's flux-accounting meeting and today's coffee hour. There was a consensus creating a fairtree library will position flux-accounting for our future work. 

- INPUT: An account-tree object loaded from the flux-accounting database (So a separate load code needs to be developed and @cmoussa1 agreed to take a crack at this)

- PROCESSING: perform "sorting depth first visit" on this tree to compute fairshare values for each and every account in the account tree

- OUTPUT: Updated account-tree object (So a separate flush code needs to be developed to update the new states into the flux-accounting database and @cmoussa1 agreed to take a crack at this)

The library approach will allow flux-accounting to remain agile until our next architectural decision is made.

dongahn:
@garlick mentioned this as a reference: https://github.com/flux-framework/flux-sched/pull/275

dongahn:
Flux architecture has significantly changed since this writeup. But FWIW: https://github.com/flux-framework/flux-sched/wiki/Job-Priority-Plugin-Implementation-Notes

garlick:
Actually I think this one:
https://github.com/flux-framework/flux-sched/wiki/Note-About-The-Flux-Fair-Share-Algorithm

garlick:
On branding, If I put "fairness" into google translate, I find lots of great sounding words!

Norwegian: rettferdighet
Swedish: rättvisa
Latin: aeque
Korean: gongpyeong
Estonian: õiglus

Anyway.

dongahn:
@garlick: 

> Actually I think this one:
https://github.com/flux-framework/flux-sched/wiki/Note-About-The-Flux-Fair-Share-Algorithm

Ah... this is indeed useful.

If I read this correctly, the main challenge with fair tree is difficulty in understanding how it calculates fair share.

I suspect this problem is two-fold: 1) poor documentation to explain the fair tree algorithm and 2) lack of user-level tools to assist users with reasoning about their current fare shares. 

Like you suggested we can significantly improve 1) by more clearly describing this algorithm in simple terms in an RFC; 2) can be improved by adding some tool support. 

To me this algorithm was described as overly "complicated," which shouldn't be the case. 

[...]

PR #91, opened 2021-01-04
association_table: add job usage column to association_table

### LINKED ISSUE #87: print-hierarchy: add usage factor to print-hierarchy
After #79 lands, one of the immediate next steps will be to add the value returned by `calc_usage_factor()` to the output of `print-hierarchy`:

```console
$ flux account print-hierarchy
Bank|User|RawShares|Usage
A||1
 B||1
  D||1
   D|user1|1|123
  E||1
 C||1
  F||1
   F|user2|1|456
   F|user3|1|789
  G||1
   G|user4|1|123
```

We can add this usage value to the `association_table` in the flux-accounting database for every user/bank combination in the database:

```sql
usage real DEFAULT 0.0,
```

And add an update_stmt to this value in the `association_table` at the end of `calc_usage_factor()`.

The other thing we'll need to change is the query used to fetch user information in **print_hierarchy.py** and **print_hierarchy.cpp**:

```python
"""
SELECT association_table.username,
association_table.shares,
association_table.usage,
association_table.bank
FROM association_table
WHERE association_table.bank=?
"""
```

cmoussa1:
~Now that #79 has landed, we can add the job usage value returned by `calc_usage_factor()` to the output of `print-hierarchy`. The output should now look like the following:~

_please see comments at the bottom of this PR thread regarding changing the structure of this PR._ 

_note: this PR also includes the re-addition of `AUTHORS`; I removed the file in #90 not realizing that it was required to build flux-accounting: so sorry about that!_

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -8,7 +8,7 @@ AC_CONFIG_AUX_DIR([config])
 AC_CONFIG_MACRO_DIR([config])
 AC_CONFIG_SRCDIR([NEWS])
 
-AM_INIT_AUTOMAKE([subdir-objects tar-ustar filename-length-max=256])
+AM_INIT_AUTOMAKE([subdir-objects tar-ustar filename-length-max=256 foreign])
 AM_SILENT_RULES([yes])
 AM_CONFIG_HEADER([config.h])
 AM_MAINTAINER_MODE([enable])
--- a/src/bindings/python/flux/accounting/create_db.py
+++ b/src/bindings/python/flux/accounting/create_db.py
@@ -87,15 +87,16 @@ def create_db(
     conn.execute(
         """
             CREATE TABLE IF NOT EXISTS association_table (
-                creation_time bigint(20)            NOT NULL,
-                mod_time      bigint(20)  DEFAULT 0 NOT NULL,
-                deleted       tinyint(4)  DEFAULT 0 NOT NULL,
-                username      tinytext              NOT NULL,
-                admin_level   smallint(6) DEFAULT 1 NOT NULL,
-                bank          tinytext              NOT NULL,
-                shares        int(11)     DEFAULT 1 NOT NULL,
-                max_jobs      int(11)               NOT NULL,
-                max_wall_pj   int(11)               NOT NULL,
+                creation_time bigint(20)              NOT NULL,
+                mod_time      bigint(20)  DEFAULT 0   NOT NULL,
+                deleted       tinyint(4)  DEFAULT 0   NOT NULL,
+                username      tinytext                NOT NULL,
+                admin_level   smallint(6) DEFAULT 1   NOT NULL,
+                bank          tinytext                NOT NULL,
+                shares        int(11)     DEFAULT 1   NOT NULL,
+                max_jobs      int(11)                 NOT NULL,
+                max_wall_pj   int(11)                 NOT NULL,
+                job_usage     real        DEFAULT 0.0 NOT NULL,
                 PRIMARY KEY   (username, bank)
         );"""
     )
--- a/src/bindings/python/flux/accounting/job_archive_interface.py
+++ b/src/bindings/python/flux/accounting/job_archive_interface.py
@@ -427,6 +427,23 @@ def calc_usage_factor(
     )
     acct_conn.commit()
 
+    # update job_usage column in association_table
+    update_usage_stmt = """
+        UPDATE association_table
+        SET job_usage=?
+        WHERE username=?
+        AND bank=?
+        """
+    acct_conn.execute(
+        update_usage_stmt,
+        (
+            usage_user_historical,
+            user,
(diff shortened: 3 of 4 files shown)


cmoussa1:
From our coffee hour discussion, banks also need to have a `RawUsage` value, which is the sum of `RawUsage` of its children.

cmoussa1:
Rebased to catch up after #92.

---

I've been playing around with adding the `RawUsage` values to the parent bank/sub bank rows this morning and trying to think of a good solution. Tallying up `RawUsage` for a bank with associations under it (i.e a leaf bank) seems pretty straightforward given the current **print_hierarchy.py** structure - just sum the `RawUsage` values for all associations under the leaf bank:

```python
# get RawUsage for this bank
select_stmt = """
        SELECT association_table.job_usage
        FROM association_table
        WHERE association_table.bank=?
        """
dataframe = pd.read_sql_query(select_stmt, conn, params=(row["bank"],))
bank_raw_usage = dataframe['job_usage'].sum()
```

And we can append the resulting `bank_raw_usage` to the current hierarchy line before we start printing out associations row by row.

What I am unsure of, however, is how to propagate the sums to its parent banks, i.e also add the summed `RawUsage` to the parent banks, so that it looks like:

```
Bank|User|RawShares|RawUsage
A||5 <- these RawUsage values here
 B||1|0 <-
  D||1|0
   D|user1|1|0.0
  E||1|0
 C||1|5 <- 
  F||1|5
   F|user2|1|3.0
   F|user3|1|2.0
  G||1|0
   G|user4|1|0.0
```

@SteVwonder do you have any ideas on how we might be able to propagate the `RawUsage` values up to parent banks and ultimately the root bank?

SteVwonder:
> @SteVwonder do you have any ideas on how we might be able to propagate the RawUsage values up to parent banks and ultimately the root bank?

You might need to traverse the entire bank hierarchy (i.e., issue all of your DB queries) and aggregate the RawUsage for all of the banks before you start printing. I think this is what you will have to do in C++ too in order to populate @dongahn's weighted tree data structure, right?

dongahn:
> @SteVwonder do you have any ideas on how we might be able to propagate the RawUsage values up to parent banks and ultimately the root bank?

Is your code in a position to do "subtree propagation towards the root" like https://github.com/flux-framework/flux-accounting/blob/master/src/fairness/weighted_tree/weighted_tree.cpp#L183?

SteVwonder:
```python
# get RawUsage for this bank
select_stmt = """
        SELECT association_table.job_usage
        FROM association_table
        WHERE association_table.bank=?
        """

dataframe = pd.read_sql_query(select_stmt, conn, params=(row["bank"],))
bank_raw_usage = dataframe['job_usage'].sum()
```
As an aside, I think you can use SQL's `SUM` function here: https://www.w3schools.com/SQl/sql_count_avg_sum.asp

```python
select_stmt = """
        SELECT SUM(job_usage)
        FROM association_table
        WHERE association_table.bank=?
        """
```

cmoussa1:
After talking some about this in yesterday's coffee hour, I feel like it might be best to adjust the direction of this PR to more closely focus on the steps outlined in #63, which is actually loading the account-tree object, instead of focusing first on the printing the hierarchical tree via a subcommand (`flux account print-hierarchy`). 

I'll describe the required work in a comment on that issue.

cmoussa1:
Thinking ahead for the end of the month release for flux-accounting; I don't think I can make the changes necessary for this PR and get it reviewed in time for the end of the month, but the first four commits are useful and in my opinion should try to be merged in so that #97 can land as well. 

I can drop the last three commits that add the usage values to the output of `print-hierarchy` and change this PR's title to **"Add usage values to association_table"**.

cmoussa1:
@SteVwonder - I've changed the structure of this PR to just include the addition of the `job_usage` column in the `association_table` in the flux-accounting database. I will focus on adding this column to the output of `print-hierarchy` in the next month's release; however, this PR probably should land before #97 gets merged since that PR is built on top of this one.

Let me know if I made things more confusing than they needed to be. 😅

AUTHORS
```diff
@@ -0,0 +1,7 @@
+Dong Ahn
+Al Chu
+Ryan Day
+Jim Garlick
+Mark Grondona
+Stephen Herbein
+Christopher Moussa
```
garlick, line 7: We got rid of the authors file in flux-core because it's a pain to maintain, and duplicates data in the commit history.  The relevant  PR is flux-framework/flux-core#3090.  Just need an autmake option to make it OK with the authotools.

cmoussa1: Oh! I didn't know about this - thank you for letting me know. I'll go ahead and change that first commit to just add the `foreign` keyword to automake.

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

PR #92, opened 2021-01-05
delete_user/delete_bank: keep job usage history after deletion of user 

### LINKED ISSUE #84: delete-user: document deletion of job usage history on delete
Currently, the `delete-bank` subcommand removes all users under that bank when it is removed. After #79 is landed, the flux-accounting database will have another table (`job_usage_factor_table`) that references both the `association_table` and `bank_table`; when a bank (and its subsequent users) get removed, their job usage history will also get removed from this table. It might be a good idea to:

- document this in the `delete_bank()` function

```python
# deleting a bank will also remove any users
# that fall under the bank that gets deleted,
# as well as their job usage history from
# job_usage_factor_table
def delete_bank(conn, bank):
```

- document this in **create_db.py** where the `job_usage_factor_table` gets created and make a note that rows in this table will get deleted when a user/bank combination is removed. 

```python
# Job Usage Factor Table
# stores past job usage factors for users
#
# when a user gets deleted via delete_user() or delete_bank(),
# their job usage history will also be removed
logging.info("Creating job_usage_factor table in DB...")
```

- present some sort of prompt when calling `delete-user` or `delete-bank` to have the user confirm a deletion of a user or bank:

```console
$ flux account delete-user fluxuser
This will remove all job usage history for fluxuser. Are you sure you want to continue? (y/n)
``` 

```console
$ flux account delete-bank some_bank
This will remove all job usage history for the users under the bank you are about to delete. Are you sure you want to continue? (y/n)
```

Then again, maybe we don't want to delete a user's job usage history in the first place! In that case, I think that would involve just removing the `ON DELETE CASCADE` statement from the creation of `job_usage_factor_table`. Then, even if a user/bank combination ever gets removed from the flux-accounting database, their job usage history will still remain in the database, which might be behavior we want.

cmoussa1:
**Problem**: As noted in #84, the `job_usage_factor_table` references both the `association_table` and `bank_table`; when a bank (and its subsequent users) get removed, their job usage history will also be removed from this table. Speaking with Ryan, we should probably keep this data in case we need it at a later date.

---

This PR removes the `ON DELETE CASCADE` option from the creation of `job_usage_factor_table` so a user's job usage history is kept even when they are removed from the flux-accounting DB. I added a test that checks to see a user's entry in the `job_usage_factor_table` is still present even when they are removed from the `association_table`. 

Fixes #84

### DIFF
--- a/src/bindings/python/flux/accounting/README.md
+++ b/src/bindings/python/flux/accounting/README.md
@@ -26,8 +26,6 @@ def apply_decay_factor(decay_factor, acct_conn, user=None, bank=None):
 
 After the current usage factor is calculated, it is written to the first usage bin in **job_usage_factor_table** along with the other, older factors. The oldest factor gets removed from the table since it is no longer needed.
 
-Then, a similar process is repeated to calculate the raw usage factor for all of the user's siblings' jobs in that same time period.
-
 ---
 
 ### An example of calculating the job usage factor
--- a/src/bindings/python/flux/accounting/create_db.py
+++ b/src/bindings/python/flux/accounting/create_db.py
@@ -128,7 +128,6 @@ def create_db(
                 FOREIGN KEY (username, bank)
                     REFERENCES association_table (username, bank)
                         ON UPDATE CASCADE
-                        ON DELETE CASCADE
         );"""
     )
     add_usage_columns_to_table(
--- a/src/bindings/python/flux/accounting/create_db.py
+++ b/src/bindings/python/flux/accounting/create_db.py
@@ -107,7 +107,7 @@ def create_db(
     conn.execute(
         """
             CREATE TABLE IF NOT EXISTS bank_table (
-                bank_id     integer PRIMARY KEY,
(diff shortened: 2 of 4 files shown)


cmoussa1:
Okay, I was able to get some feedback from Ryan on the behavior we should probably adopt on keeping the job usage history, so I edited the PR to reflect that. I think this is ready for review! Removing **[WIP]** now.

cmoussa1:
As discussed during coffee hour, we should add `AUTOINCREMENT` to the `bank_id` column to ensure that when new banks are added after one gets deleted, it does not assume the same ID number (and thus inherits old usage).

cmoussa1:
Just pushed a commit that adds `AUTOINCREMENT` to the `bank_table` so that newly added banks will have a new ID even if old ones are deleted.

SteVwonder (approved):
Thanks @cmoussa1! This LGTM!

cmoussa1:
Thanks! I'll go ahead and set MWP.

PR #81, opened 2020-12-05
ci: fix github actions

grondo:
I noticed the Github workflow for this repo was running the pr-validator as part of the matrix job, which is unnecessary, this action only should be run once.

While I was working on it, I also moved the `coverage` check into the github workflows so that Travis-CI will no longer be required for this repo.

However, after doing that it looks like more work is needed. The unittest doesn't actually run any tests, so the current GH actions aren't checking anything. It also looks like pylint is catching errors and  `black` is reformatting code, but that isn't caught as an error in the actions(?)

https://github.com/flux-framework/flux-accounting/runs/1502453306?check_suite_focus=true#step:8:61

```
./config/tap-driver.py:8:1: E302 expected 2 blank lines, found 1
./config/tap-driver.py:9:29: E261 at least two spaces before inline comment
./config/tap-driver.py:25:1: E305 expected 2 blank lines after class or function definition, found 1
./t/pycotap/pycotap/__init__.py:19:1: E302 expected 2 blank lines, found 1
./t/pycotap/test/test.py:5:11: E401 multiple imports on one line
./t/pycotap/test/test.py:9:1: E402 module level import not at top of file
./t/pycotap/test/test.py:10:1: E402 module level import not at top of file
./t/pycotap/test/test.py:17:1: E402 module level import not at top of file
./t/pycotap/example/test_example.py:4:11: E401 multiple imports on one line
./t/pycotap/example/test_example.py:8:1: E402 module level import not at top of file
./t/pycotap/example/test_example.py:9:1: E402 module level import not at top of file
./src/bindings/python/flux/accounting/flux-account.py:14:1: F401 'time' imported but unused
./src/bindings/python/flux/accounting/flux-account.py:18:1: F401 'pandas as pd' imported but unused
./src/bindings/python/flux/accounting/flux-account.py:27:1: C901 'main' is too complex (16)
./src/bindings/python/flux/accounting/print_hierarchy.py:12:1: F401 'sqlite3' imported but unused
./src/bindings/python/flux/accounting/print_hierarchy.py:19:1: E302 expected 2 blank lines, found 1
./src/bindings/python/flux/accounting/accounting_cli_functions.py:14:1: F401 'pwd' imported but unused
./src/bindings/python/flux/accounting/accounting_cli_functions.py:15:1: F401 'csv' imported but unused
./src/bindings/python/flux/accounting/accounting_cli_functions.py:69:5: F841 local variable 'last_parent_bank_seen' is assigned to but never used
[...]

PR #96, opened 2021-01-10
delete-user, delete-bank: add commit() to functions, add bank arg to delete-user subcommand

### LINKED ISSUE #95: delete-user: subcommand needs to be updated to include 'bank' argument
Trying to delete a user results in the following error:

```console
[fluxuser@30451d797d82 accounting-DB]$ flux account -p FluxAccounting.db delete-user user1
  File "/usr/libexec/flux/cmd/flux-account.py", line 249, in <module>
    main()
  File "/usr/libexec/flux/cmd/flux-account.py", line 220, in main
    aclif.delete_user(conn, args.username)
```

The `delete-user` subcommand needs to be updated to require a `bank` argument as well:

```python
subparser_delete_user = subparsers.add_parser(
    "delete-user", help="remove a user from the accounting database"
)
subparser_delete_user.set_defaults(func="delete_user")
subparser_delete_user.add_argument(
    "username", help="username", metavar=("USERNAME")
)
subparser_delete_user.set_defaults(func="delete_user")
subparser_delete_user.add_argument(
    "bank", help="bank", metavar=("BANK")
)
```

...

```python
elif args.func == "delete_user":
    aclif.delete_user(conn, args.username, args.bank)
```

cmoussa1:
As noted in #95, there were a couple of issues with the `delete-user` and `delete-bank` subcommands.

Trying to delete a user results in the following error:

```console
[fluxuser@30451d797d82 accounting-DB]$ flux account -p FluxAccounting.db delete-user user1
  File "/usr/libexec/flux/cmd/flux-account.py", line 249, in <module>
    main()
  File "/usr/libexec/flux/cmd/flux-account.py", line 220, in main
    aclif.delete_user(conn, args.username)
```

As well as the presence of a foreign key issue between the `association_table` and the `job_usage_factor_table`. Basically, when a user gets removed from the `association_table`, their entry in the `job_usage_factor_table` is no longer valid because it references the entry in the `association_table`, thus rejecting the deletion. 

The other problem was the absence of a `conn.commit()` in `both delete_user()` and `delete_bank()`. Without it, when you pass a subcommand to try and delete either a user or bank, the command succeeds, but doesn't keep the change. So I think I will need to add `conn.commit()` to both functions; one to `delete_user()` and one at the end of `delete_bank()` (See [this](https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.commit) reference for `conn.commit()`). 

---

This PR adds `BANK` as another argument to the `delete-user` subcommand, adds `conn.commit()` to `delete_user()` and `delete_bank()`, and removes the foreign key on the `job_usage_factor_table`. 

Fixes #95

### DIFF
--- a/src/bindings/python/flux/accounting/test/test_user_subcommands.py
+++ b/src/bindings/python/flux/accounting/test/test_user_subcommands.py
@@ -126,7 +126,7 @@ class TestAccountingCLI(unittest.TestCase):
 
         self.assertEqual(len(num_rows_before_delete), 1)
 
-        aclif.delete_user(acct_conn, user="fluxuser", bank="acct")
+        aclif.delete_user(acct_conn, username="fluxuser", bank="acct")
 
         cursor.execute(
             "SELECT * FROM association_table WHERE username='fluxuser' AND bank='acct'"
--- a/src/bindings/python/flux/accounting/test/test_user_subcommands.py
+++ b/src/bindings/python/flux/accounting/test/test_user_subcommands.py
@@ -116,6 +116,25 @@ class TestAccountingCLI(unittest.TestCase):
 
         self.assertRaises(ValueError)
 
+    # delete a user from the association table
+    def test_06_delete_user(self):
+        cursor = acct_conn.cursor()
+        cursor.execute(
+            "SELECT * FROM association_table WHERE username='fluxuser' AND bank='acct'"
+        )
+        num_rows_before_delete = cursor.fetchall()
+
+        self.assertEqual(len(num_rows_before_delete), 1)
+
+        aclif.delete_user(acct_conn, user="fluxuser", bank="acct")
+
+        cursor.execute(
+            "SELECT * FROM association_table WHERE username='fluxuser' AND bank='acct'"
+        )
+        num_rows_after_delete = cursor.fetchall()
+
+        self.assertEqual(len(num_rows_after_delete), 0)
+
     # remove database and log file
     @classmethod
     def tearDownClass(self):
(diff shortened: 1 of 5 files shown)


chu11:
overall LGTM, although I think some tests are necessary?

cmoussa1:
Thanks @chu11; just pushed a couple of additional commits - one adds a unit test to check `delete_user()`. The last two were small formatting commits I noticed while adding the test (`delete_user()` took a `user` arg instead of `username`, so I changed it to match the rest of the user subcommands, and the last commit formats the edited files with the Black formatter).

src/bindings/python/flux/accounting/test/test_user_subcommands.py
```diff
@@ -116,6 +116,23 @@ def test_05_edit_bad_field(self):
...
+        num_rows_before_delete = cursor.fetchall()
+
+        aclif.delete_user(acct_conn, user="fluxuser", bank="acct")
+
+        cursor.execute(
+            "SELECT * FROM association_table WHERE username='fluxuser' AND bank='acct'"
+        )
+        num_rows_after_delete = cursor.fetchall()
+
+        self.assertGreater(len(num_rows_before_delete), len(num_rows_after_delete))
```
chu11, line 134: test could be more accurate if you you calculated the exact number of rows you expect to delete?  And make sure that the exact number is deleted?

cmoussa1: Oh okay, I think I see what you're saying. Yeah, I could test to make sure the length of the returned list is an exact number. I just pushed up a commit that performs two checks on the `SELECT` query before and after the user gets deleted. Before, it should return a length of 1, and afterwards, it should return a length of 0.

chu11 (approved):
LGTM!

cmoussa1:
Thanks! I'll go ahead and set MWP then.

cmoussa1:
Rebased to catch up after #81.

PR #97, opened 2021-01-15
weighted tree: add load of flux-accounting DB to weighted tree library

cmoussa1:
_(This is a work-in-progress PR built on top of #91.)_

This PR adds a new file to the weighted tree library, **load_accounting_db.cpp** which traverses the flux-accounting database in a hierarchical manner (similar to the `print-hierarchy` subcommand) and loads it into the weighted tree library for sorting users.

For now, the file is compiled to a binary that can be run by passing in a path to a flux-accounting database:

```
$ ./load_accounting_db FluxAccounting.db
```

which will convert the database into a `.csv` format, aggregate the job usage up from leaf nodes (users) to their respective parent banks and up to the root bank, and pass it into the weighted tree library, which will sort them and output an order of users:

```
leaf.3.1
leaf.3.2
leaf.2.1
leaf.2.2
leaf.1.3
leaf.1.1
leaf.1.2
```

---

There are a number of **TODO** items that I can think of off the top of my head, but as we talked about yesterday, I thought it might be best to push it up in it's [WIP] state to get some feedback and make sure this is headed in the right direction:

**Add tests**

I am thinking I can test this program by creating fake flux-accounting databases with a similar structure to the `.csv` files located in `src/fairness/weighted_tree/test/accounts_data` and compare the order of users they are supposed to be in. I did do this manually for now, ~and they were all correct except for the `small_tie_all` case. Here is the comparison:~

```
expected    test
---------   ----
leaf.1.3    leaf.1.3
leaf.3.3    leaf.2.3  <-- flipped order
leaf.2.3    leaf.3.3  
leaf.2.2    leaf.1.2  <-- same here
leaf.1.2    leaf.2.2 
leaf.3.2    leaf.3.2
leaf.1.1    leaf.1.1
leaf.2.1    leaf.2.1
leaf.3.1    leaf.3.1
```

_EDIT: tested this on a Linux system and the order was correct; I was getting an incorrect order on my Mac_

**Change from binary to a library function?** 

Right now the program is called as an executable for easy for testing for me, but I think ultimately this needs to be changed to a library function. I could use some feedback on whether this is the right approach.

**Depth of `aggregate_usage()`**

I think the `aggregate_usage()` function I have written now only works for a database depth of 2:

```
root
 account1
  leaf.1.1
  leaf.1.2
 account2
  leaf.2.1
  leaf.2.2
```

So this might need to be changed to a recursive function that can traverse an arbitrarily deep hierarchy and aggregate job usage.


dongahn:
@cmoussa1: thank you for making an early PR available for review. I left a few minor comments above.

`print_hierarchy.cpp` strikes me as the core piece. Like you said, if you can grow the code so that it can be a standalone library, we can use the library for both tests as well as the client side of the fairshare plugin. 

Two things:

- It looks like `print_hierarchy.cpp` walks the bank hierarchy in the DB in depth-first-walk manner. So I have to believe you can easily aggregate the sum of usage of the subtree at each bank as part of this initial walk. IOW, you will visit all of your subbanks/users before you are done with the visiting bank so you can calculate the sum of usages of your subtree on a postorder event of each visit.

- You really want to directly interface with libweighted_tree in this code so that you can avoid having to create an intermediate CSV file etc. (though you may want to have an option to produce CSV.

Altogether, the algorithm could look like

```
VISIT (my_db_metadata, my_weighted_tree_node):

    Fetch my DB entree 
    Update my_weighted_tree_node //passed in from the parent

    IF leaf
        return

    FOREACH child:
      Create a child_weighted_tree_node
      my_weighted_tree_node.add_child (child_weighted_tree_node)
      Create child_db_metadata
      VISIT (child_db_metadata, child_weighted_tree_node)
      Add child_db_metadata's raw_usage to my_weighted_tree_node's raw_usage
     
   return
 ```

**Comment by cmoussa1** [comment]
Just wanted to update on my progress before today's meetings based on the feedback I received:

> IOW, you will visit all of your sub banks/users before you are done with the visiting bank so you can calculate the sum of usages of your subtree on a postorder event of each visit.

I believe I was able to do this by creating a pointer to the last seen bank, if it is in fact a parent bank:

```c++
// otherwise, this bank has sub banks, so call this helper
// function again with the first sub bank it found
else {
  std::shared_ptr<weighted_tree_node_t> subtree_ancestor = ancestors.back();
```

Then, I can add usage values to this subtree_ancestor as it descends down into each subtree with each recursive call:

```c++
for (std::string b : banks) {
  ...
  get_sub_banks(...)
  subtree_ancestor->set_usage(subtree_ancestor->get_usage() + leaf_bank_usage);
```

This should allow usage to be aggregated even for the following hierarchy:

```
Bank,Usage
root,60
 account1,30
  leaf.1.1,10
  leaf.1.2,10
  leaf.1.3,10
 account2,20
  leaf.2.1,10
  leaf.2.2,10
 account3,10
  account4,10
   account5,10
    leaf.5.1,10
```

---

> You really want to directly interface with libweighted_tree in this code so that you can avoid having to create an intermediate CSV file etc. (though you may want to have an option to produce CSV.

Agreed. I chose the `.csv` route just so I could continue to make progress just in case I wasn't able to figure out how to integrate `libweighted_tree` in time. However, I think I am really close. Similar to the code in **weighted_tree_load.cpp**, I can create a deque of `weighted_tree_node`s:

```c++
std::shared_ptr<weighted_tree_node_t> node = nullptr;
std::deque<std::shared_ptr<weighted_tree_node_t>> ancestors;
```

And add banks to the tree via the following:

```c++
// add bank to ancestors deque
node = std::make_shared<weighted_tree_node_t> (
  ancestors.empty() ? nullptr : ancestors.back().get(),
  bank_name,
  false,
  std::stoll(bank_shares),
  0 // just a temporary usage value; the usage will be aggregated and updated later
);
ancestors.push_back(node);
```

And if the bank turns out to have users in it, I add them like this:

```c++
node = std::make_shared<weighted_tree_node_t> (
  ancestors.back().get(),
  username,
  true,
  std::stoll(user_shares),
  std::stoll(user_usage)
);
ancestors.back().get()->add_child(node);
```

However, when it comes time to create a `walker` and traverse the tree, the walker reports an incorrect size of the tree and no users are output:

```c++
std::shared_ptr<weighted_tree_node_t> root = ancestors.front();
weighted_walk_t walker(root);
```

I am guessing it is the way I am adding users to the tree, but I am not totally sure. @dongahn am I supposed to `push_back(node)` even with user nodes? In your **weighted_tree_load.cpp**, you have this:

```c++
for (int i = 0; i < (-offset + 1); ++i)
  ancestors.pop_back ();
```

just before you add users, but I'm not totally sure what this is doing.

cmoussa1:
Just pushed a commit that has the changes described in my comment above.

cmoussa1:
We discussed it a little at our team meeting today, but I think that sub banks should be added via `add_child()`. This resulted in the walker reporting the correct tree size for a flux-accounting DB.

I believe what is left is I have to re-think my current approach to aggregating the usage up from leaf banks up to the root bank, as my current implementation does not report it correctly in my latest push, so I will work on that now.

cmoussa1:
@dongahn I just pushed up a commit (bc928e5) that utilizes `libweighted_tree` instead of writing to a temporary `.csv` file. 

It adds the root bank to a deque of `weighted_tree_node_t` named **ancestors**. All sub banks and users are then added via `add_child()`. The `weighted_walk_t` walker now reports a correct tree size when walking the tree. So that question I was asking about in our team meeting earlier should be answered; thanks! 

One of the things that is left is to aggregate the usage up from users to their respective banks and up to the root bank.

cmoussa1:
Thanks to some early review and feedback from @SteVwonder, he found some changes that needed to be made when I first began utilizing `lbweighted_tree`; those changes have been pushed up and can be found in commits dee56b0 and 6a79235 (Stephen, I added your name as a co-author to these commits but please let me know if you would not like them to be there). 

I think these last two commits are what I need reviewed since they are the commits that fetch data from the flux-accounting database, build a weighted tree based on the fetched data, and output a sorted list of users.

dongahn:
@cmoussa1: thank you and @SteVwonder. I will take a look as well.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
+
+#include <iostream>
+#include <fstream>
+#include <sqlite3.h>
+#include <string.h>
```
dongahn, line 28:
```suggestion
#include <string>
```

I believe it is preferred not to add `.h` in C++ when you can.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+#include <stdio.h>
```
dongahn, line 29: If you are bringing `stdio.h` only for `fprintf`, you may consider using `std::cout` type instead and remove this dependency. BTW, 

```suggestion
#include <cstdio>
```

is preferred as well if this C standard I/O routines are required.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+#include <stdlib.h>
```
dongahn, line 30:
```suggestion
#include <cstdlib>
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  accounting_db;
+
+int get_sub_banks(
+  sqlite3* DB,
+  const std::string& bank_name,
+  int level,
+  sqlite3_stmt *b_select_shares_stmt,
+  sqlite3_stmt *b_select_sub_banks_stmt,
+  sqlite3_stmt *b_select_associations_stmt
+) {
```
dongahn: For readability and making this consistent with other C or C++ codes we have, the following would be better, I think. (Notice the leading whitespace before the open parenthesis as well).

```suggestion
int get_sub_banks (sqlite3* DB,
                   const std::string& bank_name,
                   int level,
                   sqlite3_stmt *b_select_shares_stmt,
                   sqlite3_stmt *b_select_sub_banks_stmt,
                   sqlite3_stmt *b_select_associations_stmt)
{
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  int exit = 0;
```
dongahn, line 51: It think the leading indentation should be 4 per our coding guide.

dongahn: Also, avoid using `exit` as a variable name since it is a common standard routine to exit the program. How about simply `rc` as the variable name?

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+  // bind parameter to prepared SQL statement
```
dongahn, line 54:
```suggestion
  // bind parameter to prepare SQL statement
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  exit = sqlite3_bind_text(b_select_shares_stmt, 1, bank_name.c_str(), -1, NULL);
+  if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
```
dongahn: If you want, you can use a symbolic return code, in particular you are already including `#include <cstdlib>`
```suggestion
      exit (EXIT_FAILURE);
```

https://en.cppreference.com/w/cpp/utility/program/EXIT_status

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  std::string bank_shares = reinterpret_cast<char const*> (sqlite3_column_text(b_select_shares_stmt, 0));
+
+  // add sub bank to deque
+  accounting_db.emplace_back(
+    std::to_string(level),
+    bank_name,
+    "%^+_nouser",
+    bank_shares,
+    "0"
+  );
```
dongahn:
```suggestion
  accounting_db.emplace_back(std::to_string(level),
                             bank_name,
                             "%^+_nouser",
                             bank_shares,
                              "0");
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  const std::string& bank_name,
+  int level,
+  sqlite3_stmt *b_select_shares_stmt,
+  sqlite3_stmt *b_select_sub_banks_stmt,
+  sqlite3_stmt *b_select_associations_stmt
+) {
+  int exit = 0;
+
+  // bind parameter to prepared SQL statement
+  exit = sqlite3_bind_text(b_select_shares_stmt, 1, bank_name.c_str(), -1, NULL);
```
dongahn, line 55: Our code guide prefers having a leading whitespace before the open paren for a function call

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+    bank_shares,
+    "0"
+  );
+
+  exit = sqlite3_bind_text(b_select_sub_banks_stmt, 1, bank_name.c_str(), -1, NULL);
+  if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
```
dongahn: same as above

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  // bind parameter to prepared SQL statement
+  exit = sqlite3_bind_text(b_select_shares_stmt, 1, bank_name.c_str(), -1, NULL);
+  if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
+  }
+  // execute SQL statement and store result
+  exit = sqlite3_step(b_select_shares_stmt);
```
dongahn, line 63: Did you want to handle bad return codes from here?

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  // add sub bank to deque
+  accounting_db.emplace_back(
+    std::to_string(level),
+    bank_name,
+    "%^+_nouser",
+    bank_shares,
+    "0"
+  );
+
+  exit = sqlite3_bind_text(b_select_sub_banks_stmt, 1, bank_name.c_str(), -1, NULL);
```
dongahn, line 83: Did you want to handle bad return codes from here?

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+  if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
+  }
+  exit = sqlite3_step(b_select_sub_banks_stmt);
```
dongahn, line 90: Same as above

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+    exit = sqlite3_bind_text(b_select_associations_stmt, 1, bank_name.c_str(), -1, NULL);
+    if (exit != SQLITE_OK) {
+      fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg(DB));
+      sqlite3_close(DB);
+
+      return 1;
+    }
+
+    // execute SQL statement
+    exit = sqlite3_step(b_select_associations_stmt);
```
dongahn, line 110: handle bad return code

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,331 @@
...
+    while (exit == SQLITE_ROW) {
+      std::string user = reinterpret_cast<char const*> (sqlite3_column_text(b_select_associations_stmt, 0));
+      std::string single_user_shares = reinterpret_cast<char const*> (sqlite3_column_text(b_select_associations_stmt, 1));
+      std::string job_usage = reinterpret_cast<char const*> (sqlite3_column_text(b_select_associations_stmt, 3));
+      // place the user, their allocated shares, and their job_usage into vector
+      users.emplace_back(user, single_user_shares, job_usage);
+      exit = sqlite3_step(b_select_associations_stmt);
```
dongahn, line 117: return code

dongahn (changes requested):
@cmoussa1: I need to call into a series of meetings now. Here are some of reviews I have so far.

cmoussa1:
@dongahn I just pushed a series of `fixup` commits that I hope address your suggestions. They are built on top of 6a79235 and tries to fix the indentation/formatting feedback, the change from `exit` to `rc`, and the removal of some of the unused `#include` statements.

cmoussa1:
Just pushed two additional commits:

1. The first converts **load_accounting_db.cpp** to a library instead of an executable program.
2. Add tests for **load_accounting_db.cpp**. I created sample flux-accounting databases from the existing `.csv` files in this repo and ran them against the library to make sure the fairshare order is the same.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,291 @@
...
+                            -1,
+                            NULL);
+    if (rc != SQLITE_OK) {
+        fprintf (stderr, "Failed to fetch data: %s\n", sqlite3_errmsg (DB));
+        sqlite3_close (DB);
+
+        return nullptr;
+    }
+    // execute SQL statement and store result
+    rc = sqlite3_step (b_select_shares_stmt);
```
SteVwonder, line 62: This return code should be checked and an error printed if it is not `SQLITE_ROW` (or maybe `SQLITE_DONE`).

SteVwonder: Same for the other `sqlite_*` functions below that have a return code captured but not checked.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,291 @@
...
+            users.emplace_back (user, single_user_shares, job_usage);
+            rc = sqlite3_step (b_select_associations_stmt);
+        }
+        // iterate through associations and add them as children
+        // of node
+        for (unsigned int i = 0; i < users.size (); i++) {
+            std::string username = std::get<0> (users[i]);
+            std::string user_shares = std::get<1> (users[i]);
+            std::string user_usage = std::get<2> (users[i]);
+            user_node = std::make_shared<weighted_tree_node_t> (
```
SteVwonder, line 134: I would move this tree node construction up into the while loop above, and then you can remove the `users` vector and this `for` loop.  That will shorten the code a bit and reduce the amount of looping.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,291 @@
...
+            // execute SQL statement
+            std::string bank = reinterpret_cast<char const *> (
+                sqlite3_column_text (b_select_sub_banks_stmt, 0));
+            banks.push_back (bank);
+            rc = sqlite3_step (b_select_sub_banks_stmt);
+        }
+        for (std::string b : banks) {
+            // reset the prepared statements back to their initial state and
+            // clear their bindings
+            sqlite3_clear_bindings (b_select_associations_stmt);
```
SteVwonder, line 167: Same thing here. I would merge the `while` and `for` loops, so that all of the recursive calls and sqlite3_reset/clears are done within the while loop.  Then you can remove the `banks` vector.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,291 @@
...
+    sqlite3_finalize (b_select_sub_banks_stmt);
+    sqlite3_finalize (b_select_associations_stmt);
+
+    weighted_walk_t walker (root);
+    walker.run ();
+
+    const auto &users = walker.get_users ();
+
+    // close DB connection
+    sqlite3_close (DB);
```
SteVwonder, line 288: Minor nit: I'd move this up a few lines to just below the sqlite3_finalize calls

src/fairness/weighted_tree/test/load_accounting_db.hpp
```diff
@@ -0,0 +1,39 @@
...
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
+#include <sqlite3.h>
+
+#include "src/fairness/weighted_tree/weighted_walk.hpp"
+
+using namespace Flux::accounting;
+
+std::shared_ptr<weighted_tree_node_t>
+get_sub_banks (sqlite3 *DB,
```
SteVwonder, line 31: If `load_accounting_db` is the only function that we intend to ever call `get_sub_banks`, I would remove it from the public header.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+    return users;
```
SteVwonder, line 290: I would recommend refactoring this function to return `root`, and then consumers of this function can construct a walker and run it, if they so choose.

SteVwonder (changes requested):
Generally LGTM!  Just a few changes listed below, mainly about error handling and a small refactor of the load_accounting_db return value.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+    sqlite3 *DB;
+    int rc = 0;
+
+    // open FluxAccounting DB in read-write mode; if it does not exist yet,
+    // create a new database file
+    rc = sqlite3_open_v2 (path.c_str (), &DB, SQLITE_OPEN_READWRITE, NULL);
+    if (rc) {
+        std::cerr << "error opening DB: " << sqlite3_errmsg (DB) << std::endl;
+        exit (EXIT_FAILURE);
```
SteVwonder, line 200: Since this transitioned to being a library as opposed to a front-end utility, I don't think exiting is the right move.  I'd return a valid error value, like a null pointer.

src/fairness/weighted_tree/test/load_accounting_db.cpp
cmoussa1: My apologies; maybe I could use some more clarification here.

My understanding of adding the sub banks to a vector and then iterating through that vector was so the recursive call would be executed for every sub bank found in the vector. Without it, would the function not exit after it found the first leaf bank?

cmoussa1:
Thanks for your review @SteVwonder - I just pushed up a series of `fixup` commits addressing your feedback.

src/fairness/weighted_tree/test/load_accounting_db.cpp
SteVwonder: Oh. Good call.  We need to reset the prepared sql statements before the recursive call, but reseting within the loop would bork the `sqlite3_step`.  So as it is LGTM! My bad.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // function again with the first sub bank it found
+    else {
+        parent_bank = node;
+        while (rc == SQLITE_ROW) {
+            // execute SQL statement
+            std::string bank = reinterpret_cast<char const *> (
+                sqlite3_column_text (b_select_sub_banks_stmt, 0));
+            banks.push_back (bank);
+            rc = sqlite3_step (b_select_sub_banks_stmt);
+        }
```
SteVwonder, line 166: After the `while` loop exists, it might be worth checking that `rc == SQLITE_DONE` (or similar).  A sqlite error would also break the while loop but would go uncaught and without an error message.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,290 @@
...
+                            parent_bank,
+                            username,
+                            true,
+                            std::stoll (user_shares),
+                            std::stoll (user_job_usage));
+            node->add_child (user_node);
+            leaf_bank_usage += std::stoi (user_job_usage);
+
+            rc = sqlite3_step (b_select_associations_stmt);
+        }
```
SteVwonder, line 145: After the `while` loop exists, it might be worth checking that `rc == SQLITE_DONE` (or similar).  A sqlite error would also break the while loop but would go uncaught and without an error message.

SteVwonder:
Thanks @cmoussa1! Just one last comment (above).  Then this is good to me.  Also feel free to squash any fixup commits.

src/fairness/weighted_tree/test/load_accounting_db.cpp
cmoussa1: Ah, good point. I see what you're saying - I can push a `fixup` commit that adds a check similar to the ones following the other `sqlite_step()` calls.

cmoussa1:
Thanks @SteVwonder! I went ahead and pushed up a change to add some checks for `SQLITE_ERROR` after a couple of those `sqlite_step()` calls.

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

src/fairness/weighted_tree/weighted_tree.cpp
```diff
@@ -109,7 +109,7 @@ void weighted_tree_node_t::propagate_subtree_leaf_size ()
  ******************************************************************************/
 
 weighted_tree_node_t::weighted_tree_node_t (
-    weighted_tree_node_t *parent, const std::string &name,
+    std::shared_ptr<weighted_tree_node_t> parent, const std::string &name,
```
dongahn, line 112: @cmoussa1 and @SteVwonder: I vaguely remember there was some memory leak issues with using the shared pointer. I would run valgrind on a test program.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+#include "src/fairness/weighted_tree/test/load_accounting_db.hpp"
+
+using namespace Flux::accounting;
+
+std::shared_ptr<weighted_tree_node_t> node = nullptr;
+std::shared_ptr<weighted_tree_node_t> user_node = nullptr;
+std::shared_ptr<weighted_tree_node_t> root_ptr = nullptr;
+
+std::shared_ptr<weighted_tree_node_t>
+get_sub_banks (sqlite3 *DB,
```
dongahn, line 40: Don't split the function name from the return type. There are tools that assume they are in the same line, I think.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+               const std::string &bank_name,
+               std::shared_ptr<weighted_tree_node_t> parent_bank,
+               sqlite3_stmt *b_select_shares_stmt,
+               sqlite3_stmt *b_select_sub_banks_stmt,
+               sqlite3_stmt *b_select_associations_stmt) {
```
dongahn, line 45: The open parenthesis should be on the next line per our code guide.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+    int leaf_bank_usage = 0;
+
+    // bind parameter to prepare SQL statement
+    rc = sqlite3_bind_text (b_select_shares_stmt,
+                            1,
+                            bank_name.c_str (),
+                            -1,
+                            NULL);
+    if (rc != SQLITE_OK) {
+        fprintf (stderr, "Failed to fetch data: %s\n", sqlite3_errmsg (DB));
```
dongahn, line 56: Since this is C++ code and `<iostream>` was already included, I would go with: 

```suggestion
        std::cerr << "Failed to fetch data: " << sqlite3_errmsg (DB);
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+        sqlite3_close (DB);
+
+        return nullptr;
+    }
+    // execute SQL statement and store result
+    rc = sqlite3_step (b_select_shares_stmt);
+    if (rc == SQLITE_ERROR) {
+        fprintf (stderr, "%s\n", sqlite3_errmsg (DB));
```
dongahn, line 64: Same as above.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+        sqlite3_close (DB);
+
+        return nullptr;
+    }
+
+    std::string bank_shares = reinterpret_cast<char const *>(
```
dongahn, line 70:
```suggestion
    std::string bank_shares = reinterpret_cast<char const *> (
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+        sqlite3_column_text (b_select_shares_stmt, 0));
+
+    // initialize a weighted tree node
+    node = std::make_shared<weighted_tree_node_t> (parent_bank,
+                                                   bank_name,
+                                                   false,
+                                                   std::stoll (bank_shares),
```
dongahn, line 77: http://www.cplusplus.com/reference/string/stoll/
`std::stoll` can raise an exception and it would be good to add `try/catch` at least for

`invalid_argument` and `out_of_range` exceptions.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+        parent_bank->add_child (node);
+    }
+
+    rc = sqlite3_bind_text (b_select_sub_banks_stmt,
+                            1,
+                            bank_name.c_str (),
+                            -1,
+                            NULL);
+    if (rc != SQLITE_OK) {
+        fprintf (stderr, "Failed to fetch data: %s\n", sqlite3_errmsg (DB));
```
dongahn, line 93: Same as above

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+        sqlite3_close (DB);
+
+        return nullptr;
+    }
+    rc = sqlite3_step (b_select_sub_banks_stmt);
+    if (rc == SQLITE_ERROR) {
+        fprintf (stderr, "%s\n", sqlite3_errmsg (DB));
```
dongahn, line 100: Same as above

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+        return nullptr;
+    }
+
+    // vector of strings to hold sub banks
+    std::vector<std::string> banks;
+
+    // we've reached a bank with no sub banks, so print
+    // out all associations under this sub bank
+    if (rc != SQLITE_ROW) {
```
dongahn, line 111: This is a huge conditional statement. Would it be possible to refactor the logic into some functions to improve readability?

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+        rc = sqlite3_bind_text (b_select_associations_stmt,
+                                1,
+                                bank_name.c_str (),
+                                -1,
+                                NULL);
+        if (rc != SQLITE_OK) {
+            fprintf(stderr, "Failed to fetch data: %s\n", sqlite3_errmsg (DB));
```
dongahn, line 118: Same

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+            // add user as a child of the node
+            user_node = std::make_shared<weighted_tree_node_t> (
+                            parent_bank,
+                            username,
+                            true,
+                            std::stoll (user_shares),
+                            std::stoll (user_job_usage));
+            node->add_child (user_node);
+            leaf_bank_usage += std::stoi (user_job_usage);
```
dongahn, line 142: `std::stoi` also raise exceptions

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+            rc = sqlite3_step (b_select_associations_stmt);
+        }
+        if (rc == SQLITE_ERROR) {
+            fprintf (stderr, "%s\n", sqlite3_errmsg (DB));
```
dongahn, line 147: Same

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+             curr_ancestor != nullptr;
+             curr_ancestor = curr_ancestor->get_parent ()) {
+
+            curr_ancestor->set_usage (curr_ancestor->get_usage ()
+                                      + leaf_bank_usage);
+        }
+    }
+    // otherwise, this bank has sub banks, so call this
+    // function again with the first sub bank it found
+    else {
```
dongahn, line 164: I thin the following would be a bit more readable

```suggestion
    } else {
          // otherwise, this bank has sub banks, so call this
          // function again with the first sub bank it found
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+        parent_bank = node;
+        while (rc == SQLITE_ROW) {
+            // execute SQL statement
+            std::string bank = reinterpret_cast<char const *> (
+                sqlite3_column_text (b_select_sub_banks_stmt, 0));
+            banks.push_back (bank);
+            rc = sqlite3_step (b_select_sub_banks_stmt);
+        }
+        if (rc == SQLITE_ERROR) {
+            fprintf (stderr, "%s\n", sqlite3_errmsg (DB));
```
dongahn, line 174: Same

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+            sqlite3_close (DB);
+
+            return nullptr;
+        }
+        for (std::string b : banks) {
```
dongahn, line 179: Don't construct a new string for performance ( minor though)
```suggestion
        for (const std::string &b : banks) {
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+            // reset the prepared statements back to their initial state and
+            // clear their bindings
+            sqlite3_clear_bindings (b_select_associations_stmt);
+            sqlite3_reset (b_select_associations_stmt);
+            sqlite3_clear_bindings (b_select_sub_banks_stmt);
+            sqlite3_reset (b_select_sub_banks_stmt);
+            sqlite3_clear_bindings (b_select_shares_stmt);
+            sqlite3_reset (b_select_shares_stmt);
```
dongahn, line 187: Does any of these function return an error? If so please hand them.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+            get_sub_banks (DB,
```
dongahn, line 188: I think you want to handle an error case here as well? When `get_sub_banks` return `nullptr`?

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+                           b_select_shares_stmt,
+                           b_select_sub_banks_stmt,
+                           b_select_associations_stmt);
+        }
+    }
+    return root_ptr;
+}
+
+std::shared_ptr<weighted_tree_node_t>
+load_accounting_db (const std::string &path) {
```
dongahn, line 200: Please use a single line

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+    // SQL statements to retrieve data from flux-accounting database
+    sqlite3_stmt *b_select_root_bank_stmt;
+    sqlite3_stmt *b_select_shares_stmt;
+    sqlite3_stmt *b_select_sub_banks_stmt;
+    sqlite3_stmt *b_select_associations_stmt;
+
+    sqlite3 *DB;
```
dongahn, line 207: I think it would be a good practice to add `nullptr` to these since you pass them into functions to get the results. Setting the initial condition would help for debugging.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+                                        "FROM bank_table "
+                                        "WHERE parent_bank=''";
+    // compile SQL statement into byte code
+    rc = sqlite3_prepare_v2 (DB,
+                             select_root_bank_stmt.c_str(),
+                             -1,
+                             &b_select_root_bank_stmt,
+                             0);
+    if (rc != SQLITE_OK) {
+        fprintf (stderr, "Failed to fetch data: %s\n", sqlite3_errmsg (DB));
```
dongahn, line 262: `cerr`?

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+#include <tuple>
+#include <vector>
+
+#include "src/fairness/weighted_tree/test/load_accounting_db.hpp"
+
+using namespace Flux::accounting;
+
+std::shared_ptr<weighted_tree_node_t> node = nullptr;
+std::shared_ptr<weighted_tree_node_t> user_node = nullptr;
+std::shared_ptr<weighted_tree_node_t> root_ptr = nullptr;
```
dongahn, line 37: @cmoussa1: use of globals should be minimized. Do you think you can avoid using this? At first glance, I didn't see why `get_sub_banks` need globals.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+std::shared_ptr<weighted_tree_node_t>
```
dongahn, line 39: Do you think you can document what `get_sub_banks` does in terms of INPUT, Processing and OUTPUT and error conditions?

src/fairness/weighted_tree/test/load_accounting_db.hpp
```diff
@@ -0,0 +1,31 @@
...
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
+#include <sqlite3.h>
+
+#include "src/fairness/weighted_tree/weighted_walk.hpp"
+
+using namespace Flux::accounting;
+
+std::shared_ptr<weighted_tree_node_t>
+load_accounting_db (const std::string &path);
```
dongahn, line 31: A single line would be better

src/fairness/weighted_tree/test/Makefile.am
```diff
@@ -6,8 +6,8 @@ AM_LDFLAGS = $(CODE_COVERAGE_LDFLAGS)
 
 AM_CPPFLAGS = -I$(top_srcdir) $(FLUX_CORE_CFLAGS)
 
-TESTS = weighted_tree_test01.t
-check_PROGRAMS = $(TESTS) load_accounting_db
+TESTS = weighted_tree_test01.t load_accounting_db_test02.t
```
dongahn, line 9: I believe this commit can be split into at least three:

- input/output data
- test source code
- Makefile changes

src/fairness/weighted_tree/test/load_accounting_db_test02.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <iostream>
+#include <fstream>
+#include <string>
+
+#include "src/fairness/weighted_tree/test/load_accounting_db.hpp"
+#include "src/common/libtap/tap.h"
+
+using namespace Flux::accounting;
+
+std::string accounting_db_data_dir;
```
dongahn, line 40: Perhaps avoid global by passing this as a function argument?

src/fairness/weighted_tree/test/load_accounting_db_test02.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+static void test_fairshare_order (const std::string &filename,
+                           const std::vector<std::string> &expected) {
```
dongahn, line 43: Column misaligned

dongahn (changes requested):

src/fairness/weighted_tree/test/load_accounting_db.cpp
cmoussa1: If I keep the function name on the same line as the return type, the formatting (I think) would look like this:

```c++
std::shared_ptr<weighted_tree_node_t> get_sub_banks (sqlite3 *DB,
                                                     const std::string &bank_name,
                                                     std::shared_ptr<weighted_tree_node_t> parent_bank,
                                                     sqlite3_stmt *b_select_shares_stmt,
                                                     sqlite3_stmt *b_select_sub_banks_stmt,
                                                     sqlite3_stmt *b_select_associations_stmt) 
{
```

Lines 2-6 are longer than 80 characters. Is this okay?

dongahn: The line can be split at the function parameter level. No, try not to exceed 80 chars

src/fairness/weighted_tree/test/load_accounting_db.cpp
dongahn: You will need `std::endl` at the end as well.

cmoussa1:
Just pushed up a series of commits named after the suggestions and feedback I got. I am not done addressing all of it, however; just thought I would push what I had changed for now while I continue to make progress.

src/fairness/weighted_tree/weighted_tree.cpp
cmoussa1: Here is some output from running valgrind:

```
[moussa1:test]$ valgrind --leak-check=yes ./mybinary 
==53845== Memcheck, a memory error detector
==53845== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==53845== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==53845== Command: ./mybinary
==53845== 
testing with small_no_tie.db:
leaf.3.1
leaf.3.2
leaf.2.1
leaf.2.2
leaf.1.3
leaf.1.1
leaf.1.2
==53845== 
==53845== HEAP SUMMARY:
==53845==     in use at exit: 2,047 bytes in 26 blocks
==53845==   total heap usage: 306 allocs, 280 frees, 168,237 bytes allocated
==53845== 
==53845== 2,047 (136 direct, 1,911 indirect) bytes in 1 blocks are definitely lost in loss record 8 of 8
==53845==    at 0x4C2A593: operator new(unsigned long) (vg_replace_malloc.c:342)
==53845==    by 0x405A67: __gnu_cxx::new_allocator<std::_Sp_counted_ptr_inplace<Flux::accounting::weighted_tree_node_t, std::allocator<Flux::accounting::weighted_tree_node_t>, (__gnu_cxx::_Lock_policy)2> >::allocate(unsigned long, void const*) (new_allocator.h:104)
==53845==    by 0x405797: std::allocator_traits<std::allocator<std::_Sp_counted_ptr_inplace<Flux::accounting::weighted_tree_node_t, std::allocator<Flux::accounting::weighted_tree_node_t>, (__gnu_cxx::_Lock_policy)2> > >::allocate(std::allocator<std::_Sp_counted_ptr_inplace<Flux::accounting::weighted_tree_node_t, std::allocator<Flux::accounting::weighted_tree_node_t>, (__gnu_cxx::_Lock_policy)2> >&, unsigned long) (alloc_traits.h:357)
==53845==    by 0x40E10D: std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<Flux::accounting::weighted_tree_node_t, std::allocator<Flux::accounting::weighted_tree_node_t>, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool, long long, int>(std::_Sp_make_shared_tag, Flux::accounting::weighted_tree_node_t*, std::allocator<Flux::accounting::weighted_tree_node_t> const&, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool&&, long long&&, int&&) (shared_ptr_base.h:616)
==53845==    by 0x40DF4F: std::__shared_ptr<Flux::accounting::weighted_tree_node_t, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<std::allocator<Flux::accounting::weighted_tree_node_t>, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool, long long, int>(std::_Sp_make_shared_tag, std::allocator<Flux::accounting::weighted_tree_node_t> const&, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool&&, long long&&, int&&) (shared_ptr_base.h:1090)
==53845==    by 0x40DDED: std::shared_ptr<Flux::accounting::weighted_tree_node_t>::shared_ptr<std::allocator<Flux::accounting::weighted_tree_node_t>, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool, long long, int>(std::_Sp_make_shared_tag, std::allocator<Flux::accounting::weighted_tree_node_t> const&, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool&&, long long&&, int&&) (shared_ptr.h:316)
==53845==    by 0x40DC77: std::shared_ptr<Flux::accounting::weighted_tree_node_t> std::allocate_shared<Flux::accounting::weighted_tree_node_t, std::allocator<Flux::accounting::weighted_tree_node_t>, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool, long long, int>(std::allocator<Flux::accounting::weighted_tree_node_t> const&, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool&&, long long&&, int&&) (shared_ptr.h:588)
==53845==    by 0x40D9EF: std::shared_ptr<Flux::accounting::weighted_tree_node_t> std::make_shared<Flux::accounting::weighted_tree_node_t, std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool, long long, int>(std::shared_ptr<Flux::accounting::weighted_tree_node_t>&, std::string const&, bool&&, long long&&, int&&) (shared_ptr.h:604)
==53845==    by 0x40C3C9: get_sub_banks(sqlite3*, std::string const&, std::shared_ptr<Flux::accounting::weighted_tree_node_t>, sqlite3_stmt*, sqlite3_stmt*, sqlite3_stmt*) (load_accounting_db.cpp:97)
==53845==    by 0x40D622: load_accounting_db(std::string const&) (load_accounting_db.cpp:382)
==53845==    by 0x402326: main (test_load_accounting_db.cpp:41)
==53845== 
==53845== LEAK SUMMARY:
==53845==    definitely lost: 136 bytes in 1 blocks
==53845==    indirectly lost: 1,911 bytes in 25 blocks
==53845==      possibly lost: 0 bytes in 0 blocks
==53845==    still reachable: 0 bytes in 0 blocks
==53845==         suppressed: 0 bytes in 0 blocks
==53845== 
==53845== For lists of detected and suppressed errors, rerun with: -s
==53845== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
cmoussa1: Sorry, I think I might need some explanation here. How does one typically check for `nullptr` when calling a recursive function?

src/fairness/weighted_tree/weighted_tree.cpp
SteVwonder: > I vaguely remember there was some memory leak issues with using the shared pointer.

Ah! Good call.  When all references to the root tree vertex are gone, we want the whole tree to be destructed, but if the children have shared pointers to their parent, then we have a circular reference (parent -> children and children -> parent). 

> How is this typically handled?

It depends on what is causing the memory leak.  It could be a missed free or bad destructor.  In this case, it looks to be a circular reference.  One way to handle this one is to use a `weak_ptr` for the references that children have to their parent:
```diff
diff --git a/src/fairness/weighted_tree/weighted_tree.cpp b/src/fairness/weighted_tree/weighted_tree.cpp
index a1ab4c7..7e27b0b 100644
--- a/src/fairness/weighted_tree/weighted_tree.cpp
+++ b/src/fairness/weighted_tree/weighted_tree.cpp
@@ -87,18 +87,18 @@ void weighted_tree_node_t::calc_set_children_weight ()

 void weighted_tree_node_t::propagate_subtree_size ()
 {
-    if (m_parent == nullptr)
+    if (m_parent.expired())
         return;
-    m_parent->m_subtree_size++;
-    m_parent->propagate_subtree_size ();
+    m_parent.lock()->m_subtree_size++;
+    m_parent.lock()->propagate_subtree_size ();
 }

 void weighted_tree_node_t::propagate_subtree_leaf_size ()
 {
-    if (m_parent == nullptr)
+    if (m_parent.expired())
         return;
-    m_parent->m_subtree_leaf_size++;
-    m_parent->propagate_subtree_leaf_size ();
+    m_parent.lock()->m_subtree_leaf_size++;
+    m_parent.lock()->propagate_subtree_leaf_size ();
 }


@@ -150,7 +150,7 @@ std::shared_ptr<weighted_tree_node_t> weighted_tree_node_t
 std::shared_ptr<weighted_tree_node_t> weighted_tree_node_t
                                           ::get_parent () const
 {
-    return m_parent;
+    return m_parent.lock();
 }

 bool weighted_tree_node_t::is_tie_with_next () const
@@ -241,11 +241,11 @@ int weighted_tree_node_t::dprint_csv (std::ostringstream &out,
     int rc = 0;
     try {
         if (is_user ()) {
-            if (!m_parent) {
+            if (m_parent.expired()) {
                 errno = EINVAL;
                 return -1;
             }
-            out << level << "," << m_parent->get_name () << ","
+            out << level << "," << m_parent.lock()->get_name () << ","
                 << get_name () << "," << get_shares () << "," << get_usage ();
             if (long_format)
                 out << "," << get_fshare();
diff --git a/src/fairness/weighted_tree/weighted_tree.hpp b/src/fairness/weighted_tree/weighted_tree.hpp
index e4939ad..618ffc6 100644
--- a/src/fairness/weighted_tree/weighted_tree.hpp
+++ b/src/fairness/weighted_tree/weighted_tree.hpp
@@ -65,7 +65,7 @@ private:
     uint64_t m_subtree_leaf_size = 0;
     double m_weight = 0.0f;
     bool m_tie_with_next = false;
-    std::shared_ptr<weighted_tree_node_t> m_parent = nullptr;
+    std::weak_ptr<weighted_tree_node_t> m_parent = std::weak_ptr<weighted_tree_node_t>();
     std::vector<std::shared_ptr<weighted_tree_node_t>> m_children;
 };
```
I **think** that is the best way to handle this, but I'll defer to @dongahn as to what is best.

SteVwonder: PS - I didn't get a chance to run that through valgrind, so it may not completely solve the memory leak.

dongahn: > It depends on what is causing the memory leak. It could be a missed free or bad destructor. In this case, it looks to be a circular reference. One way to handle this one is to use a weak_ptr for the references that children have to their parent:

I remember using a raw pointer (as before) was the best I could come up with. I do remember I played with destructors and stuff but couldn't find a reasonable combination for this at all.

cmoussa1: Thanks for explaining this @SteVwonder and @dongahn - I went ahead and made the suggested changes and re-ran valgrind with my test program:

```
[moussa1@quartz386:test]$ valgrind --leak-check=yes ./mybinary 
==2277== Memcheck, a memory error detector
==2277== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==2277== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==2277== Command: ./mybinary
==2277== 
testing with small_no_tie.db:
leaf.3.1
leaf.3.2
leaf.2.1
leaf.2.2
leaf.1.3
leaf.1.1
leaf.1.2
==2277== 
==2277== HEAP SUMMARY:
==2277==     in use at exit: 0 bytes in 0 blocks
==2277==   total heap usage: 306 allocs, 306 frees, 168,237 bytes allocated
==2277== 
==2277== All heap blocks were freed -- no leaks are possible
==2277== 
==2277== For lists of detected and suppressed errors, rerun with: -s
==2277== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

So it looks like the leaks are gone now, so I can push those changes up.

src/fairness/weighted_tree/test/load_accounting_db.cpp
cmoussa1: I think I can do something like this:

```c++
if (get_sub_banks (DB,
                   b,
                   parent_bank,
                   b_select_shares_stmt,
                   b_select_sub_banks_stmt,
                   b_select_associations_stmt) == nullptr) {
    std::cerr << "get_sub_banks () returned a nullptr" << std::endl;
    return nullptr;
}
```

That way, if the recursive call returns a `nullptr`, an error message gets logged and a `nullptr` gets returned.

cmoussa1:
Thanks for giving this a review @dongahn - I pushed up a number of additional commits (starting at 
7319599) that I hope address your suggestions. The one I did not address yet is possibly breaking up the large `if` condition - it is not immediately clear to me how I might break that up to make it more readable just yet. I will keep thinking about it though in case you wanted to take a look at the other commits addressing the other fixes you suggested in the meantime.

cmoussa1:
Just pushed an additional commit that adds some helper functions to the large `if` condition in `get_sub_banks()` that I hope makes readability a little easier.

cmoussa1:
Force-pushed a series of changes based on the feedback I received from our flux-accounting meeting:

- changed the return value in `get_sub_banks ()` to just return `node` instead of `root_ptr`
- adjusted styling of `if-else` branches
- fixed indentation formatting (renamed variables to make them shorter helped with this)
- added a couple of helper functions to help reduce duplication in some parts of code

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -246,194 +254,133 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
...
     int rc = 0;
 
-    // open FluxAccounting DB in read-write mode; if it does not exist yet,
-    // create a new database file
+    std::string root_bank;
+    std::shared_ptr<weighted_tree_node_t> root = nullptr;
+
+    // open flux-accounting DB in read-write mode
     rc = sqlite3_open_v2 (path.c_str (), &DB, SQLITE_OPEN_READWRITE, NULL);
     if (rc) {
```
dongahn, line 321:
```suggestion
    if (rc != SQLITE_OK) {
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -246,194 +254,133 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
...
 
     sqlite3 *DB = nullptr;
     int rc = 0;
 
-    // open FluxAccounting DB in read-write mode; if it does not exist yet,
-    // create a new database file
+    std::string root_bank;
+    std::shared_ptr<weighted_tree_node_t> root = nullptr;
+
+    // open flux-accounting DB in read-write mode
```
dongahn, line 319: Could you check if `sqlite3_` functions set `errno` and if not set `errno` at this level. This will be useful when this function gets into a public API. `errno= EIO` makes sense to me but other may have a better idea.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -246,194 +254,133 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
...
 
     // call recursive function
-    std::shared_ptr<weighted_tree_node_t> root = nullptr;
-    root = get_sub_banks (DB,
-                          root_bank,
-                          nullptr,
-                          b_select_shares_stmt,
-                          b_select_sub_banks_stmt,
-                          b_select_associations_stmt);
+    root = get_sub_banks (DB, root_bank, nullptr, c_shrs, c_sub_banks, c_assoc);
```
dongahn, line 378: Perhaps we should check `root == nullptr`

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -112,50 +156,42 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
                             sqlite3 *DB,
                             const std::string &bank_name,
                             std::shared_ptr<weighted_tree_node_t> parent_bank,
-                            sqlite3_stmt *b_select_shares_stmt,
-                            sqlite3_stmt *b_select_sub_banks_stmt,
-                            sqlite3_stmt *b_select_associations_stmt)
+                            sqlite3_stmt *c_shrs,
```
dongahn, line 159: I like this function is now returning the root of the subtree, not the tree. We should revise the comment accordingly. 

```diff
> It returns a shared pointer to the root of the tree.
< It returns a shared pointer to the root of the subtree.
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -112,50 +156,42 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
...
+                            sqlite3_stmt *c_sub_banks,
+                            sqlite3_stmt *c_assoc)
 {
 
     std::shared_ptr<weighted_tree_node_t> node = nullptr;
     std::shared_ptr<weighted_tree_node_t> user_node = nullptr;
```
dongahn, line 165: I believe `user_node` is used incorrectly. Passing `nullptr` to `add_assoc` later is not used within that function. You should just remove this variable. I will leave another comment to the callosities of `add_assoc`.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -33,63 +33,107 @@
...
+add_assoc () constructs a weighted_tree_node_t object out of a
+association's data and adds it to the weighted tree.
 */
-void add_child_to_tree (std::shared_ptr<weighted_tree_node_t> parent_bank,
-                        const std::string &username,
-                        const std::string &user_shares,
-                        const std::string &user_job_usage,
-                        std::shared_ptr<weighted_tree_node_t> user_node,
-                        std::shared_ptr<weighted_tree_node_t> node)
+void add_assoc (const std::string &username,
```
dongahn, line 68:
```suggestion
int add_assoc (const std::string &username,
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -33,63 +33,107 @@
...
-                                                username,
-                                                true,
-                                                std::stoll (user_shares),
-                                                std::stoll (user_job_usage));
+    user_node = std::make_shared<weighted_tree_node_t> (node,
+                                                        username,
+                                                        true,
+                                                        std::stoll (shrs),
+                                                        std::stoll (usage));
     node->add_child (user_node);
```
dongahn, line 80:
```suggestion     
    return node->add_child (user_node);
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -167,75 +203,47 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
...
-                                   username,
-                                   user_shares,
-                                   user_job_usage,
-                                   user_node,
-                                   node);
-
-                // add single user's job usage to their bank's total job usage
-                leaf_bank_usage += std::stoi (user_job_usage);
+                // add association as a child of the node
+                add_assoc (username, shrs, usage, user_node, node);
```
dongahn, line 244:
```suggestion
                if (add_assoc (username, shrs, usage, node) < 0) {
                    // error handling
                }

File: src/fairness/weighted_tree/test/load_accounting_db.cpp [hunk 60]
```diff
@@ -33,63 +33,107 @@
...
-void add_child_to_tree (std::shared_ptr<weighted_tree_node_t> parent_bank,
-                        const std::string &username,
-                        const std::string &user_shares,
-                        const std::string &user_job_usage,
-                        std::shared_ptr<weighted_tree_node_t> user_node,
-                        std::shared_ptr<weighted_tree_node_t> node)
+void add_assoc (const std::string &username,
+                const std::string &shrs,
+                const std::string &usage,
+                std::shared_ptr<weighted_tree_node_t> user_node,
```
dongahn, line 71:
```suggestion
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -33,63 +33,107 @@
...
+                std::shared_ptr<weighted_tree_node_t> node)
```
dongahn, line 72:
```suggestion
                std::shared_ptr<weighted_tree_node_t> &node)
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -33,63 +33,107 @@
...
 {
     // add user as a child of the node
-    user_node = std::make_shared<weighted_tree_node_t> (
-                                                parent_bank,
-                                                username,
-                                                true,
-                                                std::stoll (user_shares),
-                                                std::stoll (user_job_usage));
+    user_node = std::make_shared<weighted_tree_node_t> (node,
```
dongahn, line 75:
```suggestion
    auto user_node = std::make_shared<weighted_tree_node_t> (node,
```

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -112,50 +156,42 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
...
+                            sqlite3_stmt *c_assoc)
 {
 
     std::shared_ptr<weighted_tree_node_t> node = nullptr;
     std::shared_ptr<weighted_tree_node_t> user_node = nullptr;
-    std::shared_ptr<weighted_tree_node_t> root_ptr = nullptr;
 
-    int rc = 0;
-    int leaf_bank_usage = 0;
+    int rc, bank_usg = 0;
```
dongahn, line 167: It would be a bit better to declare `bank_usg` where it is used like line #225?

dongahn (changes requested):
@cmoussa1: I like the way you restricted the code now! I added a few suggestions. Once those are addressed, this can go in I believe. Nice work!

src/fairness/weighted_tree/test/load_accounting_db.cpp
cmoussa1: Thanks for clarifying this during our team meeting. I think I understand what you are suggesting - setting something like `errno = EIO` in case the SQLite database cannot be opened. I can push up a `fixup` commit that sets this.

dongahn: @garlick or @SteVwonder: Is there any other "catch-all" errno which works better with sqlite function failures like this?

cmoussa1:
Thanks for reviewing this and for your feedback @dongahn - I pushed up a number of `fixup` commits related to your suggestions - if they look good, I can squash them up.

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -379,6 +379,8 @@ std::shared_ptr<weighted_tree_node_t> load_accounting_db (
 
     // call recursive function
     root = get_sub_banks (DB, root_bank, nullptr, c_shrs, c_sub_banks, c_assoc);
+    if (root == nullptr)
+        return nullptr;
```
dongahn, line 385: Hmm. If you return right away, I think some of the prepared statements would leak. I see some return statements above as well. You may want to make sure no resources are leaked?

src/fairness/weighted_tree/test/load_accounting_db.cpp
```diff
@@ -323,6 +324,7 @@ std::shared_ptr<weighted_tree_node_t> load_accounting_db (
     rc = sqlite3_open_v2 (path.c_str (), &DB, SQLITE_OPEN_READWRITE, NULL);
     if (rc != SQLITE_OK) {
         std::cerr << "error opening DB: " << sqlite3_errmsg (DB) << std::endl;
+        errno = EIO;
```
dongahn, line 327: Sorry I didn't leave the same comment for other places where sqlite functions are used and failed. You may want to add errno for those calls as well.

dongahn (approved):

src/fairness/weighted_tree/test/load_accounting_db.cpp
cmoussa1: Ah, sorry about that. 

I am looking at the different error codes available to return; I'm not sure `EOI` would be a good errno to use for the other `sqlite3_*` function calls. Does `EPROTO` make more sense to use in those other spots of code?

dongahn: Or EINVAL where it makes better sense...

cmoussa1:
Thanks @dongahn - I pushed up a fixup commit (`df8209c`) that makes two changes:

- I set `EPROTO` and `EINVAL` in the spots where `sqlite3_*` errors were being handled

- in `load_accounting_db()`, I added a `done` label statement to close the DB connection and destroy the prepared SQL statements in case an error is caught during the function. I think this should ensure that if an error occurs, the database connection is not left open and the prepared SQL statements are not destroyed.

dongahn:
Changes look good to me. Thanks.

cmoussa1:
Thanks! I'll go ahead and squash and set MWP.

PR #103, opened 2021-02-10
weighted_tree: make load_accounting_db public

### LINKED ISSUE #102: Refactor weighted tree loading to be "public"
#97 adds the ability to create a populated weighted tree from an accounting DB for testing purposes.  That functionality will be required moving forward to handle #63.  To make this easier, we should pull the `load_accounting_db.*` out of the `test` directory and up a level into `weighted_tree`.

cmoussa1:
_note: this PR is built on top of #97 and should not be reviewed/merged until that one lands. I'll keep it as a draft PR for now._

PR #97 adds the ability to create a populated weighted tree from a flux-accounting DB for testing purposes. That functionality will be required moving forward to handle #63.

---

To make this easier, this PR pulls the `load_accounting_db.*` files out of the "test" directory and up a level into `weighted_tree/`.

Fixes #102


dongahn:
Just a quick comment: did you want to put the new public function into the `Flux:: accounting` namespace?

cmoussa1:
Hm, I'm not sure - I hadn't thought about it. My idea in #104 was to leverage the code written in #97 and use it to create the hierarchical output. Would it then make sense to add the new public function to the namespace?

cmoussa1:
OK, now that #97 has landed, I think this is ready for review.

dongahn:
@cmoussa1: I understand you want to create a public interface that will instantiate a weighted tree object with the given DB. I have two concerns: 1) by placing this file to the weighted_tree library code, we are breaking a good separation of concern. I think it perhaps makes better  sense to keep `libweighted_tree` only contains the code that concerns "fairness algorithm code" not the source of the data; 2) the load source files are really isn't shaped well as a public API. If nothing else, It has a bunch of internal codes with global linkage etc. 

I'd suggest the following. 
1) Add a reader directory under fairness
2) Add DB-based reader as a C++ class with a clean set public APIs. This should include adding error message support which can be used to capture the error messages from sqlite as we discussed at our team meeting yesterday. For the general OOP design pattern, I believe you can model after our resource reader structure for Fluxion -- https://github.com/flux-framework/flux-sched/tree/master/resource/readers)
3) Add the DB-based unit test cases in that "reader" directory 

Hope this makes sense.

cmoussa1:
Thanks for your feedback and suggestions @dongahn - I pushed up a couple of changes that moves the load-DB code and its unit test file and sample DB's into their own directory under `fairness/`, called `reader/`. I also went ahead and changed the names of the source files to `data_reader_db.*`.

> This should include adding error message support which can be used to capture the error messages from sqlite as we discussed at our team meeting yesterday.

I think I could use some more clarification here. Are you thinking that this code could add additional context when an error occurs while interacting with the SQLite database? To the best of my knowledge, the `sqlite3_*` error message will be captured and sent to `cerr` and the appropriate `errno` will be set, such as in `get_sub_banks()`:

```c++
rc = sqlite3_bind_text (c_shrs, 1, bank_name.c_str (), -1, NULL);
if (rc != SQLITE_OK) {
    std::cerr << sqlite3_errmsg (DB) << std::endl;
    errno = EPROTO;

    return nullptr;
}
```

But maybe there is something more this code could do that I don't know about. Maybe the `"Unable to fetch data"` error messages could use more context, since there are multiple locations where a failed data fetch could occur.

dongahn:
@cmoussa1: please see the error message string support [here](https://github.com/flux-framework/flux-sched/blob/master/resource/readers/resource_reader_base.hpp#L122).

If you have something like that in the reader base class,  you can change the above code:

```c++
rc = sqlite3_bind_text (c_shrs, 1, bank_name.c_str (), -1, NULL);
if (rc != SQLITE_OK) {
    m_err_msg += sqlite3_errmsg (DB) + std::string (" ");
    errno = EPROTO;
    return nullptr;
}
```

This way, if the use code of the reader object wants to print out an error message, they can by doing something like [this code](https://github.com/flux-framework/flux-sched/blob/68d1ee5a9a135d3b13fe9047d167247f7be97848/resource/modules/resource_match.cpp#L524)

Each user of your code would have its own preferred way to log messages and it won't be pretty if your public API prints out things in stderr/stdout.

Also as I see this, I think `EINVAL` would be a better catch-all error number than `EPROTO`...

src/fairness/reader/data_reader_base.cpp
```diff
@@ -0,0 +1,39 @@
+/*****************************************************************************\
+ *  Copyright (c) 2020 Lawrence Livermore National Security, LLC.  Produced at
```
SteVwonder, line 2: `2020` -> `2021`

src/fairness/reader/data_reader_base.cpp
```diff
@@ -0,0 +1,39 @@
...
+ *  Flux is distributed in the hope that it will be useful, but WITHOUT
+ *  ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY or
+ *  FITNESS FOR A PARTICULAR PURPOSE.  See the terms and conditions of the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.
+ *  See also:  http://www.gnu.org/licenses/
+\*****************************************************************************/
```
SteVwonder, line 23: We used to use this large header, but flux-core switched to a more minimal header.  I think we should preferentially use this one: https://github.com/flux-framework/flux-core/blob/master/src/common/libflux/attr.h

src/fairness/reader/data_reader_base.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace Flux {
+namespace reader {
+
+/*!  Base data reader class.
+*/
+class data_reader_base_t {
+public:
+    virtual ~data_reader_base_t () = default;
+
+    std::shared_ptr<weighted_tree_node_t> load_accounting_db (
```
SteVwonder, line 38: Maybe consider making this name a little more generic to support reading from csvs and other file formats?

src/fairness/reader/data_reader_db.cpp
```diff
@@ -196,13 +204,17 @@ std::shared_ptr<weighted_tree_node_t> get_sub_banks (
                                                        0);
     }
     catch (const std::invalid_argument &ia) {
-        std::cerr << "Invalid argument: " << ia.what() << std::endl;
+        std::stringstream msg;
+        msg << "Invalid argument: " << ia.what ();
+        m_err_msg += msg.str ();
```
SteVwonder, line 209: `m_err_msg += "Invalid argument: " + std::string (ia.what ());`

src/fairness/reader/data_reader_db.cpp
```diff
@@ -343,7 +364,9 @@ std::shared_ptr<weighted_tree_node_t> load_accounting_db (
     // open flux-accounting DB in read-write mode
     rc = sqlite3_open_v2 (path.c_str (), &DB, SQLITE_OPEN_READWRITE, NULL);
     if (rc != SQLITE_OK) {
-        std::cerr << "error opening DB: " << sqlite3_errmsg (DB) << std::endl;
+        std::stringstream msg;
+        msg << "error opening DB: " << sqlite3_errmsg (DB);
+        m_err_msg += msg.str ();
```
SteVwonder, line 369: Per @dongahn 's verbal comment, the `+=` is useful for recursive functions, but for single-invocation functions, using `=` is preferred to ensure stale errors are not included.

src/fairness/reader/test/data_reader_db_test01.cpp
```diff
@@ -36,14 +36,16 @@
...
 
 using namespace Flux::accounting;
+using namespace Flux::reader;
 
 static void test_fairshare_order (const std::string &filename,
                                   const std::vector<std::string> &expected)
 {
     bool bo = true;
     std::shared_ptr<weighted_tree_node_t> root;
+    data_reader_db_t data_reader;
```
SteVwonder, line 46: We only have one derived class so far, so this is probably fine, but once we have additional derived classes, we will probably need a factory class like in `resource`'s `reader` directory.

cmoussa1: Ah, thanks for pointing this out. What do you think about me posting a follow-up PR after this one that adds a factory class? I also don't mind adding another commit that adds one if we'd like to have it available from the start.

cmoussa1:
Thanks for your review @SteVwonder! I pushed up a couple of `fixup` commits that I hope address your suggestions:

- changed the header to a more minimal header (along with the year)
- renamed the function definition in **data_reader_base.hpp** to just `load_accounting_info`
- changed the assignment to `m_err_msg` to be just an assignment (`=`) for single invocation functions, but an append (`+=`) for recursive functions.

src/fairness/reader/test/data_reader_db_test01.cpp
SteVwonder: > What do you think about me posting a follow-up PR after this one that adds a factory class? I can open an issue to track this.

Sounds good to me.

SteVwonder (commented):
Generally LGTM! Just one comment on the changes.  Thanks @cmoussa1!

src/fairness/reader/data_reader_db.cpp
```diff
@@ -118,13 +112,13 @@ int reset_and_clear_bindings (sqlite3 *DB,
     for (sqlite3_stmt *stmt : stmts) {
         rc = sqlite3_clear_bindings (stmt);
         if (rc != SQLITE_OK) {
-            std::cerr << sqlite3_errmsg (DB) << std::endl;
+            m_err_msg = sqlite3_errmsg (DB);
```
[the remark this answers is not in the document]
cmoussa1: Ah, yes you're right, sorry about that. It does get called by `get_sub_banks`. 

I just pushed an update that keeps it at `+=`. I also went ahead and squashed the `fixup` commits - hope that is okay!

SteVwonder (approved):
:+1: LGTM! Thanks @cmoussa1!

cmoussa1:
Thank you! I'll go ahead and set MWP.

dongahn:
@cmoussa1: just a quick thing. For error message where you use `+=`, you would need trailing whitespace or newline so if multiple messages will be separated properly. Sorry I jumped in late.

cmoussa1:
Thanks for pointing that out! I will post an update that adds the trailing whitespace before I set MWP.

dongahn:
@cmoussa1: thanks Chris. I think a newline would be a bit better for users. The user of this API can more easily split the error message using the newline as the token and decide what to do (e.g., printing only the last message, finding uniq messages etc...)

dongahn (approved):

cmoussa1:
Thank you! 🎉

Issue #82, opened 2020-12-08
ci: run `make check` as part of github actions


SteVwonder:
Right now, the github actions mainly validate commit messages and run linters.  The integration tests are not run and their status is not collected.

cmoussa1:
We should also add valgrind to any commits that add/change C++ code to the repo as well.

SteVwonder:
I believe this was addressed by #107. Closing.

cmoussa1:
Thanks for closing this manually! I thought I had this included in the commit message:

```
Fixes #59, #82
```

I guess that format is not supported for closing issues anymore? Just took a look at [this](https://stackoverflow.com/questions/3547445/closing-multiple-issues-in-github-with-a-commit-message) Stack Overflow thread.

PR #106, opened 2021-02-24
python: restructure python bindings, fix miscellaneous python errors

cmoussa1:
Working on #82 yesterday revealed a number of issues I hadn't noticed that the flux-accounting repo had with Python version 3.6, so I decided to first work on a restructure of the Python bindings first before I continue with adding `make` and `make check` to the CI.

There were a couple of small issues with some of the Python unit tests - rounding errors and naming conflicts with test DB files. Those should be addressed in the first couple commits.

The significant change that this PR brings is changing the structure of the Python bindings. The current structure could cause problems when running tests with Python version `3.6`. `src/bindings/python/flux` didn't appear to be a module because the module loading system didn't see an `__init__.py`. If I added an `__init__.py`, it could also end up breaking the `import` of other flux modules from the system path:

```
bash-4.4$ PYTHONPATH=/usr/src/src/bindings/python/:/usr/src/t/pycotap python3  -c 'import flux.constants'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'flux.constants'
```

---

This PR changes the Python bindings to be a namespace package called `fluxacct`, which itself holds a subdir called `accounting` that holds the current Python source code. This allows splitting sub-packages and modules within a single package (for now, there is just one, but perhaps in the future we might see the need to add another package). 

It also moves the front-end user command, `flux-account.py` out of this package and into `src/cmd/`. 

---

One question I had while working on this was about the following section of the `configure.ac` file:

```
AS_VAR_SET(fluxpydir, $pyexecdir/flux)
AC_SUBST(fluxpydir)
AS_VAR_SET(acctpydir, $fluxpydir/accounting)
AC_SUBST(acctpydir)
```

I'm not exactly sure what this section is doing. I left it alone for now, but I'm wondering if it also needs to be changed as a result of changing the structure of the Python bindings. Building the flux-accounting repo through `docker-run-checks.sh` and on an LC machine succeeds, so maybe no change is necessary, but I figured I'd double check.


src/bindings/python/fluxacct/accounting/Makefile.am
```diff
@@ -6,7 +6,7 @@ acctpy_PYTHON = \
 	print_hierarchy.py
 
 dist_fluxcmd_SCRIPTS = \
-	flux-account.py
+	../../../../cmd/flux-account.py
```
SteVwonder, line 9: I think it might make more sense to delete these two lines from this `Makefile.am` and add them to the `Makefile.am` in `src/cmd`.

src/bindings/python/flux/accounting/job_archive_interface.py
```diff
@@ -444,7 +444,7 @@ def calc_usage_factor(
     )
     acct_conn.commit()
 
-    return usage_user_historical
+    return round(usage_user_historical, 1)
```
SteVwonder, line 447: Do you know which part of the test is non-deterministic (or maybe time-dependent) which could be causing the fluctuation at smaller decimal values?

SteVwonder (commented):
Thanks @cmoussa1!

src/bindings/python/flux/accounting/job_archive_interface.py
cmoussa1: I _think_ the fluctuation could be here, where the historical job usage factor has a decay factor applied to it:

```python
usage_factor * math.pow(decay_factor, power)
```

Do you think it would be better to apply the rounding here?

cmoussa1: Ah, okay. After some more debugging I think I found it.

When a per-job factor is calculated, it takes two inputs: `job.nnodes` and `job.elapsed`:

```python
for job in job_totals_user:
    per_job_factors.append(round((job.nnodes * job.elapsed), 5))
```

Most of the tests use `mock.patch()` to simulate a period of time, which means this round to the 5th decimal place shouldn't affect a factor's value. However, the two tests that were consistently failing were required to use `time.time()` inside of `mock.patch()` to simulate a half-period further, thus effecting the value of `job.elapsed`, which in turn effects the historical usage factor. 

So I think this is what is causing the fluctuation. Maybe instead of rounding the final value and affecting `calc_usage_factor()`, I can instead round the `test` value in the unit test itself?

```python
self.assertEqual(round(usage_factor, 1), 4519.0)
```

SteVwonder: Awesome sleuthing @cmoussa1!

> So I think this is what is causing the fluctuation. Maybe instead of rounding the final value and affecting calc_usage_factor(), I can instead round the test value in the unit test itself?

That definitely seems like an improvement.

> However, the two tests that were consistently failing were required to use time.time() inside of mock.patch() to simulate a half-period further, thus effecting the value of job.elapsed, which in turn effects the historical usage factor.

Is it possible to use the mock patch to deterministically simulate the movement of time into the next half-period?  For example, maybe the mock return value for `time.time` can be changed within the test using [`Mock.return_value`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.return_value) or by passing a list of times to [`Mock.side_effect`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect).

cmoussa1: > Is it possible to use the mock patch to deterministically simulate the movement of time into the next half-period?

Ah, that's a good point - yes I can. At first I thought no, because the time period of the current half-life period is set when the database is initially created using `time.time()`, but I forgot to realize I can also simulate that as well in the unit test file with an update statement:

```python
# simulate end of half life period in FluxAccounting database
update_stmt = """
    UPDATE t_half_life_period_table SET end_half_life_period=?
    WHERE cluster='cluster'
    """
    acct_conn.execute(update_stmt, ("10000000",))
    acct_conn.commit()
```

Now, I think that none of the unit tests in **test_job_archive_interface.py** should be dependent on `time.time()`. I'll go ahead and push up a commit that integrates this change. If it looks good to you, I'll go ahead and squash this one and the previous commit.

SteVwonder (commented):
Thanks @cmoussa1.  These changes look great to me!

Like we discussed during the ☕ hour, if you could post the contents of the install prefix after running `make install` as a sort of "check" for the changes, that would be awesome!

cmoussa1:
Sure thing @SteVwonder! This on the Docker container where I am installing alongside a system install of flux-core.

Here is where `flux-account.py` gets installed:

```
[moussa1@docker-desktop src]$ ls -l /usr/libexec/flux/cmd/
total 3144
-rwxr-xr-x 1 root root  399440 Feb 22 17:35 flux-R
-rwxr-xr-x 1 root root    9177 Mar  3 15:59 flux-account.py <-- flux-accounting front end
-rwxr-xr-x 1 root root    1739 Feb 22 17:35 flux-admin.py
-rwxr-xr-x 1 root root 1018352 Feb 22 17:35 flux-broker
-rwxr-xr-x 1 root root   16392 Feb 22 17:35 flux-cron
-rwxr-xr-x 1 root root   55120 Feb 22 17:35 flux-event
-rwxr-xr-x 1 root root  136944 Feb 22 17:35 flux-exec
-rwxr-xr-x 1 root root  333696 Feb 22 17:35 flux-job
-rwxr-xr-x 1 root root    9552 Feb 22 17:35 flux-jobs.py
-rwxr-xr-x 1 root root    7974 Feb 22 17:35 flux-jobspec.py
-rwxr-xr-x 1 root root    2597 Feb 22 17:35 flux-jobtap.py
-rwxr-xr-x 1 root root   31672 Feb 22 17:35 flux-keygen
-rwxr-xr-x 1 root root  410032 Feb 22 17:35 flux-kvs
-rwxr-xr-x 1 root root   51264 Feb 22 17:35 flux-logger
-rwxr-xr-x 1 root root   51462 Feb 22 17:35 flux-mini.py
-rwxr-xr-x 1 root root   84336 Feb 22 17:35 flux-module
-rwxr-xr-x 1 root root   62072 Feb 22 17:35 flux-ping
-rwxr-xr-x 1 root root   54144 Feb 22 17:35 flux-queue
-rwxr-xr-x 1 root root   14110 Feb 22 17:35 flux-resource.py
-rwxr-xr-x 1 root root  158352 Feb 22 17:35 flux-start
-rwxr-xr-x 1 root root  263112 Feb 22 17:35 flux-terminus
```

And the result of `flux account -h`:

```
[moussa1@docker-desktop src]$ flux account -h
usage: flux-account.py [-h] [-p PATH] [-o OUTPUT_FILE]
                       {view-user,add-user,delete-user,edit-user,view-job-records,create-db,add-bank,view-bank,delete-bank,edit-bank,print-hierarchy}
                       ...

Flux Accounting Database.

positional arguments:
  {view-user,add-user,delete-user,edit-user,view-job-records,create-db,add-bank,view-bank,delete-bank,edit-bank,print-hierarchy}
                        sub-command help
    view-user           view a user's information in the accounting database
    add-user            add a user to the accounting database
    delete-user         remove a user from the accounting database
    edit-user           edit a user's value
    view-job-records    view job records
    create-db           create the flux-accounting database
    add-bank            add a new bank
    view-bank           view bank information
    delete-bank         remove a bank
    edit-bank           edit a bank's allocation
    print-hierarchy     print accounting database

optional arguments:
  -o OUTPUT_FILE, --output-file OUTPUT_FILE
                        specify location of output file
```

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

PR #107, opened 2021-02-25
github: add make, make check to GitHub Actions

cmoussa1:
_note: this PR is built on top of #106._

Currently, the flux-accounting repo's GitHub Actions performs checks on the formatting of the Python source code as well as the formatting of the commit messages, but not much else, noted in #59 and #82. It does not run the tests in this repo or try to build it.

---

This PR adds a build/`make check` step to the GitHub Actions for this repo. It pulls a lot from the `docker-run-checks.sh`, `generate-matrix.py`, `checks-lib.sh`, and `checks_run.sh` files from the flux-sched repo, and builds against two different Dockerfiles: `centos8` and `bionic`. I am leaving this as [WIP] for now because I'm not sure how close I am to actually getting this right. 😅 

One question that came up while I was working on this: I had some trouble adding another build to the matrix that configured with python3.7. I tried adding another build in `generate-matrix.py` like the following:

```python
# Centos8
matrix.add_build(
    name="centos8 - py3.7",
    image="centos8",
    env=dict(PYTHON_VERSION="3.7"),
    docker_tag=True,
)
```

But the build fails. Do I need to install python3.7 somewhere in the `centos8` Dockerfile before I can add a build to configure against Python 3.7?


grondo:
> One question that came up while I was working on this: I had some trouble adding another build to the matrix that configured with python3.7

The `bionic` base image should have python3.7 installed, so could you try with that one instead of centos8?

cmoussa1:
Thanks @grondo! I think I was able to get that working - I added another build based off the `bionic` base image.

It looks like Codecov is reporting a huge drop in coverage for the Python bindings though... I'll have to take a look at this and see what's going on.

cmoussa1:
> It looks like Codecov is reporting a huge drop in coverage for the Python bindings though... I'll have to take a look at this and see what's going on.

We discussed this a bit during coffee hour, and it actually looks like the Python unit tests weren't being picked up by Codecov before (only the C++ tests), thus the large decrease in coverage. The two files that need a closer look are **accounting_cli_functions.py** and **job_archive_interface.py**.

I can open an issue to track this and submit a follow-up PR to this one that increases the coverage for both of these files.

.github/workflows/main.yml
```diff
@@ -15,12 +15,12 @@ jobs:
     - run: git fetch origin master
     - uses: flux-framework/pr-validator@master
 
-
-  run-checks:
+  check-python-formatting:
+    name: check python formatting
```
grondo, line 19: It appears this job runs more than just python formatting (I'm not sure exactly what kind of style `flake8` enforces though, so apologies if I'm wrong), so maybe a more appropriate name would `python style and format`?

Also, I didn't see it, but are you running `pylint` as part of CI?

grondo (commented):
Just one minor comment for now, otherwise looks good. It will be easier to review once #106 is merged.

.github/workflows/main.yml
cmoussa1: > It appears this job runs more than just python formatting (I'm not sure exactly what kind of style `flake8` enforces though, so apologies if I'm wrong), so maybe a more appropriate name would be `python style and format`?

That seems like a more appropriate name - I can change it. Thanks for the feedback! 

> Also, I didn't see it, but are you running `pylint` as part of CI?

I am not running `pylint` as of now, but I can add it.

cmoussa1: Ah, sorry for the noise - I didn't know what `pylint` was but I just looked it up. Looks like it is also a style checker? Which is, to my understanding, very similar to what `flake8` does.

grondo: Ah, maybe @SteVwonder can comment since I'm not really familiar with the difference. However, in my mind it would be nice if we used the same version of pylint between flux-framework projects that use Python to ensure a kind of uniformity. However, that is just my opinion and I'm willing to go with whatever you and @SteVwonder decide.

SteVwonder: I don't really have a preference for pylint vs flake8 in general, but I agree with @grondo that it would be good to keep things consistent across flux-framework projects.

cmoussa1:
OK, now that #106 is merged, I think this is ready for review. I'll take it out of Draft status.

.mergify.yml
```diff
@@ -3,7 +3,11 @@ pull_request_rules:
     conditions:
       - base=master
       - status-success="validate commits"
-      - status-success="run-checks (3.7)"
+      - status-success="check python formatting (3.6)"
+      - status-success="check python formatting (3.7)"
```
grondo, line 7: I think these status names are now out  of date. There appears to be only one format status `python format (3.7)`

To check the status names, look at the list of names on the main PR page under `Show all checks`.

You may also want to add the `coverage` check as a required status.

.github/workflows/main.yml
```diff
@@ -15,8 +15,8 @@ jobs:
     - run: git fetch origin master
     - uses: flux-framework/pr-validator@master
 
-
-  run-checks:
+  python-format:
+    name: python format
```
grondo, line 19: Up to you, but since you are only using a single python version in this check, you could remove the use 
 of a `matrix` here to simplify things. If you do, then remember to update the name of the status check in mergify since it would no longer contain the python version.

grondo (commented):
Looks good!

I think the mergify status names just need to be updated to match the actual status check names, and then this would be ready to merge.

I noted some other optional improvements inline as well.

.github/workflows/main.yml
```diff
@@ -34,22 +34,81 @@ jobs:
...
+
+  ci-checks:
+    needs: [ generate-matrix ]
+    runs-on: ubuntu-latest
+    env:
+      TAP_DRIVER_QUIET: 1
+      FLUX_TEST_TIMEOUT: 300
+      DOCKER_REPO: fluxrm/flux-core
+      DOCKER_USERNAME: travisflux
+      DOCKER_PASSWORD: ${{ secrets.DOCKER_HUB_TRAVISFLUX_TOKEN }}
```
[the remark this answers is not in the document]
cmoussa1: Oh whoops! Thank you for pointing this out. I'll push up a change that removes those lines.

.github/workflows/main.yml
cmoussa1: I agree with you, I can just remove `matrix`. I just pushed up a change that should remove the use of `matrix` and change the name of the check in mergify to just `python format`.

grondo (approved):
@cmoussa1, this looks good now I think. Be sure to remove the `[WIP]` prefix from the issue description. Then I think someone will have to manually merge since the mergify config is edited.

cmoussa1:
Thanks for the review and feedback @grondo!

grondo:
If you're ready, I can push the merge button.

cmoussa1:
This should be ready to go! Thanks for pushing the button. 🙂

SteVwonder:
Awesome! 🥳 🎆 🎉 

Feels good to have CI on this repo now.

PR #109, opened 2021-03-10
print-hierarchy: add job usage values to output of flux-accounting DB hierarchy

cmoussa1:
We discussed this a little while ago, but as this repo matures, it feels like the C++ side of flux-accounting deals with the the actual hierarchy of a flux-accounting DB, its banks and users' weights and relationships with one another, and its sorting as it relates to fair share. The Python side deals with the administrative side of the flux-accounting database (adding/removing users and banks and editing user/bank information).

---

This PR makes two changes:

The first change removes the Python version of the `print-hierarchy` subcommand, along with its unit tests. 

The second adds job usage values to the output of the C++ version of `print-hierarchy`. I leveraged the code written in **data_reader_db.cpp** in #103 to load in a flux-accounting DB and parse/tokenize each line to create hierarchical output.

One thing I wasn't sure about was if we wanted to keep the same type of test for **print_hierarchy.cpp**. Before, we had a test that generated both C++ and Python output and compared the two (**t1000-print-hierarchy.t**). Now that we just have C++ output, for now, I kept the same style of test; instead of comparing it to Python output, I compared it to an expected `.txt` file. Is it best practice to keep it this style? Would it be better to write unit tests in the `print-hierarchy` directory?


cmoussa1:
Just pushed a couple of changes to `print_hierarchy/`:

**print_hierarchy.cpp** was changed to **flux_shares.cpp**. This file is then compiled into an `flux-shares` executable; my thought process is that `shares` could be picked up as a subcommand, i.e `flux shares`:

```console
[moussa1@docker-desktop src]$ flux shares -h
usage: flux shares [-P] [DB_PATH]
optional arguments:
	-h,--help		Show this help message
	-P,			Print the database hierarchy in a parsable format
```

`flux shares` has an optional argument `-P`, which when passed in will output the database hierarchy in a parsable format:

```console
[moussa1@docker-desktop src]$ flux shares -P /tmp/FluxAccounting.db 
Account|Username|RawShares|RawUsage
root||1000|133
 account1||1000|121
  account1|leaf.1.1|10000|100
  account1|leaf.1.2|1000|11
  account1|leaf.1.3|100000|10
 account2||100|11
  account2|leaf.2.1|100000|8
  account2|leaf.2.2|10000|3
 account3||10|1
  account3|leaf.3.1|100|0
  account3|leaf.3.2|10|1
```

If `-P` is not passed, it will print the hierarchy in column format:

```console
[moussa1@docker-desktop src]$ flux shares /tmp/FluxAccounting.db 
Account                         Username           RawShares            RawUsage
root                                                    1000                 133
 account1                                               1000                 121
  account1                      leaf.1.1               10000                 100
  account1                      leaf.1.2                1000                  11
  account1                      leaf.1.3              100000                  10
 account2                                                100                  11
  account2                      leaf.2.1              100000                   8
  account2                      leaf.2.2               10000                   3
 account3                                                 10                   1
  account3                      leaf.3.1                 100                   0
  account3                      leaf.3.2                  10                   1
```

The other optional argument that can be passed to `flux shares` is a path to the flux-accounting database. If no filepath is specified, `flux shares` will default to `/usr/var/FluxAccounting.db`, which is also the default path for all subcommands under `flux account`. This is set with a constant `std::string` variable named `DB_PATH`.

cmoussa1:
Just pushed an update that adds `x_ac_expand_install_dirs.m4` to the `config` directory. This should allow me to use `X_LOCALSTATEDIR` for setting the path to `FluxAccounting.db` when configuring with a custom `localstatedir`.

cmoussa1:
I think this should be ready for your review whenever you get the chance. :-)

src/fairness/print_hierarchy/flux_shares.cpp
```diff
@@ -0,0 +1,156 @@
...
+print_hierarchy () takes two arguments: a filepath to a flux-accounting DB, and
+a boolean value 'parsable'. It will call load_accounting_db () to get a
+hierarchy of users. It will then parse each line, fetching tokens delimited by
+','. The formatting for lines will be different if a line contains a bank or if
+it contains an association.
+
+If parsable is false, the output of the database hierarchy will be
+formatted into columns.
+*/
+int print_hierarchy (const std::string &filename, bool parsable)
```
SteVwonder, line 65: Rather than parsing the output of the `dprint_csv` walker method, can you implement your own walker method?  Something like `print_csv` to print the parsable format (with an optional argument for the delimiter) and `print_pretty` for the human readable version?

src/fairness/print_hierarchy/flux_shares.cpp
```diff
@@ -0,0 +1,156 @@
...
+    int rc = 0;
+
+    for (int i = 1; i < argc; ++i) {
+        std::string arg = argv[i];
+        if (arg == "-h" || arg == "--help") {
+            show_usage ();
+            return rc;
+        } else if (arg == "-P")
+            parsable = true;
+    }
```
SteVwonder, line 145: Might be worth adding an `else` to print an error on an unknown argument.

src/fairness/print_hierarchy/Makefile.am
```diff
@@ -8,16 +8,21 @@ AM_CPPFLAGS = $(FLUX_CORE_CFLAGS)
 
 SUBDIRS = .
 
-noinst_PROGRAMS = print_hierarchy
+noinst_PROGRAMS = flux-shares
 
-print_hierarchy_SOURCES = \
-    print_hierarchy.cpp
+flux_shares_SOURCES = flux_shares.cpp \
+                      ../reader/data_reader_db.cpp
```
SteVwonder, line 14: Might be worth including the reader functions in the `libweighted_tree` shared library, then this inclusion would not be necessary.  And users of the installed `libweighted_tree` shared library could take advantage of the readers too.

SteVwonder (changes requested):
Thanks @cmoussa1!  Generally looks good.  The default path is a big improvement.  A few suggested changes below.  The big one is structuring the printing as a new walker method as opposed to a function that parses output from `dprint_csv`.

src/fairness/print_hierarchy/flux_shares.cpp
cmoussa1: I think I've got a solution. After some fiddling around, I ended up with two different functions: `print_csv ()` and `pretty_print ()`. Both of these start at the root node of a weighted tree and traverses it, printing all children. It should traverse the tree recursively until it reaches all leaf nodes. 

`print_csv ()` can take a custom delimiter if one is passed in via the `-P` option. I added a test to demonstrate this with `,`:

```
Account,Username,RawShares,RawUsage
root,,1000,133
 account1,,1000,120
  account1,leaf.1.1,10000,100
  account1,leaf.1.2,1000,10
  account1,leaf.1.3,100000,10
 account2,,100,12
  account2,leaf.2.1,10000,10
  account2,leaf.2.2,1000,1
  account2,leaf.2.3,100000,1
 account3,,10,1
  account3,leaf.3.1,100,0
  account3,leaf.3.2,10,1
```

src/fairness/print_hierarchy/flux_shares.cpp
cmoussa1: Ah, yeah, thank you for pointing this out. I pushed up a fix that should print the usage for `flux shares` and then sets `rc` to -1 if a bad argument is passed in.

src/fairness/print_hierarchy/flux_shares.cpp
```diff
@@ -0,0 +1,182 @@
...
+                  << std::setw(20) << std::right << ""
+                  << std::setw(20) << std::right << node->get_shares ()
+                  << std::setw(20) << std::right << node->get_usage ()
+                  << std::endl;
+    }
+
+    // recur on subtree
+    int num_children = node->get_num_children ();
+    if (num_children > 0) {
+        for (int i = 0; i < num_children; i++) {
```
SteVwonder, line 115: What is the intended order of printing for the accounts?  Should they be sorted alphabetically or by usage/priority?  If priority, do the vertices in this weighted-tree need to have `calc_and_sort_weighted_children` on them?  It isn't clear to me what the default traversal order is here (maybe the order that the DB returned them it?).

cmoussa1: The order for printing the accounts is based on the order they are fetched from `get_sub_banks()` in **data_reader_db.cpp**. After the root bank is found, subsequent sub banks are fetched via this SQLite statement:

```c++
s_sub_banks = "SELECT bank_table.bank FROM bank_table WHERE parent_bank=?";
```

And associations with this statement:

```c++
s_assoc = "SELECT association_table.username, association_table.shares, "
          "association_table.bank, association_table.job_usage "
          "FROM association_table WHERE association_table.bank=?"
          "ORDER BY association_table.username";
```

I believe SQLite does not order rows by default (in fact, I don't even think return order is guaranteed? 😬), but if `ORDER BY` is passed in to the statement, I think you can sort by one column, either in ascending or descending order, which I think applies to strings as well. Just doing a quick `sshare` on my machine, it looks like it sorts banks by alphabetical order? So I think I could make that change here. 

I think that would involve changing the `s_sub_banks` and `s_assoc` queries in **data_reader_db.cpp**. I tested this out, and I think I was able to get it working. Here is a super quick demo:

```console
[moussa1@docker-desktop src]$ flux account add-bank root 1
[moussa1@docker-desktop src]$ flux account add-bank --parent-bank=root C 1
[moussa1@docker-desktop src]$ flux account add-bank --parent-bank=root B 1
[moussa1@docker-desktop src]$ flux account add-bank --parent-bank=root A 1
[moussa1@docker-desktop src]$ flux shares
Account                         Username           RawShares            RawUsage
root                                                       1                   0
 A                                                         1                   0
 B                                                         1                   0
 C                                                         1                   0
[moussa1@docker-desktop src]$ flux account add-user --username=userA --bank=C
[moussa1@docker-desktop src]$ flux account add-user --username=userZ --bank=C
[moussa1@docker-desktop src]$ flux account add-user --username=userF --bank=C
[moussa1@docker-desktop src]$ flux shares
Account                         Username           RawShares            RawUsage
root                                                       1                   0
 A                                                         1                   0
 B                                                         1                   0
 C                                                         1                   0
  C                                userA                   1                   0
  C                                userF                   1                   0
  C                                userZ                   1                   0
```
So, above, I inserted three sub banks under `root` in reverse alphabetical order: **C**, **B**, **A**. As for the associations, the insert order was **userA**, **userZ**, **userF**. 

I can push up this change in an additional commit in this PR.

src/fairness/print_hierarchy/Makefile.am
cmoussa1: After some fiddling around, I think I've got it working. I pushed up some changes and squashed them into 1c30f35, adding **data_reader_db.*** to `libweighted_tree`.

src/fairness/print_hierarchy/flux_shares.cpp
SteVwonder: Awesome! Looks good.

SteVwonder: Can you add this out-of-order insertion and then in-order printing as a test case (if it isn't already covered by another test)?

cmoussa1: Sure. I went ahead and added this as another test database that **t1000-print-hierarchy.t** compares output with and pushed it up.

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

cmoussa1:
Thanks! :-) I'll go ahead and set MWP.

PR #113, opened 2021-04-06
fairness: update headers of source files

cmoussa1:
The headers for some of the files in the fairness directory need to be updated to the new minimal header and made consistent with the other files in this repo.

This PR updates the headers of some of the files in the `fairness` directory.

PR #114, opened 2021-04-06
docker: change module/dependency installation in bionic image, add a README

cmoussa1:
While working inside of the `bionic` Docker image the other day/this morning, I ran into some trouble trying to install flux-accounting and use it. More specifically, I continued to run into dependency issues both with the `accounting` module and the `pandas` dependency itself:

```console
moussa1@docker-desktop:/usr/src$ flux account -h
  File "/usr/libexec/flux/cmd/flux-account.py", line 15, in <module>
    import fluxacct.accounting
ModuleNotFoundError: No module named 'fluxacct'
```

After some fiddling around, I think I found that the answer included not adding the location of `/usr/lib/python3.7/site-packages/` to my `PYTHONPATH`. I think the `bionic` image could be improved by not having to perform these steps manually after the user starts up an interactive session.

---

This PR changes the `pandas` installation to install to the default Flux Python version on the bionic image, which is Python `3.6`. This matches the version on the other Docker image in flux-accounting, `centos8`. It also adds a new line in the `bionic` Dockerfile which adds `/usr/lib/python3.6/site-packages/` (the location of the `fluxacct.accounting` module) to `PYTHONPATH` so that the user does not have to add it themselves when running inside the container.

It also changes the Python version of the `coverage` build when running CI; the three builds, `bionic`, `centos8`, and `coverage` should now all configure and build with Python `3.6` by default.

The last thing I added was a `README` to the `docker` directory inside flux-accounting with instructions on how to build the images and run an interactive session, along with specific instructions on how to configure and build with a different version of Python if they so choose.

### DIFF
--- a/src/test/docker/bionic/Dockerfile
+++ b/src/test/docker/bionic/Dockerfile
@@ -18,7 +18,7 @@ RUN \
  fi
 
 RUN \
- sudo python3.7 -m pip install pandas==0.24.1
+ sudo pip3 install pandas==0.24.1
 
 USER $USER
 WORKDIR /home/$USER
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -137,7 +137,6 @@ matrix = BuildMatrix()
 # Ubuntu: coverage
 matrix.add_build(
     name="coverage",
-    env=dict(PYTHON_VERSION="3.7"),
     coverage=True,
     jobs=2,
 )
@@ -146,15 +145,13 @@ matrix.add_build(
 matrix.add_build(
     name="centos8 - py3.6",
     image="centos8",
-    env=dict(PYTHON_VERSION="3.6"),
     docker_tag=True,
 )
 
 # Bionic
 matrix.add_build(
-    name="bionic - py3.7",
+    name="bionic - py3.6",
     image="bionic",
-    env=dict(PYTHON_VERSION="3.7"),
     docker_tag=True,
 )
 
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -5,7 +5,7 @@ pull_request_rules:
       - status-success="validate commits"
       - status-success="python format"
       - status-success="centos8 - py3.6"
-      - status-success="bionic - py3.7"
+      - status-success="bionic - py3.6"
       - status-success="python lint"
       - status-success="coverage"
       - label="merge-when-passing"
--- a/configure.ac
+++ b/configure.ac
@@ -88,6 +88,13 @@ AC_SUBST(fluxrc1dir)
 AS_VAR_SET(fluxrc3dir, $sysconfdir/flux/rc3.d)
 AC_SUBST(fluxrc3dir)
 
+# Target of PYTHONPATH set by flux(1) cmddriver, so flux(1)
+# doesn't inadvertently insert system python paths (or any
+# other python path for that matter) first in PYTHONPATH.
+#
+AS_VAR_SET(fluxpylinkdir, $fluxlibdir/python$PYTHON_VERSION)
+AC_SUBST(fluxpylinkdir)
+
 AS_VAR_SET(fluxpydir, $pyexecdir/fluxacct)
 AC_SUBST(fluxpydir)
 AS_VAR_SET(acctpydir, $fluxpydir/accounting)
--- a/src/bindings/python/fluxacct/Makefile.am
+++ b/src/bindings/python/fluxacct/Makefile.am
@@ -1,1 +1,15 @@
 SUBDIRS = accounting
+
+install-data-hook:
+	$(AM_V_at)echo Linking python modules in non-standard location... && \
+	  $(INSTALL) -d -m 0755 "$(DESTDIR)$(fluxpylinkdir)" && \
+	  target=$(fluxpydir) && \
+	  f=$${target##*/} && \
+	  cd "$(DESTDIR)$(fluxpylinkdir)" && \
+	  rm -f $$f && \
+	  $(LN_S) $$target .
+
+uninstall-local:
+	$(AM_V_at)target=$(fluxpydir) && f=$${target##*/} && \
+	  echo "Removing $(fluxpylinkdir)/$$f" && \
+	  rm -rf $(fluxpylinkdir)/$$f
--- a/src/test/docker/README.md
+++ b/src/test/docker/README.md
@@ -1,0 +1,28 @@
+## Docker images for flux-accounting
+
+The Dockerfiles, resulting docker images, and the `docker-run-checks.sh` script contained herein are used as part of the strategy for CI testing of flux-accounting.
+
+Docker is used under CI to speed up deployment of an environment with correct build dependencies. flux-accounting builds against the latest or a tagged version of flux-core by including `FROM fluxrm/flux-core:` at the top of each Dockerfile.
+
+### Local Testing
+
+Developers can test the docker images themselves. If new dependencies are needed, they can update the `$image` Dockerfiles manually (where `$image` is one of `bionic` or `centos8`). To run inside a local Docker image, run the command:
+
+```console
+docker-run-checks.sh -i $image [options] -- [arguments]
+```
+
+### Interactive Testing
+
+While running in an interactive Docker container, you can build, test, and interact with flux-accounting. The `bionic` image has multiple versions of Python installed, which you can configure and build against (the default version for both images is Python version `3.6`).
+
+Remember to install the required dependencies before you build and add the appropriate install location to your `PYTHONPATH`. Below is an example of configuring and building against Python version `3.7` while running inside the Docker container.
+
+```console
+sudo python3.7 -m pip install pandas==0.24.1
+./autogen.sh
+PYTHON_VERSION=3.7 ./configure
+make
+make check
+export PYTHONPATH=$PYTHONPATH:/usr/lib/python3.7/site-packages/
+```


src/test/docker/bionic/Dockerfile
```diff
@@ -18,7 +18,10 @@ RUN \
  fi
 
 RUN \
- sudo python3.7 -m pip install pandas==0.24.1
+ sudo pip3 install pandas==0.24.1
+
+ENV \
+ PYTHONPATH=$PYTHONPATH:/usr/lib/python3.6/site-packages/
```
grondo, line 24: Is this `ENV` line really necessary (Shouldn't `/usr/lib/python3.6/site-packages` be in the default `PYTHONPATH` on bionic?)

src/test/generate-matrix.py
```diff
@@ -137,7 +137,7 @@ def __str__(self):
 # Ubuntu: coverage
 matrix.add_build(
     name="coverage",
-    env=dict(PYTHON_VERSION="3.7"),
+    env=dict(PYTHON_VERSION="3.6"),
```
grondo, line 140: if `PYTHON_VERSION` is not required by flux-accounting, I wonder if you can just drop this line? Since 3.6 is the default python3, it should be picked up by default by the flux-accounting configure?

src/test/generate-matrix.py
```diff
@@ -152,9 +152,9 @@ def __str__(self):
 
 # Bionic
 matrix.add_build(
-    name="bionic - py3.7",
+    name="bionic - py3.6",
     image="bionic",
-    env=dict(PYTHON_VERSION="3.7"),
+    env=dict(PYTHON_VERSION="3.6"),
```
grondo, line 157: Same as above, maybe this line could actually be dropped?

grondo (approved):
LGTM. I just ask a few questions inline, but the questions are about redundant `ENV` and `env` lines, and you may have a good reason for keeping those...

src/test/docker/bionic/Dockerfile
cmoussa1: If I leave this line out, I get the following `ModuleNotFoundError` after running `sudo make install`:

```console
moussa1@docker-desktop:/usr/src$ flux account -h
  File "/usr/libexec/flux/cmd/flux-account.py", line 15, in <module>
    import fluxacct.accounting
ModuleNotFoundError: No module named 'fluxacct'
``` 

even though `fluxacct/accounting/` gets installed here:

```console
moussa1@docker-desktop:/usr/src$ ls -l /usr/lib/python3.6/site-packages/
total 12
drwxr-xr-x 3 root root 4096 Feb 25 16:33 _flux
drwxr-xr-x 6 root root 4096 Feb 25 16:33 flux
drwxr-xr-x 3 root root 4096 Apr  8 15:29 fluxacct
```

And if I `echo $PYTHONPATH` I don't seem to get anything:

```console
moussa1@docker-desktop:/usr/src$ echo $PYTHONPATH

moussa1@docker-desktop:/usr/src$
```

Only when I explicitly add that `site-packages` path to my `PYTHONPATH` do I get flux-accounting output. Maybe I am messing up somewhere else?

src/test/generate-matrix.py
cmoussa1: I can try this now!

src/test/docker/bionic/Dockerfile
grondo: I think you can see python's default path by something like:
```
python3 -c 'import sys; print(sys.path)'
```

cmoussa1: Ah, okay. Yes, when I run this, I see `/usr/lib/python3.6/site-packages`:

```console
moussa1@docker-desktop:/usr/src$ python3 -c 'import sys; print(sys.path)'
['', '/usr/src', '/usr/lib/python3.6/site-packages', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/usr/local/lib/python3.6/dist-packages', '/usr/lib/python3/dist-packages']
```

Interesting. I can't imagine the default version of `python` on `bionic` would affect anything right? 

```console
moussa1@docker-desktop:/usr/src$ python --version
Python 2.7.17
```

Since the version of `flux python` is `3.6`:

```console
moussa1@docker-desktop:/usr/src$ flux python --version
Python 3.6.9
```

---

FWIW, the default version of Python on `centos8`:

```console
[moussa1@docker-desktop src]$ python --version
Python 3.6.8
```

SteVwonder: I think the problem here is that the python packages are being installed to `site-packages` whilst ubuntu bionic wants/expects them to be installed in `dist-packages`, which also affects flux-core and is being tracked here: https://github.com/flux-framework/flux-core/issues/2617.  If so, I think we will need some autotools magic to make this work.

SteVwonder: Sorry, it is more subtle than my original post.  It is true that if things were installed in `dist-packages`, the module would be found, but flux-accounting is also missing the symlink thingy that we do in other projects so that the `flux` starter can safely add a path to PYTHONPATH without also pulling in a bunch of system modules:
```
herbein1@docker-desktop:/usr/src$ printenv PYTHONPATH
:/usr/lib/python3.6/site-packages/
herbein1@docker-desktop:/usr/src$ unset PYTHONPATH
herbein1@docker-desktop:/usr/src$ printenv PYTHONPATH
herbein1@docker-desktop:/usr/src$ python3 -c 'import sys; print(sys.path)'
['', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/usr/local/lib/python3.6/dist-packages', '/usr/lib/python3/dist-packages']
herbein1@docker-desktop:/usr/src$ flux python -c 'import sys; print(sys.path)'
['', '/usr/lib/flux/python3.6', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/usr/local/lib/python3.6/dist-packages', '/usr/lib/python3/dist-packages']
herbein1@docker-desktop:/usr/src$ ls /usr/lib/flux/python3.6
_flux  flux
herbein1@docker-desktop:/usr/src$ ls -la /usr/lib/flux/python3.6/
total 8
drwxr-xr-x 2 root root 4096 Feb 26 23:57 .
drwxr-xr-x 5 root root 4096 Feb 26 23:57 ..
lrwxrwxrwx 1 root root   38 Feb 26 23:57 _flux -> /usr/lib/python3.6/site-packages/_flux
lrwxrwxrwx 1 root root   37 Feb 26 23:57 flux -> /usr/lib/python3.6/site-packages/flux
```

So if we copy over that symlink thing from flux-core autotools, things should start working again in flux-accounting and we can punt on fixing #2617 for now.

cmoussa1: Thanks for your help @SteVwonder. I was able to add the autotools support (copied over from flux-core) this morning and have this symlink be generated when flux-accounting is built and installed. Here it is shown on `bionic`:

```console
moussa1@docker-desktop:/usr/src$ ls -l /usr/lib/flux/python3.6/
total 0
lrwxrwxrwx 1 root root 38 Feb 25 16:33 _flux -> /usr/lib/python3.6/site-packages/_flux
lrwxrwxrwx 1 root root 37 Feb 25 16:33 flux -> /usr/lib/python3.6/site-packages/flux
lrwxrwxrwx 1 root root 41 Apr  9 14:50 fluxacct -> /usr/lib/python3.6/site-packages/fluxacct
```

And I can run `flux account` commands after running `sudo make install` without having to manually append to my `PYTHONPATH`:

```console
moussa1@docker-desktop:/usr/src$ echo $PYTHONPATH

moussa1@docker-desktop:/usr/src$ flux account -h
usage: flux-account.py [-h] [-p PATH] [-o OUTPUT_FILE]
                       {view-user,add-user,delete-user,edit-user,view-job-records,create-db,add-bank,view-bank,delete-bank,edit-bank}
                       ...

Flux Accounting Database.

positional arguments:
  {view-user,add-user,delete-user,edit-user,view-job-records,create-db,add-bank,view-bank,delete-bank,edit-bank}
                        sub-command help
    view-user           view a user's information in the accounting database
    add-user            add a user to the accounting database
    delete-user         remove a user from the accounting database
    edit-user           edit a user's value
    view-job-records    view job records
    create-db           create the flux-accounting database
    add-bank            add a new bank
    view-bank           view bank information
    delete-bank         remove a bank
    edit-bank           edit a bank's allocation

optional arguments:
  -o OUTPUT_FILE, --output-file OUTPUT_FILE
                        specify location of output file
```

I also built and installed on the `centos8` image to make sure it worked there as well. So I'll go ahead and remove the manual append to `PYTHONPATH` in the `bionic` image and squash the changes made to **configure.ac** and **src/bindings/python/fluxacct/Makefile**.

cmoussa1:
OK, just pushed a change based on the feedback I got on how to add support for installing the flux-accounting package on the `bionic` image (squashed into commit f020b66), and also rebased to catch up after #113.

SteVwonder (approved):
Thanks @cmoussa1 ! LGTM!

SteVwonder:
@mergifyio refresh

SteVwonder:
Looks like this PR needs to update the Mergify config too. It is looking for `bionic - py 3.7` but this PR changes the name to 3.6. Then I think we will need to push the button manually.

cmoussa1:
Thanks for letting me know! I just pushed a commit that changes the name for the `bionic` check in **.mergify.yml** to `bionic - py3.6`.

SteVwonder:
LGTM! Thanks @cmoussa1

SteVwonder:
@mergifyio rebase

Issue #115, opened 2021-04-08
weighted tree library: breakdown of OUTPUT objective


cmoussa1:
In #63, I gave an update on what objectives had been completed and which ones still needed work. I believe both INPUT and PROCESSING are completed, and OUTPUT is the one that still has some outstanding items. I detailed them in that issue thread:

> > OUTPUT: Updated account-tree object (So a separate flush code needs to be developed to update the new states into the flux-accounting database and @cmoussa1 agreed to take a crack at this)

> This objective is handled partly by #109: an updated account-tree object can now be output with a new subcommand: `flux > shares`. One of the last steps will consist of a couple of sub-steps that come to mind right away:
> - Updating new fairshare values into the flux-accounting database when new values are calculated
> - Adding this fairshare value to the output of `flux shares`

I figured it would help me if I continued to outline some of these sub-objectives in more detail to keep me on track and organized. 

---

> Updating new fairshare values into the flux-accounting database when new values are calculated

This will involve a couple of sub-steps, which I think could reasonably go into a couple of small PR's:

1. Add a **fairshare** field to the flux-accounting database under the `association_table` for every association. This will be a `REAL` number that represents each association's fairshare value.

```
fairshare     real        DEFAULT 0.0 NOT NULL
```

2. Update the `association_table` with these new fairshare values after the values are calculated. Perhaps I could add another function under `src/fairness/reader/data_reader_db.cpp` that takes a couple of arguments: a path to the SQLite database and the root of the weighted tree, and updates the table with these new values.

```c++
update_fairshare_values (const std::string &path, std::shared_ptr<weighted_tree_node_t> root)
```

I think I can perform the same traversal as I did in `flux_shares.cpp` to go to each sub bank and update users' new fairshare values.

```SQL
UPDATE association_table SET fairshare=? WHERE username=? AND bank=?
```

> Adding this fairshare value to the output of `flux shares`

This should involve adding another field to the `print` functions in `flux_shares.cpp` (`get_fshare ()`).

cmoussa1:
After talking some offline with @SteVwonder about this, we discussed that it might be best to create a new class in flux-accounting called `writer` in order to update fairshare values in the `association_table`. So I'll propose a sketch of what I'm thinking here: 

(*note: the image quality of the sketch is pretty poor for some reason, but if you click on the image you can expand it and it should be presented at a higher resolution*)

[...]

PR #116, opened 2021-04-08
add "fairshare" field to association_table, data_writer class to weighted_tree library

cmoussa1:
### Background 

One of the things required to fulfill the objectives outlined in #115 objective **(a)** is to: 

> Update new fairshare values into the flux-accounting database when new values are calculated.

This involves adding a fairshare field to the `association_table` in the flux-accounting database, a field which will store the fairshare value for each association entry. It also needs some sort of interface to write these calculated values to the database.

### Proposal

This PR adds a new `fairshare` field to the `association_table`.

It also adds a new class to the `weighted_tree` library: a `data_writer` class, one that is responsible for writing fairshare values to their appropriate end location.

The `data_writer` subclass added in this PR is `data_writer_db`, which is capable of traversing a `weighted_tree` object and writing user nodes' `fshare` values to their corresponding rows in the `association_table` in a flux-accounting SQLite database. Its place in the project structure is proposed as follows (located in the dark red ellipse):

data writer class sketch

###  data_writer_db

The primary function in **data_writer_db.cpp** is `update_fairshare_values ()`, a function which recursively traverses a `weighted_tree` object and looks for user nodes. When it finds one, it grabs its `fshare` value and prepares an `UPDATE` SQL statement to be pushed to a flux-accounting database, where the `fairshare` column gets updated. The function returns an `integer`: `0` indicating success, `-1` if the `UPDATE` statement fails to compile or have its parameters bound, or the respective `SQL` error code if the `UPDATE` statement fails while trying to write to the database or while getting destroyed. 

### Testing

I structured the tests as follows: each test database gets loaded in and placed into a `weighted_tree` object. A `walker` runs and generates fairshare values for the users in the `weighted_tree`. Those fairshare values then get `UPDATE`d to the flux-accounting database using `data_writer_db`. The fairshare values for the users in the `weighted_tree` then get compared with their respective entries in the flux-accounting database via a helper function `fetch_fshare ()`. Since they are doubles, I compared them using an epsilon value:

```c++
double epsilon = 0.000001f;
fabs (users[i]->get_fshare () - fshare) < epsilon
```

### Future Work

My idea for this `data_writer` class is that it can be used (in combination with the `data_reader` class) in some sort of cron job or other periodically-run script that calculates user fairshare values and stores them in a flux-accounting database. A rough example is shown in the `flux-calc-fshare.cpp` gold ellipse in the diagram above.


cmoussa1:
OK, now that I've posted the addition of the `data_writer` class, I think this is getting close to reviewable.

src/fairness/writer/data_writer_base.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace Flux {
+namespace writer {
+
+/*!  Base data writer class.
+*/
+class data_writer_base_t {
+public:
+    virtual ~data_writer_base_t () = default;
+
+    int write_accounting_info (const std::string &path);
```
SteVwonder, line 22: Is this function ever implemented?  Maybe it is worth refactoring this so that the db writer subclass has a function with the same name/arguments (i.e., this function should grow a `root` argument and the db subclass should drop the `db` part of its method's name).

SteVwonder: Might also be instructive to think if you could implement the `csv` writer and the `stdout` writers in terms of this base class.  In which case, one option might be to move some arguments (e.g., db/csv path) into the constructor.

src/fairness/writer/data_writer_db.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    ud = "UPDATE association_table SET fairshare=? WHERE username=? AND bank=?";
+
+    if (node->is_user ()) {
+        // get parameters for UPDATE statement
+        fshare = std::to_string (node->get_fshare ()).c_str ();
+        username = node->get_name ().c_str ();
+        bank = node->get_parent ()->get_name ().c_str ();
+
+        c_ud = compile_stmt (DB, ud);
```
SteVwonder, line 95: Rather than recompiling the sqlite statement on each call, could you compile it once, pass it in as an argument, and call `sqlite3_reset` on it here before re-binding parameters to it?

src/fairness/writer/test/data_writer_db_test01.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    s = "SELECT fairshare FROM association_table WHERE username=? AND bank=?";
+
+    sqlite3_prepare_v2 (DB, s.c_str (), -1, &cs, 0);
+
+    sqlite3_bind_text (cs, 1, u.c_str (), -1, NULL);
+    sqlite3_bind_text (cs, 2, b.c_str (), -1, NULL);
+
+    sqlite3_step (cs);
+
```
SteVwonder, line 54: Not that it matters all that much since this is a test, but does the prepared statement need to have `finalize` called on it?

src/fairness/weighted_tree/Makefile.am
```diff
@@ -13,17 +13,20 @@ noinst_HEADERS = \
...
     weighted_tree.cpp \
     weighted_walk.cpp \
     ../reader/data_reader_db.cpp \
+    ../writer/data_writer_db.cpp \
     ../account/account.hpp \
     weighted_tree.hpp \
     weighted_walk.hpp \
-    ../reader/data_reader_db.hpp
+    ../reader/data_reader_db.hpp \
+    ../reader/data_writer_db.hpp
```
SteVwonder, line 29: `reader` -> `writer`

SteVwonder (changes requested):
Very nicely done @cmoussa1! This PR was a pleasure to review :) Nice clean code, nice precise commits with good commit messages, and tests for the new functionality.  Couple of small things here and there, but overall very solid!

src/fairness/writer/data_writer_db.cpp
cmoussa1: Yes, I can do that! I went ahead and pushed a `fixup` commit 40009bb added a small helper function `clear_and_reset_bindings ()` that calls `sqlite3_reset` before passing the compiled SQL statement back to the recursive function.

src/fairness/writer/test/data_writer_db_test01.cpp
cmoussa1: Yes it does, thank you for pointing that out. I pushed up a `fixup` commit 0eb2e05 that calls `sqlite3_finalize` before returning the fairshare value.

src/fairness/writer/data_writer_base.hpp
cmoussa1: I hope I understood this correctly; apologies if I didn't! I looked at flux-sched's [resource readers class](https://github.com/flux-framework/flux-sched/tree/master/resource/readers) for some clarification. 

For my own understanding, is the concept here that sub-classes **override** the base class's functions when a subclass object is created and its member functions are called?

cmoussa1:
Thank you for the review @SteVwonder! I just pushed up some `fixup` commits based on your feedback and suggestions.

- added a `reset` and `clear_bindings` helper function to prevent the compiled SQL statement from getting re-compiled every time the recursive `update_fairshare_values ()` function is called
- added a `sqlite3_finalize ()` call to the compiled SQL statement in the test file
- corrected the path to the `data_writer` files in the `weighted_tree` Makefile
- changed the function names of the **base** and **sub** `data_writer` classes to match

src/fairness/writer/data_writer_base.hpp
SteVwonder: > For my own understanding, is the concept here that sub-classes override the base class's functions when a subclass object is created and its member functions are called?

Yeah, exactly.  Check out [this wiki article](https://en.wikipedia.org/wiki/Method_overriding) for more info.

SteVwonder: Thanks @cmoussa1! The changes you made are exactly what I recommended in my first comment.  So I think this PR is good to go and I can resolve this comment.  As a future refactor (certainly doesn't need to be done in this PR), it would be cool if the `flux shares` stdout output writer could be re-written as a subclass of `data_writer_base_t`, which would probably require some refactoring of the `write_acct_info` method's arguments (that's what I was poorly hinting at in my second comment).

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

cmoussa1:
Thank you @SteVwonder!

Issue #117, opened 2021-04-15
job_archive_interface: add call to calc_usage_factor() to flux-account.py


cmoussa1:
While creating the diagram that outlines the project structure of flux-accounting, I noticed that there is no call to the function in **job_archive_interface.py** that performs the job usage calculation for a user/bank association in a flux-accounting database, `calc_usage_factor()`. 

I should add a subcommand to **flux-account.py** that performs this job usage calculation for every user/bank combination in the flux-accounting database.

cmoussa1:
Fixed by #118. Closing.

PR #118, opened 2021-04-21
job_archive_interface: refactor calc_usage_factor(), add job usage calculation subcommand

cmoussa1:
### Background 

Noted in #117, there is currently no call to the `calc_usage_factor()` function, which is responsible for calculating an association's job usage value. This is another prerequisite for our server-side plugin work, where we are periodically re-calculating both usage and fairshare values and updating them, where they'll then be used to calculate a user's job priority.

But as I was adding a new subcommand that calls `calc_usage_factor()`, I noticed myself struggling to read the function itself. The function has lots of long comments, long variable names, and lots of calls to the flux-accounting database inside the function. I didn't like how much time was spent trying to figure out what I wrote. The other issue I found was with the calculation itself. I believe some of the calculations for job usage values were incorrect; I think a refactor is in order.

### Solution

The first half of this PR is a pretty significant cleanup and refactor of the `calc_usage_factor()` function. It involves the following:

- removing/editing a lot of long comments
- renaming long variable names
- adding helper functions that read from/write to the flux-accounting database
- changing the behavior of returning decayed usage values
- changing the unit tests (adding `unittest.mock()` and fixing the expected values) to reflect the calculation changes in `calc_usage_factor()`)

To visualize the refactor to `calc_usage_factor()`, I created a flowchart to represent a sample call to the function (you should be able to click on the image to expand it):

calc_usage_factor flowchart

The second part of this PR adds a new subcommand to `flux account`: `update-usage`, which fetches all associations from the flux-accounting database and calls `calc_usage_factor()` for each one. It then checks to see if the current half-life period needs to be updated. It takes one positional argument: a path to the job-archive DB. An optional argument can also be passed: the **PriorityDecayHalfLife** parameter, which if not specified, defaults to 1 (this mirrors the default value when the flux-accounting database is initially created as well).

```console
[moussa1@docker-desktop src]$ flux account update-usage -h
usage: flux-account.py update-usage [-h]
                                    [--priority-decay-half-life PRIORITY DECAY HALF LIFE]
                                    JOB-ARCHIVE_DB_PATH

positional arguments:
  JOB-ARCHIVE_DB_PATH   job-archive DB location

optional arguments:
  --priority-decay-half-life PRIORITY DECAY HALF LIFE
                        contribution of historical usage in weeks on the
                        composite usage value
```

### Testing

I added two new unit tests to `test_job_archive_interface.py` that call `update_job_usage()` at two different periods: one in the current half-life period (to ensure that current historical usage factors do NOT get updated), and one in the next half-life period (to ensure that current historical usage factors DO get updated). 

### Future Work/Impact

I was thinking that this PR would enable job usage values to be periodically calculated via a `cron` job to regularly update historical job usage values for users, and that it could be used in combination with the `data_reader` and `data_writer` classes in the weighted tree library to re-calculate fairshare values (based on the newly-calculated job usage values) and periodically update the flux-accounting database.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -291,6 +291,23 @@ def apply_decay_factor(decay, acct_conn, user=None, bank=None):
     return sum(usg_past_decay)
 
 
+def get_curr_usg_bin(acct_conn, user, bank):
+    # append current usage to the first usage factor bin
+    s_usg = """
+        SELECT usage_factor_period_0 FROM job_usage_factor_table
+        WHERE username=? AND bank=?
+        """
+    dataframe = pd.read_sql_query(
+        s_usg,
+        acct_conn,
+        params=(
+            user,
+            bank,
+        ),
+    )
+    return float(dataframe.iloc[0])
+
+
 def calc_usage_factor(jobs_conn, acct_conn, pdhl, user, bank):
 
     # hl_period represents the number of seconds that represent one usage bin
@@ -354,25 +371,7 @@ def calc_usage_factor(jobs_conn, acct_conn, pdhl, user, bank):
         usg_historical = apply_decay_factor(0.5, acct_conn, user, bank)
     elif (last_t_inactive - float(end_hl)) < hl_period:
         # found new jobs in the current half-life period
-
-        # append current usage to the first usage factor bin
-        fetch_current_usage_factor = """
-            SELECT usage_factor_period_0
-            FROM job_usage_factor_table
-            WHERE username=?
-            AND bank=?
-            """
-        dataframe = pd.read_sql_query(
-            fetch_current_usage_factor,
-            acct_conn,
-            params=(
-                user,
-                bank,
-            ),
-        )
-        usage_factor_period_0 = dataframe.iloc[0]
-
-        usg_current += float(usage_factor_period_0)
+        usg_current += get_curr_usg_bin(acct_conn, user, bank)
 
         # usage_user_past = sum of the older usage factors
         usg_past = fetch_old_usage_factors(acct_conn, user, bank)
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -254,7 +254,7 @@ def get_last_job_ts(acct_conn, user, bank):
     return float(timestamp.iloc[0])
 
 
-def fetch_old_usage_factors(acct_conn, user=None, bank=None):
+def fetch_usg_bins(acct_conn, user=None, bank=None):
     past_usage_factors = []
 
     select_stmt = "SELECT * from job_usage_factor_table WHERE username=? AND bank=?"
@@ -278,7 +278,7 @@ def apply_decay_factor(decay, acct_conn, user=None, bank=None):
     usg_past = []
     usg_past_decay = []
 
-    usg_past = fetch_old_usage_factors(acct_conn, user, bank)
+    usg_past = fetch_usg_bins(acct_conn, user, bank)
 
     # apply decay factor to past usage periods of a user's jobs
     for power, usage_factor in enumerate(usg_past, start=1):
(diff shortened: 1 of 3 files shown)


src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -222,6 +222,22 @@ def append_to_where(where_stmt, conditional):
     return job_records
 
 
+def update_t_inactive(acct_conn, last_t_inactive, user, bank):
```
SteVwonder, line 225: This is a really good catch and cleanup.  Can you also add a test case for this to ensure it doesn't regress?

src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -304,7 +336,7 @@ def apply_decay_factor(decay, acct_conn, user=None, bank=None):
         acct_conn.commit()
         period += 1
 
-    return sum(usg_past_decay)
+    return sum(usg_past_decay[:-1])
```
SteVwonder, line 339: Can you add a comment here to make clear why the `[-1]` is being used (I think your commit message makes it clear, but it'll be good to make it clear here too).

src/bindings/python/fluxacct/accounting/test/test_job_archive_interface.py
```diff
@@ -261,7 +261,7 @@ def test_13_check_usage_factor_in_table(self):
         select_stmt = "SELECT usage_factor_period_0 FROM job_usage_factor_table WHERE username='1002' AND bank='C'"
         dataframe = pd.read_sql_query(select_stmt, acct_conn)
         usage_factor = dataframe.iloc[0]
-        self.assertEqual(usage_factor["usage_factor_period_0"], 17044.0)
+        self.assertEqual(usage_factor["usage_factor_period_0"], 16956.0)
```
SteVwonder, line 266: Do these test cases cover all three edges cases that are enumerated in the commit message? Or do we need to add more test cases to cover all three?

src/bindings/python/fluxacct/accounting/test/test_job_archive_interface.py
```diff
@@ -338,10 +338,26 @@ def test_15_recalculate_usage_after_half_life_period(self):
 
         self.assertEqual(usage_factor, 2199.5)
 
+    # calling update_job_usage() in the same half-life period should NOT
+    # update usage factors for users
+    @mock.patch("time.time", mock.MagicMock(return_value=(10604799)))
```
SteVwonder, line 343: It might make more sense to set the `return_value` in an similar way to the above and below tests (e.g., `return_value=(10000000 + (604800 * 0.99))`)

SteVwonder (commented):
Thanks @cmoussa1!  This is a great cleanup and nice addition of functionality.  Couple of small comments below:

src/bindings/python/fluxacct/accounting/job_archive_interface.py
cmoussa1: Yeah, absolutely. I went ahead and added a test (`test_13_update_t_inactive_success`) that checks the last seen timestamp for a user before and after their job usage factor gets calculated to ensure that it gets updated.

src/bindings/python/fluxacct/accounting/job_archive_interface.py
cmoussa1: Sure thing. I pushed a `--fixup` commit that adds a comment just above the `return` statement that explains why we don't return the last factor.

src/bindings/python/fluxacct/accounting/test/test_job_archive_interface.py
cmoussa1: Yes: 

- test 15 checks _Problem (1)_, where new jobs are found in a new half-life period.
- tests 11-13 checks _Problem (2)_, where new jobs are found in the current half-life period.
- test 17 checks _Problem (3)_, where no new jobs are found in a new half-life period or new jobs are found.

src/bindings/python/fluxacct/accounting/test/test_job_archive_interface.py
cmoussa1: Agreed. I pushed up a `--fixup` commit that expands the `return_value` to match how `@mock.patch` is used amongst the other tests.

SteVwonder (approved):
Thanks @cmoussa1! LGTM!

cmoussa1:
Thanks @SteVwonder! 😁

PR #120, opened 2021-05-03
flux-shares: add data_writer_stdout subclass, add fairshare field to output

### LINKED ISSUE #63: Weighted tree library

### LINKED ISSUE #115: weighted tree library: breakdown of OUTPUT objective

cmoussa1:
#### Background

`flux_shares.cpp` currently contains its own convenience functions to print out information from a weighted tree object. 

Now that flux-accounting has the `data_writer` class, it might make more sense to move those convenience functions to a `data_writer` subclass that `flux_shares.cpp` can utilize.

#### Solution

This PR can be separated into three different parts.

Part 1 adds the newly-added fairshare field to `data_reader_db` so that fairshare info also gets added during the construction of a weighted tree object.

Part 2 adds a new subclass to the `data_writer` class (and to the weighted tree library): `data_writer_stdout`, which is responsible for writing flux-accounting information to stdout. It essentially moves the helper functions that used to be in `flux_shares.cpp` into this new subclass, which in turn can be called by `flux_shares.cpp`.

Part 3 changes the expected output in the sharness tests to include the new fairshare column.

#### Fulfillment

This PR should satisfy the remaining outstanding objectives in both #63 and #115:

> Adding this fairshare value to the output of flux shares

Fixes #63 
Fixes #115

### DIFF
--- a/src/fairness/writer/data_writer_stdout.hpp
+++ b/src/fairness/writer/data_writer_stdout.hpp
@@ -1,0 +1,81 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#include <sqlite3.h>
+#include <cerrno>
+
+#include "src/fairness/weighted_tree/weighted_walk.hpp"
+#include "src/fairness/reader/data_reader_db.hpp"
+#include "src/fairness/writer/data_writer_base.hpp"
+
+using namespace Flux::accounting;
+
+namespace Flux {
+namespace writer {
+
+/*!  Base data writer class.
+ */
+class data_writer_stdout_t : public data_writer_base_t {
+public:
+    data_writer_stdout_t (std::string indent = "",
+                          bool parsable = false,
+                          std::string delimiter = "")
+    {
+        m_indent = indent;
+        m_parsable = parsable;
+        m_delimiter = delimiter;
+    }
+    virtual ~data_writer_stdout_t () = default;
+
+    /*! Write fairshare values from a weighted tree.
+     *
+     * \param path          path to a flux-accounting database
+     * \param node          node of the weighted tree of the bank/user hierarchy
+     */
+    int write_acct_info (const std::string &path,
+                         std::shared_ptr<weighted_tree_node_t> node);
+
+private:
+    std::string m_indent = "";
+    bool m_parsable = false;
+    std::string m_delimiter = "";
+
+    /*! Print flux-accounting information with a custom delimiter.
+     *
+     * \param delimiter     custom delimiter for output
+     */
+    void print_csv_header (const std::string& delimiter="|");
+
+    /*! Bind a parameter to a compiled SQL statement
+     *
+     * \param node          node of the weighted tree of the bank/user hierarchy
+     * \param indent        level of indent in printing out the hierarchy
+     * \param delimiter     custom delimiter for output
+     */
+     void print_csv (std::shared_ptr<weighted_tree_node_t> node,
+                     const std::string& indent,
+                     const std::string& delimiter="|");
+
+    /*! Print the header in a nice format.
+     *
+     */
+    void pretty_print_header ();
+
+    /*! Print the user/bank hierarchy in a nice format.
+     *
+     * \param node      node of the weighted tree of the bank/user hierarchy
+     * \param indent    level of indent in printing out the hierarchy
+     */
+     void pretty_print (std::shared_ptr<weighted_tree_node_t> node,
+                        const std::string& indent);
+};
+
+} // namespace writer
+} // namespace Flux
--- a/src/fairness/print_hierarchy/flux_shares.cpp
+++ b/src/fairness/print_hierarchy/flux_shares.cpp
@@ -16,12 +16,10 @@ extern "C" {
 #endif
 }
 
-#include "src/fairness/reader/data_reader_db.hpp"
+#include "src/fairness/writer/data_writer_stdout.hpp"
 
 using namespace Flux::accounting;
-using namespace Flux::reader;
-
-const std::string DBPATH = std::string (X_LOCALSTATEDIR) + "/FluxAccounting.db";
+using namespace Flux::writer;
 
 static void show_usage ()
 {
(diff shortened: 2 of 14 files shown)


src/fairness/writer/data_writer_stdout.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "src/fairness/writer/data_writer_base.hpp"
+
+using namespace Flux::accounting;
+
+namespace Flux {
+namespace writer {
+
+/*!  Base data writer class.
+ */
+class data_writer_stdout_t : public data_writer_base_t {
```
dongahn, line 24: @cmoussa1: what is the abstract interface `data_writer_stdout_t` implement? Is that `int write_acct_info` in the base class? If so, that method in `data_writer_base_t` should be marked as `virtual` with `=0`? 

As is, it is not clear why `data_writer_stdout_t` is a kind of `data_writer_base_t`. A typical pattern I have seen is a basic set of `writer` operations are abstracted out in the base class and each derived writer subclasses implement those.

cmoussa1: Yeah, I understand - thanks for this clarification. I suppose I should have appended a `TODO` to the description of the PR, as I still had to implement the `write_acct_info ()` method from the sub class - I still had that work left to do, I'm sorry for not making that more clear.

My plan was to overload the `write_acct_info ()` method in `data_writer_stdout_t` with the parameters specific to this subclass, like the `delimiter` argument. Then I could make the helper functions in `data_writer_stdout_t` private methods and have `flux_shares.cpp` just call `write_acct_info ()`. Does this implementation make sense and follow that pattern?

cmoussa1:
OK, I pushed up my remaining `TODO` items and I think this is getting close to ready, so I'll mark it as ready for review.

src/fairness/writer/data_writer_stdout.hpp
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * \param node          node of the weighted tree of the bank/user hierarchy
+     * \param indent        indent amount when printing out parsable hierarchy
+     * \param parsable      boolean determining parsable or pretty output
+     * \param delimiter     custom delimiter for output
+     */
+    int write_acct_info (std::shared_ptr<weighted_tree_node_t> node,
+                         const std::string& indent,
+                         bool parsable,
+                         const std::string& delimiter="|");
```
SteVwonder, line 38: The signature of this `write_acct_info` differs from the signature in the base class.  Might make sense to refactor this class so they are the same.  I'll defer to @dongahn on the best way to do this, but maybe push the `indent`, `parsable`, `delimiter` configuration into the constructor of the writer (so the constructors of the various classes will differ) but the write function has the same signature.

cmoussa1: Thanks for pointing this out - I just pushed up a change that I hope works better. I moved `indent`, `parsable`, and `delimiter` to be public members of the `data_writer_stdout` subclass and kept the signature the same as the base class. This allows me to access these public members of a `data_writer_stdout` subclass with `this->member`.

I also moved some of the helper functions to `data_writer_stdout.cpp`, so `flux_shares.cpp` now really only parses the command line arguments and then calls `write_acct_info ()` with an optional filename.

src/fairness/writer/data_writer_stdout.hpp
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * \param path          path to a flux-accounting database
+     * \param node          node of the weighted tree of the bank/user hierarchy
+     */
+    int write_acct_info (const std::string &path,
+                         std::shared_ptr<weighted_tree_node_t> node);
+
+    std::string indent = "";
+    bool parsable = false;
+    std::string delimiter = "";
```
SteVwonder, line 39: Rather than making these publically modifiable attributes, maybe make them private and then override the default constructor to take these as arguments and set them within the constructor?

SteVwonder (approved):
Just one comment, otherwise, this LGTM!

src/fairness/writer/data_writer_stdout.hpp
cmoussa1: Thanks for pointing this out @SteVwonder - I wanted to make sure I understood your suggestion correctly (apologies if I didn't), so I pushed up a couple of `fixup` commits. If they look good, then I can go ahead and squash them. Basically, I moved those attributes to `private` and added a new constructor with parameters that will override the default:

```c++
data_writer_stdout_t (std::string indent = "",
                      bool parsable = false,
                      std::string delimiter = "")
{
    m_indent = indent;
    m_parsable = parsable;
    m_delimiter = delimiter;
}
```

Then, in other places of code where I declare a `data_writer_stdout` object (like in `flux_shares.cpp`), I would initialize the object with these attributes. 

```c++
data_writer_stdout_t data_writer (indent, parsable, delimiter);
```

Hopefully I understood what you were recommending.

cmoussa1:
Just force-pushed an update to squash the `fixup` commits that hopefully address the feedback above.

src/fairness/writer/data_writer_stdout.hpp
SteVwonder: Yeah. That looks great! Thanks!

cmoussa1: Woohoo! Thanks. 🙂 I'll set MWP then.

PR #122, opened 2021-05-26
plugin: add multifactor priority plugin

cmoussa1:
This is a very early first-cut, work-in-progress PR that adds multi-factor priority calculation to flux-accounting. It is still missing a couple of integral pieces, but I thought it might be best to post it early for the following reasons:

- Since this is a somewhat important milestone for flux-accounting as a whole, I wanted to make sure I have been headed in the right direction so far with my plugin
- I think I might benefit from some collaboration and feedback on the pieces that are still missing from this plugin (testing strategy, the data structure in the plugin, managing users who belong to more than one bank, etc.)

---

#### Background

A multifactor priority plugin is responsible for calculating a job's priority using a number of different factors that may help or hurt a job's overall priority. This PR starts with adding just two factors: **fairshare** and **urgency**. The intent is that once this PR becomes mergeable, we can start to add additional factors in follow-up PR's to create a more complete and solidified priority calculator.

The **fairshare** factor represents the ratio between the amount of allotted resources and the amount of resources actually consumed. It should range between 0.0 and 1.0, where 0.0 represents the lowest possible value (indicating a user who is using way more resources than allocated) and 1.0 represents the highest. When a user first gets added to the database, their fairshare value should be 0.5, which should mean the user is using exactly the amount of resources they were allocated.

The **urgency** factor is similar to the **nice** factor in Slurm in that it allows users to prioritize their own jobs with an integer in the range [1,16], where 16 is the highest priority and 1 is the lowest.

Together, these two factors make up a priority calculation using the following formula:

```
priority = | (fshare_weight * fshare_factor) + (urgency - 16) |
```

where:

- **fshare_weight**: represents the integer weight that fairshare has on the overall priority calculation
- **fshare_factor**: represents a user's fairshare value stored in the flux-accounting database (this value gets periodically updated as a user submits jobs over time)
- **urgency**: represents the user's urgency when submitting a job (default value is 16)

As additional factors get added (**job size**, **age**, **queue**, etc.), they will be added to this equation to generate a more representative multi-factor priority. 

#### Implementation

This PR adds a new folder to `flux-accounting/src/` called `plugins/` which stores both the priority plugin itself and the external service that pushes data to the plugin.

**bulk_update.py** is the external service that pushes user accounting data from a flux-accounting DB to the priority plugin, where it is stored in a multimap data structure to be looked up when a job priority needs to be calculated.

It is a Python script that can be run from the command line:

```console
[moussa1@docker-desktop plugins]$ flux python bulk_update.py -h
usage: bulk_update.py [-h] [-p PATH]

Description: Send a bulk update of user information from a flux-accounting
database to the multi-factor priority plugin.

optional arguments:
```

It has one optional argument, `--path`, a path to a flux-accounting DB. If no `--path` is specified, it will look for a default location of the flux-accounting DB, similar to the front-end Python command line tool **flux-account.py**.

**bulk_update.py** runs one query to the flux-accounting DB to fetch every user row. Specifically, it queries the username and their respective fairshare value. It packs this data in a dictionary object and sends an RPC request to the plugin. 

Just before it begins sending user data to the plugin, it sends an empty message to the plugin, telling it to clear its multimap's contents to prepare for an update.

**mf_priority.cpp** is the priority plugin file. It contains a number of callbacks and a helper function that is responsible for calculating the priority itself.

`clear_map_cb ()` and `rec_update_cb ()` are the functions responsible for clearing and updating the multimap data structure in the plugin and interacting with the external service, **bulk_update.py**.

When a job gets submitted, `validate_cb ()` runs and performs a lookup in the multimap to look for the userid that submitted the job. If the userid cannot be found, the job is rejected with an error message explaining that the user couldn't be found in the flux-accounting DB.

If a user is indeed in the flux-accounting DB, the `userid` and `urgency` are extracted using `flux_plugin_arg_unpack ()` and sent to a helper function `priority_calculation ()`, where its priority is calculated and returned, to be packed using `flux_plugin_arg_pack ()`.

`priority_calculation ()` is responsible for calculating a job's priority using the formula described in the _Background_. It currently covers two corner cases:

1. If a user submits a job with an urgency of `0`, the priority returned is `FLUX_JOB_PRIORITY_MIN` = 0. A job will remain as pending.

2. If a job is submitted (or updated) with a max urgency of `31`, the priority returned is `FLUX_JOB_PRIORITY_MAX` = 4294967295. This is the highest priority a job can have. 

If the submitted job is neither of these, then the calculation proceeds normally with the formula stated above. I think a note should be made that the calculated priority is an absolute value. I chose this in case the priority calculated turns out to be a negative value (I think the scenario in which this happens is if a user's fairshare value is so low that, when subtracted from the urgency, it results in a small negative integer; taking an absolute value _should_ result in a very low integer priority).

#### Examples

<details>

Since this PR does not have testing yet, I thought it might be beneficial to provide at least a couple examples to show how the plugin works in its current state.

Here is how I compile the plugin, load it, and push data to it:

```console
[moussa1@docker-desktop plugins]$ g++ -fPIC -shared -o mf_priority.so mf_priority.cpp
[moussa1@docker-desktop plugins]$ flux jobtap load mf_priority.so
Loaded:
mf_priority
Previously loaded:
builtin.priority.default
[moussa1@docker-desktop plugins]$ flux python bulk_update.py -p test/accounting_db_data/small_no_tie.db
```

* submitting a job with no specification of urgency:

```console
[moussa1@docker-desktop plugins]$ flux mini submit -n 1 hostname
ƒZ8pD9HtK
[moussa1@docker-desktop plugins]$ flux job wait-event -v ƒZ8pD9HtK priority
1622047983.174637 submit userid=58985 urgency=16 flags=0
1622047983.190122 depend
1622047983.190164 priority priority=45321
```

* submitting a job with `--urgency=15`

```console
[moussa1@docker-desktop plugins]$ flux mini submit -n 1 --urgency=15 hostname
ƒZuN5hafZ
[moussa1@docker-desktop plugins]$ flux job wait-event -v ƒZuN5hafZ priority
1622048084.122513 submit userid=58985 urgency=15 flags=0
1622048084.137835 depend
1622048084.137883 priority priority=45320
```

* submitting a job with `--urgency=0` and updating it later with `--urgency=31`

```console
[moussa1@docker-desktop plugins]$ flux mini submit -n 1 --urgency=0 hostname
ƒaQNfiUvB
[moussa1@docker-desktop plugins]$ flux job wait-event -v ƒaQNfiUvB priority
1622048149.879270 submit userid=58985 urgency=0 flags=0
1622048149.894008 depend
1622048149.894067 priority priority=0
[moussa1@docker-desktop plugins]$ flux job urgency ƒaQNfiUvB 31
[moussa1@docker-desktop plugins]$ flux job wait-event -v ƒaQNfiUvB start
1622048149.879270 submit userid=58985 urgency=0 flags=0
1622048149.894008 depend
1622048149.894067 priority priority=0
1622048183.573533 urgency userid=58985 urgency=31
1622048183.573644 priority priority=4294967295
1622048183.579611 alloc annotations={"sched":{"resource_summary":"rank0/core0"}}
1622048183.588152 start
```
</details>

#### Assumptions/Gaps/Testing

<details>

I'll use this section to document the gaps for this PR I can think of so far, the assumptions made, and any other questions or feedback I think I could use.

* Testing

~~This is one of the larger gaps I can think of for this PR. I looked at the tests for the plugins in flux-core and am wondering how I should approach testing in the flux-accounting repo. Are sharness tests a good approach for my plugin?~~ 

~~Maybe I could populate a fake database with some fake users and fairshare values and have them submit similar jobs and make sure their calculated priorities are as expected, but I guess I wasn't sure how I would do this if I am not in a flux instance.~~ 

* Only Using User/Fairshare

~~One of the assumptions I've made so far in this PR is that a user only belongs to one bank; with this assumption, bank information doesn't need to be passed from the DB to the plugin. However, users can belong to multiple banks, so I need to handle this case.~~ 

* Clearing the map for an update

In its current structure, a job could be submitted at the time just after the map data structure gets cleared and just before an update is pushed to the plugin. In this case, the job would be rejected saying that the user wasn't found in the flux-accounting DB. Would the correct way to handle this would be to instead `return flux_jobtap_priority_unavail ()` so that the submitted job will remain in the `PRIORITY` state until a priority is assigned using `flux_jobtap_reprioritize_job ()`?

* Declaring the map as a global variable

The plugin declares the map as a global variable. I declared it as global because multiple functions and callbacks interact with and update this map. Is this okay behavior? Should it instead be declared and initialized in one of the callbacks and passed to other functions when it is needed?

These are the `TODOS` and clarifications I can think of so far, but maybe there are things I missed. 
</details>

#### PR Structure

I think this PR can be broken into the following parts: 

- Part One adds a `userid` field to the flux-accounting DB. 
- Parts Two and Three add the priority plugin and the external bulk update service respectively.
- Part Four adds some sharness tests for the plugin as well and adds some helper scripts required by the new sharness test (e.g. add an `rc` directory)

### DIFF
--- a/src/plugins/bulk_update.py
+++ b/src/plugins/bulk_update.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2020 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import flux
+import argparse
+import sys
+import os
+import sqlite3
+
+import fluxacct.accounting
+
+
+def set_db_loc(args):
+    path = args.path if args.path else fluxacct.accounting.db_path
+
+    return path
+
+
+def est_sqlite_conn(path):
+    # try to open database file; will exit with -1 if database file not found
+    if not os.path.isfile(path):
+        print(f"Database file does not exist: {path}", file=sys.stderr)
+        sys.exit(1)
+
+    db_uri = "file:" + path + "?mode=rw"
+    try:
+        conn = sqlite3.connect(db_uri, uri=True)
+        # set foreign keys constraint
+        conn.execute("PRAGMA foreign_keys = 1")
+    except sqlite3.OperationalError:
+        print(f"Unable to open database file: {db_uri}", file=sys.stderr)
+        sys.exit(1)
+
+    return conn
+
+
+def bulk_update(path):
+    conn = est_sqlite_conn(path)
+    cur = conn.cursor()
+
+    # fetch all rows from association_table (will print out tuples)
+    for row in cur.execute("SELECT userid, bank, fairshare FROM association_table"):
+        # create a JSON payload with the results of the query
+        data = {"userid": str(row[0]), "bank": str(row[1]), "fairshare": str(row[2])}
+
+        flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="""
+        Description: Send a bulk update of user information from a
+        flux-accounting database to the multi-factor priority plugin.
+        """
+    )
+
+    parser.add_argument(
+        "-p", "--path", dest="path", help="specify location of database file"
+    )
+    args = parser.parse_args()
+
+    path = set_db_loc(args)
+
+    bulk_update(path)
+
+
+if __name__ == "__main__":
+    main()
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1,0 +1,230 @@
+/************************************************************\
+ * Copyright 2021 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+/* mycpppriority.cpp - custom basic job priority plugin
+ *
+ */
+
+extern "C" {
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+}
+#include <flux/core.h>
+#include <flux/jobtap.h>
+#include <map>
+#include <iterator>
+#include <cmath>
+#include <cassert>
+#include <algorithm>
+#include <cinttypes>
+
+std::map<int, std::map<std::string, double>> users;
+
+/******************************************************************************
+ *                                                                            *
+ *                           Helper Functions                                 *
+ *                                                                            *
+ *****************************************************************************/
+
+/*
+ * Calculate a user's job priority using the following factors:
+ *
+ * fairshare: the ratio between the amount of resources allocated vs. resources
+ *     consumed.
+ * urgency: a user-controlled factor to prioritize their own jobs.
+ */
+int64_t priority_calculation (flux_plugin_t *p,
+                              flux_plugin_arg_t *args,
+                              int userid,
+                              char *bank,
+                              int urgency)
+{
+    double fshare_factor = 0.0, priority = 0.0;
+    int fshare_weight;
+
+    std::map<int, std::map<std::string, double>>::iterator it;
+    std::map<std::string, double>::iterator inner_it;
+
+    fshare_weight = 100000;
+
+    if (urgency == FLUX_JOB_URGENCY_HOLD)
+        return FLUX_JOB_PRIORITY_MIN;
+
+    if (urgency == FLUX_JOB_URGENCY_EXPEDITE)
+        return FLUX_JOB_PRIORITY_MAX;
+
+    // search element in map of maps by key
+    it = users.find (userid);
+
+    // search for correct fshare value using passed-in bank; otherwise, use
+    // a default bank
+    if (bank != NULL) {
+        inner_it = it->second.find (bank);
+        fshare_factor = inner_it->second;
+    }
+    else {
+        inner_it = it->second.find ("default");
+        fshare_factor = inner_it->second;
+    }
+
+    priority = (fshare_weight * fshare_factor) + (urgency - 16);
+
+    return abs (round (priority));
+}
+
+
+/******************************************************************************
+ *                                                                            *
+ *                               Callbacks                                    *
+ *                                                                            *
+ *****************************************************************************/
+
+/*
+ * Unpack a payload from an external bulk update service and place it in the
+ * multimap datastructure.
+ */
+static void rec_update_cb (flux_t *h,
+                           flux_msg_handler_t *mh,
+                           const flux_msg_t *msg,
+                           void *arg)
+{
+    char *uid, *fshare, *bank;
+
+    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s}",
+                             "userid", &uid,
+                             "bank", &bank,
+                             "fairshare", &fshare) < 0) {
+        flux_log_error (h, "failed to unpack custom_priority.trigger msg");
+        goto error;
+    }
+
+    if (flux_respond (h, msg, NULL) < 0)
+        flux_log_error (h, "flux_respond");
+
+    // if the user being added to the does not yet have any entries in the map,
+    // treat their first bank as the "default" bank
+    if (users.count (std::atoi (uid)) == 0)
+        users[std::atoi (uid)]["default"] = std::stod (fshare);
+
+    users[std::atoi (uid)][bank] = std::stod (fshare);
+
+    return;
+error:
+    flux_respond_error (h, msg, errno, flux_msg_last_error (msg));
+}
+
+
+/*
+ * Unpack the urgency and userid from a submitted job and call calc_priority (),
+ * which will return a new job priority to be packed.
+ */
+static int priority_cb (flux_plugin_t *p,
+                        const char *topic,
+                        flux_plugin_arg_t *args,
+                        void *data)
+{
+    int urgency, userid;
+    char *bank = NULL;
+    int64_t priority;
+
+    flux_t *h = flux_jobtap_get_flux (p);
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:i, s:i, s{s{s{s?s}}}}",
+                                "urgency", &urgency,
+                                "userid", &userid,
+                                "jobspec", "attributes", "system",
+                                "bank", &bank) < 0) {
+        flux_log (h,
+                  LOG_ERR,
+                  "flux_plugin_arg_unpack: %s",
+                  flux_plugin_arg_strerror (args));
+        return -1;
+    }
+
+    priority = priority_calculation (p, args, userid, bank, urgency);
+
+    if (flux_plugin_arg_pack (args,
+                              FLUX_PLUGIN_ARG_OUT,
+                              "{s:I}",
+                              "priority",
+                              priority) < 0) {
+        flux_log (h,
+                  LOG_ERR,
+                  "flux_plugin_arg_pack: %s",
+                  flux_plugin_arg_strerror (args));
+        return -1;
+    }
+    return 0;
+}
+
+
+/*
+ * Look up the userid of the submitted job in the multimap; if user is not found
+ * in the map, reject the job saying the user wasn't found in the
+ * flux-accounting database.
+ */
+static int validate_cb (flux_plugin_t *p,
+                        const char *topic,
+                        flux_plugin_arg_t *args,
+                        void *data)
+{
+    int userid;
+    char *bank = NULL;
+    std::map<int, std::map<std::string, double>>::iterator it;
+    std::map<std::string, double>::iterator inner_it;
+
+    flux_t *h = flux_jobtap_get_flux (p);
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:i, s{s{s{s?s}}}}",
+                                "userid", &userid,
+                                "jobspec", "attributes", "system",
+                                "bank", &bank) < 0) {
+        return flux_jobtap_reject_job (p, args, "unable to unpack bank arg");
+    }
+
+    // searching element in std::map by key
+    it = users.find (userid);
+    if (it == users.end ())
+        return flux_jobtap_reject_job (p, args,
+                                       "user not found in flux-accounting DB");
+
+    if (bank != NULL) {
+        inner_it = it->second.find (std::string (bank));
+        if (inner_it == it->second.end ())
+            return flux_jobtap_reject_job (p, args,
+                                    "user does not belong to specified bank");
+    }
+
+    return 0;
+}
+
+
+static const struct flux_plugin_handler tab[] = {
+    { "job.validate", validate_cb, NULL },
+    { "job.state.priority", priority_cb, NULL },
+    { "job.priority.get", priority_cb, NULL },
+    { 0 },
+};
+
+
+extern "C" int flux_plugin_init (flux_plugin_t *p)
+{
+    if (flux_plugin_register (p, "mf_priority", tab) < 0
+        || flux_jobtap_service_register (p, "rec_update", rec_update_cb, p) < 0)
+        return -1;
+    return 0;
+}
+
+/*
+ * vi:tabstop=4 shiftwidth=4 expandtab
+ */
--- a/src/test/docker/bionic/Dockerfile
+++ b/src/test/docker/bionic/Dockerfile
@@ -1,4 +1,4 @@
-FROM fluxrm/flux-core:bionic
+FROM fluxrm/flux-core:bionic-v0.28.0
 
 ARG USER=flux
 ARG UID=1000
--- a/t/t1001-mf-priority.t
+++ b/t/t1001-mf-priority.t
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+test_description='Test multi-factor priority plugin'
+
+. `dirname $0`/sharness.sh
+MULTI_FACTOR_PRIORITY=${FLUX_BUILD_DIR}/src/plugins/.libs/mf_priority.so
+# BULK_UPDATE=${FLUX_BUILD_DIR}/t/scripts/send_fake_payloads.py
+
+export TEST_UNDER_FLUX_NO_JOB_EXEC=y
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 1 job
+
+flux setattr log-stderr-level 1
+
+test_expect_success 'load multi-factor priority plugin' '
+	flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
+'
+
+test_expect_success 'check that mf_priority plugin is loaded' '
+	flux jobtap list | grep mf_priority
+'
+
+test_expect_success 'try to submit a job when user does not exist in DB' '
+	test_must_fail flux mini submit -n1 hostname > failure.out 2>&1 &&
+	test_debug "cat failure.out" &&
+	grep "user not found in flux-accounting DB" failure.out
+'
+
+test_expect_success 'send an empty payload to make sure unpack fails' '
+	cat <<-EOF >bad_payload.py &&
+	import flux
+
+	#create a JSON payload
+	flux.Flux().rpc("job-manager.mf_priority.rec_update", {}).get()
+	EOF
+	test_must_fail flux python bad_payload.py &&
+	flux dmesg | grep "failed to unpack custom_priority.trigger msg: Protocol error"
+'
+
+test_expect_success 'create fake_payload.py' '
+	cat <<-EOF >fake_payload.py
+	import flux
+	import pwd
+	import getpass
+
+	username = getpass.getuser()
+	userid = pwd.getpwnam(username).pw_uid
+	# create a JSON payload
+	data = {"userid": str(userid), "bank": "account3", "fairshare": "0.45321"}
+	flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
+	data = {"userid": str(userid), "bank": "account2", "fairshare": "0.11345"}
+	flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
+	EOF
+'
+
+test_expect_success 'update plugin with sample test data' '
+	flux python fake_payload.py
+'
+
+test_expect_success 'submit a job with default urgency' '
+	jobid=$(flux mini submit --setattr=system.bank=account3 -n1 hostname) &&
+	flux job wait-event -f json $jobid priority | jq '.context.priority' > job1.test &&
+	cat <<-EOF >job1.expected &&
+	45321
+	EOF
+	test_cmp job1.expected job1.test
+'
+
+test_expect_success 'submit a job with custom urgency' '
+	jobid=$(flux mini submit --setattr=system.bank=account3 --urgency=15 -n1 hostname) &&
+	flux job wait-event -f json $jobid priority | jq '.context.priority' > job2.test &&
+	cat <<-EOF >job2.expected &&
+	45320
+	EOF
+	test_cmp job2.expected job2.test
+'
+
+test_expect_success 'submit a job with urgency of 0' '
+	jobid=$(flux mini submit --setattr=system.bank=account3 --urgency=0 -n1 hostname) &&
+	flux job wait-event -f json $jobid priority | jq '.context.priority' > job3.test &&
+	cat <<-EOF >job3.expected &&
+	0
+	EOF
+	test_cmp job3.expected job3.test &&
+	flux job cancel $jobid
+'
+
+test_expect_success 'submit a job with urgency of 31' '
+	jobid=$(flux mini submit --setattr=system.bank=account3 --urgency=31 -n1 hostname) &&
+	flux job wait-event -f json $jobid priority | jq '.context.priority' > job4.test &&
+	cat <<-EOF >job4.expected &&
+	4294967295
+	EOF
+	test_cmp job4.expected job4.test
+'
+
+test_expect_success 'submit a job with other bank' '
+	jobid=$(flux mini submit --setattr=system.bank=account2 -n1 hostname) &&
+	flux job wait-event -f json $jobid priority | jq '.context.priority' > job5.test &&
+	cat <<-EOF >job5.expected &&
+	11345
+	EOF
+	test_cmp job5.expected job5.test
+'
+
+test_expect_success 'submit a job using default bank' '
+	jobid=$(flux mini submit -n1 hostname) &&
+	flux job wait-event -f json $jobid priority | jq '.context.priority' > job6.test &&
+	cat <<-EOF >job6.expected &&
+	45321
+	EOF
+	test_cmp job6.expected job6.test
+'
+
+test_expect_success 'submit a job using a bank the user does not belong to' '
+	test_must_fail flux mini submit --setattr=system.bank=account1 -n1 hostname > bad_bank.out 2>&1 &&
+	test_debug "cat bad_bank.out" &&
+	grep "user does not belong to specified bank" bad_bank.out
+'
+
+test_expect_success 'reject job when invalid bank format is passed in' '
+	test_must_fail flux mini submit --setattr=system.bank=1 -n1 hostname > invalid_fmt.out 2>&1 &&
+	test_debug "cat invalid_fmt.out" &&
+	grep "unable to unpack bank arg" invalid_fmt.out
+'
+
+test_done
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -106,3 +106,8 @@ jobs:
     - name: coverage report
       if: success() && matrix.coverage
       run: DOCKER_REPO= bash <(curl -s https://codecov.io/bash)
+
+    - name: annotate errors
+      if: failure() || cancelled()
+      env: ${{matrix.env}}
+      run: src/test/checks-annotate.sh
--- a/configure.ac
+++ b/configure.ac
@@ -107,6 +107,9 @@ AC_SUBST(acct_db_path)
 AC_CONFIG_MACRO_DIR([config])
 X_AC_EXPAND_INSTALL_DIRS
 
+fluxplugin_ldflags="-avoid-version -export-symbols-regex '^flux_plugin_init\$\$' --disable-static -shared -export-dynamic"
+AC_SUBST(fluxplugin_ldflags)
+
 AC_CONFIG_FILES([Makefile
   src/Makefile
   src/common/Makefile
@@ -125,6 +128,7 @@ AC_CONFIG_FILES([Makefile
   src/bindings/python/fluxacct/accounting/Makefile
   src/bindings/python/fluxacct/accounting/__init__.py
   src/cmd/Makefile
+  src/plugins/Makefile
   t/Makefile
   ])
 AC_OUTPUT
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,1 +1,1 @@
-SUBDIRS = common cmd fairness bindings
+SUBDIRS = common cmd fairness bindings plugins
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -87,15 +87,16 @@ def create_db(
     conn.execute(
         """
             CREATE TABLE IF NOT EXISTS association_table (
-                creation_time bigint(20)              NOT NULL,
-                mod_time      bigint(20)  DEFAULT 0   NOT NULL,
-                deleted       tinyint(4)  DEFAULT 0   NOT NULL,
-                username      tinytext                NOT NULL,
-                admin_level   smallint(6) DEFAULT 1   NOT NULL,
-                bank          tinytext                NOT NULL,
-                shares        int(11)     DEFAULT 1   NOT NULL,
-                job_usage     real        DEFAULT 0.0 NOT NULL,
-                fairshare     real        DEFAULT 0.0 NOT NULL,
+                creation_time bigint(20)                NOT NULL,
+                mod_time      bigint(20)  DEFAULT 0     NOT NULL,
+                deleted       tinyint(4)  DEFAULT 0     NOT NULL,
+                username      tinytext                  NOT NULL,
+                userid        int(11)     DEFAULT 65534 NOT NULL,
+                admin_level   smallint(6) DEFAULT 1     NOT NULL,
+                bank          tinytext                  NOT NULL,
+                shares        int(11)     DEFAULT 1     NOT NULL,
+                job_usage     real        DEFAULT 0.0   NOT NULL,
+                fairshare     real        DEFAULT 0.0   NOT NULL,
                 PRIMARY KEY   (username, bank)
         );"""
     )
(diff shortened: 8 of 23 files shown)


grondo:
> This is one of the larger gaps I can think of for this PR. I looked at the tests for the plugins in flux-core and am wondering how I should approach testing in the flux-accounting repo. Are sharness tests a good approach for my plugin?

> Maybe I could populate a fake database with some fake users and fairshare values and have them submit similar jobs and make sure their calculated priorities are as expected, but I guess I wasn't sure how I would do this if I am not in a flux instance.

I would definitely start with a sharness test. If you have not already, pull in sharness.sh and `sharness.d/flux-sharness.sh` from flux-core. You can then use the `test_under_flux` "function" to re-run sharness test(s) under a flux instance.

There are a few existing priority-order tests you can look to as examples in flux-core:
```
t2216-job-manager-priority-order-single.t
t2217-job-manager-priority-order-limited.t
t2218-job-manager-priority-order-unlimited.t
```

You could also easily create your fake database, and/or have bulk_update.py push contrived values into your plugin, etc. Basically anything you can script can be tested.

cmoussa1:
OK, I've updated this PR with the following changes:

- added handling for users who belong to more than one bank
- added sharness tests that load the plugin and submit jobs using the custom plugin

So I'll go ahead and bring this PR out of [WIP] and post it for official review. I handled the multiple-bank case in the following way:

When a user submits a job, the validate callback will try to unpack a bank from the jobspec and store it in a `bank` variable if it finds one. If not, `bank` is left NULL. In the case where `bank` is left `NULL` but the user submitting the job belongs to more than one bank, the validate callback will reject the job, explaining that the user must specify a bank to charge since they belong to more than one bank. 

Eventually, I think the plugin should handle this case by not rejecting the job and just using a default bank (e.g setting the default to the first bank the user was added to in the DB), and I'll open an issue to track this so I can add it in the near future.

grondo:
> Eventually, I think the plugin should handle this case by not rejecting the job and just using a default bank (e.g setting the default to the first bank the user was added to in the DB), and I'll open an issue to track this so I can add it in the near future.

This sounds fine for now. I would think eventually administrators would want to be able to tweak the default bank for a user outside of the order in which the user is added to banks, so I would consider making the default bank a property of each user which could be separately set in the db.

> Oh, one other thing I wanted to make sure I was doing correctly was compiling the plugin into a .so using Makefile.am. I think I came up with a solution in this PR, as a .libs directory containing the .so gets created during make. Is this the correct way to compile the plugin?

You got most of the way there! You'll want to also add `-avoid-version` to avoid adding a DSO version to the plugin (e.g. it is now called `mf_priority.so.0.0.0`, and use `-export-symbols-regex '^flux_plugin_init\$\$'` so that only the `flux_plugin_init()` symbol is exported from the DSO.

You can just copy the `fluxplugin_ldflags` from the `configure.ac` in flux-core:`
```
fluxplugin_ldflags="-avoid-version -export-symbols-regex '^flux_plugin_init\$\$' --disable-static -shared -export-dynamic"
AC_SUBST(fluxplugin_ldflags)
```

And then use that on your plugin's LDFLAGS.

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,271 @@
...
+        return FLUX_JOB_PRIORITY_MIN;
+
+    if (urgency == FLUX_JOB_URGENCY_EXPEDITE)
+        return FLUX_JOB_PRIORITY_MAX;
+
+    // search element in multimap by key
+    it = users.find (userid);
+    if (it == users.end ())
+        return flux_jobtap_reject_job (p, args,
+                                       "user not found in flux-accounting DB");
```
grondo, line 66: `flux_jobtap_reject_job()` can only be used from `job.validate` (or `job.dependency.*`), but `priority_calculation()` looks like it can be called from `job.priority.get`. In this case, you can return `flux_jobtap_priority_unavail()` if the job is still in PRIORITY state. It looks like this can't happen though -- if the user does not exist in the db then the job would have been rejected by `validate_cb`.

grondo (changes requested):
@cmoussa1, I made a first pass with some comments. I mainly reviewed the multifactor plugin, since I'm not the right person to review the remainder of the code.

I don't understand enough of how the user+bank is used to get the fairshare factor to know if the use of a multimap is the right data structure here, it seems like it requires a little extra work currently. 

It may be easier if you had two maps -- one that maps banks to fairshare value, and one that maps userid to a list of banks in which the user is a member. Then you can easily test bank membership, and get the bank fairshare value in a single lookup call.

I'd also suggest caching the bank along with the job, perhaps with `flux_jobtap_job_aux_set()` so that you don't have to continually lookup the bank name for each job when you've already done it in the `job.validate` callback.

I could be misunderstanding the role of the users map though, so just correct me if I'm wrong.

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,271 @@
...
+                                       "user belongs to two or more banks; " \
+                                       "please specify which bank to charge");
+
+    return 0;
+}
+
+
+char v[] = "job.validate";
+char p[] = "job.state.priority";
+char g[] = "job.priority.get";
```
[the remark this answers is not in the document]
cmoussa1: Thanks for pointing this out - the reason I defined them is because C++ gave me a warning about converting strings to `char *` if I just use the strings directly in `tab[]`:

```console
warning: ISO C++ forbids converting a string constant to ‘char*’
```

Is it OK to ignore this warning?

src/plugins/mf_priority.cpp
cmoussa1: Ah, yes I think you're right - `validate_cb ()` would have rejected the job first if the user does not exist in the DB, so I don't think there is a need for the `flux_jobtap_reject_job ()` in `priority_calculation ()`; I can remove that.

src/plugins/mf_priority.cpp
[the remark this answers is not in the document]
grondo: Ah, this is because the topic string in `struct flux_plugin_handler` is `char *topic` instead of `const char *topic`. This is a bug in the `flux_plugin_t` interface. I'll open a bug in flux-core and fix it, and then those warnings should be resolved.

Sorry about that!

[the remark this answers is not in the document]
cmoussa1: No worries! Thanks for helping me walk through that. 🙂

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,271 @@
...
+                  LOG_ERR,
+                  "flux_plugin_arg_unpack: %s",
+                  flux_plugin_arg_strerror (args));
+        return -1;
+    }
+
+    // searching element in std::map by key
+    it = users.find (userid);
+    if (it == users.end ())
+        return flux_jobtap_reject_job (p, args,
```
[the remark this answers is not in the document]
cmoussa1: Thank you for the reminder - I'll go ahead and add a test for this.

cmoussa1:
Thanks for your feedback so far @grondo - maybe it makes sense to continue talking out my design decision regarding the data structure used in the plugin before I make more changes.

> It may be easier if you had two maps -- one that maps banks to fairshare value, and one that maps userid to a list of banks in which the user is a member. Then you can easily test bank membership, and get the bank fairshare value in a single lookup call.
>
> I'd also suggest caching the bank along with the job, perhaps with flux_jobtap_job_aux_set() so that you don't have to continually lookup the bank name for each job when you've already done it in the job.validate callback.
>
> I could be misunderstanding the role of the users map though, so just correct me if I'm wrong.

So the idea of using a multimap data structure was to make use of the ability to have multiple entries with the same key in the data structure, i.e there could potentially be two entries in the multimap that have a key of `userA`, reflecting the scenario where a user who belongs to two banks `bank1` and `bank2`, each with their own respective fairshare values. 

```
key      value
userA -> <'bank1', 0.12345>
userA -> <'bank2', 0.54321>
.
.
.
userB -> <'bank1', 0.98765>
```

Using a multimap is the reason I needed to implement a `clear_map_cb ()`, because if I didn't clear it, old entries wouldn't be overwritten during an update. I hope this explanation makes sense. 

Maybe it would be better to use your suggestion and use two maps, although I am trying to think of how to look up an individual user's fairshare value just using the bank they belong to.

src/plugins/mf_priority.cpp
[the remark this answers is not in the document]
grondo: A fix is up for this at flux-framework/flux-core#3720

grondo:
> Maybe it would be better to use your suggestion and use two maps, although I am trying to think of how to look up an individual user's fairshare value just using the bank they belong to.

Ah, does each user,bank pair have a different fairshare value? Sorry, I'm not familiar with the overall architecture here so I probably shouldn't comment on specific design decisions!

> Using a multimap is the reason I needed to implement a clear_map_cb (), because if I didn't clear it, old entries wouldn't be overwritten during an update. I hope this explanation makes sense.

Yes, however, I was mainly pointing out that having separate "clear" and "update" RPCs makes the plugin racy, since a number of things could happen in between the two RPCs (e.g. a job submission), whereas if you perform the clear in the update rpc, then you are safe from races since the job manager module/thread cannot do any asynchronous work while in your callback.

Does that make sense? I would just combine the separate clear and update service callbacks into a single clear+update callback. You could make the clear optional in the future if necessary by adding a `clear` boolean flag to your update RPC.

cmoussa1:
> Ah, does each user,bank pair have a different fairshare value? Sorry, I'm not familiar with the overall architecture here so I probably shouldn't comment on specific design decisions!

No, I appreciate you asking these questions and helping me think through these things! Yes, user-bank pairs should each have their own fairshare values.

> Yes, however, I was mainly pointing out that having separate "clear" and "update" RPCs makes the plugin racy, since a number of things could happen in between the two RPCs (e.g. a job submission), whereas if you perform the clear in the update rpc, then you are safe from races since the job manager module/thread cannot do any asynchronous work while in your callback.
>
> Does that make sense? I would just combine the separate clear and update service callbacks into a single clear+update callback. You could make the clear optional in the future if necessary by adding a clear boolean flag to your update RPC.

And yeah, I think this makes sense. I should combine the two into just one clear-and-update RPC. 

I believe I'll need to think about how the `clear ()` occurs in this callback, because as of now, the external service Python script sends one RPC for every user-bank pair, so if I only put a `clear ()` at the top of the update RPC, the map will get cleared every time a payload gets sent and will only contain the most recent JSON payload.

grondo:
> I believe I'll need to think about how the clear () occurs in this callback, because as of now, the external service Python script sends one RPC for every user-bank pair, so if I only put a clear () at the top of the update RPC, the map will get cleared every time a payload gets sent and will only contain the most recent JSON payload.

It will be much better if all updates are sent in a single RPC if we can figure out how to do that. The mass of RPCs for a large system will be inefficient, plus the race condition you've noted with clearing the multimap first.

One thing you could do is instead of using a multimap, use a map of maps (or hash of hashs). The outer map is would be by userid, and the inner map would be by bank name. This would allow you to _update_ existing bank entries instead of clearing and re-adding them. Then the bulk-update tool could send a single RPC with a JSON object which contained all updated user/bank pairs, and the update RPC would just iterate over those and do the updates.

cmoussa1:
> One thing you could do is instead of using a multimap, use a map of maps (or hash of hashes). The outer map is would be by userid, and the inner map would be by bank name. This would allow you to update existing bank entries instead of clearing and re-adding them. Then the bulk-update tool could send a single RPC with a JSON object which contained all updated user/bank pairs, and the update RPC would just iterate over those and do the updates.

Yup, absolutely. I don't think a `clear_map` callback would be required at all if we use a map of maps. And I think accessing values from this map of map feels more straightforward, so this might be the way to go.

The one drawback that I believe comes with using a map of maps arises when trying to access a "default" bank for a user. I think C++ maps are internally ordered by the key, so defaulting to the first bank found under the `userid` key could potentially change when new banks are added. 

I wonder if adding a field in the user table denoting which bank should be treated as the default bank would be a good way to handle this. I could then send that data in the external update service, and in the internal map where the key is `bank`, I could lookup this default boolean value to determine if the bank is the default one.

grondo:
> I wonder if adding a field in the user table denoting which bank should be treated as the default bank would be a good way to handle this. I could then send that data in the external update service, and in the internal map where the key is bank, I could lookup this default boolean value to determine if the bank is the default one.

Or, you could reserve the bank name "default" and store a copy or reference of the default bank under `users[userid]["default"]`.

cmoussa1:
Thanks for the helpful review so far - I've just pushed up a round of `fixup` commits that I hope address the feedback given:

- `fluxplugin_ldflags` was copied over from flux-core for compiling the plugin
- the data structure in the plugin was changed from a `multimap` to a `map` of `map`s; the outer layer has a key of `userid`, and the inner layer has a key of `bank` and respective `fairshare` value in that bank. A `"default"` bank entry is also created for the user when they are added to the plugin's data structure. The `"default"` bank uses the first bank the user was added to in the flux-accounting database. I think a TODO for this PR would be to add support for changing `"default"` banks both in the flux-accounting database and in the plugin itself.
- Because of the change to the data structure used in the plugin ⬆️, the `clear_map` callback is no longer needed, since the map can simply be updated with new values without having to clear out the keys first.
- `flux_jobtap_reject_job ()` was removed from `job.priority.get` since the plugin already checks for the user's existence in the flux-accounting database during the `job.validate` state. 
- Two new tests were added: one to confirm that a job would be rejected in the `job.validate` state if they did not belong in the flux-accounting DB, and one that uses the default bank when submitting a job

cmoussa1:
Just squashed my `fixup` commits and re-pushed.

t/t1001-mf-priority.t
```diff
@@ -0,0 +1,90 @@
...
+'
+
+test_expect_success 'update plugin with sample test data' '
+        flux python fake_payload.py
+'
+
+test_expect_success 'submit a job with default urgency' '
+        jobid=$(flux mini submit --setattr=system.bank=account3 -n1 hostname) &&
+        flux job wait-event -f json $jobid priority | jq '.context.priority' > test_job1.txt &&
+        test_cmp test_job1.txt ${FLUX_BUILD_DIR}/t/expected/job_1_priority.txt
```
grondo, line 56: Instead of storing expected output separately from the test under version control, it is probably easier and more maintainable to create the expected output within the test itself (you'll see the pattern quite often in flux-core tests), e.g. here:
```
test_expect_success 'submit a job with default urgency' '
￼        jobid=$(flux mini submit --setattr=system.bank=account3 -n1 hostname) &&
￼        flux job wait-event -f json $jobid priority | jq '.context.priority' > test_job1.txt &&
         cat >job1.expected <<-EOF
         4096
         EOF
￼        test_cmp job1.expected test_job1.txt
```

Also note that usage of `test_cmp` is `test_cmp expected actual`.

I would use this method for almost all tests unless the expected output is very large.

t/t1001-mf-priority.t
```diff
@@ -0,0 +1,90 @@
...
+        flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
+'
+
+test_expect_success 'list all loaded plugins' '
+        flux jobtap list -a > test.txt
+'
+
+test_expect_success 'make sure multi-factor priority plugin is loaded' '
+        test_cmp test.txt ${FLUX_BUILD_DIR}/t/expected/loaded_priority.txt
+'
```
grondo, line 25: Suggestion: do not use `test_cmp` to test output of `flux jobtap list` because you never know, some default plugin besides just your priority plugin could be loaded. Instead, combine these two tests into one and use `grep` to ensure your plugin was loaded:

```
test_expect_success 'make sure multi-factor priority plugin is loaded' '
        flux jobtap list | grep mf_priority
'
```

Remove `loaded_priority.txt` from version control

grondo (commented):
Just a quick few comments on the tests. I'd remove all "expected output" from version control and generate or check expected output within the tests (see comments inline)

t/t1001-mf-priority.t
cmoussa1: Thanks for pointing this out. I've just pushed a change to use the above pattern to compare the test result to the expected value. This feels much cleaner than the approach I was originally using.

t/t1001-mf-priority.t
cmoussa1: Ah, good point. Thank you. I've just pushed up a change in the test to just use `grep` to make sure my plugin is loaded.

cmoussa1:
Just pushed to catch up after #124 and #126.

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,230 @@
...
+    assert (sizeof arg == sizeof p);
+    memcpy (&p, &arg, sizeof arg);
+
+    char *uid, *fshare, *bank;
+
+    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s}",
+                             "userid", &uid,
+                             "bank", &bank,
+                             "fairshare", &fshare) < 0) {
+        flux_log_error (h, "failed to unpack custom_priority.trigger msg");
```
grondo, line 109: It would probably be good to add a test for this error condition. Just send an empty payload or other invalid input to the mf_priority plugin's service and ensure that the RPC fails expectedly, and that the job manager or instance doesn't crash ;-)

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,230 @@
...
+        flux_log (h,
+                  LOG_ERR,
+                  "flux_plugin_arg_unpack: %s",
+                  flux_plugin_arg_strerror (args));
+        return -1;
+    }
+
+    // searching element in std::map by key
+    it = users.find (userid);
+    if (it == users.end ())
```
grondo, line 204: Should `job.validate` also check that the `bank`, if provided, is valid for the given user?

src/plugins/mf_priority.cpp
cmoussa1: Does sending an empty or invalid input payload causes an error to happen just before it gets to the `flux_log_error` message? For example, if an empty payload is sent:

```
  File "bad_payload.py", line 8, in <module>
    flux.Flux().rpc("job-manager.mf_priority.rec_update", {}).get()
  File "/usr/lib64/flux/python3.6/flux/rpc.py", line 78, in get
    resp_str = self.get_str()
  File "/usr/lib64/flux/python3.6/flux/util.py", line 76, in func_wrapper
    retval = func(future, *args, **kwargs)
  File "/usr/lib64/flux/python3.6/flux/rpc.py", line 72, in get_str
    self.pimpl.flux_rpc_get(payload_str)
  File "/usr/lib64/flux/python3.6/flux/util.py", line 52, in func_wrapper
    raise EnvironmentError(error.errno, errmsg.decode("utf-8")) from None
OSError: [Errno 71] Object item not found: userid
```

Or if an invalid payload gets sent:

```
  File "bad_payload.py", line 8, in <module>
    flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
  File "/usr/lib64/flux/python3.6/flux/rpc.py", line 78, in get
    resp_str = self.get_str()
  File "/usr/lib64/flux/python3.6/flux/util.py", line 76, in func_wrapper
    retval = func(future, *args, **kwargs)
  File "/usr/lib64/flux/python3.6/flux/rpc.py", line 72, in get_str
    self.pimpl.flux_rpc_get(payload_str)
  File "/usr/lib64/flux/python3.6/flux/util.py", line 52, in func_wrapper
    raise EnvironmentError(error.errno, errmsg.decode("utf-8")) from None
OSError: [Errno 71] Expected string, got integer
```

I could very well be doing something wrong though. Is there a proper way to handle this failed unpack when a payload is empty or of invalid format?

grondo: That looks like the expected result to me, since you send the result of `flux_msg_last_error()` in the error response. You may have to run `flux dmesg` to see the logged error.

grondo (changes requested):
This is looking good! Just a few pretty minor comments in line. This time I focused on the job manager plugin specifically.

I think this is very close to ready for a merge :+1:

src/plugins/mf_priority.cpp
cmoussa1: Ah, okay, thanks for pointing that out. I haven't used `flux dmesg` before, so apologies if I am totally using it wrong. I just pushed a `--fixup` commit (6c27d7a) that adds a test that uses `flux dmesg` to make sure that `flux_log_error` reports the right message.

src/plugins/mf_priority.cpp
cmoussa1: Yes, you're right. I just pushed up a `fixup` commit (e92974b) that adds another check in `validate_cb ()` where the passed-in bank is checked under the user, and if it is not found, the job will be rejected with an error message saying that the user does not belong to the passed-in bank. I also added a test (4cc116c) that should trigger this error message.

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,230 @@
...
+ * multimap datastructure.
+ */
+static void rec_update_cb (flux_t *h,
+                           flux_msg_handler_t *mh,
+                           const flux_msg_t *msg,
+                           void *arg)
+{
+    flux_plugin_t *p;
+    assert (sizeof arg == sizeof p);
+    memcpy (&p, &arg, sizeof arg);
```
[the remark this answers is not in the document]
cmoussa1: Thanks for letting me know about this. Just using:

```
flux_plugin_t *p = arg;
```

was giving me an error; I _think_ it was because the C++ compiler complains about an implicit conversion type to `void *`. I referenced [this](https://stackoverflow.com/questions/30088025/malloc-invalid-conversion-from-void-to-double) SO post. Let me know if I am way off base here. 

Regardless, I dropped those lines because, like you mentioned, they are not used anywhere in the function.

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    char *uid, *fshare, *bank;
+
+    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s}",
```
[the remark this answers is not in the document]
cmoussa1: Ah, I don't have an issue open on this, but I should open one so I can track it. Maybe I can work on this along with the multi-user testing for the planned July release.

src/plugins/mf_priority.cpp
```diff
@@ -0,0 +1,230 @@
...
+    std::map<int, std::map<std::string, double>>::iterator it;
+
+    flux_t *h = flux_jobtap_get_flux (p);
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:i, s{s{s{s?s}}}}",
+                                "userid", &userid,
+                                "jobspec", "attributes", "system",
+                                "bank", &bank) < 0) {
+        flux_log (h,
```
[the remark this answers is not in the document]
cmoussa1: This sounds like a good idea. I pushed up a `fixup` commit (0b80bd2) that just rejects the job if unpacking the bank fails instead of logging an error. I also added a test (c9d2587) that should trigger this rejection message when an invalid format is passed in, like your example above where `setattr=system.bank=1`.

src/plugins/mf_priority.cpp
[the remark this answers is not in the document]
grondo: Ah, sorry, C++ must be less forgiving of type conversion. You would want to do something like this instead:
```C
    flux_plugin_t *p = (flux_plugin_t *) arg;
```

Though I didn't test that with a c++ compiler so I could be wrong again 😉

src/plugins/mf_priority.cpp
grondo: LGTM. I think you are using it correctly. In this case, you are getting an error back from the RPC which indicates the incorrect payload was caught by your plugin, so IMO it would be sufficient to ensure the `bad_payload.py` script fails with an expected error message.

grondo:
Looks good! Feel free to squash all the fixup commits and I can give the PR one more pass.

cmoussa1:
Thanks! Just squashed my `fixup` commits and re-pushed.

grondo (approved):

src/plugins/bulk_update.py
```diff
@@ -0,0 +1,65 @@
+import flux
```
grondo, line 1: Should this file have a copyright header? I'm actually unsure, but might as well throw it in here just in case.

src/test/docker/bionic/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fluxrm/flux-core:bionic
+FROM fluxrm/flux-core:bionic-v0.27.0
```
grondo, line 1: Update to v0.28.0 tag?

t/t1001-mf-priority.t
```diff
@@ -0,0 +1,127 @@
...
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 1 job
+
+flux setattr log-stderr-level 1
+
+test_expect_success 'load multi-factor priority plugin' '
+        flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
+'
+
+test_expect_success 'list all loaded plugins' '
```
grondo, line 19:
```suggestion
test_expect_success 'check that mf_priority plugin is loaded' '
```

t/t1001-mf-priority.t
```diff
@@ -0,0 +1,127 @@
...
+'
+
+test_expect_success 'send an empty payload to make sure unpack fails' '
+cat <<-EOF >bad_payload.py
+import flux
+
+#create a JSON payload
+flux.Flux().rpc("job-manager.mf_priority.rec_update", {}).get()
+EOF
+flux python bad_payload.py
```
grondo, line 36: Check that `bad_payload.py` fails with `test_must_fail`.

Also, missing `&&` between these two commands. (You might try running tests with `--chain-lint` to see if there are any other missing `&&`-chains.)


```suggestion
test_must_fail flux python bad_payload.py &&
```

t/t1001-mf-priority.t
```diff
@@ -0,0 +1,127 @@
...
+'
+
+test_expect_success 'update plugin with sample test data' '
+        flux python fake_payload.py
+'
+
+test_expect_success 'submit a job with default urgency' '
+        jobid=$(flux mini submit --setattr=system.bank=account3 -n1 hostname) &&
+        flux job wait-event -f json $jobid priority | jq '.context.priority' > job1.test &&
+        cat <<-EOF >job1.expected
```
grondo, line 63: missing `&&`
```suggestion
        cat <<-EOF >job1.expected &&
```

t/t1001-mf-priority.t
```diff
@@ -0,0 +1,127 @@
...
+        45321
+        EOF
+        test_cmp job1.expected job1.test
+'
+
+test_expect_success 'submit a job with custom urgency' '
+        jobid=$(flux mini submit --setattr=system.bank=account3 --urgency=15 -n1 hostname) &&
+        flux job wait-event -f json $jobid priority | jq '.context.priority' > job2.test &&
+        cat <<-EOF >job2.expected
```
grondo, line 72:
```suggestion
        cat <<-EOF >job2.expected &&
```

grondo (commented):
Some more very minor comments this round.

You should ensure that `./t1001-mf-priority --chain-lint` runs correctly. I think there are some missing `&&` after uses of `cat` in most of the tests (an easy fix) I just called out a couple cases.

src/plugins/bulk_update.py
cmoussa1: Oh yeah, thank you for pointing this out. I just pushed up a fix that adds a copyright header. Sorry for forgetting to put one. 🤦‍♂️

src/test/docker/bionic/Dockerfile
cmoussa1: Yup, right on time with flux-core! Just pushed up a change that updates this to `v0.28.0`.

t/t1001-mf-priority.t
cmoussa1: It looks like `--chain-lint` fails unexpectedly when appending `&&` after these lines (`cat <<-EOF > job.expected`), but doesn't fail when I leave them out. The most recently pushed of `t1001-mf-priority.t` passes with `--chain-lint` though:

```
[moussa1@docker-desktop t]$ ./t1001-mf-priority.t --chain-lint
ok 1 - load multi-factor priority plugin
ok 2 - check that mf_priority plugin is loaded
ok 3 - try to submit a job when user does not exist in DB
ok 4 - send an empty payload to make sure unpack fails
ok 5 - create fake_payload.py
ok 6 - update plugin with sample test data
ok 7 - submit a job with default urgency
ok 8 - submit a job with custom urgency
ok 9 - submit a job with urgency of 0
ok 10 - submit a job with urgency of 31
ok 11 - submit a job with other bank
ok 12 - submit a job using default bank
ok 13 - submit a job using a bank the user does not belong to
ok 14 - reject job when invalid bank format is passed in
# passed all 14 test(s)
1..14
```

grondo: Hm, sorry about the apparent misinformation, but let me check into this.

cmoussa1: No worries - it could definitely be something I'm doing wrong. I was pretty confused that `--chain-lint` was complaining about the following line:

```
cat <<-EOF >bad_payload.py &&
```

which, if no `&&` is written, dumps out this error:

```
error: bug in the test script: broken &&-chain: 
cat <<-EOF >bad_payload.py
import flux
```

but not 

```
cat <<-EOF >job1.expected
```

which, if `&&` is written, dumps out this:

```
FATAL: Unexpected exit with code 2
flux-start: 0 (pid 16667) exited with rc=1
```

grondo: Ok, the problem at least in some of these cases is that you are using 8 spaces instead of tabs.
Specifically, `<<-EOF` tells the shell to ignore TABS up to the next `<tab>EOF`. If you convert to tabs (which I think we've settled on for sharness tests) then I think you'll be able to add the `&&` to the `cat` lines.

The problem with leaving this `&&` off is that the commands before the `cat <<-EOF` could fail, but the test itself could succeed.

cmoussa1: Ahhh, okay. Yeah, I must have been converting tabs to spaces when I was saving the `.t` in my IDE. I'm so sorry for the confusion! The most recent push should have the `&&`'s added to the sharness tests.

cmoussa1: That would also probably explain why it would complain about an unexpected indent when creating the Python scripts inside the sharness tests. I should push up and squash up a fix that properly indents those sharness tests.

grondo: yeah, the indents between the `cat <<-EOF` and the `EOF` have to be tabs. I think without that the `cat` will read all the way to the closing quote of the `test_expect_success()` -- so you aren't actually running the test! :-)

cmoussa1: Oh my goodness 😳, well, thank you for catching that and informing me why that happens. This latest push should have replaced the spaces with tabs to prevent the tests from not running.

t/t1001-mf-priority.t
```diff
@@ -0,0 +1,127 @@
...
+# BULK_UPDATE=${FLUX_BUILD_DIR}/t/scripts/send_fake_payloads.py
+
+export TEST_UNDER_FLUX_NO_JOB_EXEC=y
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 1 job
+
+flux setattr log-stderr-level 1
+
+test_expect_success 'load multi-factor priority plugin' '
+		flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
```
grondo, line 16: Hm, now all the indents seem to be 2 tabs? 
```suggestion
	flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
```

cmoussa1: Oops, I had always had the indents at 2 tabs. These tests should now all be indented at one tab (4 spaces).

grondo: Yeah, it is confusing. I have my editor set to 8 space tabs (at least for the sharness tests), so for me the lines were indented 16 spaces! A single tab was probably plenty. :slightly_smiling_face:

grondo (approved):
This LGTM! Feel free to set MWP when ready as far as I'm concerned. Thanks for all the work!

cmoussa1:
Thanks @grondo for all of your feedback and help!

cmoussa1:
Oh, I think this might have to be manually merged since I made a change to `.github/workflows/main.yml`.

grondo:
Ah yes, sorry I missed that. I'll go ahead and press the merge button.

cmoussa1:
Thank you!

PR #124, opened 2021-06-24
association_table: remove unused fields

cmoussa1:
##### Problem

The `association_table` currently contains two fields that have not been used with any features of flux-accounting: `max_jobs` and `max_wall_pj`. As a result of creating a set of smaller issues to handle per-user, per-job, and per-queue limits, it looks like these kinds of limits will either be handled by a mechanism other than the flux-accounting DB (I opened issue [#3652](https://github.com/flux-framework/flux-core/issues/3652) in flux-core to discuss part of these per-job limits) or changed in some way to be handled within flux-accounting. 

##### Solution

This PR just removes the `max_jobs` and `max_wall_pj` fields from the `association_table` to prepare for adding per-user limit support in flux-accounting.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/accounting_cli_functions.py
+++ b/src/bindings/python/fluxacct/accounting/accounting_cli_functions.py
@@ -156,7 +156,7 @@ def view_user(conn, user):
         print(e_database_error)
 
 
-def add_user(conn, username, bank, admin_level=1, shares=1, max_jobs=1, max_wall_pj=60):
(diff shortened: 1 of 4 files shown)


cmoussa1:
OK, I think I can take this PR out of **[WIP]** and post it for review now.

SteVwonder (approved):
LGTM! Thanks @cmoussa1!

cmoussa1:
Thanks! 🙂

PR #126, opened 2021-06-29
python: break up accounting_cli_functions.py

cmoussa1:
##### Problem

`accounting_cli_functions.py` contains functions for both users and banks, resulting in a pretty large source file. It would probably be cleaner to break this up into two smaller files: one that contains user subcommands and one that contains bank subcommands.

---

This PR does just that; it breaks up `accounting_cli_functions.py` into two files: `user_subcommands.py` and `bank_subcommands.py` and reflects the new changes wherever the Python modules are used.


SteVwonder (approved):
LGTM! Thanks @cmoussa1!

cmoussa1:
Thanks! 🙂

PR #130, opened 2021-07-13
association_table: add "default_bank" column, change plugin's default bank behavior

### LINKED ISSUE #123: support "default" bank for users
Something that has come up while working on #122 is the need for setting a `"default"` bank for users when they submit jobs. It's probably unrealistic for the user to specify the bank they want to charge _every time_ they want to run a job, especially if they only belong to one bank.

The `"default"` bank support in #122 currently treats the first bank the user belonged to in the flux-accounting database as the default bank. It cannot be changed once the user gets added to the DB. 

It'd be useful to add support in both the DB and the plugin for changing the `"default"` bank for a user after they are initially added - this would probably come in the form of a `flux account` admin tool, maybe via the command line.

cmoussa1:
### Background

As discussed in #123, the "default" bank support added in #122 currently treats the first bank the user belonged to in the flux-accounting database as the default bank. It cannot be changed once the user gets added to the DB. It would be good to add more robust default bank support in both the flux-accounting database and the priority plugin, where a user's default bank could be changed by a sys admin even after they are initially added.

### Implementation

This PR adds a `default_bank` column to the `association_table` in the flux-accounting database. When a user/bank combination gets added to the database, a lookup is performed to see if the user already exists in the table and, if they do, what their default bank is. It will set the `default_bank` to the value of the first bank they are added to; however, this value can now be changed like most other fields in the `association_table` with the `edit-user` subcommand.

The `default_bank` column is now also added to the payload information that is sent to the multi-factor priority plugin. When the plugin extracts this data, it will look at the `default_bank` key-value pair and compare it to the `bank` key-value pair in the same payload using `strcmp`. If the values are the same, it will add the fairshare information to the "default" key in the map data structure in the plugin. 

This is similar to the behavior previously present in the plugin, except now, if there is ever an update to a user's default bank in the flux-accounting DB, that change should be reflected in 1) the new payloads sent by `bulk_update.py`, and 2) the map data structure that caches user-bank-fairshare information.

Fixes #123

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -96,11 +96,12 @@ static void rec_update_cb (flux_t *h,
                            const flux_msg_t *msg,
                            void *arg)
 {
-    char *uid, *fshare, *bank;
+    char *uid, *fshare, *bank, *default_bank;
 
-    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s}",
+    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s, s:s}",
                              "userid", &uid,
                              "bank", &bank,
+                             "default_bank", &default_bank,
                              "fairshare", &fshare) < 0) {
         flux_log_error (h, "failed to unpack custom_priority.trigger msg");
         goto error;
@@ -109,9 +110,7 @@ static void rec_update_cb (flux_t *h,
     if (flux_respond (h, msg, NULL) < 0)
         flux_log_error (h, "flux_respond");
 
-    // if the user being added to the does not yet have any entries in the map,
-    // treat their first bank as the "default" bank
-    if (users.count (std::atoi (uid)) == 0)
+    if (strcmp (bank, default_bank) == 0)
         users[std::atoi (uid)]["default"] = std::stod (fshare);
 
     users[std::atoi (uid)][bank] = std::stod (fshare);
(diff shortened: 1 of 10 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -96,11 +96,12 @@ static void rec_update_cb (flux_t *h,
                            const flux_msg_t *msg,
                            void *arg)
 {
-    char *uid, *fshare, *bank;
+    char *uid, *fshare, *bank, *default_bank;
 
-    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s}",
+    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s, s:s}",
```
grondo, line 101: Just an idea/question: It might be slightly awkward here that the protocol requires that the `default_bank` is sent along with every `bank` update. Unless I've missed something below, if `default_bank` != `bank`, then `default_bank` is unused. 

Another idea would be to have `default_bank` be a boolean, but I'm not sure if this would make the client side more difficult.

Also, I realize you have an issue open to turn this simpler protocol into a "bulk update" protocol, so that you can update multiple users in one RPC. If that is the case, then maybe a cleanup of the RPC design could be done at that time. No reason to fix the current design if it works for now.

grondo: Ok, actually I see now above that this protocol simplifies the current bulk_update.py client, so it is probably fine to leave as is for now.

grondo (approved):
This looks good to me! Nice and simple approach.

I made one comment inline, but after re-reviewing the code I noted that it seems fine as is. I left the comment just to get us thinking about what a real "bulk" update protocol might look like.

cmoussa1:
Thanks for the review @grondo - I should copy your thoughts over to #133 so that I don't forget them, and then I will set MWP.

Issue #121, opened 2021-05-13
enforcing per-user limits


cmoussa1:
_If I need to instead open this issue in a different repository, please let me know and I'll move it._

The second issue that came up as a result of breaking up [flux-sched #638](https://github.com/flux-framework/flux-sched/issues/638) is the need to enforce per-user limits in the following ways:

* max number of nodes used across all of a user's running jobs
* max number of running jobs (in practice, the "max number of nodes used across all of a user's running jobs" limit is used)
* max number of submitted jobs per user

Currently, these limits are not defined in the flux-accounting database, so a good first step would be to add additional fields for every association in the database:

```sql
max_nodes           int(11)     DEFAULT 5   NOT NULL,
max_submitted_jobs  int(11)     DEFAULT 5   NOT NULL
```

I've just picked a default value of `5` for both limits, but we should probably come to a consensus as to what makes sense for a default value for users.

We'll also need a way to pass this information where appropriate where these limits can be:

a) compared to a user's current usage both in terms of how many nodes they are using and how many jobs they have submitted, and
b) enforced if a user has reached either of these limits.

grondo:
Max submitted jobs per-user _could_ be enforced by a jobtap plugin (either the multifactor fairshare plugin or a standalone plugin). The plugin can keep a count of currently active jobs per user and reject any job that would exceed the maximum in the `job.validate` callback.

cmoussa1:
#131 has now landed, so this issue is halfway to being completed. An active jobs count is tracked for users in the flux-accounting DB and will reject any job that would exceed the maximum. Now what is left is to add a `max_nodes` limit, which is the max number of nodes used across all of a user/bank combination's running jobs.

cmoussa1:
I briefly started to take a look at the requirements for adding a `max_nodes`  limit today and I figured I'd write some initial thoughts down so I don't forget them later:

- thanks to the great and detailed feedback I got in #131, adding a `max_nodes` should be _fairly_ straightforward since that PR did a lot of the leg work to create a `bank_info` struct for each user/bank combination in the flux-accounting DB. After adding a `max_nodes` column in the DB, I can add a new item to the struct (similarly named `max_nodes`) and send/receive it with the other user/bank information  via `flux_jobtap_job_aux_set/get ()`.
[...]

PR #131, opened 2021-07-20
plugin: add a per-user/bank max jobs limit

cmoussa1:
~~_note: this PR is built off of the changes proposed in #130, so I will leave this one as [WIP] until that one lands._~~

~~_#130 has landed, but there are a couple of questions that might alter the direction this PR takes, so I think I'll leave this PR as [WIP] until those questions are sorted out._~~

### Background

The multi-factor priority plugin added in #122 currently just calculates a user's job priority and does not enforce any kind of per-user limits like the ones laid out in #121. Some of the limits that flux-accounting has a direct impact on enforcing are the following:

**max_jobs:** max number of submitted jobs per user

**max_nodes:** max number of nodes used across all of a user's running jobs

This PR looks to add enforcement of the **max_jobs** per-user/bank combination limit to the priority plugin.

### Implementation

This PR looks to add limit support to the priority plugin by enforcing a limit on the maximum number of submitted jobs per user/bank combination.

In the flux-accounting DB, a new column is added to the `association_table` called **max_jobs**, which represents the maximum number of submitted jobs per user:

```sql
max_jobs  int(11)     DEFAULT 5   NOT NULL
```

In the plugin, this new **max_jobs** value is unpacked and stored in a new map-of-maps data structure, very similar to the one that stores user/bank/fairshare information. This value is then compared to the current number of submitted jobs, represented by another integer in the data structure. When a job is submitted and enters `job.validate`, the current number of submitted jobs is compared to the user's limit and is incremented if the user is under their limit. Once a job finishes running and enters `job.state.inactive`, the user's active count of submitted jobs is decremented. The decrement happens in a new callback added to the plugin called `inactive_cb ()`.

If a user has reached their limit and tries to submit another job, the job will be rejected with an error message saying that their limit has been reached.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -26,7 +26,14 @@ extern "C" {
 #include <algorithm>
 #include <cinttypes>
 
-std::map<int, std::map<std::string, double>> users;
+std::map<int, std::map<std::string, struct bank_info>> users;
+std::map<int, std::string> users_def_bank;
+
+struct bank_info {
+    double fairshare;
+    int max_jobs;
+    int current_jobs;
+};
 
 /******************************************************************************
  *                                                                            *
@@ -49,9 +56,7 @@ int64_t priority_calculation (flux_plugin_t *p,
 {
     double fshare_factor = 0.0, priority = 0.0;
     int fshare_weight;
-
-    std::map<int, std::map<std::string, double>>::iterator it;
-    std::map<std::string, double>::iterator inner_it;
+    struct bank_info *b;
 
     fshare_weight = 100000;
 
@@ -61,19 +66,19 @@ int64_t priority_calculation (flux_plugin_t *p,
     if (urgency == FLUX_JOB_URGENCY_EXPEDITE)
         return FLUX_JOB_PRIORITY_MAX;
 
-    // search element in map of maps by key
-    it = users.find (userid);
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
 
-    // search for correct fshare value using passed-in bank; otherwise, use
-    // a default bank
-    if (bank != NULL) {
-        inner_it = it->second.find (bank);
-        fshare_factor = inner_it->second;
+    if (b == NULL) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
+                                     3, "mf_priority: bank info is missing; "
+                                        "holding job");
+        return 0;
     }
-    else {
-        inner_it = it->second.find ("default");
-        fshare_factor = inner_it->second;
-    }
+
+    fshare_factor = b->fairshare;
 
     priority = (fshare_weight * fshare_factor) + (urgency - 16);
 
@@ -96,13 +101,14 @@ static void rec_update_cb (flux_t *h,
                            const flux_msg_t *msg,
                            void *arg)
 {
-    char *uid, *fshare, *bank, *default_bank;
+    char *uid, *fshare, *bank, *default_bank, *max_jobs;
 
-    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s, s:s}",
+    if (flux_request_unpack (msg, NULL, "{s:s, s:s, s:s, s:s, s:s}",
                              "userid", &uid,
                              "bank", &bank,
                              "default_bank", &default_bank,
-                             "fairshare", &fshare) < 0) {
+                             "fairshare", &fshare,
+                             "max_jobs", &max_jobs) < 0) {
         flux_log_error (h, "failed to unpack custom_priority.trigger msg");
         goto error;
     }
@@ -110,10 +116,13 @@ static void rec_update_cb (flux_t *h,
     if (flux_respond (h, msg, NULL) < 0)
         flux_log_error (h, "flux_respond");
 
-    if (strcmp (bank, default_bank) == 0)
-        users[std::atoi (uid)]["default"] = std::stod (fshare);
+    struct bank_info *b;
+    b = &users[std::atoi(uid)][bank];
 
-    users[std::atoi (uid)][bank] = std::stod (fshare);
+    b->fairshare = std::atof (fshare);
+    b->max_jobs = std::atoi (max_jobs);
+
+    users_def_bank[std::atoi (uid)] = default_bank;
 
     return;
 error:
@@ -178,8 +187,11 @@ static int validate_cb (flux_plugin_t *p,
 {
     int userid;
     char *bank = NULL;
-    std::map<int, std::map<std::string, double>>::iterator it;
-    std::map<std::string, double>::iterator inner_it;
+    int current_jobs, max_jobs = 0;
+    double fairshare = 0.0;
+
+    std::map<int, std::map<std::string, struct bank_info>>::iterator it;
+    std::map<std::string, struct bank_info>::iterator bank_it;
 
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
@@ -191,18 +203,84 @@ static int validate_cb (flux_plugin_t *p,
         return flux_jobtap_reject_job (p, args, "unable to unpack bank arg");
     }
 
-    // searching element in std::map by key
+    // make sure user belongs to flux-accounting DB
     it = users.find (userid);
     if (it == users.end ())
         return flux_jobtap_reject_job (p, args,
                                        "user not found in flux-accounting DB");
 
+    // make sure user belongs to bank they specified; if no bank was passed in,
+    // look up their default bank
     if (bank != NULL) {
-        inner_it = it->second.find (std::string (bank));
-        if (inner_it == it->second.end ())
+        bank_it = it->second.find (std::string (bank));
+        if (bank_it == it->second.end ())
             return flux_jobtap_reject_job (p, args,
-                                    "user does not belong to specified bank");
+                                     "user does not belong to specified bank");
+    } else {
+        bank = const_cast<char*> (users_def_bank[userid].c_str ());
+        bank_it = it->second.find (std::string (bank));
+        if (bank_it == it->second.end ())
+            return flux_jobtap_reject_job (p, args,
+                                     "user/default bank entry does not exist");
     }
+
+    max_jobs = bank_it->second.max_jobs;
+    current_jobs = bank_it->second.current_jobs;
+    fairshare = bank_it->second.fairshare;
+
+    // if a user's fairshare value is 0, that means they shouldn't be able
+    // to run jobs on a system
+    if (fairshare == 0)
+        return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
+
+    // make sure user has not already hit their max active jobs count
+    if (max_jobs > 0 && current_jobs >= max_jobs)
+        return flux_jobtap_reject_job (p, args,
+                                       "user has max number of jobs submitted");
+
+    if (flux_jobtap_job_aux_set (p,
+                                 FLUX_JOBTAP_CURRENT_JOB,
+                                 "mf_priority:bank_info",
+                                 &bank_it->second,
+                                 NULL) < 0)
+        flux_log_error (h, "flux_jobtap_job_aux_set");
+
+    bank_it->second.current_jobs++;
+
+    return 0;
+}
+
+
+static int inactive_cb (flux_plugin_t *p,
+                        const char *topic,
+                        flux_plugin_arg_t *args,
+                        void *data)
+{
+    int userid;
+    struct bank_info *b;
+
+    flux_t *h = flux_jobtap_get_flux (p);
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:i}",
+                                "userid", &userid) < 0) {
+        flux_log (h,
+                  LOG_ERR,
+                  "flux_plugin_arg_unpack: %s",
+                  flux_plugin_arg_strerror (args));
+        return -1;
+    }
+
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL)
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
+                                     3, "mf_priority: bank info is missing");
+
+    b->current_jobs--;
 
     return 0;
 }
@@ -212,6 +290,7 @@ static const struct flux_plugin_handler tab[] = {
     { "job.validate", validate_cb, NULL },
     { "job.state.priority", priority_cb, NULL },
     { "job.priority.get", priority_cb, NULL },
+    { "job.state.inactive", inactive_cb, NULL },
     { 0 },
 };
 
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -98,6 +98,7 @@ def create_db(
                 shares        int(11)     DEFAULT 1     NOT NULL,
                 job_usage     real        DEFAULT 0.0   NOT NULL,
                 fairshare     real        DEFAULT 0.5   NOT NULL,
+                max_jobs      int(11)     DEFAULT 5     NOT NULL,
                 PRIMARY KEY   (username, bank)
         );"""
     )
--- a/src/plugins/bulk_update.py
+++ b/src/plugins/bulk_update.py
@@ -48,7 +48,7 @@ def bulk_update(path):
 
     # fetch all rows from association_table (will print out tuples)
     for row in cur.execute(
-        "SELECT userid, bank, default_bank, fairshare FROM association_table"
+        "SELECT userid, bank, default_bank, fairshare, max_jobs FROM association_table"
     ):
         # create a JSON payload with the results of the query
         data = {
@@ -56,6 +56,7 @@ def bulk_update(path):
             "bank": str(row[1]),
             "default_bank": str(row[2]),
             "fairshare": str(row[3]),
+            "max_jobs": str(row[4]),
         }
 
         flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -5,7 +5,8 @@ TESTSCRIPTS = \
 	t1001-mf-priority-basic.t \
 	t1002-mf-priority-small-no-tie.t \
 	t1003-mf-priority-small-tie.t \
-	t1004-mf-priority-small-tie-all.t
+	t1004-mf-priority-small-tie-all.t \
+	t1005-max-jobs-limits.t
 
 dist_check_SCRIPTS = \
 	$(TESTSCRIPTS) \
--- a/t/t1001-mf-priority-basic.t
+++ b/t/t1001-mf-priority-basic.t
@@ -45,9 +45,9 @@ test_expect_success 'create fake_payload.py' '
 	username = getpass.getuser()
 	userid = pwd.getpwnam(username).pw_uid
 	# create a JSON payload
-	data = {"userid": str(userid), "bank": "account3", "default_bank": "account3", "fairshare": "0.45321"}
+	data = {"userid": str(userid), "bank": "account3", "default_bank": "account3", "fairshare": "0.45321", "max_jobs": "10"}
 	flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
-	data = {"userid": str(userid), "bank": "account2", "default_bank": "account3", "fairshare": "0.11345"}
+	data = {"userid": str(userid), "bank": "account2", "default_bank": "account3", "fairshare": "0.11345", "max_jobs": "10"}
 	flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
 	EOF
 '
--- a/t/t1001-mf-priority-basic.t
+++ b/t/t1001-mf-priority-basic.t
@@ -123,4 +123,28 @@ test_expect_success 'reject job when invalid bank format is passed in' '
 	grep "unable to unpack bank arg" invalid_fmt.out
 '
 
+test_expect_success 'create a fake payload with an empty fairshare key-value pair' '
+	cat <<-EOF >empty_fairshare.py
+	import flux
+	import pwd
+	import getpass
+
+	username = getpass.getuser()
+	userid = pwd.getpwnam(username).pw_uid
+	# create a JSON payload
+	data = {"userid": str(userid), "bank": "account4", "default_bank": "account3", "fairshare": "", "max_jobs": "10"}
+	flux.Flux().rpc("job-manager.mf_priority.rec_update", data).get()
+	EOF
+'
+
+test_expect_success 'update plugin with sample test data' '
+	flux python empty_fairshare.py
+'
+
+test_expect_success 'submit a job with new bank and 0 fairshare should result in a job rejection' '
+	test_must_fail flux mini submit --setattr=system.bank=account4 hostname > zero_fairshare.out 2>&1 &&
+	test_debug "zero_fairshare.out" &&
+	grep "user fairshare value is 0" zero_fairshare.out
+'
+
 test_done
--- a/t/t1002-mf-priority-small-no-tie.t
+++ b/t/t1002-mf-priority-small-no-tie.t
@@ -25,13 +25,13 @@ test_expect_success 'create a group of users with unique fairshare values' '
 	cat <<-EOF >fake_small_no_tie.json
 	{
 		"users" : [
-			{"userid": "5011", "bank": "account1", "default_bank": "account1", "fairshare": "0.285714"},
-			{"userid": "5012", "bank": "account1", "default_bank": "account1", "fairshare": "0.142857"},
-			{"userid": "5013", "bank": "account1", "default_bank": "account1", "fairshare": "0.428571"},
-			{"userid": "5021", "bank": "account2", "default_bank": "account2", "fairshare": "0.714286"},
-			{"userid": "5022", "bank": "account2", "default_bank": "account2", "fairshare": "0.571429"},
-			{"userid": "5031", "bank": "account3", "default_bank": "account3", "fairshare": "1.0"},
-			{"userid": "5032", "bank": "account3", "default_bank": "account3", "fairshare": "0.857143"}
+			{"userid": "5011", "bank": "account1", "default_bank": "account1", "fairshare": "0.285714", "max_jobs": "5"},
+			{"userid": "5012", "bank": "account1", "default_bank": "account1", "fairshare": "0.142857", "max_jobs": "5"},
+			{"userid": "5013", "bank": "account1", "default_bank": "account1", "fairshare": "0.428571", "max_jobs": "5"},
+			{"userid": "5021", "bank": "account2", "default_bank": "account2", "fairshare": "0.714286", "max_jobs": "5"},
+			{"userid": "5022", "bank": "account2", "default_bank": "account2", "fairshare": "0.571429", "max_jobs": "5"},
+			{"userid": "5031", "bank": "account3", "default_bank": "account3", "fairshare": "1.0", "max_jobs": "5"},
+			{"userid": "5032", "bank": "account3", "default_bank": "account3", "fairshare": "0.857143", "max_jobs": "5"}
 		]
 	}
 	EOF
--- a/t/t1003-mf-priority-small-tie.t
+++ b/t/t1003-mf-priority-small-tie.t
@@ -25,14 +25,14 @@ test_expect_success 'create a group of users with some ties in fairshare values'
 	cat <<-EOF >fake_small_tie.json
 	{
 		"users" : [
-			{"userid": "5011", "bank": "account1", "default_bank": "account1", "fairshare": "0.5"},
-			{"userid": "5012", "bank": "account1", "default_bank": "account1", "fairshare": "0.5"},
-			{"userid": "5013", "bank": "account1", "default_bank": "account1", "fairshare": "0.75"},
-			{"userid": "5021", "bank": "account2", "default_bank": "account2", "fairshare": "0.5"},
-			{"userid": "5022", "bank": "account2", "default_bank": "account2", "fairshare": "0.5"},
-			{"userid": "5023", "bank": "account2", "default_bank": "account2", "fairshare": "0.75"},
-			{"userid": "5031", "bank": "account3", "default_bank": "account3", "fairshare": "1.0"},
-			{"userid": "5032", "bank": "account3", "default_bank": "account3", "fairshare": "0.875"}
+			{"userid": "5011", "bank": "account1", "default_bank": "account1", "fairshare": "0.5", "max_jobs": "5"},
+			{"userid": "5012", "bank": "account1", "default_bank": "account1", "fairshare": "0.5", "max_jobs": "5"},
+			{"userid": "5013", "bank": "account1", "default_bank": "account1", "fairshare": "0.75", "max_jobs": "5"},
+			{"userid": "5021", "bank": "account2", "default_bank": "account2", "fairshare": "0.5", "max_jobs": "5"},
+			{"userid": "5022", "bank": "account2", "default_bank": "account2", "fairshare": "0.5", "max_jobs": "5"},
+			{"userid": "5023", "bank": "account2", "default_bank": "account2", "fairshare": "0.75", "max_jobs": "5"},
+			{"userid": "5031", "bank": "account3", "default_bank": "account3", "fairshare": "1.0", "max_jobs": "5"},
+			{"userid": "5032", "bank": "account3", "default_bank": "account3", "fairshare": "0.875", "max_jobs": "5"}
 		]
 	}
 	EOF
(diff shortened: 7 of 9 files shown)


grondo:
@cmoussa1, before I start a code review, what is the requirement being solved by this PR? (Sorry I haven't paid attention so need some help catching up)

This appears (to me) to apply a limit of _active_ jobs per user,bank pair. Will we also have need of a global per _user_ limit (this could be done in a different plugin). Will banks be hierarchical like in Slurm, and will we need hierarchical bank limits?

Sorry for being behind on these concepts! Thanks.

cmoussa1:
No need to apologize @grondo! Please let me know if there's anything else I can help clarify along the way.

This PR aims to add support for a max jobs limit per user/bank pair (ex. I could have a `max-jobs` limit of 5 active jobs under bank A, but a `max-jobs` limit of 3 active jobs under bank B). I'll need to follow up if there is a need to enforce a _global_ per-user max jobs limit; that is a good question. Banks are currently structured in a hierarchical fashion like Slurm, but I haven't been informed of a per-bank limit (at least yet, I should follow up on that as well).

cmoussa1:
OK, I've confirmed that the PR proposed does align with intended goal of enforcing a per-user max jobs limit on just a user/bank combination, so I'll go ahead and take this out of [WIP].

grondo:
> OK, I've confirmed that the PR proposed does align with intended goal of enforcing a per-user max jobs limit on just a user/bank combination, so I'll go ahead and take this out of [WIP].

Suggestion, then: retitle the PR to "plugin: add a per user,bank max jobs limit"

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +212,56 @@ static int validate_cb (flux_plugin_t *p,
...
+    mj_inner_it = (bank == NULL) ?
+        mj_it->second.find ("default") :
+        mj_it->second.find (bank);
+
+    if (mj_inner_it->second[1] >= mj_inner_it->second[0])
+        return flux_jobtap_reject_job (p, args,
+                                    "user has max number of jobs submitted");
+
+    // increment user's max_jobs count
+    mj_inner_it->second[1]++;
```
grondo, line 226: It looks like there may be  a problem, though I could be misunderstanding the code since I'm not familiar with C++ `std::map`:

It _appears_ that current job counts are tracked separately for a default bank "name" versus literally the "default" bank. E.g. if a user has a default bank of "A" with a limit of 4 jobs, and submits 4 jobs to bank "A", then submits a job with no bank, then the "default" bank max-jobs will be checked which is currently zero, so the job will be accepted.

Sorry if I misunderstood the code.

In any event, I'd suggest a test to ensure that jobs submitted with a default bank definitely have the default bank limit applied when jobs have a mix of explicit bank versus no bank.

grondo (commented):
Haven't made a full review yet, however I ran across one question below.

src/plugins/mf_priority.cpp
cmoussa1: Ah, yes. I think you are right. Two separate counts would be tracked in both the "default" key-value entry and the bank "name" key-value entry. So I'll need to re-evaluate my approach, and probably just keep one count of active jobs, regardless if the user is submitting under a default bank or explicitly setting their default bank with the bank "name."

grondo: What I would suggest is to have a map of userid->default_bank. When a bank is not provided, then you look up the default bank name and use that name in place of a provided bank name. Everything else then works the same, except there is never a `userid,"default"` user/bank pair.

One gotcha will be if the user's default bank changes between when a job starts and when it ends. You can't look up the default bank name at the end of the job from `default_banks[userid]` since it will have changed, and you will decrement the wrong entry. So, the default bank for a job should be somehow set  in the `job.validate` callback, and used for that job from that point on.

You could do this at first by stashing the bank name in the job "aux" hash, i.e. `flux_jobtap_job_aux_set()`. If you do this for all cases (i.e. whether a bank is set of your are looking up the default bank), then in later callbacks you can fetch the bank name directly with `flux_jobtap_job_aux_get()`. Or, you could instead stash a "bank_struct" with information like:
```
struct bank_info {
    const char *name;
    double fairshare;
    int max_jobs;
    int current_jobs;
};
```

Then your `inactive_cb` could be simplified to something like:
```c
static int inactive_cb (flux_plugin_t *p,
                        const char *topic,
                        flux_plugin_arg_t *args,
                        void *data)
{   
    struct bank_info *bank;
    
    if ((bank = flux_jobtap_job_aux_get (p,
                                         FLUX_JOBTAP_CURRENT_JOB,
                                         "mf_priority:bank")))
        bank->current_jobs--;
    
}
```

cmoussa1: > What I would suggest is to have a map of userid->default_bank. When a bank is not provided, then you look up the default bank name and use that name in place of a provided bank name. Everything else then works the same, except there is never a `userid,"default"` user/bank pair.

Thanks for this suggestion. I think I understand what you are suggesting. With this change, I'm able to test the use case of submitting a mix of both "explicitly-set bank" and "default bank" jobs that both contribute to one active job count with the following:

```shell
test_expect_success 'submit max number of jobs with a mix of default bank and explicity set bank' '
	jobid1=$(flux python ${SUBMIT_AS} 5011 sleep 60) &&
	jobid2=$(flux python ${SUBMIT_AS} 5011 --setattr=system.bank=account3 sleep 60) &&
	jobid3=$(flux python ${SUBMIT_AS} 5011 sleep 60)
'

test_expect_success 'submit a job while already having max number of active jobs' '
	test_must_fail flux python ${SUBMIT_AS} 5011 sleep 60 > max_jobs.out 2>&1 &&
	test_debug "cat max_jobs.out" &&
	grep "user has max number of jobs submitted" max_jobs.out &&
'
```

I can push up a fix that contains this change. 

> You could do this at first by stashing the bank name in the job "aux" hash, i.e. flux_jobtap_job_aux_set(). If you do this for all cases (i.e. whether a bank is set of your are looking up the default bank), then in later callbacks you can fetch the bank name directly with flux_jobtap_job_aux_get(). 

I'm sorry that I don't understand this enough, so this is probably a dumb question. The concept makes sense to me; in the case where a user's default bank changes during the course of an active job, we'll need to correctly decrement the appropriate bank. Is it possible to only pass the bank name with `flux_jobtap_job_aux_set ()` and `get ()`? Is there a reason for passing other information like `fairshare` and `max jobs` through a struct?

grondo: > Is there a reason for passing other information like fairshare and max jobs through a struct?

The only reason is that this avoid the extra lookup into the user/bank map to find the current job count. If you instead kept a `bank_info` structure in the map instead of raw counts, then you can attach a _reference_ to the bank information for that job to the job with `job_aux_set()`. That may help you down the line, (e.g. it will avoid the extra lookup when reprioritizing jobs as well).

Storing the bank name will work just as well and this was only a suggestion for another way to do it.

cmoussa1: Ah, okay. I could see this also reducing the extra lookup for fetching the bank values for calculating job priority as well.

Would you be okay with solely passing in the default bank information through `job_aux_set ()` and `get ()` in this PR, and posting a follow-up refactoring PR to use a `bank_info` structure to improve the information passing for both the max jobs limit and the priority calculation?

grondo: Sure! I was just suggesting another way to do it.

src/plugins/mf_priority.cpp
```diff
@@ -25,8 +25,11 @@ extern "C" {
 #include <cassert>
 #include <algorithm>
 #include <cinttypes>
+#include <vector>
 
 std::map<int, std::map<std::string, double>> users;
+std::map<int, std::map<std::string, std::vector<int>>> users_mj;
```
grondo, line 31: Suggestion: consider renaming `users_mj` to something more descriptive, since it is a global variable. Perhaps just `user_max_jobs`. Add a comment describing the global variable (e.g. "Mapping of user,bank to max jobs for that user,bank pair")

(You could also separately add a comment for the other globals here)

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +216,74 @@ static int validate_cb (flux_plugin_t *p,
                                     "user does not belong to specified bank");
     }
 
+    mj_it = users_mj.find (userid);
+
+    if (bank == NULL) {
+        def_bank_it = users_def_bank.find (userid);
+        mj_inner_it = mj_it->second.find (def_bank_it->second);
```
grondo, line 216: Should you check for failure from `std::map::find` here? If these functions are guaranteed to succeed due to checks in `job.validate` then a comment to that effect might help. (Though I still wonder if the error condition should be checked here to avoid a broker crash. Someone with more C++ experience might be able to tell you)

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +216,74 @@ static int validate_cb (flux_plugin_t *p,
...
+
+        if (flux_jobtap_job_aux_set (p,
+                                     FLUX_JOBTAP_CURRENT_JOB,
+                                     "mf_priority:bank",
+                                     &def_bank_it->second[0],
+                                     NULL) < 0)
+            flux_log_error (h, "flux_jobtap_job_aux_set");
+    } else {
+        mj_inner_it = mj_it->second.find (bank);
```
grondo, line 225: Again, do you need to check for failure here, or add a comment to say why `bank` will always be found in `mj_it->second`?

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +216,74 @@ static int validate_cb (flux_plugin_t *p,
...
+
+        if (flux_jobtap_job_aux_set (p,
+                                     FLUX_JOBTAP_CURRENT_JOB,
+                                     "mf_priority:bank",
+                                     &bank[0],
+                                     NULL) < 0)
+            flux_log_error (h, "flux_jobtap_job_aux_set");
+    }
```
grondo, line 233: Suggestion: It might be cleaner to refactor the bank lookup, if necessary, into a separate function. Then you can do something like:
```c
if (!bank
    && !(bank = lookup_default_bank (user))
    return flux_jobtap_reject_job (p, args, "unable to find default bank");

// proceed with "bank" as usual...

```

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +216,74 @@ static int validate_cb (flux_plugin_t *p,
...
+                                     &def_bank_it->second[0],
+                                     NULL) < 0)
+            flux_log_error (h, "flux_jobtap_job_aux_set");
+    } else {
+        mj_inner_it = mj_it->second.find (bank);
+
+        if (flux_jobtap_job_aux_set (p,
+                                     FLUX_JOBTAP_CURRENT_JOB,
+                                     "mf_priority:bank",
+                                     &bank[0],
```
grondo, line 230: I can't think of a good reason to use `&bank[0]` here, I think that is the same address as just `bank`.

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +209,74 @@ static int validate_cb (flux_plugin_t *p,
...
+    }
+
+    bank = flux_jobtap_job_aux_get (p,
+                                    FLUX_JOBTAP_CURRENT_JOB,
+                                    "mf_priority:bank");
+
+    mj_it = users_mj.find (userid);
+    mj_inner_it = mj_it->second.find (static_cast<char *> (bank));
+
+    mj_inner_it->second[1]--;
```
grondo, line 278: This is just a general observation: This code is hard to read. It is difficult to follow what `mj_it` and `mj_inner_it` point to and (not your fault) the `first` and `second` members of the iterator make things even more obtuse. Then the fact that max and current job count is stored as a tuple/array(?) the end result is a little confusing. (Disregard this comment if you still plan to collect user/bank information into a `bank_info` struct because I think that will resolve the issue)

Since this is a map type structure, can you not just index into the map, e.g. `users_mj[userid][bank][1]--`? (Sorry if that is a dumb question) You can see it would be even nicer if `users[userid][bank]` contained a struct bank_info because you could just reference members by name:
```
 users[userid][bank]->current_jobs--;
```

Or, if you stored the address of the bank_info struct as the aux item for the job, just:
```
bankinfo->current_jobs--;
```

grondo (commented):
Nice job on the testing!

Just a few more comments. It might be useful to have someone more familiar with C++ have a review as well, as I think I may be getting confused by the `std::map` interface (which seems awfully difficult to follow)

I made some suggestions inline, most of which should be considered optional, and will have another pass later today.

src/plugins/mf_priority.cpp
cmoussa1: Yeah, I would agree with you. I took the weekend to take a step back and try and re-evaluate adding the max jobs limit support. I think using a `bank_info` struct would make the most sense, so I just pushed one `fixup` commit that attempts to use it. I don't expect it to be perfect but the commit itself contains a few significant changes. 

- the main proposed change is the use of the map data structure containing the user information - instead of adding one new map for every user information component (fairshare, max jobs, any future limits we decide to add in the future), a new map is created that maps a userid and bank combination to a `bank_info` struct:

```c++
struct bank_info {
    double fairshare;
    int max_jobs;
    int current_jobs;
};
```

I kept the default bank map for now because I wasn't sure how else to handle the default bank case.

- the validate callback now fetches `current_jobs` and `max_jobs` information from the `bank_info` struct of the user/bank combination using the C++ map's [`at ()`](http://www.cplusplus.com/reference/map/map/at/) function. This throws an `out_of_range` exception if the value cannot be found.
- the use of the `bank_info` struct I think also simplifies the syntax of incrementing/decrementing the current job count, which no longer makes use of a map iterator:

```c++
users[userid][bank].current_jobs++;
```
- using the struct also removes the need of using an iterator in the priority calculation as well, as we can fetch the fairshare value in a similar way:

```c++
fshare_factor = users[userid][bank].fairshare;
```

Let me know if these changes make sense and help improve the readability at all (I hope they do).

grondo: This sounds great! I'll try to go through another review and testing today.

src/plugins/mf_priority.cpp
```diff
@@ -61,19 +69,12 @@ int64_t priority_calculation (flux_plugin_t *p,
     if (urgency == FLUX_JOB_URGENCY_EXPEDITE)
         return FLUX_JOB_PRIORITY_MAX;
 
-    // search element in map of maps by key
-    it = users.find (userid);
+    bank = static_cast<char *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank"));
```
grondo, line 75: Be sure to check for failure here. It is unlikely, since all jobs have `mf_priority:bank` set before they pass validation. However, if `bank == NULL`, my guess that will lead to a later crash, which will take out the flux broker, so better to catch "unlikely" errors.

My guess is if `mf_priority:bank` isn't found, then the job should have an exception raised.

src/plugins/mf_priority.cpp
```diff
@@ -61,19 +69,12 @@ int64_t priority_calculation (flux_plugin_t *p,
...
-    // a default bank
-    if (bank != NULL) {
-        inner_it = it->second.find (bank);
-        fshare_factor = inner_it->second;
-    }
-    else {
-        inner_it = it->second.find ("default");
-        fshare_factor = inner_it->second;
-    }
+    fshare_factor = users[userid][bank].fairshare;
```
grondo, line 77: This is where I'm out of my depth: Should this be wrapped in a try/except block? Will the program abort if an exception is uncaught?

grondo: I'm not sure if it is even possible, but it would be good to test that failure is handled gracefully here if an invalid bank or userid is used at this point, so that either `users[userid]` or `users[userid][bank]` do not exist. (No problem though if that would be awkward to test)

src/plugins/mf_priority.cpp
```diff
@@ -191,19 +199,76 @@ static int validate_cb (flux_plugin_t *p,
...
+                                 FLUX_JOBTAP_CURRENT_JOB,
+                                 "mf_priority:bank",
+                                 bank,
+                                 NULL) < 0)
+        flux_log_error (h, "flux_jobtap_job_aux_set");
+
+    // make sure user has not already hit their max active jobs count
+    if (current_jobs >= max_jobs)
+        return flux_jobtap_reject_job (p, args,
+                                       "user has max number of jobs submitted");
```
grondo, line 237: Just had a thought: We'll probably want a way to set `max_jobs == UNLIMITED`. Maybe `0` or `-1` would work for that? Either way, maybe:
```suggestion
    // make sure user has not already hit their max active jobs count
    if (max_jobs > 0 && current_jobs >= max_jobs)
        return flux_jobtap_reject_job (p, args,
                                       "user has max number of jobs submitted");
```

src/plugins/mf_priority.cpp
```diff
@@ -191,19 +199,76 @@ static int validate_cb (flux_plugin_t *p,
...
+                                "{s:i}",
+                                "userid", &userid) < 0) {
+        flux_log (h,
+                  LOG_ERR,
+                  "flux_plugin_arg_unpack: %s",
+                  flux_plugin_arg_strerror (args));
+        return -1;
+    }
+
+    bank = static_cast<char *> (flux_jobtap_job_aux_get (
```
grondo, line 265: Again, check return of `flux_jobtap_job_aux_get()` here to avoid a crash.

src/plugins/mf_priority.cpp
```diff
@@ -191,19 +199,76 @@ static int validate_cb (flux_plugin_t *p,
...
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank"));
+
+    users[userid][bank].current_jobs--;
```
grondo, line 270: same question about catching any exception on this line (honestly above my head though)

grondo (commented):
The changes look great @cmoussa1! Much easier to follow the code now.

I just have some comments inline about error and exception handling.

src/plugins/mf_priority.cpp
cmoussa1: Yeah, if there is no entry for `.fairshare` it will throw an `out_of_range` exception, so I should probably wrap this in a `try/except` block. 

I'm not sure the best strategy is to reject the job at this point; what do you think? Is there a more appropriate way to return an error and stop the job from running other than returning `flux_jobtap_reject_job ()`? The reason I ask is if there is in fact no `.fairshare` value, then the priority calculation will be incorrect.

grondo: Yeah, I think you would have to actually raise an exception on the job, `flux_jobtap_raise_exception()`, since it is too late to call `flux_jobtap_reject_job ()`.

You could raise an non sev 0 exception, and set the priority to 0 to put the job on hold? Then if the problem if later fixed the job does not have to resubmitted?

cmoussa1: I think that makes sense. If the exception is a non sev 0 exception, should the `severity` be set to a positive integer? Is there a typical integer used when a value is not available? Something like this:

```c++
flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
                             2, "plugin: fairshare value not available");
```

Sorry if this is a dumb question.

grondo: I've used severity 4 for "warning" exceptions since this corresponds to syslog `LOG_WARNING`.
Here you may want to use 3 which is `LOG_ERR`? It is kind of arbitrary.

If the job is also held, you can note that in the exception. Also instead using `plugin: ` for the prefix, I'd suggest `mf_priority: ` so users/admins know which plugin logged the exception.

cmoussa1: > I'm not sure if it is even possible, but it would be good to test that failure is handled gracefully here if an invalid bank or userid is used at this point, so that either users[userid] or users[userid][bank] do not exist. (No problem though if that would be awkward to test)

I think what I can do is change the way `fairshare` is accessed from the `users` map to instead use the `at ()` function, which will throw an `out_of_range` exception if either the user or the bank cannot be found:

```c++
fshare_factor = users.at (userid).at (bank).fairshare;
```

If I wrap this in a try-catch, if an exception is thrown, I can raise a an exception on the job with an error message saying that it can't find a fairshare value and just return a priority of 0.

I think I can also confirm that this behavior occurs by sending a fake payload in the sharness tests with a null fairshare key-value pair:

```
data = {"userid": str(userid), "bank": "account4", "default_bank": "account3", "fairshare": "", "max_jobs": "10"}
```

and then confirming the priority returned for a submitted job under that bank is 0.

grondo: Clever testing strategy!

One formatting note. I know that our coding style requests a space between `function` and parens `()`. However for this C++ case of chaining function calls, I think you could drop that if you want, e.g. instead of 
```
fshare_factor = users.at (userid).at (bank).fairshare;
```

you could do

```
fshare_factor = users.at(userid).at(bank).fairshare;
```

The second form seems more readable to me, but I leave the final decision to you!

src/plugins/mf_priority.cpp
cmoussa1: Sounds good. I just pushed up a `fixup` commit that checks if `bank == NULL`, and if it is, it raises an exception on the job saying that the bank could not be found, and sets the priority to 0. The same exception is raised in `inactive_cb ()`.

src/plugins/mf_priority.cpp
cmoussa1: Thanks for this suggestion! I just pushed up a `fixup` commit that adds this check for `max_jobs > 0`.

src/plugins/mf_priority.cpp
cmoussa1: I agree with you. The space with chaining function calls looks little funky. I'll push up a quick fix to get rid of that space. 👍

cmoussa1:
Just pushed up a change to squash the `fixup` commits that make changes to the plugin based on the feedback into just one `fixup` commit.

grondo:
As far as I'm concerned, @cmoussa1, feel free to squash the final fixup commit into your history.

Then I'll make another quick pass, but you may also want a C++ developer feedback on the C++ bits before a final merge.

cmoussa1:
Thanks @grondo.

---

@dongahn: if you have some time, would you mind looking at commit 1345833? More specifically, I think some feedback on how key-value entries are fetched from the `users` map would be most helpful, especially on how to handle a failed lookup if it were to occur. My approach for these map lookups were to wrap the [`.at ()` function](https://www.cplusplus.com/reference/map/map/at/) in a try-catch block since it will throw an `out_of_range` exception if the lookup fails:

```c++
    try {
        fshare_factor = users.at(userid).at(bank).fairshare;
    } catch (const std::out_of_range& oor) {
        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
                                     3, "mf_priority: cannot find fairshare; "
                                     "holding job");
        return 0;
    }
```

In the priority calculation function, it will raise a job exception and return a priority of 0. It will also raise a job exception if it cannot decrement the user's active jobs count after a job has completed.

grondo:
If you can create a series of tests to actually trigger those exceptions and exercise the code in the `catch` block, that will go a long way to giving confidence that your approach is working!

cmoussa1:
> If you can create a series of tests to actually trigger those exceptions and exercise the code in the `catch` block, that will go a long way to giving confidence that your approach is working!

Is there a way to write sharness tests to look for those raised job exceptions?

grondo:
Yes, `flux job wait-event -t TIMEOUT [OPTIONS] exception JOBID` will wait for an `exception` event in the eventlog for `JOBID`. For an example of usage see `t2271-job-dependency-after.t`:

https://github.com/flux-framework/flux-core/blob/fac664ff9d1746fab720ec22a730cc050c4f9ce5/t/t2271-job-dependency-after.t#L75-L83

cmoussa1:
Ah, okay, thanks. :-) Why don't I add some tests that trigger this exception before I ask @dongahn to review this, just in case things change slightly.

dongahn:
@cmoussa1: 

I think the code looks reasonable to me and I have to say I have struggled with exception handling in C/C++ mixed code myself as well :-)

Just as a general rule of thumb,  though, I would try to use C++ exception handling control flows as much as possible for cases where an exception truly occurs. Also if you can try to set exception handling in a more controlled/structured fashion, that'd be generally better than sprinkling the same catch clauses redundantly to too many different  code locations. (Although I admit, there are many practical cases that this cannot be done easily.)

Let's pick your example. 

```c++
    try {
        fshare_factor = users.at(userid).at(bank).fairshare;
    } catch (const std::out_of_range& oor) {
        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
                                     3, "mf_priority: cannot find fairshare; "
                                     "holding job");
        return 0;
    }
```

For this try-catch block, you can actually check whether key exists or not using `users.find()` and call `flux_jobtap_raise_exception` more deterministically. If this condition can actually occur frequently outside exceptional conditions. A benefit of adding explicitly checks like this in this case would be you will know if this condition occurs because you don't have `userid` key or `bank` key.  In addition, if you structure it that way, you can handle those *true* exceptions (like `userid` got erased in a much more twisted manner) at a much higher level in your abstraction. 

Now, because of a unique case where `flux-core`'s C code calls your C++ callback, the highest level where you can handle a C++ exception should be, I think, at the each callback level, in this case `priority_cb`. This is because once an C++ exception gets to the C land in the runtime stack, it will be propagated all the way up the stack to main and crash the program. 

If you think applying a exception catch at that top most callback level makes sense, you may find the following somewhat useful. https://github.com/flux-framework/flux-sched/pull/646

`eh_wrapper` uses a pretty neat feature of modern C++ called perfect forwarding. Feel free to pull that into `flux-accounting`, if that's useful.

dongahn:
BTW, if `errno` is used by the caller of your callback, we may need to set it. I don't know if, for example, `flux_jobtap_raise_exception` uses it or not though.

cmoussa1:
Thanks for the detailed explanation @dongahn - that is really useful information to know and I'll try to remember this in the future when dealing with exception handling. 

It turns out I might not actually need to raise an exception at all, due to my own misunderstanding of how C++ converts a string to a float using `atof`:

When `flux_request_unpack` unpacks the fairshare data from the external Python service (which, although I try to stray away from using absolute terms), it should always have a value because the `fairshare` column in the flux-accounting DB can't be `NULL` (in fact, none of the key-value pairs sent in the external service should be `NULL` because all of those columns can't be `NULL` in the database). 

In the case where the value is in fact somehow  `NULL` (which I simulate in the last test case in `t1001-mf-priority-basic.t`), the value gets unpacked and then converted to a `double` using `std::atof`. What I didn't know was that converting an empty string to a double resulted in a value of `0`. So, in the implementation I have most recently pushed up, the `out_of_range` exception wouldn't be raised, because there always is a value associated with the `fairshare` key (as well as every value in the `users` map). That would also explain why when I tried changing the test to wait for a `priority` exception, it never occurred and the test failed.

So when a user gets added to the `users` map, every key-value pair should be initialized to some value. By the time the plugin gets to `priority_calculation ()`, the user and their bank combination should have already been confirmed to exist because the presence of the key-value pair is checked in `job.validate`.

Maybe what I _can_ do instead in `job.validate` is check that if `fairshare` is exactly 0, and if it is, reject the job (since a value of 0 indicates the user has been a bad apple on the system and shouldn't be allowed to run jobs). After the checks in `job.validate` are complete, then I think it would be safe to access those elements in the map using `.at ()`, since we know the entries exist in the map. 

I hope this explanation makes sense, and sorry in advance for leading us down this exception-handling path - it is my fault for not understanding how C++ converts from one data type to the other (or misunderstanding another concept).

dongahn:
Good to know. 

> So when a user gets added to the users map, every key-value pair should be initialized to some value. By the time the plugin gets to priority_calculation (), the user and their bank combination should have already been confirmed to exist because the presence of the key-value pair is checked in job.validate.

If a bug creeps in the future in such a way that this user existence guarantee is broken, however, an out-of-range exception can be raised from `map::at()` which can lead to program crash? I think having high level exception handling before an exception reach C would be generally good if this can be done.

cmoussa1:
> If a bug creeps in the future in such a way that this user existence guarantee is broken, however, an out-of-range exception can be raised from map::at() which can lead to program crash? I think having high level exception handling before an exception reach C would be generally good if this can be done.

OK, that's a good suggestion. If I understood your recommendation above, do you recommend I instead use `map::find()`? That way, if it fails to find, I can do one of two things:

1) reject the job if it can't find the necessary information to validate a job, or
2) raise a job exception if it can't find the necessary information to calculate a job priority or decrement a user's active job count after a job transitions to `job.state.inactive`?

cmoussa1:
I just pushed up a `fixup` commit (269a5e2) that I think does a better job of handling the map lookups in `users` and will fail more safely than before.

Previously, there were map lookups scattered in a number of the callbacks, in `validate_cb()`, `priority_calculation()`, and `inactive_cb()`, which could leave a point-of-failure in any of those functions that might be hard to track down and result in a program crash. I think it finally clicked what @grondo and I had a discussion about creating a `bank_info` struct, and then passing it via `flux_jobtap_job_aux_set()`. This reduces the map lookups (using `map::find()`) in the plugin to just occur in `job.validate`, and if any lookup happens to fail there, the job will be rejected. All other accessing of bank information after a job gets validated can then be fetched with `flux_jobtap_job_aux_get()` with the following:

```c++
struct bank_info *b;

b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
                                                p,
                                                FLUX_JOBTAP_CURRENT_JOB,
                                                "mf_priority:bank_info"));

...

fshare_factor = b->fairshare;
```

Hopefully this cleans up the code some more and makes it easier to track down bugs related to the map access if they do occur.

dongahn:
> OK, that's a good suggestion. If I understood your recommendation above, do you recommend I instead use map::find()? That way, if it fails to find, I can do one of two things:

This sounds reasonable to me. 

I think it is actually outside the scope of this PR. But if it applicable, I would later also consider adding C++ exception safe callbacks for any other exceptions that could be raised from your callback. One example would be to use `eh_wrapper_t` class:  https://github.com/flux-framework/flux-sched/blob/master/qmanager/modules/qmanager_callbacks.cpp#L455.

cmoussa1:
The `fixup` commits should be squashed now and now the plugin shouldn't make any use of C++ exception handling (just a job exception via `flux_jobtap_raise_exception ()`), and all of the map lookups should occur only in `validate_cb ()` instead of in multiple callbacks in the plugin. Information needed for job priority calculation and active job count tracking should be set/accessed with `flux_jobtap_job_aux_set/get ()`.

src/plugins/mf_priority.cpp
```diff
@@ -96,24 +101,29 @@ static void rec_update_cb (flux_t *h,
...
     if (flux_respond (h, msg, NULL) < 0)
         flux_log_error (h, "flux_respond");
 
-    if (strcmp (bank, default_bank) == 0)
-        users[std::atoi (uid)]["default"] = std::stod (fshare);
+    struct bank_info b;
+
+    b.fairshare = std::atof (fshare);
+    b.max_jobs = std::atoi (max_jobs);
+    b.current_jobs = 0;
```
grondo, line 123: Would it save a couple assignments to do something like:
```c
    struct bank_info *b;
    b = &users[std:atoi(uid)][bank];
    b->fairshare = std::atof (fshare);
    b->max_jobs = std::atoi (max_jobs);
    ...
```

Then you do not have to assign `b` to `users[std::atoi (uid)][bank]` below

grondo, line 123: Also, I'm probably confused, but are you sure you want to reset `current_jobs` to zero when a user/bank is updated? It seems like if there are current active jobs for this user/bank the count should be left alone.

grondo (commented):
A couple questions inline after a quick skim. It could be that I'm confused about use of C++ map here, and if so I apologize for the noise!

src/plugins/mf_priority.cpp
cmoussa1: Oh, no, I think you're absolutely right. Thanks for pointing this out. Every time there is an update, the `current_jobs` count would be reset to 0. I failed to realize that it already gets initialized to 0 when the `bank_info` object is created. So I should leave the `current_jobs` count alone.

Maybe I can add another payload update towards the end of the sharness test to confirm that `current_jobs` doesn't get reset when another update occurs. Maybe between the last two tests... 🤔

grondo: A good functionality test might be to send an update and make sure the max_jobs limit is applied, then send another update increasing the max_jobs limit, e.g. by 1, and ensure the new limit is applied. This will check that limits can be updated, and also that sending an update doesn't reset the current job count in the plugin.

cmoussa1: I like that idea. I think the latest update I pushed up addresses that suggestion. A new payload update gets sent to the plugin with a new `max_jobs` limit of 4 (+1 to the user's previous max jobs limit of 3). This update occurs while the user has 3 active jobs running, so when the user submits another job (to increase their active count to 4), it should be submitted and run successfully. I added another job submission attempt as well to make sure that the user can't submit 5 or more active jobs. I think this should test two things:

1) that the user's `max_jobs` count can be increased while the user has active jobs running, and
2) that the user's `current_jobs` count is not reset when new payload data is sent to the plugin.

src/plugins/mf_priority.cpp
```diff
@@ -191,19 +203,85 @@ static int validate_cb (flux_plugin_t *p,
...
     }
 
+    max_jobs = bank_it->second.max_jobs;
+    current_jobs = bank_it->second.current_jobs;
+    fairshare = bank_it->second.fairshare;
+
+    // if a user's fairshare value is 0, that means they shouldn't be able
+    // to run jobs on a system
+    if (fairshare == 0)
+        return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
```
grondo, line 234: Question: if a users fairshare value is 0 should they still be able to submit a job, but it would be held until such time as fairshare > 0 and max jobs limit is not exceeded?

grondo (approved):
This seems good to merge to me!

Just one question inline, but the answer and any discussion could come in a future issue or PR.

cmoussa1:
Thanks @grondo! I'll go ahead an open an issue per your question in the latest round of review to track it, and I'll set MWP in a bit.

src/plugins/mf_priority.cpp
grondo: Sounds good to me!

Issue #135, opened 2021-08-03
fairness: add a front-end interface to recalculate/update fairshare values


cmoussa1:
Currently there is a way to re-calculate fairshare values for users in a flux-accounting database through the `data_writer_db` subclass, but there is no way to call it on a periodic time scale (so that it can be automated to be called through a `cron` job). So I'll need to add a front-end command that will call the `write_acct_info ()` function inside this subclass.

cmoussa1:
#138 has now landed, so I can close this issue.

PR #127, opened 2021-06-30
[WIP] NEWS.md: add release notes for v0.9.0

cmoussa1:
This PR adds release notes for `v0.9.0`. For now, it just includes the changes from #122, since that is probably the more important PR's among those that are open. I'll keep this **[WIP]** until that PR is good to go (and I can always add the other fixes/features from the other open PR's if those land beforehand :-) ).

Once this gets merged, I'll be sure to create an annotated tag using the following instruction:

```
git tag -a v0.9.0 -m "Tag v0.9.0" && git push upstream v0.9.0
```

PR #138, opened 2021-08-23
cmd: add a front-end fairshare update script

cmoussa1:
#### Background

As noted in #135, there is a way to calculate fairshare values for users in a flux-accounting database using the `weighted_tree` library, but there is no implementation to call it with a front-end command and on a periodic time scale.

#### Implementation

This is an implementation of a front-end script that will load flux-accounting information from a DB, calculate fairshare values, and write the updated information back to the DB. It utilizes both the `data_reader` and `data_writer` classes to read/write the DB information. It takes one optional argument, a path to the flux-accounting DB. If no path is passed in, it will try to load a DB path using `X_LOCALSTATEDIR`, the path set for **FluxAccounting.db** when configuring with a custom `localstatedir`.

#### Testing

The tests for this front-end script utilize two helper scripts: one to create the test DB and populate it with users and banks, and one to simulate a user's job usage value being updated. The front-end fairshare update script will then run twice in the sharness test, before and after the job usage update. This is to ensure the fairshare values are correctly updated and reflected in the DB after the script runs.

---

My idea for this script's implementation is to run this as a `cron` job in conjunction with the job usage update subcommand in flux-accounting: `flux account update-usage`. After the job usage update script is run and new job usage values are calculated and updated in the DB, this fairshare update script will run and update users' fairshare values.

### DIFF
--- a/src/fairness/weighted_tree/Makefile.am
+++ b/src/fairness/weighted_tree/Makefile.am
@@ -13,7 +13,9 @@ noinst_HEADERS = \
     ../account/account.hpp \
     weighted_tree.hpp \
     weighted_walk.hpp \
+    ../reader/data_reader_base.hpp \
     ../reader/data_reader_db.hpp \
+    ../writer/data_writer_base.hpp \
     ../writer/data_writer_db.hpp \
     ../writer/data_writer_stdout.hpp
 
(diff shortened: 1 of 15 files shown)


dongahn:
Actually, we don't have mergefy.io for flux-accounting yet. I will manually merge when the minor issue is corrected. Thanks.

cmoussa1:
> Actually, we don't have mergefy.io for flux-accounting yet. I will manually merge when the minor issue is corrected. Thanks.

Hm, did you mean for this comment to land in #127? Regardless, flux-accounting should have mergify support - that's what gives us the ability to add `merge-when-passing` labels to PR's, right? We have been using that for a little while now.

cmoussa1:
Just rebased to catch up after #144.

src/fairness/weighted_tree/Makefile.am
```diff
@@ -13,21 +13,27 @@ noinst_HEADERS = \
     ../account/account.hpp \
     weighted_tree.hpp \
     weighted_walk.hpp \
+	../reader/data_reader_base.hpp \
```
SteVwonder, line 16: Looks like a mixing of tabs and spaces

SteVwonder (approved):
LGTM! Nice and straightforward command.  Thanks @cmoussa1!

src/fairness/weighted_tree/Makefile.am
cmoussa1: Oops. Thanks for pointing that out. I just force-pushed a change to fix that.

cmoussa1:
Thanks @SteVwonder! I'll go ahead and set MWP then. :-)

PR #140, opened 2021-08-31
flux-account.py: fix add-user, edit-user, edit-bank, add sharness tests for python subcommands

### LINKED ISSUE #137: t: add tests for flux-account.py
I noticed yesterday that I don't think there are any tests for **flux-account.py**, the front-end script for all of the accounting commands for the flux-accounting database. It would be nice to have a sharness test or the like that confirmed all of the subcommands in this script were working, especially if the functionality in any of the Python bindings change.

cmoussa1:
~~_note: this PR is built on top of #131, so I will leave this as [WIP] until that PR is merged and then I will rebase._~~

_note: this PR will round out the `0.9.0` release for flux-accounting._

I noticed while doing some interactive testing with **flux-account.py** that a couple subcommands were broken/not working as expected, mostly because of some broken function headers that came partly as a result from #124 and from a general lack of testing. This PR includes some various minor fixes to a couple of the subcommands and user/bank functions.

This PR also adds a set of tests for **flux-acount.py** to hopefully prevent bugs like this from creeping into the code in the future. These tests use the basic subcommands that interact with user and bank accounts in the flux-accounting DB, such as adding a user account, editing some fields, adding a bank, and removing a bank, just to name a few.

note: A successful `view-user` call includes a `created` timestamp that will change with every run of the tests, so to test for success I am just `grep`-ing for a `creation_time` column. 

Fixes #137

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -124,7 +124,7 @@ def edit_bank(conn, bank, shares):
     print(shares)
     # if user tries to edit a shares value <= 0,
     # raise an exception
-    if shares <= 0:
+    if int(shares) <= 0:
         raise Exception("New shares amount must be >= 0")
     try:
         # edit value in bank_table
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -45,6 +45,7 @@ def add_user(
     uid=65534,
     admin_level=1,
     shares=1,
+    max_jobs=5,
 ):
 
     # get uid of user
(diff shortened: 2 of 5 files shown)


grondo (changes requested):
Overall LGTM.

Some of the commit messages could use message bodies. These really help in the future when you are trying to understand the context of why a change was made.
In flux-core, we've adopted the practice of using a `Problem:` statement, stating the problem that the commit fixes, and then in a second paragraph describe the solution. This isn't necessarily a requirement, but does seem to help when looking back at commits, and helps you ensure you are making a commit that addresses a single, stateable problem.

For this PR, I'd encourage at least some commit message body for all commits though.

cmoussa1:
Thanks @grondo - that's totally reasonable, and a good idea for me to adopt here too. Thanks for pointing this out. I'll go ahead and add some message bodies to the commits and force push those changes up.

grondo (approved):
Nice job updating the commit messages! LGTM!

cmoussa1:
Woohoo! Thanks @grondo. I'll go ahead and set MWP.

PR #143, opened 2021-09-14
flux-accounting DB: add Quality of Service table, field in `association_table`

cmoussa1:
This is an early [WIP] PR for defining and configuring Quality of Service (further referred to as QoS) in the flux-accounting database. In a subsequent PR, I plan to integrate this new table with the multi-factor priority plugin, where it will be used to affect the prioritization of jobs.

#### Background

QoS is a mechanism for further prioritizing submitted jobs. Each QoS has a unique name and an associated priority that can positively or negatively impact the priority of a job in a queue. Jobs that request (and are permitted) a QoS will incorporate the priority associated with that QoS in the multi-factor priority calculation. 

#### Implementation

This PR introduces the definition and configuration of QoS in the flux-accounting DB.

QoS is primarily defined and configured via its own table in the DB, named `qos_table`. It has two columns, one for the name of the QoS and one for the associated priority with it. The primary key of this table is the name of the QoS to prevent duplication. 

This table can be interacted with a new set of subcommands defined in `qos_subcommands.py`, which hosts basic functionality to add a new QoS(s), edit their associated priorities, view its information, and remove them from the table. The new subcommands introduced to interact with this new table are: `add-qos`, `view-qos`, `edit-qos`, and `delete-qos`. 

The other component of defining the QoS is to add it to each user/bank row in the `association_table`. This new `qos` column represents the available QoS(s) that the user is permitted to access when submitting jobs. This column will hold a text value that is comma-separated, with each available QoS under this column. It can be set when the user is first added to the DB or edited later on down the road. There is a helper function, `validate_qos()`, which will split the QoS(s) passed in to both the `add-user` and `edit-user` subcommands. It will check to see if the QoS passed in exist in the `qos_table` before adding them to a user/bank row in the `association_table`. A `ValueError` exception is raised if the QoS does not exist in the table.

_note: if a QoS is removed from the `qos_table`, those changes are not propagated to the `qos` column in the `association_table`, so a warning comment is left over the function to make note of this behavior._

#### Testing

Basic tests are added for both the functions defined in `qos_subcommands.py` as well as calling them from the front-end Python script `flux-account.py`.

~~`TODO`: add tests to ensure `ValueError` is raised when adding/editing a non-existent QoS in a user row.~~

EDIT: Those `ValueError` tests have been added now to both the unit tests for `qos_subcommands.py` and the sharness tests in `t1007-flux-account.t`.

### DIFF
--- a/scripts/.pylintrc
+++ b/scripts/.pylintrc
@@ -44,7 +44,7 @@ symbols=no
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=missing-docstring, too-few-public-methods, too-many-arguments, star-args, arguments-differ, no-member, import-error, no-name-in-module, useless-object-inheritance, fixme, bad-continuation, too-many-lines, duplicate-code, too-many-instance-attributes, too-many-locals
+disable=missing-docstring, too-few-public-methods, too-many-arguments, star-args, arguments-differ, no-member, import-error, no-name-in-module, useless-object-inheritance, fixme, bad-continuation, too-many-lines, duplicate-code, too-many-instance-attributes, too-many-locals, too-many-branches
 
 
 [REPORTS]
--- a/src/bindings/python/fluxacct/accounting/Makefile.am
+++ b/src/bindings/python/fluxacct/accounting/Makefile.am
@@ -2,6 +2,7 @@ acctpy_PYTHON = \
 	__init__.py \
 	user_subcommands.py \
 	bank_subcommands.py \
+	qos_subcommands.py \
 	job_archive_interface.py \
 	create_db.py
 
@@ -14,7 +15,8 @@ TESTSCRIPTS = \
 	test/test_create_db.py \
 	test/test_example.py \
 	test/test_job_archive_interface.py \
-	test/test_user_subcommands.py
+	test/test_user_subcommands.py \
+	test/test_qos_subcommands.py
 
 dist_check_SCRIPTS = \
 	$(TESTSCRIPTS)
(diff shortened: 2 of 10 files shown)


cmoussa1:
OK, I think I've finished my remaining `TODO` items, so I'll pull this out of **[WIP]** and mark this as ready for review.

ryanday36:
This looks good to me @cmoussa1. In addition to priority we use QoS to override job limits (time limits and resource limits) and to preempt jobs submitted with low priority (standby) QoS. I think those are both beyond the scope of this PR, but it's worth keeping in mind as you implement things.

dongahn:
Thanks @cmoussa1 and @ryanday36! 

@cmoussa1: if you have specific areas you want  me or others to review, feel free to call out.

cmoussa1:
Thanks @ryanday36 and @dongahn!

> In addition to priority we use QoS to override job limits (time limits and resource limits) and to preempt jobs submitted with low priority (standby) QoS. I think those are both beyond the scope of this PR, but it's worth keeping in mind as you implement things.

Thanks for this reminder. And yes, this PR is really just setting the foundation with defining and configuring a QoS in the database; not much else. I expect in future and subsequent PRs to actually tie in QoS to the actual priority calculation, overriding limits, job preemption. So that work will be coming in the near future. :-)

> if you have specific areas you want me or others to review, feel free to call out.

Thanks! I'll ping someone for a Python-specific review since it seems like the high-level functionality has been reviewed and looked at.

cmoussa1:
Just rebased to catch up after #144.

SteVwonder (approved):
LGTM! Thanks @cmoussa1.

cmoussa1:
Thanks @SteVwonder! I'll set MWP then. :-)

PR #150, opened 2021-09-24
create_test_db: change path to a command line arg

### LINKED ISSUE #145: t1006-update-fshare: '[Errno 2] No such file or directory' failure when running make check
Ran into this failure on an LC system when running `make check` on `t1006-update-fshare.t`:

```console
  File "/usr/tce/packages/python/python-3.7.2/lib/python3.7/pathlib.py", line 1241, in mkdir
    self._accessor.mkdir(self, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/usr/src/t/expected'
```

This is because I hard-coded the path to the test DB in the helper script, `create_test_db.py`. 🤦‍♂️ The fix is easy though - I should instead pass in a path in the test file, similar to `update_usage_column.py`. I'll have a fix posted today.

cmoussa1:
**Problem**: The path to the test DB in `create_test_db.py` is hard-coded in, which results in a `FileNotFoundError` when running the test on LC systems because I don't have permission to write files to `/usr/src`.

Change the helper script to take a command line argument which will represent a path to the DB file.

**Solution**: Use `${FLUX_BUILD_DIR}` in the sharness test when setting the path to the DB file, and set the path inside the flux-accounting directory.

Fixes #145

### DIFF
--- a/t/scripts/create_test_db.py
+++ b/t/scripts/create_test_db.py
@@ -9,6 +9,7 @@
 # SPDX-License-Identifier: LGPL-3.0
 ###############################################################
 import sqlite3
+import sys
 
 import fluxacct.accounting
 from fluxacct.accounting import user_subcommands as u
@@ -30,7 +31,11 @@ def edit_usage_col(conn, username, value):
 
 
 def main():
-    filename = "/usr/src/t/expected/t_small_no_tie.db"
+    if len(sys.argv) < 2:
+        print("You must pass a path to the flux-accounting DB")
+        sys.exit(-1)
+
+    filename = sys.argv[1]
     c.create_db(filename)
     conn = sqlite3.connect(filename)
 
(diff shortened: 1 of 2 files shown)


t/scripts/create_test_db.py
```diff
@@ -30,7 +31,7 @@ def edit_usage_col(conn, username, value):
 
 
 def main():
-    filename = "/usr/src/t/expected/t_small_no_tie.db"
+    filename = sys.argv[1]
```
chu11, line 34: To be on the safe side, it'd be a good idea to check that the user input `sys.argv[1]`.  Perhaps output an error if the user didn't.

cmoussa1: Thanks for this suggestion; I just squashed and force-pushed a change that performs a simple check that the user input `sys.argv[1]`.

chu11 (approved):
LGTM

cmoussa1:
Thanks, setting MWP.

PR #151, opened 2021-09-27
NEWS.md: add release notes for v0.10.0

cmoussa1:
This PR adds release notes for `v0.10.0`. I'll keep this **[WIP]** until the fix from #150 has landed.

Once this gets merged, I'll be sure to create an annotated tag using the following instruction:

```
git tag -a v0.10.0 -m "Tag v0.10.0" && git push upstream v0.10.0
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,22 @@
+flux-accounting version 0.10.0 - 2021-09-30
+-------------------------------------------
+
+#### Fixes
+
+* Fix bug in add-user where wrong number of arguments were passed to function (#140)
+
+* Fix bug in edit-user to ensure an edit made in one user/bank row was only made in just that one row instead of in multiple rows in the flux-accounting database (#140)
+
+#### Features
+
+* Add a new front-end update script that will re-calculate users' fairshare values and update them in the flux-accounting database (#138)
+
+* Add new Quality of Service table in the flux-accounting database, which will hold Quality of Services and their associated priorities (#143)
+
+* Add new sharness tests for Python subcommands (#140)
+
+* Remove pandas dependency from flux-accounting, which was required to build/install (#144)
+
 flux-accounting version 0.9.0 - 2021-09-07
 ------------------------------------------
 


NEWS.md
```diff
@@ -1,3 +1,22 @@
+flux-accounting version 0.10.0 - 2021-09-30
+-------------------------------------------
+
+#### Fixes
+
+* Fix bug in add-user where wrong number of arguments were passed to function (#140)
+
+* Fix bug in edit-user where edit made in one user/bank row was reflected in all of a user's rows in the flux-accounting database (#140)
```
dongahn, line 8: "was reflected" or "wasn't reflected?" 

You are probably right but I just wanted to make sure. I looked at #140 but from its high level description I wasn't sure.

dongahn (approved):
LGTM. I just added one question.

NEWS.md
cmoussa1: "was reflected" should be the right description - the fix was to make it so that an edit made in one user/bank row was _only_ made in just that one row instead of in multiple rows.

dongahn: Great! I find this additional description is a bit clearer than the original. Perhaps we use that instead? Thoughts?

```suggestion
* Fix bug in edit-user to ensure an edit made in one user/bank row was only made in just that one row instead of in multiple rows in the flux-accounting database (#140)
```

cmoussa1: Sure! That's a great suggestion; I'll push that change up. Thanks!

grondo:
If v0.10.0 is going to be the first version for which we will build an RPM, you might want to start adding a `make dist` tarball to the releases. Though not absolutely required, it will make use of the `tar.gz` from the release easier within build systems, since `autogen.sh` has already been run, so `rpmbuild` (for example) can go straight to `./configure`.

You may be able to just copy the "deploy" steps from flux-core CI.  It might be helpful if one of the CI builds rank `make distcheck` to create the distribution tarball and ensure that it works.

dongahn:
@grondo and @cmoussa1: this actually reminds me a problem that I recently saw when I installed a release version of flux on LC machine. There are systems which don't have the automake version compatible with the version used to create our release. In those machines (I think one was Lassen), the config/make errors out with something to the effect:  "automake-1.14: command not found"

So I tried to regenerate `configure` but [this line](https://github.com/flux-framework/flux-sched/blob/master/configure.ac#L6) gave me a trouble. 

Perhaps we can modify the configure.ac such that if you are running outside `git` or within a release environment, you can use some other token for versioning.

grondo:
Yeah, good point, regenerating `configure` from a release tarball is not supported.

I'm confused why a `make dist` tarball was trying to run `automake` though. That should never happen, which is the point of `make dist` to create a tarball that is not dependent on autotools.

dongahn:
>I'm confused why a make dist tarball was trying to run automake though. That should never happen, which is the point of make dist to create a tarball that is not dependent on autotools.

Good point. ¯\_(ツ)_/¯ 

I will repro and create a new Issue ticket since that can affect flux-account release as well.

grondo:
Thanks, if you create a flux-core issue I can look into it. Maybe during `make dist` we can capture the version in a `.version` file and use that to get the current version when we're not in a `git` repo.

dongahn:
@grondo: I actually look at my logs and actually this should not be an issue. I realized I modify the Makefile.am to enable perfflow aspect integration for our experiments last weekend. Sorry for the noise. 

I rebuilt the release for flux-core and flux-sched latest and they went smoothly on Lassen. 

Sorry for the noise.

cmoussa1:
OK, now that #150 has landed, I can pull this out of [WIP]. Thanks again for the approval.

@grondo: thanks for the recommendation above for adding a `make dist` tarball to the releases. Is there any additional work that's required in this PR to make that possible? Or should this work be done separately? Sorry that I am unfamiliar with this build process.

grondo:
> @grondo: thanks for the recommendation above for adding a make dist tarball to the releases. Is there any additional work that's required in this PR to make that possible? Or should this work be done separately? Sorry that I am unfamiliar with this build process.

Looks like you already opened #154 on this one :+1:.

PR #157, opened 2021-10-18
README: update docs with configuration instructions

cmoussa1:
Now that flux-accounting is starting to take shape as a package that can be built and installed alongside the other Flux projects, it's time for me to take some time to update some documentation wherever necessary. The first step would be to update the install and configuration documentation for setting up flux-accounting. 

This PR adds quick instructions on setting up and configuring the flux-accounting DB, loading the priority plugin, and configuring the number of fairshare and job usage update scripts to be run as cron jobs to the top-level README. 

I plan to add documentation like the ones proposed in this PR to the Flux Administrator's Guide on the official flux-docs as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,39 @@ To configure flux-accounting with a specific version of Python, pass the `PYTHON
 PYTHON_VERSION=3.7 ./configure
 ```
 
-Run flux-accounting's commands:
+### Configuring flux-accounting background scripts
+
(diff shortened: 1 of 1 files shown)


grondo:
Seems like it would be fine to add Ryan to the core team. Or you could update mergify to allow approved review by `@flux-framework/accounting` or `@flux-framework/core`

PR #160, opened 2021-10-19
mergify: replace strict mode merge with queue+rebase

cmoussa1:
Problem: Mergify.io has deprecated the strict merge mode action:

https://blog.mergify.io/strict-mode-deprecation/

Replace the strict mode configuration with a default queue with a rebase action which should be equivalent. For this PR I've copied what was done in flux-core in flux-framework/flux-core#3907.


cmoussa1:
Ah, yeah, I think you're right @grondo. I ended up copying the `queue_rules` section as well like you mentioned. I think Mergify is happy now, although under the check it still makes note of the deprecated use of the `strict` mode (not sure if that is referencing the old `.mergify.yml`?)

grondo (approved):
LGTM. I'll press the button if mergify is happy

PR #170, opened 2021-11-03
database: change default DB path, update README build instructions

### LINKED ISSUE #168: database: change default DB path location
While reviewing flux-framework/flux-docs#109, @garlick made a suggestion to change the default database path location to `@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db`. This should be a relatively straightforward fix, as I think this just involves changing the value of `db_dir` and `db_path` in `/src/bindings/python/fluxacct/accounting/__init__.py.in`:

```make
db_dir = "@X_LOCALSTATEDIR@/lib/flux/"
db_path = "@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db"
``` 

When building the project from source, I can use the following:

```shell
./autogen && ./configure --localstatedir=/var/ && make -j && make check
```

to create a default DB location for `/var/lib/flux/`. I'm sorry in advance that I do not know autotools very well; does this approach seem reasonable or make sense @garlick?

Regardless, as long as the person or people administering the DB has permissions to this directory, then they should be able to interact with this DB, i.e adding banks and users, editing their information, and more.

cmoussa1:
#### Problem

Noted in #168, the default path for the FluxAccounting.db should be changed to `@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db`.

---

This PR changes the default path for the FluxAccounting.db to `@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db` since it is already set up to be owned by user flux. 

When building the project from source, one can use the following:

```shell
./autogen.sh && ./configure --localstatedir=/var/ && make -j && make check
```

to create a default DB location for `/var/lib/flux/`.

As long as the person or people administering the DB has permissions to this directory, then they should be able to interact with this DB, i.e adding banks and users, editing their information, and more. 

It also updates the default DB path for `flux account update-fshare` and adds a default DB path for `flux account account-shares`. 

Finally, it updates the build instructions in the top-level README with the new default DB location and removes the optional path arguments from the sample code blocks. 

---

Fixes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Development for a bank/accounting interface for the Flux resource manager. Write
 
 ```console
 ./autogen.sh
-./configure
+./configure --localstatedir=/var/
 make -j
 make check
 ```
@@ -21,7 +21,7 @@ make check
 To configure flux-accounting with a specific version of Python, pass the `PYTHON_VERSION` environment variable on the `./configure` line (_note: flux-accounting needs to be configured against the same version of Python as flux-core that it is configured against; this is the default behavior of `./configure` if you choose the same prefix for flux-core and flux-accounting_):
 
 ```console
-PYTHON_VERSION=3.7 ./configure
+PYTHON_VERSION=3.7 ./configure --localstatedir=/var/
 ```
 
 ### Configuring flux-accounting background scripts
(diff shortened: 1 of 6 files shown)


cmoussa1:
Just rebased to catch up after #171.

garlick (approved):
Yep that looks right!

cmoussa1:
Thanks @garlick! Based on your suggestion about `create-db` as well, I have removed the positional argument so that the database should be able to be created with just `flux account create-db`.

garlick (approved):
Great, I think the change to let create-db use the default db path makes sense.  Thanks!

cmoussa1:
Thank you! Setting MWP

PR #171, opened 2021-11-03
cmd: unify optional path argument

### LINKED ISSUE #169: commands: unify DB path optional argument
While reviewing flux-framework/flux-docs#109, @garlick noted that several of the flux-accounting commands use different arguments for specifying a database path. These arguments should really be unified so that there is no confusion if a command is being run where the default path is not being used, i.e all commands should have the same optional path argument, `-p`.

cmoussa1:
#### Problem

Noted in #169, the optional path argument for several of the flux-accounting commands are inconsistent and should be unified.

---

This small PR changes the optional path arguments for the differing flux-accounting commands to be consistent across all the commands, `-p`.

I also snuck in a commit that removes a dangling comment in this PR as well, I hope that is okay. 

---

Fixes #169

### DIFF
--- a/src/cmd/flux_account_update_fshare.cpp
+++ b/src/cmd/flux_account_update_fshare.cpp
@@ -44,8 +44,6 @@ int main (int argc, char** argv)
     std::string filepath;
     int rc;
 
-    // const std::string *err_msg;
-
     for (int i = 1; i < argc; ++i) {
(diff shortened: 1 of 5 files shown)


garlick (approved):
LGTM!  Thanks!

cmoussa1:
Thank you! Setting MWP

PR #173, opened 2021-11-04
build: drop ax_python_devel.m4

### LINKED ISSUE #172: configure fails to accept Python 3.10.0
I haven't tested this in flux-accounting, but presumably the same bug that @andre-merzky found in flux-framework/flux-sched#880 exists here.

garlick:
Problem: ax_python_devel.m4 is not used.

Since flux-accounting doesn't use PYTHON_CLAGS or PYTHON_LIBS,
it would seem that there is no need for ax_python_devel.m4.
Furthermore, it has a bug that causes it to reject python
3.10.0 as too old.

Just drop ax_python_devel.m4 and adjust configure.ac
accordingly.

Fixes #172


cmoussa1 (approved):
Thanks so much @garlick! This LGTM.

garlick:
I dropped some extra stuff from configure.ac - hopefully its OK?  LMK if you've had a close look at that and if it's OK I'll set MWP.  I assume since CI passed it's probably ok!

cmoussa1:
Yup! I think it is fine. I tested those changes in my `centos8` docker image and was able to build successfully, so feel free to set MWP when you are ready. Thanks again!

garlick:
OK thanks!  Here goes then.

Issue #136, opened 2021-08-03
priority plugin: handle reloaded plugin scenario


cmoussa1:
We have mentioned it a few times during PR's that are related to the priority plugin, but one case we'll need to handle in the plugin is when it is reloaded. 

In that scenario, the plugin gets a `job.new` callback for every active job. We might have to have a special case `job.new` handler to handle this case since `job.validate` won't be called on these jobs. At first, maybe we could check if `job state != NEW`.

cmoussa1:
This should be resolved by #227 now, so I'll close this.

PR #177, opened 2021-11-24
plugin: change max jobs limit enforcement

cmoussa1:
#### Problem

A little while ago it was brought up that the current enforcement of a max jobs limit used on our machines counts running jobs towards a user/bank row's max_jobs limit. Currently, the multi-factor priority plugin increments for all active jobs, not just running jobs.

---

This PR changes the way the plugin increments the `current_jobs` count for a user/bank row in the flux-accounting DB by incrementing the count when a job enters `job.state.priority` instead of `job.validate`. 

It also adds a new callback, `depend_cb ()`, which will tap into a job when it reaches `job.state.depend`. A check is performed in this callback to see if the user/bank already has `max_jobs` running; if it does, it adds a dependency on the recently submitted job and its job ID is pushed to the back of a list of held jobs for that user/bank. Once a currently running job for that user/bank reaches `job.state.inactive` (thus decrementing the user/bank's `current_jobs` count), a check is performed on the list of held jobs to see if there are any; if so, the dependency is removed from the first held job in the list and it can transition to the RUN state.

Since there is somewhat significant new behavior in the priority plugin, some changes were required to also be made in `t1005-max-jobs-limits.t`. The sharness tests now test that jobs submitted when a user/bank already has `max_jobs` running correctly stay in the DEPEND state until one job transitions to `job.state.inactive`, where it is then checked to be in the RUN state.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -75,10 +75,10 @@ int64_t priority_calculation (flux_plugin_t *p,
                                                     "mf_priority:bank_info"));
 
     if (b == NULL) {
-        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
-                                     3, "mf_priority: bank info is missing; "
-                                        "holding job");
-        return 0;
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.state.priority: bank info is " \
+                                     "missing");
+        return -1;
     }
 
     fshare_factor = b->fairshare;
@@ -200,8 +200,9 @@ static int priority_cb (flux_plugin_t *p,
                                                     "mf_priority:bank_info"));
 
     if (b == NULL) {
-        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
-                                     3, "mf_priority: bank info is missing");
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.state.priority: bank info is " \
+                                     "missing");
 
         return -1;
     }
@@ -270,6 +271,20 @@ static int validate_cb (flux_plugin_t *p,
     if (fairshare == 0)
         return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
 
+    // special case where the user/bank bank_info struct is set to NULL; used
+    // for testing the "if (b == NULL)" checks
+    if (max_jobs == -1) {
+        if (flux_jobtap_job_aux_set (p,
+                                     FLUX_JOBTAP_CURRENT_JOB,
+                                     "mf_priority:bank_info",
+                                     NULL,
+                                     NULL) < 0)
+            flux_log_error (h, "flux_jobtap_job_aux_set");
+
+        return 0;
+    }
+
+
     if (flux_jobtap_job_aux_set (p,
                                  FLUX_JOBTAP_CURRENT_JOB,
                                  "mf_priority:bank_info",
@@ -308,8 +323,9 @@ static int depend_cb (flux_plugin_t *p,
                                                     "mf_priority:bank_info"));
 
     if (b == NULL) {
-        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
-                                     3, "mf_priority: bank info is missing");
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.state.depend: bank info is " \
+                                     "missing");
 
         return -1;
     }
@@ -318,8 +334,8 @@ static int depend_cb (flux_plugin_t *p,
     // dependency to hold job until an already running job has finished
     if ((b->max_running_jobs > 0) && (b->current_jobs == b->max_running_jobs)) {
         if (flux_jobtap_dependency_add (p, id, "max-jobs-limit") < 0) {
-            flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
-                                         3, "mf_priority: failed to add " \
+            flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                         "mf_priority", 0, "failed to add " \
                                          "job dependency");
 
             return -1;
@@ -357,8 +373,9 @@ static int inactive_cb (flux_plugin_t *p,
                                                     "mf_priority:bank_info"));
 
     if (b == NULL) {
-        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
-                                     3, "mf_priority: bank info is missing");
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.state.inactive: bank info is " \
+                                     "missing");
 
         return -1;
     }
@@ -371,9 +388,8 @@ static int inactive_cb (flux_plugin_t *p,
         long int jobid = b->held_jobs.front ();
 
         if (flux_jobtap_dependency_remove (p, jobid, "max-jobs-limit") < 0)
-            flux_jobtap_raise_exception (p, jobid, "plugin",
-                                         3, "mf_priority: failed to remove" \
-                                         "job dependency");
+            flux_jobtap_raise_exception (p, jobid, "mf_priority",
+                                         0, "failed to remove job dependency");
 
         b->held_jobs.erase (b->held_jobs.begin ());
     }
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -199,9 +199,12 @@ static int priority_cb (flux_plugin_t *p,
                                                     FLUX_JOBTAP_CURRENT_JOB,
                                                     "mf_priority:bank_info"));
 
-    if (b == NULL)
+    if (b == NULL) {
         flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
                                      3, "mf_priority: bank info is missing");
+
+        return -1;
+    }
 
     b->current_jobs++;
 
@@ -353,9 +356,12 @@ static int inactive_cb (flux_plugin_t *p,
                                                     FLUX_JOBTAP_CURRENT_JOB,
                                                     "mf_priority:bank_info"));
 
-    if (b == NULL)
+    if (b == NULL) {
         flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
                                      3, "mf_priority: bank info is missing");
+
+        return -1;
+    }
 
     b->current_jobs--;
 
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -26,14 +26,16 @@ extern "C" {
 #include <cassert>
 #include <algorithm>
 #include <cinttypes>
+#include <vector>
 
 std::map<int, std::map<std::string, struct bank_info>> users;
 std::map<int, std::string> users_def_bank;
 
 struct bank_info {
     double fairshare;
-    int max_jobs;
+    int max_running_jobs;
     int current_jobs;
+    std::vector<long int> held_jobs;
 };
 
 /******************************************************************************
@@ -138,7 +140,7 @@ static void rec_update_cb (flux_t *h,
         b = &users[uid][bank];
 
         b->fairshare = fshare;
-        b->max_jobs = max_jobs;
+        b->max_running_jobs = max_jobs;
 
         users_def_bank[uid] = def_bank;
     }
@@ -161,6 +163,7 @@ static int priority_cb (flux_plugin_t *p,
     int urgency, userid;
     char *bank = NULL;
     int64_t priority;
+    struct bank_info *b;
 
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
@@ -190,6 +193,18 @@ static int priority_cb (flux_plugin_t *p,
                   flux_plugin_arg_strerror (args));
         return -1;
     }
+
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL)
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
+                                     3, "mf_priority: bank info is missing");
+
+    b->current_jobs++;
+
     return 0;
 }
 
@@ -243,7 +258,7 @@ static int validate_cb (flux_plugin_t *p,
                                      "user/default bank entry does not exist");
     }
 
-    max_jobs = bank_it->second.max_jobs;
+    max_jobs = bank_it->second.max_running_jobs;
     current_jobs = bank_it->second.current_jobs;
     fairshare = bank_it->second.fairshare;
 
@@ -252,11 +267,6 @@ static int validate_cb (flux_plugin_t *p,
     if (fairshare == 0)
         return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
 
-    // make sure user has not already hit their max active jobs count
-    if (max_jobs > 0 && current_jobs >= max_jobs)
-        return flux_jobtap_reject_job (p, args,
-                                       "user has max number of jobs submitted");
-
     if (flux_jobtap_job_aux_set (p,
                                  FLUX_JOBTAP_CURRENT_JOB,
                                  "mf_priority:bank_info",
@@ -264,7 +274,55 @@ static int validate_cb (flux_plugin_t *p,
                                  NULL) < 0)
         flux_log_error (h, "flux_jobtap_job_aux_set");
 
-    bank_it->second.current_jobs++;
+    return 0;
+}
+
+
+static int depend_cb (flux_plugin_t *p,
+                      const char *topic,
+                      flux_plugin_arg_t *args,
+                      void *data)
+{
+    int userid;
+    long int id;
+    struct bank_info *b;
+
+    flux_t *h = flux_jobtap_get_flux (p);
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:i, s:I}",
+                                "userid", &userid, "id", &id) < 0) {
+        flux_log (h,
+                  LOG_ERR,
+                  "flux_plugin_arg_unpack: %s",
+                  flux_plugin_arg_strerror (args));
+        return -1;
+    }
+
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
+                                     3, "mf_priority: bank info is missing");
+
+        return -1;
+    }
+
+    // if user has already hit their max running jobs count, add a job
+    // dependency to hold job until an already running job has finished
+    if ((b->max_running_jobs > 0) && (b->current_jobs == b->max_running_jobs)) {
+        if (flux_jobtap_dependency_add (p, id, "max-jobs-limit") < 0) {
+            flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
+                                         3, "mf_priority: failed to add " \
+                                         "job dependency");
+
+            return -1;
+        }
+        b->held_jobs.push_back (id);
+    }
 
     return 0;
 }
@@ -301,6 +359,19 @@ static int inactive_cb (flux_plugin_t *p,
 
     b->current_jobs--;
 
+    // if the user/bank combo has any currently held jobs and the user is now
+    // under their max jobs limit, remove the dependency from first held job
+    if ((b->held_jobs.size () > 0) && (b->current_jobs < b->max_running_jobs)) {
+        long int jobid = b->held_jobs.front ();
+
+        if (flux_jobtap_dependency_remove (p, jobid, "max-jobs-limit") < 0)
+            flux_jobtap_raise_exception (p, jobid, "plugin",
+                                         3, "mf_priority: failed to remove" \
+                                         "job dependency");
+
+        b->held_jobs.erase (b->held_jobs.begin ());
+    }
+
     return 0;
 }
 
@@ -310,6 +381,7 @@ static const struct flux_plugin_handler tab[] = {
     { "job.state.priority", priority_cb, NULL },
     { "job.priority.get", priority_cb, NULL },
     { "job.state.inactive", inactive_cb, NULL },
+    { "job.state.depend", depend_cb, NULL },
     { 0 },
 };
 
--- a/src/test/docker/bionic/Dockerfile
+++ b/src/test/docker/bionic/Dockerfile
@@ -1,4 +1,4 @@
-FROM fluxrm/flux-core:bionic-v0.28.0
+FROM fluxrm/flux-core:bionic
 
 ARG USER=flux
 ARG UID=1000
--- a/src/test/docker/centos8/Dockerfile
+++ b/src/test/docker/centos8/Dockerfile
@@ -1,4 +1,4 @@
-FROM fluxrm/flux-core:centos8-v0.28.0
+FROM fluxrm/flux-core:centos8
 
 ARG USER=flux
 ARG UID=1000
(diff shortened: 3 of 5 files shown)


cmoussa1:
OK, I think this might be ready for some initial review whenever convenient. I believe the drop in coverage comes from the calls to `flux_log ()` and `flux_jobtap_raise_exception ()`.

grondo:
@cmoussa1: Sorry I've let this one languish.

Unfortunately, I've lost all my context here. Technically, an "active" job in Flux is one that is pending or running, not just running. I am probably just not recalling the specifics here, and the summary above isn't sufficient for me to understand the issue with the current approach. Could you reiterate what is wrong with counting all active (pending and running) jobs against the user's _max_jobs_ limit?

cmoussa1:
No problem @grondo! Hopefully the below explanation makes some sort of sense, but let me know if there's anything else I can try and clarify.

Currently, the priority plugin increments a user/bank's active jobs count every time a job is submitted and enters the `job.validate` state. To my understanding, this means a job is currently counted as "active" before it enters `job.state.run`, so technically this active jobs number represents both queued+running jobs (which you mentioned in your comment above). In offline conversations with @ryanday36, it sounds like we want to primarily limit the number of running jobs rather than the number of queued+running jobs.

What I was hoping to accomplish in this PR was to change the way the active jobs count was incremented by adding a new callback to increase the count when the job enters `job.state.run` instead of `job.validate`. The active jobs number is still decremented when the job reaches `job.state.inactive` and is still compared to a user/bank's limit in `job.validate`. 

Let me know what you think.

grondo:
> What I was hoping to accomplish in this PR was to change the way the active jobs count was incremented by adding a new callback to increase the count when the job enters job.state.run instead of job.validate. The active jobs number is still decremented when the job reaches job.state.inactive and is still compared to a user/bank's limit in job.validate.

That makes sense! My only suggestion would be to change the terminology used here to indicate that the limit is for _running_ jobs not _active_ jobs, since in Flux we call "active" jobs those that are pending _or_ running.

That is, your problem statement above:

> A little while ago it was brought up that the current enforcement of a max jobs limit used on our machines counts active jobs towards a user/bank row's max_jobs limit. Currently, the multi-factor priority plugin increments both submitted and active jobs.

Should be rewritten as:

> A little while ago it was brought up that the current enforcement of a max jobs limit used on our machines counts running jobs towards a user/bank row's max_jobs limit. Currently, the multi-factor priority plugin increments for all active jobs, not just running jobs.

Also, in the code, references to "active jobs" should be updated to "running jobs".

Does that make sense?

I'm still a little confused on how this will work in principle, though. Will it really work to reject submitted jobs based on the number of currently running jobs? (Or is that not how this works? Sorry if I got it wrong)
E.g., a user could submit jobs rapidly and get 1000 pending jobs before the first one is running, even if their max_jobs limit is 5, what prevents the scheduler from then running all 1000 jobs?

cmoussa1:
I think that makes sense, thanks @grondo - I went back and updated the original PR description as well as the commits and their respective messages to use _running_ instead of _active_ - hope I didn't miss anything on the way. 

> I'm still a little confused on how this will work in principle, though. Will it really work to reject submitted jobs based on the number of currently running jobs? (Or is that not how this works? Sorry if I got it wrong) 

This is a good question, and you are right. With the implementation I have currently, the plugin will reject submitted jobs based on the number of currently running jobs. There is an example of this in `t1005-max-jobs-limits.t`, where a user's `max_jobs` limit is 2. The user submits 2 jobs that are both running, submits one more, and gets a rejection message saying that the `user already has max number of running jobs`.

> E.g., a user could submit jobs rapidly and get 1000 pending jobs before the first one is running, even if their max_jobs limit is 5, what prevents the scheduler from then running all 1000 jobs?

Yeah, this is a good point, and I hadn't thought of this. The jobs would already have been validated before they start running and could circumvent the `max_jobs` limit. 

I vaguely remember seeing another limit called `max_submit_jobs`, a separate limit which prevents the user from _submitting_ more than _x_ number of jobs. In this case, the two limits might work in conjunction with each other to enforce the following:

- `max_jobs` enforces the total number of able to run at any given time for a user/bank row. Additional jobs will be **held** until after existing jobs are completed.
- `max_submit_jobs` enforces the maximum number of jobs able to be submitted at any given time for a user/bank row. Additional jobs will be **rejected** until existing jobs are completed. 

Perhaps that limit should also be implemented to prevent the above case from happening? Is it appropriate work for this PR? Maybe a follow-up? What do you think?

grondo:
> Perhaps that limit should also be implemented to prevent the above case from happening? Is it appropriate work for this PR? Maybe a follow-up? What do you think?

Ah, your analysis makes sense thanks!

I am not an expert in this area, perhaps this is more a question for @ryanday36. However, in my opinion, if this PR is for a limit on the number of _running_ jobs, then it should be modified to *hold* jobs until the number of running jobs for that user/bank drops below the limit (as you suggest above), rather than rejecting submitted jobs as is done now. I'm not sure rejecting submitted jobs based on the count of running jobs makes any sense.

Then you can add a follow on PR that implements the submit jobs limit as you describe above.

cmoussa1:
> However, in my opinion, if this PR is for a limit on the number of running jobs, then it should be modified to hold jobs until the number of running jobs for that user/bank drops below the limit (as you suggest above), rather than rejecting submitted jobs as is done now. I'm not sure rejecting submitted jobs based on the count of running jobs makes any sense.

Thanks. I think this makes sense for this PR. 

I think holding the job until the number of running jobs for that user/bank drops below the limit would be pretty straightforward (maybe returning `flux_jobtap_priority_unavail()` or something?). When the time comes to re-calculate the job's priority, what does that process look like? Will the job manager automatically call `job.priority.get` on that job again since it originally returned `flux_jobtap_priority_unavail()`? 

I'm sorry I'm not super knowledgable about this part of jobtap plugins.

grondo:
> I'm sorry I'm not super knowledgable about this part of jobtap plugins.

No, that is ok.

There are two methods that come to mind which could be used to tackle this problem:

1. When a job is submitted and there are already `max_jobs` running, then return `flux_jobtap_priority_unavail()` as you surmised _and_ add the job to a list of jobs "held" for that user/bank (instead of returning priority unavailable, you could also set the priority to 0 which would also hold the job. I'm not sure right now which is better). Then,  in the `inactive` callback for jobs, if `max_jobs` is adjusted such that a free slot has opened up, you can check to see if there is a job pending on the "held" list for that user/bank and "release" it (by setting a non-zero priority)

2. Similar to the above, but instead of holding a job in the PRIORITY state or with 0 priority, set a job dependency on the job with `flux_jobtap_dependency_add (p, id, "max-jobs-limit")` or similar. Then, similar to 1 above, place the job on a list for that user and bank and remove the dependency with `flux_jobtap_dependency_remove()` after a job for that user/bank has exited.

I don't have a clear idea of the benefits of each of these approaches at this time. I kind of like the job dependency approach, since there is then a record of the dependency (and thus reason the job is not in the SCHED state) is recorded in the job eventlog and also available in `flux jobs` output. On the other hand, setting a 0 or unavailable priority seems like the more "natural" approach for this use case.

grondo:
Oh, I did just think of a benefit to approach 2. above. If the plugin relies on priority of 0 or "unavailable" to hold jobs, then I think the plugin will _also_ have to handle job urgency callbacks to ensure that the priority _stays_ set to 0 or unavailable for the duration that the job is on the "held" list.

If the plugin instead uses dependencies and the DEPEND state, then you do not have to worry about the priority, since it is irrelevant until the dependency is removed.

cmoussa1:
Thanks for this input. I agree some additional input would be helpful! I think I'm going to sign off soon (still recovering from being sick) but I'll definitely think about the potential solutions you've proposed and see if I can make some progress somewhere.

ryanday36:
Lots of good discussion here. I agree that holding new jobs once max_jobs has been reached by running jobs makes sense. It sounds like that the DEPEND approach allows you to put in a reason that I assume would be viewable by the user. It's always nice to give them some idea of why their job isn't going to start, so I like that aspect of it. If you went with the priority approach, would you be able to add a reason or comment? My other concern would be whether any of these things would be modifiable by the user. I.e. if the jobtap sets the priority to 0 or adds a dependency, can the user then go and unset that?

grondo:
>  If you went with the priority approach, would you be able to add a reason or comment?

I don't think we currently have a way to add a reason why a job is held or the priority is unavailable.

> My other concern would be whether any of these things would be modifiable by the user. I.e. if the jobtap sets the priority to 0 or adds a dependency, can the user then go and unset that?

Very good insight. 

A user is allowed to set the _urgency_ of a job to anything from 0 (held) to 16. This causes a call to the `job.priority.get` jobtap callback, from which a priority plugin expected to return a new priority for the job. The default priority plugin always sets the job priority to urgency. In the case of the multifactor priority plugin, the urgency is just one factor used in calculation of the priority, but in the case where the job is held due to `max_jobs` being exceeded, the plugin would need to be cognizant of the limit and return either a priority of 0 or `flux_jobtap_priority_unavail()`. Otherwise, the user could end up releasing their job by setting an urgency (though they can't directly influence priority in this case)

grondo:
> but I'll definitely think about the potential solutions you've proposed and see if I can make some progress somewhere.

Maybe as an experiment you could try one of the approaches and see how it goes.

cmoussa1:
Agreed. I'll experiment with this and report back if I get anywhere. 👍

cmoussa1:
Per our coffee-hour discussion today, I've pushed up a [WIP] commit to the multi-factor priority plugin that includes the addition of the jobtap dependencies for submitted jobs when a user/bank row has reached their max running jobs limit. Note that I've just commented out the sharness tests in `t1005-max-jobs-limits.t` for now since these changes break the current tests. This [WIP] commit is meant to be for further debugging, as the following issue arises with the proposed changes:

With this plugin, it seems a "double decrement" occurs on a job when a job is submitted and a dependency is added. Take the following example, where a user/bank who has a `max-jobs` limit of 2 jobs submits 3 jobs in succession of each other:

```console
$ flux mini submit -n 1 sleep 10
ƒABhjboKuZ
$ flux mini submit -n 1 sleep 10
ƒABibSmBRy
$ flux mini submit -n 1 sleep 10
``` 

Job 1 and Job 2 will be accepted and run right away, while Job 3 gets a dependency added to it with the explanation that the user has reached their max running jobs limit. When Job 1 enters `job.state.inactive`, the check to see if there are any other jobs with a dependency is performed; if there are, the dependency is removed and Job 3 can now run. However, it seems Job 1 "re-enters" `job.state.inactive` and another decrement is performed from the same job to the `current-jobs` counter. This will result in a negative `current-jobs` count by the time Job 3 enters `job.state.inactive`. Below is a dummy log file I've created by inserting some print statements to the various job states tapped into by the priority plugin from the above scenario:

```
entering depend_cb from job 1
leaving depend_cb from job 1

job 1 is now in job.state.active
current jobs: 1

entering depend_cb from job 2
leaving depend_cb from job 2

job 2 is now in job.state.active
current jobs: 2

entering depend_cb from job 3
user has max current running jobs. current jobs: 2 max jobs: 2
adding dependency for job 3
leaving depend_cb from job 3

job 1 is now in job.state.inactive
current jobs: 1
removing dependency for job 3
leaving inactive_cb () from job 1

job 1 is now in job.state.inactive
current jobs: 0
leaving inactive_cb () from job 1

job 3 is now in job.state.active
current jobs: 1

job 2 is now in job.state.inactive
current jobs: 0
leaving inactive_cb () from 2

job 3 is now in job.state.inactive
current jobs: -1
leaving inactive_cb () from job 3
```

grondo:
I realized once I started experimenting with this branch that the dependency approach can't work as currently designed.

This is because if active_jobs is only incremented in RUN state but is checked in DEPEND state, then more than max_jobs will be allowed proceed through DEPEND, and since you can't go back to DEPEND state (and `job.state.depend` is only called once per job), max_jobs is not correctly applied.

Instead, I think this approach could still work if active_jobs is incremented in SCHED state instead of RUN state. This will limit the number of jobs in SCHED state to max_jobs, which should still have the intended effect, but will eliminate the race between SCHED->RUN state so that jobs > max_jobs will always be held in DEPEND state.

Hope this makes sense!

@cmoussa1, Let me know when you've made the change above and I'll test out this branch again if there's still issues with inactive callback being called more than once.

cmoussa1:
Thanks for this recommendation @grondo - I think this makes sense, and it did in fact fix the race between PRIORITY --> RUN. I went ahead and removed the callback for `job.state.run` and instead moved the running jobs increment to `job.state.priority`. 

Unfortunately, I still ran into the issue where the `job.state.inactive` callback gets called more than twice. I went ahead and pushed up these [WIP] changes into d485bef0c917b88b54717de5e2ed2f764df1f4ab; I left the `fout` statements in there to hopefully help demonstrate where the double calls are occurring. The `fout` statements are written to a dummy `.txt` file called `output.txt` as jobs transition from one state to another. An example where I notice this "double decrement" of inactive jobs is when a user/bank row has a `max_jobs` limit of 2, and 3 jobs are submitted in succession of each other, e.g. 3 `sleep 10` jobs, each submitted 1 second in succession of each other.

grondo:
@cmoussa1 - I was able to reproduce the issue you describe. To me it looks like a bug, since the `job.state.*` callbacks should never be called more than once for any job/state combination.

However, I noticed that flux-accounting is using flux-core v0.28.0 by default for the docker image, which is by now a little old. I tried the latest flux-core and didn't see the same problem, so it appears the bug must have been fixed along the way. Can you try with flux-core v0.33.0 or master and verify that you do not see the problem any more?

cmoussa1:
I think you are absolutely right @grondo - I just tried this branch again with flux-core `v0.33.0` and was not able to reproduce the issue. From first glance it looks like it is working as intended. I'll continue to test it and write some tests so I can clean up and update this branch. I'm so sorry I didn't notice this first! 🤦‍♂️

cmoussa1:
OK, I've gone ahead and updated the sharness test file `t1005-max-jobs-limits.t`  as well as the original PR description with the new changes. I think this might be ready for an initial review whenever convenient.

cmoussa1:
Just rebased to catch up after #193.

src/test/docker/bionic/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fluxrm/flux-core:bionic-v0.28.0
+FROM fluxrm/flux-core:bionic-v0.33.0
```
grondo, line 1: Just a thought - if you just use `fluxrm/flux-core:bionic`, then builds will always use flux-core@master. For the time being, this may be easier than trying to track the latest official release.

src/plugins/mf_priority.cpp
```diff
@@ -252,19 +267,57 @@ static int validate_cb (flux_plugin_t *p,
...
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL)
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
+                                     3, "mf_priority: bank info is missing");
+
+    // if user has already hit their max running jobs count, add a job
+    // dependency to hold job until an already running job has finished
+    if ((b->max_jobs > 0) && (b->current_jobs == b->max_jobs)) {
```
grondo, line 313: Suggestion: change `b->max_jobs` to `b->max_running_jobs` or `b->max_runnable_jobs`. Later, when you want to limit the total number of jobs for a user in any state, that could be called `max_jobs`.

src/plugins/mf_priority.cpp
```diff
@@ -252,19 +267,57 @@ static int validate_cb (flux_plugin_t *p,
...
+
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL)
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
+                                     3, "mf_priority: bank info is missing");
+
```
grondo, line 310: `return -1` if `b == NULL` to avoid NULL deref below.

src/plugins/mf_priority.cpp
```diff
@@ -252,19 +267,57 @@ static int validate_cb (flux_plugin_t *p,
...
+    // if user has already hit their max running jobs count, add a job
+    // dependency to hold job until an already running job has finished
+    if ((b->max_jobs > 0) && (b->current_jobs == b->max_jobs)) {
+        if (flux_jobtap_dependency_add (p, id, "max-jobs-limit") < 0) {
+            flux_log_error (flux_jobtap_get_flux (p),
+                            "flux_jobtap_dependency_add (%lu)",
+                            id);
+        }
+        b->held_jobs.push_back (id);
```
grondo, line 319: If addition of the dependency fails, then the job will proceed to the next state, therefore the `push_back()` at the very least should be skipped.

I think the right approach would be to raise a job exception and `return -1` instead of just logging an error, since otherwise a job could be run and exceed the max-running-jobs limit.

grondo (changes requested):
This is looking pretty good to me @cmoussa1!

I tested it out and it is working as designed. I ran with a limit of 2 jobs something like `flux mini submit --wait --progress --cc=1-1024 sleep 0` and ensured that the maximum number of running jobs was never over 2.

Just found a couple issues in the code.

Also, are there any tests for some corner cases like submitting a job when the bank info for the user doesn't exist, etc?

src/test/docker/bionic/Dockerfile
cmoussa1: Maybe I am doing something wrong... just using `fluxrm/flux-core:bionic` is what I used to do in this repo, but for some reason it builds with an old version of flux-core, so I switched to manually specifying the flux-core version. For example, this is my output of `./configure` when using `FROM fluxrm/flux-core:bionic` and launching an interactive Docker container with the following command:

```console
$ ./src/test/docker/docker-run-checks.sh --no-cache --no-home -i bionic -I --
```

```
  flux-accounting version 0.12.0-24-g5eeb904
  Prefix...........: /usr
  Debug Build......: 
  C Compiler.......: gcc
  C++ Compiler.....: g++
  CFLAGS...........: -g -O2
  CPPFLAGS.......... 
  CXXFLAGS.......... -g -O2
  FLUX.............: /usr/bin/flux
  FLUX_VERSION.....: 0.24.0-8-g7c7c85128
  FLUX_CORE_CFLAGS.: 
  FLUX_CORE_LIBS...: -lflux-core
  LIBFLUX_VERSION..: 0.24.0-8-g7c7c85128
  FLUX_PREFIX......: /usr
  LDFLAGS..........: 
  LIBS.............: 
  Linker...........: /usr/bin/ld -m elf_x86_64
  PYTHON...........: /usr/bin/python3
```

grondo: By default `docker` doesn't grab the latest `flux-core:bionic` if one already exists locally.
I think if you first run `docker pull flux-core:bionic`, then the latest built docker image will be fetched before you run `docker-run-checks.sh`. Can you try that and let me know how it goes?

cmoussa1: Oh, I see! 🤦‍♂️ You're 100% right. I need to pull down the latest versions to my machine locally before building. I always thought it tried to pull a latest version automatically when running `docker-run-checks.sh`.

Running `docker pull fluxrm/flux-core:bionic` _and then_ running `docker-run-checks.sh` built with the latest version of flux-core. I'll edit this commit with a useful message accordingly. Sorry for the noise!

src/plugins/mf_priority.cpp
cmoussa1: Thanks! Yeah, I think `return -1` would appropriately belong in both `job.state.priority` and `job.state.inactive`, so I'll go ahead and add those checks and probably put them in a separate commit. 

> Then, you may want to see if you can add these corner cases to the testsuite somehow to ensure the broker doesn't crash if a job from an unknown bank or other corner case is submitted.

I know for cases where:

- the user submitting the job doesn't belong in the flux-accounting DB, or 
- the user submits a job against a bank they don't belong to

The job gets rejected in the `job.validate` state before the struct is set and subsequently opened in the other `job.state` callbacks. These tests are handled in `t1001-mf-priority-basic.t`. Sending empty/incorrectly formatted payloads to the plugin are also tested in `t1001-mf-priority-basic.t`. 

I'll keep thinking about it and see if there's a corner case or scenario I'm missing.

grondo: what about a case where a job is submitted while the bank/user exists, but then before the job reaches one of the `job.state` callbacks the user/bank pair is removed by an update? (Is that even possible?)

Even if that is not possible, it would be good to enable that in testing to get better coverage and ensure the plugin is well behaved.

cmoussa1:
Thanks for your great suggestions so far @grondo! I've squashed and pushed up some changes based on your feedback:

- removed specifying the versions of flux-core in the Dockerfiles
- renamed `max_jobs` --> `max_running_jobs` (this will definitely help prepare for a future PR to add a `max_submit_jobs` type limit later on)
- added a `return -1` in the other callbacks in the plugin where appropriate

I am still trying to think of how to best approach testing for `NULL` user/bank information in the plugin. 

> what about a case where a job is submitted while the bank/user exists, but then before the job reaches one of the job.state callbacks the user/bank pair is removed by an update? (Is that even possible?)

I don't think that will be possible, as the plugin never removes user/bank data when an update is performed, it only adds new information if there is any. I think the only time user/bank information would be removed from the plugin would be if it was reloaded, which is case not yet handled by the plugin (but is noted in issue #136).

I wonder if it is possible to nullify that user/bank struct that gets passed to the various callbacks throughout the course of a job without having to unload/re-load the plugin?

grondo:
You could also add a special value in the bank update record that indicates the entry should be deleted instead of updated (for testing only for now, but at some point we'll need a way to delete information from the plugin right?)

cmoussa1:
> (for testing only for now, but at some point we'll need a way to delete information from the plugin right?)

@grondo Agreed. I think this will be especially handy when edits to user/bank information come into play instead of just straight additions/deletions from the flux-accounting DB. I could see this becoming useful work alongside the plugin reload scenario.

---

Thanks for the tips on increasing coverage. I hope I understood you correctly and my solution isn't too kludgy and will be viable:

I added a commit to the end of this PR (7eacc1d) that does the following:

I tried to improve the exception messages that the plugin sends when an exception occurs; specifically, note in which callback the exception occurred by prepending the state the exception occurred in, e.g. `mf_priority - job.state.inactive: bank info is missing`.

I also added a special test case in job.validate that will set the `bank_info` struct for a user/bank's job to `NULL` if a specific value is passed in for `max_jobs` (-1) when sending an update to the plugin. This special case will be used to test the exception messages that get raised throughout the various callbacks.

Finally, I added a set of tests in `t1001-mf-priority-basic.t` which send the special value for max_jobs so the jobtap exception messages get raised. Test that exceptions get raised in `job.state.depend`, `job.state.priority` (and a priority of 0 is assigned to the job), and `job.state.inactive`. I added a short test after all of this to ensure when a proper update follows this special test case, the same user/bank row who previously experienced jobtap exceptions when trying to submit a job can successfully submit jobs.

This did in fact raise coverage, but of course let me know what you think of this when you get the chance.

grondo:
That sounds good to me @cmoussa1! I'll take another quick pass in a bit. Nice work!

src/plugins/mf_priority.cpp
```diff
@@ -264,7 +292,56 @@ static int validate_cb (flux_plugin_t *p,
...
+        return -1;
+    }
+
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
```
grondo, line 326: Same comment as above about raising a fatal exception and the type name for the exception.

src/plugins/mf_priority.cpp
```diff
@@ -264,7 +292,56 @@ static int validate_cb (flux_plugin_t *p,
...
+                                     "bank info is missing");
+
+        return -1;
+    }
+
+    // if user has already hit their max running jobs count, add a job
+    // dependency to hold job until an already running job has finished
+    if ((b->max_running_jobs > 0) && (b->current_jobs == b->max_running_jobs)) {
+        if (flux_jobtap_dependency_add (p, id, "max-jobs-limit") < 0) {
+            flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
```
grondo, line 337: Same here (fatal exception) If we can't add a dependency to the job then something bad has happened and we shouldn't let the job continue.

src/plugins/mf_priority.cpp
```diff
@@ -295,12 +372,29 @@ static int inactive_cb (flux_plugin_t *p,
                                                     FLUX_JOBTAP_CURRENT_JOB,
                                                     "mf_priority:bank_info"));
 
-    if (b == NULL)
+    if (b == NULL) {
         flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "plugin",
-                                     3, "mf_priority: bank info is missing");
```
grondo, line 300: Sorry I didn't notice this before, but lets use fatal exception severity for now.

grondo (changes requested):
Getting very close @cmoussa1! I think I missed before (or forgot) that the exception severity for these job exceptions was 3 instead of 0. (Did we talk about that before? Sorry if I forgot).

See inline comments below

cmoussa1:
Thanks @grondo! Yeah, I can't remember either if you ever mentioned the severity level for jobtap exceptions. Maybe you mentioned it in a previous PR and I didn't think of it! So thank you for the explanation (and the tip on naming the exceptions). 

I pushed up and squashed your suggestions to the last commit:

It changes the exceptions to set the severity to 0 instead of 3, which would result in a job cancelation. Since the job cancelation occurs in `job.state.depend`, I removed the new tests which checked for exceptions in `job.state.priority` and `job.state.inactive`.

grondo (approved):
LGTM! Nice work @cmoussa1!

cmoussa1:
Thanks so much for your review and help @grondo! I'll go ahead and set MWP

PR #180, opened 2021-12-03
cmd, t: move `flux_account_shares.cpp` to `src/cmd/`, use FLUX_EXEC_PATH in sharness tests

cmoussa1:
#### Problem(s)

a) One of the points of feedback given in #167 was the ability to append the `cmd/` directory to `FLUX_EXEC_PATH` in the sharness tests, which enables directly calling `flux account` commands instead of very long commands including substitution of environment variables. This could definitely clean up the appearance of the sharness tests and help clarify exactly _what_ commands are being run in each test. 

b) `flux_account_shares.cpp` is currently located in its own directory under `src/fairness/`. When trying to include `FLUX_EXEC_PATH` in the sharness tests that call the `flux account-shares` command, the tests fail with a message that this command could not be found. I _believe_ this is because the `flux-account-shares` binary is not picked up until after a `make install`. 

---

This PR does two things.

1) Move the `flux_account_shares.cpp` files to the `src/cmd/` directory, where it will reside with the other front-end flux-accounting commands. It then removes the `print_hierarchy/` directory from the repository since it is no longer needed.

2) In the sharness tests, use `FLUX_EXEC_PATH` where possible and substitute the very long sharness test commands with the actual `flux account` commands that would be used after a flux-accounting installation.


grondo (commented):
@cmoussa1 - Nice improvement!

Just a couple inline comments regarding testsuite changes.

t/t1000-print-hierarchy.t
```diff
@@ -1,87 +1,92 @@
...
 
 test_description='Test print-hierarchy command'
+
 . `dirname $0`/sharness.sh
-PRINT_HIERARCHY=${FLUX_BUILD_DIR}/src/fairness/print_hierarchy/flux-account-shares
 SMALL_NO_TIE=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/small_no_tie.db
 SMALL_TIE=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/small_tie.db
 SMALL_TIE_ALL=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/small_tie_all.db
 OUT_OF_INSERT_ORDER=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/out_of_insert_order.db
+FLUX_EXEC_PATH=${SHARNESS_TEST_SRCDIR}/../src/cmd:${FLUX_EXEC_PATH}
```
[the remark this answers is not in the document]
cmoussa1: Ah, I did not know about this, thank you for pointing it out! I went ahead and added a `flux-accounting.sh` file which does just that; sets `FLUX_EXEC_PATH_PREPEND`. That enabled me to remove all of the individually-set `FLUX_EXEC_PATH` variables in a number of the sharness tests. Let me know what you think and if it's any better.

t/sharness.d/flux-accounting.sh
```diff
@@ -0,0 +1,5 @@
+FLUX_EXEC_PATH_PREPEND=${SHARNESS_TEST_SRCDIR}/../src/cmd:${FLUX_EXEC_PATH}
```
grondo, line 1: No need to add `FLUX_EXEC_PATH` to `FLUX_EXEC_PATH_PREPEND` here.

t/t1000-print-hierarchy.t
```diff
@@ -1,87 +1,89 @@
...
+
 . `dirname $0`/sharness.sh
-PRINT_HIERARCHY=${FLUX_BUILD_DIR}/src/fairness/print_hierarchy/flux-account-shares
 SMALL_NO_TIE=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/small_no_tie.db
 SMALL_TIE=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/small_tie.db
 SMALL_TIE_ALL=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/small_tie_all.db
 OUT_OF_INSERT_ORDER=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/out_of_insert_order.db
 
+flux setattr log-stderr-level 1
+
```
grondo, line 12: This will likely fail when not after `test_under_flux` (or, it will attempt to setattr in the enclosing instance if there is one, e.g. the system instance. Probably not what you want)

t/t1000-print-hierarchy.t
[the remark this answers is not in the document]
grondo: Yeah, looks good. I noted a couple other issues inline.

t/sharness.d/flux-accounting.sh
cmoussa1: Oops, sorry about that - I squashed and pushed a fix to remove `FLUX_EXEC_PATH` from `FLUX_EXEC_PATH_PREPEND`.

t/t1000-print-hierarchy.t
cmoussa1: Ah, TIL. Thanks for explaining what that line does - I just squashed and pushed a fix to remove the `flux setattr log-stderr-level 1` line from `t1000-print-hierarchy.t`.

grondo (approved):
LGTM!

cmoussa1:
Thanks @grondo! I'll go ahead and add MWP

PR #176, opened 2021-11-23
database: add queue_table, subcommands for interacting with queue_table

cmoussa1:
#### Problem

As discussed in #164, there is a need for being able to define and configure queue information within the flux-accounting database.

---

#### Solution

This is a [WIP] PR that I just wanted to post before the Thanksgiving break that adds such support within the DB. It adds a new table, `queue_table`, which holds just 4 columns - the name of the queue and a couple of required limits associated with each queue:

```
queue               tinytext        NOT NULL,
min_nodes_per_job   int(11),
max_nodes_per_job   int(11),
max_time_per_job    int(11),
PRIMARY KEY (queue)
```

_Note that the only column that is required when adding a new queue is the name; if not specified, all of the limit columns could just be empty, with no default values._

Support for interacting with this table from the command line with a number of subcommands added to `flux account`: `add-queue`, `view-queue`, `delete-queue`, and `edit_queue`. These functions allow a DB admin to add, edit, remove, and view queue information to the flux-accounting DB.

#### Usage

```console
[moussa1@docker-desktop src]$ flux account add-queue -h
positional arguments:
  QUEUE                 queue name

optional arguments:
  -h, --help            show this help message and exit
  --min-nodes-per-job MIN NODES PER JOB    min nodes per job
  --max-nodes-per-job MAX NODES PER JOB    max nodes per job
  --max-time-per-job  MAX TIME PER JOB     max time per job
```

```console
[moussa1@docker-desktop src]$ flux account view-queue -h
positional arguments:
  QUEUE       queue name

optional arguments:
  -h, --help  show this help message and exit
```

```console
[moussa1@docker-desktop src]$ flux account edit-queue -h
optional arguments:
  -h, --help         show this help message and exit
  --queue queue      queue name
  --field FIELD      column name
  --new-value VALUE  new value
```

```console
[moussa1@docker-desktop src]$ flux account delete-queue -h
positional arguments:
  QUEUE       queue name

optional arguments:
  -h, --help  show this help message and exit
```
Both unit tests and sharness tests were added with the new subcommands and module.

Fixes #164

PR #181, opened 2021-12-07
`edit-user`: change positional and optional arguments to match `edit-qos` subcommand

### LINKED ISSUE #178: edit-user: edit arguments to subcommand
While working on #176, a suggestion was made to keep the same arguments as the `add` subcommand instead of using the `--field` and `--new-value` optional args. The same improvement should probably also made to the `edit-user` subcommand. This will allow for multiple fields to be edited at the same time, allow for resets on optional fields, and will improve the overall usability of the command.

cmoussa1:
#### Problem

While working on #176, a suggestion was made to keep the same arguments as the `add` subcommand instead of using the `--field` and `--new-value` optional args. The same improvement should probably also made to the `edit-user` subcommand. This will allow for multiple fields to be edited at the same time, allow for resets on optional fields, and will improve the overall usability of the command. 

---

This PR changes the `edit-user` subcommand to take the same positional and optional arguments as the `add-user` subcommand. This includes the following improvement to the command:

##### old usage

```console
$ flux account edit-user --username=user1234 --field="max_jobs" --new-value=1234
```

##### new usage

```console
$ flux account edit-user user1234 --max-jobs 100
```

It also allows for multiple values of a user/bank row to be edited at the same time:

```console
$ flux account edit-user user1234 --max-jobs 100 --qos "standby,expedite"
```

This new behavior should now match the behavior in the `edit-qos`, `edit-bank`, and `edit-queue` subcommands.

I also snuck in a tiny cleanup commit that removes the `--parent-acct` optional argument from the `add-user` subcommand. This is an obsolete argument that was left over from some previous cleanup; I hope that is okay. 

Fixes #178

### DIFF
--- a/scripts/.pylintrc
+++ b/scripts/.pylintrc
@@ -44,7 +44,7 @@ symbols=no
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=missing-docstring, too-few-public-methods, too-many-arguments, star-args, arguments-differ, no-member, import-error, no-name-in-module, useless-object-inheritance, fixme, bad-continuation, too-many-lines, duplicate-code, too-many-instance-attributes, too-many-locals, too-many-branches, unused-argument, unused-variable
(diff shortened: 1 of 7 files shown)


ryanday36 (approved):
looks good @cmoussa1.

cmoussa1:
Thank you! Setting MWP

PR #187, opened 2021-12-16
`view-user`: add capability to view multiple rows

### LINKED ISSUE #185: view-user: subcommand doesn't list all banks when a user is added to multiple banks
A user can be added to multiple banks with the `flux account add-user` command, but `flux account view-user` only lists one bank.

cmoussa1:
#### Problem

As mentioned in #185, the `view-user` subcommand only prints the first result of a SQLite query fetching user information from the `association_table` even if the user belongs to multiple banks (and thus has multiple rows in the table).

---

This PR uses `cursor.fetchall()` to fetch all possible rows and iterate through them to print out all rows instead of just one. It also makes use of `.ljust()` to print out rows in a table-like format, similar to the output format of `pandas`, a data formatting module I removed from the project a while ago.

```console
$ flux account view-user fluxuser
creation_time   mod_time        deleted         username        userid          admin_level     bank            default_bank    shares          job_usage       fairshare       max_jobs        qos             
1639680148      1639680148      0               fluxuser        1002            1               acct            acct            10              0.0             0.5             5                               
1639680148      1639680148      0               fluxuser        1002            1               other_acct      acct            10              0.0             0.5             5
```

Fixes #185

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -16,16 +16,36 @@ import pwd
 
 def view_user(conn, user):
     cur = conn.cursor()
+    headers = [
+        "creation_time",
+        "mod_time",
+        "deleted",
+        "username",
+        "userid",
+        "admin_level",
+        "bank",
+        "default_bank",
+        "shares",
+        "job_usage",
+        "fairshare",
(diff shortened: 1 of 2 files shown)


ryanday36 (approved):
This looks good.

cmoussa1:
Thanks! Setting MWP

PR #188, opened 2021-12-21
README: add database permissions note, remove `admin_level` column

### LINKED ISSUE #183: database: mechanism to allow users to edit database using `flux account` command
Currently, only the 'flux' user can edit the database, so 'flux account add|edit|delete-*' commands only work as the flux user. It would be convenient if there was a mechanism to allow certain users (e.g. members of a specific unix group) to run those commands.

As a side note on this, it might be worth making the create-db command create the database with 644 permissions so that normal users can run 'flux account view-*' commands.

cmoussa1:
#### Problem

Opened in #183, it is not clear in the flux-accounting documentation how read/write access to the flux-accounting DB is configured, and the `admin_level` column in the `association_table` can cause confusion since it does not actually configure any sort of access to the database - it was a leftover column that I copied over when initially designed the database and should probably be removed.

---

This PR adds a note in the top-level README in flux-accounting describing how permissions to the database work (as well as a link to the official SQLite documentation). I should plan on also submitting a similar PR to the flux-docs admin guide in the near future.

This PR also removes the `admin_level` column from the `association_table` within the DB and from all subcommands and tests that utilize said column. 

Fixes #183


ryanday36 (approved):
looks good to me

cmoussa1:
Thank you! Setting MWP

PR #189, opened 2021-12-29
view-bank: improve command, add more sharness tests

### LINKED ISSUE #184: view-bank: subcommand doesn't list users / children
It would be useful if the 'flux account view-bank ...' command listed the child banks or users for a given bank to facilitate traversing the bank tree from the root.

cmoussa1:
#### Problem

Mentioned in #184, the `view-bank` command does not list children banks or users for a given bank. Moreover, the output when running this command can be sort of clunky and not pleasing. This PR looks to improve the command and add some more tests that make use of this command.

---

This is a small [WIP] PR looking to improve the command by changing the format of the output when viewing a bank within the flux-accounting database. Specifically, it looks to improve the output in the following ways:

- improve the output when viewing bank information to be formatted like a table, similar to the format of the `pandas` module before it was removed from flux-accounting:

```console
$ flux account view-bank A
bank_id         bank            parent_bank     shares          
2               A               root            1   
```

- viewing a bank that contains sub banks will also print a hierarchy with the passed-in bank acting as the root of the tree, i.e viewing a the root bank will print out the entire user/bank hierarchy, or viewing a bank with sub banks will print out a smaller user/bank hierarchy. A couple of examples are shown below:

```console
$ flux account view-bank root
root
 A
  user5011
  user5012
 B
  user5013
 C
  user5014
 D
  E
  F
```

```console
$ flux account view-bank D
D
 E
 F
```

- viewing a bank with users under that bank will also print all subsequent user information from the `association_table`. For example, a leaf bank `A` with users under this bank will show the following output:

```console
$ flux account view-bank A
bank_id         bank            parent_bank     shares          
2               A               root            1               

Users Under Bank A:

username        userid          default_bank    shares          job_usage       fairshare       max_jobs        qos             
user5011        5011            A               1               0.0             0.5             5               expedite        
user5012        5012            A               1               0.0             0.5             5               expedite,standby 
```

Note that some columns from the `association_table` are omitted under this output, as I deemed them not totally relevant to any information concerning banks - the columns presented here are more or less the most crucial from the table. 

#### Testing

New tests are added in **t1007-flux-account.t** to test the new output features of the `view-bank` command. All expected output is added into a new folder inside of the `t/expected/` directory called `flux_account`, which for now just hosts all `view-bank` output. Eventually, I think it would be nice to move all large output from the tests within **t1007-flux-account.t** to this directory, but that's probably best for a future PR.

---

Fixes #184

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -13,6 +13,67 @@ import sqlite3
 
 from fluxacct.accounting import user_subcommands as u
 
+# helper function to print user information in a table format
+def print_user_rows(rows, bank):
+    print("\nUsers Under Bank {bank_name}:\n".format(bank_name=bank))
+    user_headers = [
+        "username",
+        "userid",
+        "default_bank",
+        "shares",
+        "job_usage",
+        "fairshare",
+        "max_jobs",
+        "qos",
+    ]
+    # print column names of association_table
+    for header in user_headers:
+        print(header.ljust(15), end=" ")
+    print()
+    for row in rows:
+        for col in list(row):
+            print(str(col).ljust(15), end=" ")
+        print()
+
+
+# helper function to print bank information in a table format
+def print_bank_row(rows, bank):
+    bank_headers = [
+        "bank_id",
+        "bank",
+        "parent_bank",
+        "shares",
+    ]
+    # bank has sub banks, so list them
+    for header in bank_headers:
+        print(header.ljust(15), end=" ")
+    print()
+    for row in rows:
+        for col in list(row):
+            print(str(col).ljust(15), end=" ")
+        print()
+
+
+# helper function to traverse the bank table and delete all sub banks and users
+def print_sub_banks(conn, bank, indent=""):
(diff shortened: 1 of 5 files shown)


ryanday36 (approved):
This is looking really good @cmoussa1! You're making the view-bank command much more powerful / useful. My one suggestion for future work would be to allow the user to toggle things like the tree view and detailed user view with flags. e.g. make the default behavior just show the basic bank info, but add something like a '-t/--tree' and '-u/--users' flag for printing the full tree or detailed user info. Alternately, those views could be the default and you could add a '-s/--short' flag to just list info about the current bank.

cmoussa1:
Thanks for taking a look at this @ryanday36!

> My one suggestion for future work would be to allow the user to toggle things like the tree view and detailed user view with flags. e.g. make the default behavior just show the basic bank info, but add something like a '-t/--tree' and '-u/--users' flag for printing the full tree or detailed user info. Alternately, those views could be the default and you could add a '-s/--short' flag to just list info about the current bank.

Thanks for this suggestion. Let me open a separate issue on this improvement - I think this is totally reasonable and a great additional feature for the command.

Setting MWP.

PR #193, opened 2022-01-07
user_subcommands.py: fix miscellaneous bugs

### LINKED ISSUE #191: view-user: remove `admin_level` column
While doing some interactive testing with the flux-accounting DB, I noticed that the `admin_level` column still remains in the output of the `view-user` command. This results in incorrect output and should be removed. 

I noticed I've introduced this bug a couple of times (left over a deleted column in a SQLite table) because up to this point I had been hard-coding the column names in the command functions. I just learned [this](https://stackoverflow.com/questions/7831371/is-there-a-way-to-get-a-list-of-column-names-in-sqlite) neat Python trick to fetch the column names using the same `Cursor` object when executing the query, so I think I can remove my manual list of column names altogether.

```python
column_names = [description[0] for description in cursor.description]
```

### LINKED ISSUE #192: add-user: `UNIQUE` constraint fails when re-adding user/bank row to database
While doing some interactive testing with the flux-accounting database, I noticed the following bug:

If I add a row to the `association_table` with a user/bank combination (e.g. `user1234`, `bankA`), remove the row, and then re-add it, a `UNIQUE` constraint fails in the `job_usage_factor_table`.

```console
[moussa1@docker-desktop src]$ sudo flux account add-user --username=user1234 --bank=bankA
[moussa1@docker-desktop src]$ sudo flux account delete-user --username=user1234 --bank=bankA
[moussa1@docker-desktop src]$ sudo flux account add-user --username=user1234 --bank=bankA
UNIQUE constraint failed: job_usage_factor_table.username, job_usage_factor_table.bank
```

This is because by design, the rows in the `job_usage_factor_table` are deleted (since we use the data in this table to calculate fairshare values). To get rid of this `UNIQUE constraint` failure, I can add a `OR IGNORE` clause to the SQL statement. This should safely ignore adding any rows which already contain the same `PRIMARY KEY`:

```sqlite
INSERT OR IGNORE INTO job_usage_factor_table (username, userid, bank) VALUES (?, ?, ?)
```

cmoussa1:
#### Problem

While working on some of the improvements discussed to PR #177, I noticed a couple small miscellaneous bugs in the `view-user` and `add-user` subcommands while running them in my Docker container. The bugs are listed as follows:

- Mentioned in #191, the `admin_level` column still remains in the output of the `view-user` command. This results in incorrect output and should be removed. 
- Mentioned in #192, a user/bank row is added to the `association_table` with a user/bank combination (e.g. `user1234`, `bankA`), the row is removed, and then re-added, a `UNIQUE` constraint fails in the `job_usage_factor_table`.

---

This PR fixes both issues in these subcommands and adds small improvements to try and help mitigate similar bugs in the future. 

In the `view-user` command, I've changed the way headers for the table output are fetched. Previously, I had handwritten each header in the table, which would quickly become inconsistent if I made a change to the `association_table` in the database. The `view-user` subcommand now uses `cursor.description` to fetch the headers directly from the table itself. This should ensure consistency in the output even if and when the table schema changes in the future.

In the `add-user` command, I've added an `OR IGNORE` clause to the `INSERT` SQL statement for the `job_usage_factor_table`. This ensures that previously deleted user/bank rows do not encounter a `UNIQUE constraint` failure when being re-added to the `association_table`.

In the `view-qos` and `delete-qos` commands, I've converted the `qos` argument into positional arguments for both of the commands.

Fixes #191
Fixes #192

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -16,25 +16,11 @@ import pwd
 
 def view_user(conn, user):
     cur = conn.cursor()
-    headers = [
-        "creation_time",
-        "mod_time",
-        "deleted",
-        "username",
-        "userid",
-        "admin_level",
-        "bank",
-        "default_bank",
-        "shares",
-        "job_usage",
-        "fairshare",
-        "max_jobs",
-        "qos",
-    ]
     try:
         # get the information pertaining to a user in the DB
         cur.execute("SELECT * FROM association_table where username=?", (user,))
         rows = cur.fetchall()
+        headers = [description[0] for description in cur.description]
         if not rows:
             print("User not found in association_table")
         else:
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -201,7 +201,21 @@ def edit_user(
 
             conn.execute(update_stmt, tup)
 
-            # commit changes
-            conn.commit()
+    # update mod_time column
+    mod_time_tup = (
+        int(time.time()),
+        username,
+    )
+    if bank is not None:
+        update_stmt = """UPDATE association_table SET mod_time=?
+                         WHERE username=? AND bank=?"""
+        mod_time_tup = mod_time_tup + (bank,)
+    else:
+        update_stmt = "UPDATE association_table SET mod_time=? WHERE username=?"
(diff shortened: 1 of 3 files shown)


ryanday36 (commented):
looks good @cmoussa1. Nice work cleaning those up.

ryanday36 (approved):

cmoussa1:
Thanks @ryanday36! I'll go ahead and add MWP.

PR #195, opened 2022-01-19
cmd: add bulk DB populate command

### LINKED ISSUE #139: flux-accounting DB: support bulk insertion of user/bank information
A random thought that came up while working on some testing for the `flux account` subcommands:

Whenever the time comes to add a bunch of user/bank information to the flux-accounting DB, it is probably going to be super tedious/annoying to have to add all of this information one-by-one via the command line or within the SQLite shell. It would probably be a good idea to support a "mass add" of information, where data could be read from a `.csv` file (or some other format) and then exported to the flux-accounting DB. Maybe this could be its own subcommand, where the person adding all of this information could pass in a file to be read and converted into entries in the flux-accounting DB.

cmoussa1:
#### Problem

Populating the DB with many users and banks can be tedious, as currently there are really only two options:

- using the `add-bank` and `add-user` commands one at a time to add users and banks
- going into an interactive SQLite shell and using `INSERT` commands, one a time, to add both users and banks to the database

This can be cumbersome and slow, especially when the time comes to install flux-accounting on a system and we need to migrate a large set of users.

---

This PR adds a new command to flux-accounting, `flux account-pop-db`, which allows an admin to bulk populate a flux-accounting DB's `bank_table` and `association_table` using `.csv` files. The path to the file can be passed via the command line, and this command will iterate through the file and call `add_user ()` or `add_bank ()`, depending on which file is used. You can also pass both options to populate both the `bank_table` and the `association_table` at the same time, like the following:

```console
$ flux account-pop-db -b bank_info.csv -u user_info.csv
```

The structure of the bank `.csv` file must be in the following format: `Bank,ParentBank,Shares`. The **ParentBank** slot can be left blank if needed. An example is below:

```
root,,1
A,root,1
B,root,1
```

The structure of the user `.csv` file must be in the following format: `Username,UserID,Bank,Shares,MaxJobs,QOS`. The **Shares** and **MaxJobs** slots can be left blank; if so, they will be filled in with the default values of their respective columns in the `association_table`. An example is below:

```
user1005,1005,B,1,5,""
user1006,1006,B,,,""
user1007,1007,B,1,7,""
user1008,1008,B,,,""
user1009,1009,B,1,9,""
```

Fixes #139

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -64,16 +64,17 @@ def add_user(
     qos="",
 ):
 
-    # get uid of user
-    fetched_uid = get_uid(username)
+    if uid == 65534:
+        # get uid of user
+        fetched_uid = get_uid(username)
 
-    try:
-        if isinstance(fetched_uid, int):
-            uid = fetched_uid
-        else:
-            raise KeyError
-    except KeyError as key_error:
-        print(key_error)
+        try:
+            if isinstance(fetched_uid, int):
+                uid = fetched_uid
+            else:
+                raise KeyError
+        except KeyError:
+            print("could not find UID for user; adding default UID")
 
     # check for a default bank of the user being added; if the user is new, set
     # the first bank they were added to as their default bank
--- a/src/cmd/Makefile.am
+++ b/src/cmd/Makefile.am
@@ -46,4 +46,5 @@ dist_fluxcmd_SCRIPTS = \
 	flux-account.py \
 	flux-account-update-fshare \
 	flux-account-priority-update.py \
-	flux-account-shares
+	flux-account-shares \
+	flux-account-pop-db.py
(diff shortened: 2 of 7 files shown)


cmoussa1:
OK, I'll go ahead and mark this as ready for review.

@ryanday36 If you don't mind, let me know what you think of this tool. As it stands, it is pretty lightweight and I certainly don't mind changing or trying to add features that you might see helpful to have. I figured I would present the tool with just the necessities and build off it based on what might be most useful.

ryanday36 (approved):
This looks good. It also looks like it should be pretty straightforward to add more fields as, for example, limits get added to the DB.

cmoussa1:
Thank you! I'll go ahead and add MWP

> It also looks like it should be pretty straightforward to add more fields as, for example, limits get added to the DB.

Yes, it should be straightforward to add more fields to either the `bank_table` or the `association_table`, as well as adding additional options for bulk populating other tables in the flux-accounting DB, like the `queue_table` for example.

PR #200, opened 2022-02-02
[WIP] plugin: add max_submit_jobs per user limit

### LINKED ISSUE #196: plugin: enforce a max_submit_jobs limit
The changes proposed in #177 help more accurately track a user/bank row's running jobs count by only incrementing a job counter when a job enters the RUN state, and adding a dependency to additionally submitted jobs if a user/bank has reached their `max_jobs` limit. However, this does not prevent a user/bank from just submitting a bunch of jobs and having them flood the queue with their pending jobs, e.g. if `userA` has a `max_jobs` limit of 2, there is currently nothing stopping that user from having 2 running jobs and 98 pending jobs in the queue...

A solution to this would be to enforce an additional limit in the priority plugin that limits the number of _both_ running and pending jobs a user/bank can have at any given time, called `max_submit_jobs`. This new limit will need to be enforced in the `association_table` and sent to the plugin when an update is performed. There, it can be stored in every `bank_info` struct and keep track of a user/bank's job counts. When this limit is reached, subsequent jobs will be rejected with a message explaining that the `max_submit_jobs` limit has been reached.

cmoussa1:
#### Problem

The changes introduced in #177 left somewhat of a "bad behavior" loophole in the priority plugin when it comes to handling job submissions from a user/bank row. As a quick refresher, #177 introduced a change to the way the plugin kept track of a "running jobs" for a user/bank row, so it only counts jobs that had entered the RUN state. So, for example. if a user/bank row had a `max_jobs` limit of 2, and they submitted 4 jobs at once, the first two jobs could enter the RUN state, but the other 2 would have dependencies added to them (thus, holding the job), until one of the first two jobs entered the INACTIVE state, which would remove the dependency on one of the held jobs and allow it to enter the RUN state.

However, there is currently no way to enforce the amount jobs _submitted_ from a user/bank row. So, technically, someone could with a `max_jobs` limit of 2 could submit 100 jobs at once, leaving 2 jobs to run at a time while 98 others sit in the queue. This behavior should be prevented and enforced for every user/bank combination.

---

This [WIP] PR introduces a new user/bank row limit to both the `association_table` in the flux-accounting database and to the multi-factor priority plugin, called `max_submit_jobs`. This limit represents the maximum number of jobs able to be submitted at any given time. This limit can be set at create time for a row and can be modified at any time for a row in the `association_table`.

In the priority plugin, this limit is integrated into every `bank_info` struct for ever user/bank. Its limit is enforced when a job enters `job.validate`. If the number of pending + running jobs reaches the `max_submit_jobs` limit, subsequently submitted jobs will be rejected with a message saying that the limit has been reached, which I believe lines up with current behavior on our systems.

This counter is incremented as a job exits `job.validate` and is decremented as a job enters `job.state.inactive`.

Additional tests were added to `t1005-max-jobs-limits.t` to test the new limit by submitting a max number of jobs and ensuring an error message is output for jobs submitted while the `max_submit_jobs` limit is reached.

Fixes #196


grondo:
Before I review the changes to the plugin, I wanted to clarify what is meant by `max_submit_jobs`. 

First, to clarify some Flux terminology as I understand it (please someone correct if I'm wrong)

These bullets refer to the job state diagram [here](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html#state-diagram).

 * _active_ jobs - jobs in any state besides NEW or INACTIVE. These jobs are resident in the job-manager
 * _inactive_ jobs - jobs that have transitioned to INACTIVE state. In most cases the job-manager no longer tracks these jobs, but their final state is preserved by `job-list` and `job-archive`. These are essentially all completed, canceled, etc jobs.
 
In addition we define some meta-states for _active_ jobs:

 * _pending_ jobs - active jobs that are in states before the RUN state, i.e. they have not yet started running
 * _running_ jobs - active jobs in RUN or CLEANUP state

As you can see, we do not really have a concept of "submit jobs", so I would be wary of introducing that terminology here because it could cause confusion.

Given the above terminology, my main question is:

Is the intent of this feature to limit the number of _active_ jobs for a given user/bank, or to limit the number of _pending_ jobs for a user/bank?

Thanks!

cmoussa1:
Thanks for this feedback @grondo!

> As you can see, we do not really have a concept of "submit jobs", so I would be wary of introducing that terminology here because it could cause confusion.

I could be misremembering, but I think I remember us having a similar conversation in #177 about the difference between an _active_ job and a _running_ job. In the end, #177 introduced a change to limit the number of _running_ jobs.

I completely agree that new limits should not introduce new terminology and should instead leverage and build off the design work already done in the state diagram you provided (thanks for sharing that by the way), so I'm sorry for causing any additional confusion.

Looking at the definitions you provided above, the eventual goal of this PR would be to enforce a limit on the number of _active_ jobs, i.e. both pending and running jobs:

> In addition we define some meta-states for active jobs:
>
> - pending jobs - active jobs that are in states before the RUN state, i.e. they have not yet started running
> - running jobs - active jobs in RUN or CLEANUP state

If this sounds alright, then I should redefine the limit in the plugin code, the commits, and the PR description to make things clearer (maybe call the limit `max_active_jobs` or something).

Let me know if that helped answer your question at all! Sorry again for the confusion and not using the right terminology.

grondo:
Yes, _max_active_jobs_ is much more clear. Thanks!

Your proposal to rename the limit to `max_active_jobs` seems right to me. This will make it clear that jobs will be rejected for a give user/bank combo when the number of _running_ + _pending_ jobs reaches the _active_jobs_ limit.

cmoussa1:
Perfect, thanks for the feedback! I'll get to work on it then. :-)

PR #201, opened 2022-02-02
plugin: add max_active_jobs per-user/bank limit

### LINKED ISSUE #196: plugin: enforce a max_submit_jobs limit

cmoussa1:
#### Problem

The changes introduced in #177 left somewhat of a "bad behavior" loophole in the priority plugin when it comes to handling job submissions from a user/bank row. As a quick refresher, #177 introduced a change to the way the plugin kept track of a _running_ jobs for a user/bank row, so it only counts jobs that had entered the RUN state. So, for example, if a user/bank row had a `max_jobs` limit of 2, and they submitted 4 jobs at once, the first two jobs could enter the RUN state, but the other 2 would have dependencies added to them (thus, holding the jobs), until one of the first two jobs entered the INACTIVE state, which would remove the dependency on one of the held jobs and allow it to enter the RUN state.

However, there is currently no way to enforce the amount of _active_ jobs from a user/bank row. So, technically, someone with a `max_jobs` limit of 2 could submit 100 jobs at once, leaving 2 jobs to run at a time while 98 others sit in the queue. This behavior should be prevented and enforced for every user/bank combination.

---

This ~~[WIP]~~ PR introduces a new user/bank row limit to both the `association_table` in the flux-accounting database and to the multi-factor priority plugin, called `max_active_jobs`. This limit represents the maximum number of jobs able to be _pending + running_. This limit can be set at create time for a row and can be modified at any time for a row in the `association_table`.

In the priority plugin, this limit is integrated into every `bank_info` struct for every user/bank. Its limit is enforced when a job enters `job.validate`. If the number of _pending + running_ jobs reaches the `max_active_jobs` limit, subsequently submitted jobs will be rejected with a message saying that the limit has been reached, which I believe lines up with current behavior on our systems.

This counter is incremented as a job exits `job.validate` and is decremented as a job enters `job.state.inactive`.

Additional tests were added to `t1005-max-jobs-limits.t` to test the new limit by submitting a max number of jobs and ensuring an error message is output for jobs submitted while the `max_active_jobs` limit is reached.

Fixes #196

### DIFF
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -98,6 +98,7 @@ def create_db(
                 job_usage        real        DEFAULT 0.0   NOT NULL,
                 fairshare        real        DEFAULT 0.5   NOT NULL,
                 max_running_jobs int(11)     DEFAULT 5     NOT NULL    ON CONFLICT REPLACE DEFAULT 5,
+                max_active_jobs  int(11)     DEFAULT 7     NOT NULL    ON CONFLICT REPLACE DEFAULT 7,
                 qos              tinytext    DEFAULT ''    NOT NULL    ON CONFLICT REPLACE DEFAULT '',
                 PRIMARY KEY   (username, bank)
         );"""
--- a/src/cmd/flux-account.py
+++ b/src/cmd/flux-account.py
@@ -78,6 +78,12 @@ def add_add_user_arg(subparsers):
         metavar="MAX_RUNNING_JOBS",
     )
     subparser_add_user.add_argument(
+        "--max-active-jobs",
+        help="max number of both pending and running jobs",
+        default=7,
+        metavar="max_active_jobs",
+    )
+    subparser_add_user.add_argument(
         "--qos",
         help="quality of service",
         default="",
@@ -127,6 +133,12 @@ def add_edit_user_arg(subparsers):
         help="max number of jobs that can be running at the same time",
         default=None,
         metavar="MAX_RUNNING_JOBS",
+    )
+    subparser_edit_user.add_argument(
+        "--max-active-jobs",
+        help="max number of both pending and running jobs",
+        default=7,
+        metavar="max_active_jobs",
     )
     subparser_edit_user.add_argument(
         "--qos",
@@ -439,6 +451,7 @@ def select_accounting_function(args, conn, output_file, parser):
             args.userid,
             args.shares,
             args.max_running_jobs,
+            args.max_active_jobs,
             args.qos,
         )
     elif args.func == "delete_user":
@@ -451,6 +464,7 @@ def select_accounting_function(args, conn, output_file, parser):
             args.default_bank,
             args.shares,
             args.max_running_jobs,
+            args.max_active_jobs,
             args.qos,
         )
     elif args.func == "view_job_records":
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -33,8 +33,8 @@ std::map<int, std::string> users_def_bank;
 
 struct bank_info {
     double fairshare;
-    int max_running_jobs;
-    int current_jobs;
+    int max_run_jobs;
+    int cur_run_jobs;
     std::vector<long int> held_jobs;
 };
 
@@ -140,7 +140,7 @@ static void rec_update_cb (flux_t *h,
         b = &users[uid][bank];
 
         b->fairshare = fshare;
-        b->max_running_jobs = max_running_jobs;
+        b->max_run_jobs = max_running_jobs;
 
         users_def_bank[uid] = def_bank;
     }
@@ -207,7 +207,7 @@ static int priority_cb (flux_plugin_t *p,
         return -1;
     }
 
-    b->current_jobs++;
+    b->cur_run_jobs++;
 
     return 0;
 }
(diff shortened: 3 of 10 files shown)


src/bindings/python/fluxacct/accounting/create_db.py
```diff
@@ -87,18 +87,19 @@ def create_db(
...
+                mod_time        bigint(20)  DEFAULT 0     NOT NULL,
+                deleted         tinyint(4)  DEFAULT 0     NOT NULL,
+                username        tinytext                  NOT NULL,
+                userid          int(11)     DEFAULT 65534 NOT NULL,
+                bank            tinytext                  NOT NULL,
+                default_bank    tinytext                  NOT NULL,
+                shares          int(11)     DEFAULT 1     NOT NULL    ON CONFLICT REPLACE DEFAULT 1,
+                job_usage       real        DEFAULT 0.0   NOT NULL,
+                fairshare       real        DEFAULT 0.5   NOT NULL,
+                max_jobs        int(11)     DEFAULT 5     NOT NULL    ON CONFLICT REPLACE DEFAULT 5,
```
grondo, line 100: Sorry, I know this is still a draft, but I have one suggestion: is `max_jobs` actually `max_running_jobs`? If so, it might be nice to rename now rather than later since `max_jobs` and `max_active_jobs` could be easily confused (at least for me). By renaming `max_jobs` to `max_running_jobs` or `max_runnable_jobs` or whatever, it would make it clear throughout which limit applies to the total number of active jobs vs just runnable jobs.

cmoussa1: You're one step ahead of me! I was pushing that work off until I got the `max_active_jobs` named throughout the project; changing `max_jobs` to `max_running_jobs` or something is still left to do (and I agree it's probably a necessary change) and will most likely result in a number of changes throughout the project as well, not just in the plugin. Sorry, I probably should have written that as a **TODO** in the PR description.

grondo: Ok, I apologize for jumping the gun!

cmoussa1:
OK, I think I'll take this out of [WIP] and mark it as ready for an initial review whenever convenient. 

I prepended the set of commits that actually add the `max_active_jobs` limits with a series of commits that rename `max_jobs` to `max_running_jobs` throughout the flux-accounting project: in the DB, the Python bindings and commands, the existing sharness tests, and the priority plugin. I tried to separate them where I felt appropriate, but if you think they should just be squashed into one larger "renaming" commit, I'm totally fine with that and can squash.

grondo:
If it would make more sense, you could propose the renaming as a separate PR. This might help users browsing the release notes since we have an entry for each PR, generally.

One reason to squash the rename into one big commit is so that no one commit breaks the build/test of the project. I would leave it up to your discretion whether you think there is value in keeping the commits separate. (I would personally lean toward just one commit if all the commit messages are largely the same)

cmoussa1:
OK! I like that suggestion. I'll squash those commits and propose them in a separate PR. Thanks!

src/cmd/flux-account.py
```diff
@@ -77,6 +77,12 @@ def add_add_user_arg(subparsers):
         default=5,
         metavar="MAX_RUNNING_JOBS",
     )
+    subparser_add_user.add_argument(
+        "--max-active-jobs",
+        help="max number of both submitted and running jobs",
```
grondo, line 82: s/submitted/pending/?

grondo (changes requested):
This looks good! Nice work.

Just some minor suggestions inline, and one possible bug (though I may have read the code wrong, apologies if so!)

src/plugins/mf_priority.cpp
```diff
@@ -264,12 +268,19 @@ static int validate_cb (flux_plugin_t *p,
...
+    max_active_jobs = bank_it->second.max_active_jobs;
 
     // if a user's fairshare value is 0, that means they shouldn't be able
     // to run jobs on a system
     if (fairshare == 0)
         return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
 
+    // if a user/bank has reached their max_active_jobs limit, subsequently
+    // submitted jobs will be rejected
+    if (max_run_jobs > 0 && cur_active_jobs >= max_active_jobs)
```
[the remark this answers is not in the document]
cmoussa1: Ah, yes, you're absolutely right. This check should be for `max_active_jobs`. I just squashed and pushed up a change that fixes this `if` condition.

grondo (approved):
Great! LGTM now!

cmoussa1:
Woohoo! Thanks for the feedback and review. Setting MWP

PR #202, opened 2022-02-03
database: rename max_jobs --> max_running_jobs

cmoussa1:
#### Problem

The `max_jobs` column is confusing and doesn't clearly represent what the column is for: limiting the number of _running_ jobs a user/bank combo can have running at any given time. When the `max_active_jobs` limit is added, it would be clearer to rename `max_jobs` to something more distinct.

---

This PR contains work split off from #201 that renames `max_jobs` to `max_running_jobs` throughout the flux-accounting project to more clearly describe the difference between a "running" jobs count and an "active" jobs count for a user/bank combo. It contains renaming changes to:

- the database
- the Python commands
- the multi-factor priority plugin
- the existing sharness tests

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -23,16 +23,16 @@ def print_user_rows(rows, bank):
         "shares",
         "job_usage",
         "fairshare",
-        "max_jobs",
+        "max_running_jobs",
         "qos",
     ]
     # print column names of association_table
     for header in user_headers:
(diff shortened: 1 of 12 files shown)


grondo (approved):
Seems straightforward to me!

You may want to retitle the commit message now that the rename applies to the whole project, not just `database:`

cmoussa1:
Oh, right. 🤦‍♂️ Thanks for pointing that out. Just edited the commit message so it doesn't include just `database`.

cmoussa1:
Thanks for approving! I'll set MWP

PR #203, opened 2022-02-04
view-bank: remove manual list of column names

cmoussa1:
This is a small PR that removes the handwritten list of column names from the `print_user_rows()` and the `print_bank_rows()` helper functions and instead fetches the column names of the `association_table` and `bank_table` using `cursor.description`
so that the headers will automatically be updated every time a change is made to the tables.


ryanday36 (approved):
looks good.

cmoussa1:
Thanks! Setting MWP

PR #206, opened 2022-02-09
docker: add distcheck builder

### LINKED ISSUE #153: docker: add `DISTCHECK=t` to one of the builders
This will ensure `make distcheck` won't break in the future when changes are made or things are added.

cmoussa1:
This is a small PR that adds a `make distcheck` target to the CI using the `el8` image. This should help prevent issues from leaking into the project when adding new features and files to the code. 

Fixes #153


src/fairness/weighted_tree/Makefile.am
```diff
@@ -42,3 +42,5 @@ libweighted_tree_la_CXXFLAGS = \
     $(CODE_COVERAGE_CFLAGS) \
     $(SQLITE_CFLAGS) \
     $(AM_CXXFLAGS)
+
+EXTRA_DIST = libweighted_tree.la
```
grondo, line 46: Hm, `libweighted_tree.la` should be a product of the build and not a source that needs to be packed up with the distribution. What led you to adding this file to `EXTRA_DIST`?

grondo (changes requested):
`.la` files should not be added to the dist with `EXTRA_DIST` AFAIK. Let's figure out why this seemed to be required before merging this one.

src/fairness/weighted_tree/Makefile.am
cmoussa1: Thanks for pointing this out. I wasn't sure myself why `make distcheck` wasn't working. If I leave out the `EXTRA_DIST` assignment, this is the error that gets generated:

```
make[6]: *** No rule to make target '../../src/fairness/weighted_tree/libweighted_tree.la', needed by 'flux-account-update-fshare'.  Stop.
```

src/Makefile.am
```diff
@@ -1 +1,131 @@
...
+    $(WARNING_CXXFLAGS) \
+    $(CODE_COVERAGE_CXXFLAGS)
+
+AM_LDFLAGS = $(CODE_COVERAGE_LDFLAGS) $(SQLITE_LIBS)
+
+AM_CPPFLAGS = -I$(top_srcdir) $(FLUX_CORE_CFLAGS)
+
+noinst_LTLIBRARIES = libweighted_tree.la
+noinst_HEADERS = \
+	fairness/account/account.hpp \
```
grondo, line 13: Appears to be a whitespace error here (at least in Github view) (tab vs spaces?)

src/Makefile.am
```diff
@@ -1 +1,131 @@
...
+
+AM_CXXFLAGS = \
+    $(WARNING_CXXFLAGS) \
+    $(CODE_COVERAGE_CXXFLAGS)
+
+AM_LDFLAGS = $(CODE_COVERAGE_LDFLAGS) $(SQLITE_LIBS)
+
+AM_CPPFLAGS = -I$(top_srcdir) $(FLUX_CORE_CFLAGS)
+
+noinst_LTLIBRARIES = libweighted_tree.la
```
grondo, line 11: Suggestion: I _think_ you can still keep `libweighted_tree.la` in `src/fairness` if you instead did

```
noinst_LTLIBRARIES = fairness/libweighted_tree.la
```

Then later refer to this library as `fairness_libweighted_tree.la`?

Same with the tests, etc under `src/fairness`. (It looks like you already took this approach with the files under `src/cmd`, just use the same approach here)

Otherwise, this lo

grondo (commented):
Much improved! Nice work. Just one suggestion inline

src/Makefile.am
cmoussa1: Oops! Thanks for pointing that out - looked like there were multiple whitespace errors so I force pushed a fix to convert tabs to spaces.

src/Makefile.am
cmoussa1: Oh great! Thanks for letting me know about this, and sorry I didn't apply this to `libweighted_tree` at first. I just force pushed a change to this `Makefile.am` to keep the library in `src/fairness`.

Do you think I should keep the Makefiles that I've completely commented out? Should I remove them from the repository altogether?

src/Makefile.am
grondo: I think in a `Makefile` you may want to make sure to use tabs? It may work out ok for a Makefile.am but generally tabs are required in some places in Makefiles, so it might be safer to stick with tabs.

src/Makefile.am
grondo: Remove the Makefiles you've commented out. They won't be lost forever if you are worried about losing some information, since they'll be preserved in git history.

cmoussa1: Got it, thanks. I just broke up that initial [WIP] commit into two; one that moves the compilation to `src/Makefile.am`, and another commit to remove all the commented out Makefiles.

grondo (approved):
Looks great! Big improvement in the build system, nice work!

grondo:
Ah, don't forget to add the `el8 - distcheck` build to mergify config and/or protected branches!

cmoussa1:
Ah, right! Thank you for reminding me @grondo. I just pushed an additional commit that adds the `el8` distcheck builder to mergify. I think this will then require a manual merge?

grondo:
Yes, merging now!

cmoussa1:
Thanks!

Issue #208, opened 2022-02-16
job-archive interface: TypeError when not specifying `--priority-decay-half-life`


cmoussa1:
In an offline discussion with @ryanday36, he reported a `TypeError` when running the `flux account update-usage` command:

```console
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```

This is likely due to the fact that the `--priority-decay-half-life` optional argument does not have a default value associated with it in neither the `flux-account.py` script or in the `calc_usage_factor()` helper function in `job_archive_interface.py`, so when this argument is not specified, `None` gets passed as the value for this parameter.

I should add a default value to this optional argument to match its default value in the database, `1`. I'll probably also add a couple of tests in `t1007-flux-account.py` to make sure default values are captured.

PR #209, opened 2022-02-17
job-archive interface: update optional `--priority-decay-half-life` argument

cmoussa1:
#### Problem

As described in #208, a `TypeError` occurs when not specifying a value for the `--priority-decay-half-life` optional argument in the `update-usage` command. This is because the argument has no default value, so if no value is specified, `None` is passed to the `calc_usage_factor()` function and it fails.

---

This PR fixes the `TypeError` in the `update-usage` command by adding a default value for the `--priority-decay-half-life` optional argument, 1, which matches the default value in the flux-accounting database. 

It also adds a new sharness test file, `t1010-update-usage.t`, which creates both a sample flux-accounting DB and job-archive DB, populated with a set of fake job records. It then runs the `update-usage` command twice, one using the default value and another specifying a value on the command line. It then checks to see that the job usage values do in fact get updated in the flux-accounting database by running the `flux account-shares` command.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -438,7 +438,7 @@ def check_end_hl(acct_conn, pdhl):
         acct_conn.commit()
 
 
-def update_job_usage(acct_conn, jobs_conn, pdhl):
+def update_job_usage(acct_conn, jobs_conn, pdhl=1):
     s_assoc = "SELECT username, bank FROM association_table"
     cur = acct_conn.cursor()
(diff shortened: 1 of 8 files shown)


ryanday36 (approved):
Looks good. My only minor gripe is that I feel like the help message for --priority-decay-half-life in src/cmd/flux-account.py could be clearer. Maybe something like 'number of weeks for a job's usage contribution to decay by half'.

cmoussa1:
Thanks @ryanday36 for reviewing! I think that's a great suggestion - I'll update the help message per your suggestion and squash a fix in this branch, and add MWP

PR #211, opened 2022-02-22
job-archive interface: remove `count_ranks()` helper function, import ResourceSet class

### LINKED ISSUE #210: job-archive interface: count_ranks doesn't parse ranges of ranks
In an offline discussion with @ryanday36, he reported that there was an error when trying to parse a job record with a range of ranks in the following format: `[98, 100]`. This is most likely due to an issue in the `count_ranks()` helper function, which parses the `ranks` column in the job-archive database. This will need some rework to handle a range of cases:

1) single number
2) blank line (if a job was cancelled)
3) ranges of values delimited by commas

While I am opening this issue, though, I should probably ask and double check: is `ranks` still an okay method for counting `nnodes` for a user/bank's job? When I first implemented the job usage factor calculation, I think I asked and went off of the assumption that 1 rank was equal to 1 node. Is this assumption still true? Do the `jobspec` or `R` columns in the job-archive DB contain an `nnodes` value somewhere?

cmoussa1:
#### Problem

As noted in #210, the `count_ranks()` helper function incorrectly parses the `ranks` column when there is a range of ranks. This leads to a `ValueError` when counting the number of nodes used in a job.

---

This PR removes the `count_ranks()` helper function and instead imports the `ResourceSet` class from flux-core, which can then extract `nnodes` from a job's `R`. Note that in both the unit test file for **job_archive_interface.py** and the sharness test file **t1010-update-usage.t**, the calculated usage values checked in the tests remain the same for all users, which ensures consistency after the change to the way the job-archive interface extracted the `nnodes` values from job records.

It also contains a small improvement to the `update-usage` command to add a special option to the `print_job_records()` function. Previously, when the `update-usage` command was run, all new job records were printed to `stdout` after the command was run. The command should no longer produce output to `stdout` once run.

A new sharness test, `t1011-job-archive-interface.t` is also added, which loads the job-archive module and submits a number of `sleep 1` jobs under one user. The `update-usage` command is then called to ensure that the job usage and fairshare values get updated for the user who submitted jobs as well as the other users in that same bank.

Fixes #210

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -136,7 +136,7 @@ def add_job_records(rows):
     return job_records
 
 
-def view_job_records(conn, output_file, **kwargs):
+def get_job_records(conn, **kwargs):
     job_records = []
 
     # find out which args were passed and place them in a dict
@@ -179,13 +179,19 @@ def view_job_records(conn, output_file, **kwargs):
     cur = conn.cursor()
     cur.execute(select_stmt, (*tuple(params_list),))
     rows = cur.fetchall()
-    # if the length of dataframe is 0, that means
-    # no job records were found in the jobs table,
-    # so just return an empty list
+    # if the length of dataframe is 0, that means no job records were found
+    # in the jobs table, so just return an empty list
     if len(rows) == 0:
         return job_records
 
     job_records = add_job_records(rows)
+
+    return job_records
+
+
+def output_job_records(conn, output_file, **kwargs):
+    job_records = get_job_records(conn, **kwargs)
+
     if output_file is None:
         print_job_records(job_records)
     else:
@@ -350,9 +356,8 @@ def calc_usage_factor(jobs_conn, acct_conn, pdhl, user, bank):
 
     # get jobs that have completed since the last seen completed job
     last_j_ts = get_last_job_ts(acct_conn, user, bank)
-    user_jobs = view_job_records(
+    user_jobs = get_job_records(
         jobs_conn,
-        output_file=None,
         user=user,
         after_start_time=last_j_ts,
     )
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -14,13 +14,7 @@ import pwd
 import csv
 import math
 
-
-def count_ranks(ranks):
-    if "-" in ranks:
-        ranks_count = ranks.replace("-", ",").split(",")
-        return int(ranks_count[1]) - int(ranks_count[0]) + 1
-
-    return int(ranks) + 1
+from flux.resource import ResourceSet
 
 
 def get_username(userid):
@@ -127,6 +121,8 @@ def add_job_records(rows):
     job_records = []
 
     for row in rows:
+        rset = ResourceSet(row[6])  # fetch R
+
         job_record = JobRecord(
             row[0],  # userid
             get_username(row[0]),  # username
@@ -134,7 +130,7 @@ def add_job_records(rows):
             row[2],  # t_submit
             row[3],  # t_run
             row[4],  # t_inactive
-            count_ranks(row[5]),  # nnodes
+            rset.nnodes,  # nnodes
             row[6],  # resources
         )
         job_records.append(job_record)
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -65,7 +65,7 @@ def write_records_to_file(job_records, output_file):
 
 def print_job_records(job_records):
     print(
-        "{:<10} {:<10} {:<10} {:<10} {:<10} {:<10} {:<10} {:<10}".format(
+        "{:<10} {:<10} {:<10} {:<15} {:<15} {:<15} {:<10}".format(
             "UserID",
             "Username",
             "JobID",
@@ -73,12 +73,11 @@ def print_job_records(job_records):
             "T_Run",
             "T_Inactive",
             "Nodes",
-            "R",
         )
     )
     for record in job_records:
         print(
-            "{:<10} {:<10} {:<10} {:<10} {:<10} {:<10} {:<10} {:<10}".format(
+            "{:<10} {:<10} {:<10} {:<15} {:<15} {:<15} {:<10}".format(
                 record.userid,
                 record.username,
                 record.jobid,
@@ -86,7 +85,6 @@ def print_job_records(job_records):
                 record.t_run,
                 record.t_inactive,
                 record.nnodes,
-                record.resources,
             )
         )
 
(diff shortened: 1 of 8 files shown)


src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -127,14 +122,17 @@ def add_job_records(rows):
     job_records = []
 
     for row in rows:
+        rdict = json.loads(row[6])  # fetch R
```
grondo, line 123: Is `rdict` unused?

grondo (commented):
I realize this PR is still in Draft, but I took a quick look at usage of `ResourceSet` and it looks good to me! I had one question about the need for the `json` module (I probably just missed something reading by patch instead of looking at the whole file.)

src/bindings/python/fluxacct/accounting/job_archive_interface.py
cmoussa1: Ah, you are absolutely right, `rdict` is unused. I must have forgotten to remove it when I found out I could just load R as a string to `rset` instead of converting it to a dictionary first; sorry about that, and thanks for pointing it out. I'll push up a fix.

grondo: No problem! It surprised me at first that the `json` module was required to use `ResourceSet`. thanks!

cmoussa1:
Thanks again for starting to take a look at this yesterday @grondo. If you don't mind, do you think you could also give [7c3fffa](https://github.com/flux-framework/flux-accounting/pull/211/commits/7c3fffad43f207adc045d9bede8ed61a8786a0d5) a look? This commit adds a new sharness test which loads the `job-archive` module, submits a couple of `sleep 1` jobs, and then runs the `update-usage` command to make sure the job-archive interface can actually read the job-archive DB and generate a usage value for user/bank combos. I copied some of the helper functions in [flux-core/t/t2250-job-archive.t](https://github.com/flux-framework/flux-core/blob/master/t/t2250-job-archive.t) to help.

grondo:
I have a few questions about the test:

 * Do the jobs have to be `sleep 1` for accounting purposes? If they could be `hostname` or `sleep 0` jobs the test would run a little faster.
 * I don't really understand why you have to wait for the jobs to be inactive (`clean` event posted to job eventlog), then also wait for the jobs to be inactive in `flux job list`. Is the real potential for a race condition between the job `clean` event and the job being stored in the job-archive db? Maybe what you really need is the `waid_db()` shell function (and required `t/job-archive/query.py` script), to ensure that job-archive has recorded the job's completion before moving on?
 * Since @chu11 wrote and maintains `job-archive` and its test, we should probably get his review as well.

cmoussa1:
Thanks @grondo!

> Do the jobs have to be sleep 1 for accounting purposes? If they could be hostname or sleep 0 jobs the test would run a little faster.

Yeah, part of the way the job-usage factor is calculated for a user/bank is looking at how much time a job runs for, so I thought `sleep 1` jobs would essentially be the minimum time required to generate at least a tiny usage factor for a user/bank. If there's a way to somehow populate the job-archive with potentially fake job records that take up more time, then that might work too? I'm not sure if that's possible.

> I don't really understand why you have to wait for the jobs to be inactive (clean event posted to job eventlog), then also wait for the jobs to be inactive in flux job list. Is the real potential for a race condition between the job clean event and the job being stored in the job-archive db? Maybe what you really need is the waid_db() shell function (and required t/job-archive/query.py script), to ensure that job-archive has recorded the job's completion before moving on?

Oh, I didn't know I could just use the `wait_db()` and `query.py` scripts. That would be better. Thanks for pointing that out. I can push up a fix to just use `wait_db()` and `query.py`. 

I also agree it might be helpful to get @chu11's review on this test. Thanks!

src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -136,7 +136,7 @@ def add_job_records(rows):
     return job_records
 
 
-def view_job_records(conn, output_file, **kwargs):
+def view_job_records(conn, output_file, cuf=False, **kwargs):
```
chu11, line 139: with the addition of this `cuf` parameter the functions logic gets a bit more confusing.  Perhaps it would it be worthwhile to split this function?  Perhaps something like `get_job_records()`and `output_job_records(job_records)`?  So there's no need for this extra parameter and the places that just need the job records returned can just call `get_job_records()`?

t/t1011-job-archive-interface.t
```diff
@@ -0,0 +1,95 @@
...
+
+DB_PATH=$(pwd)/FluxAccountingTest.db
+ARCHIVEDIR=`pwd`
+ARCHIVEDB="${ARCHIVEDIR}/jobarchive.db"
+QUERYCMD="flux python ${SHARNESS_TEST_SRCDIR}/scripts/query.py"
+
+export FLUX_CONF_DIR=$(pwd)
+test_under_flux 4 job
+
+wait_db() {
```
chu11, line 15: might be worthwhile to add a small comment on what this function does, took a little time to grok

chu11 (commented):
overall LGTM, just a few comments below

t/t1011-job-archive-interface.t
grondo: That function was lifted from flux-core job-archive tests. Maybe we could add the comment there and then copy that to this PR verbatim?

chu11: oh haha, i hadn't even realized I wrote it :P

chu11: just pushed PR in flux-core, @cmoussa1, you can copy and paste comments from there.  https://github.com/flux-framework/flux-core/pull/4167

cmoussa1: Great, thanks! I'll copy and paste here and force-push a fix.

cmoussa1: Just force-pushed a fix to [b8ac7d3](https://github.com/flux-framework/flux-accounting/pull/211/commits/b8ac7d313209b7326be1421de8b02c09647baf83) to add a function description that matches the one you wrote for flux-framework/flux-core#4167.

src/bindings/python/fluxacct/accounting/job_archive_interface.py
cmoussa1: Thanks for this suggestion. I agree, it would probably be best to split this up. I force-pushed a change to [a4b468a](https://github.com/flux-framework/flux-accounting/pull/211/commits/a4b468a5d82292874748430a29e2e7c619110ac9) that spits this function into two, `get_job_records()` and `output_job_records()`.

`get_job_records()` has pretty much all of the previous functionality of fetching job records from the job-archive DB, but does not print results or write them to an output file - it just returns a list of job records. This will be used by `calc_usage_factor()` when updating the job-usage values for every user/bank combo in the flux-accounting DB.

`output_job_records()` just calls `get_job_records()` and, based on the original argument `output_file`, decides whether to print the results to `stdout` or write them to an output file. Let me know what you think of this change.

chu11: change LGTM

chu11 (approved):

cmoussa1:
Thanks for approving @chu11! I'll set MWP

PR #214, opened 2022-03-03
build: ensure autogen.sh updates package version

cmoussa1:
I'm copying this PR from flux-framework/flux-core#4174 because I think I run into the same problem when it comes time to build the `.tar.gz` for flux-accounting. 

#### Problem 

`PACKAGE_VERSION` is not updated to match `git describe` in a development tree when `autogen.sh` is run.

The previous value is retained in the autom4te.cache directory, even when configure is regenerated, which seems counterintuitive.

---

This PR adds the `--force` option to autoreconf (as called from `autogen.sh`) to ensure that nothing is cached from previous builds.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -6,6 +6,6 @@
 #
 echo "Running libtoolize --automake --copy ... "
 libtoolize --automake --copy || exit
-echo "Running autoreconf --verbose --install"
-autoreconf --verbose --install || exit
+echo "Running autoreconf --force --verbose --install"
+autoreconf --force --verbose --install || exit
 echo "Now run ./configure."


garlick (approved):
Oh interesting, I hadn't considered that this might affect `make dist`.

cmoussa1:
Thanks @garlick! I should have probably double checked with you beforehand that this actually corrects the issue I was running into; sorry about that. Do the changes proposed in this PR affect the following lines in `configure.ac`?

```
AC_INIT([flux-accounting],
        m4_esyscmd([git describe --always | awk '/.*/ {sub(/^v/, ""); printf "%s",$1; exit}']))
```

What used to happen is when I would run `./autogen.sh && ./configure` after a new release of flux-accounting, the version number would not be updated, e.g after the most recent `v0.14.0` release, it would still show `v0.13.0`. Manually setting the current version number in that `AC_INIT` line used to be the way I would correct it.

grondo:
Another way to fix it might have been `make maintainer-clean`, but @garlick's fix is better by far.

Also: instead of manually uploading tarballs on release, now that one of your ci builds is running `make distcheck`, you could look into copying the automatic deploy ci actions from flux-core.

cmoussa1:
Ah, I hadn't realized using the `make maintainer-clean` command, thanks @grondo.

> Also: instead of manually uploading tarballs on release, now that one of your ci builds is running `make distcheck`, you could look into copying the automatic deploy ci actions from flux-core.

Oh that's right. I think I opened an issue on it not too long ago - I should definitely look into copying the automatic deploy. Thanks for the reminder!

cmoussa1:
Setting MWP. Thanks again for the quick review!

PR #219, opened 2022-03-10
job-archive interface: improve bank-specific usage calculation

### LINKED ISSUE #218: job-archive interface: only query jobs for a specific user/bank combo
In an offline discussion with @ryanday36, he reported that job usage values were updated for all rows of a user in the `association_table` even when jobs were only submitted under one bank.

I am pretty sure this is because in the `calc_usage_factor()` function, `SELECT` statement simply fetches all jobs under a specific user instead of a user/bank combination:

```sql
SELECT userid,id,t_submit,t_run,t_inactive,ranks,R FROM jobs WHERE userid=?
```

The job-archive interface needs to be edited to only query and count jobs from a specific user/bank combo.

cmoussa1:
#### Problem

As noted in #218, the job usage calculation function `calc_usage_factor()` calculates the same usage value for a user's default bank row and all secondary bank rows in the `association_table`. This results in the same usage value for all banks which the user belongs to, which is incorrect since jobs should only count towards usage for the bank they submitted jobs under.

---

This PR looks to improve the job-usage calculation for a user/bank row by only counting jobs that are submitted under the intended bank. 

The query used in the `update_job_usage()` function (the function responsible for iterating through all user/bank rows in the `association_table`) is  changed to include the `default_bank` column to help determine whether jobs should be counted towards a user's default bank or one of their secondary banks.

Two helper functions are added to `job_archive_interface.py`: `bank_specific_jobs()` and `check_jobspec()`. `bank_specific_jobs()` iterates through a user's job records and parses the jobspec of each record using `check_jobspec()` to look for a `bank` attribute. There are two cases to handle when looking at a jobspec:

1) A user is submitting a job under a secondary bank. In this case, `bank_specific_jobs()` only counts jobs that have the same secondary bank name attribute in the jobspec, or

2) A user is submitting a job under their default bank. `bank_specific_jobs()` will check and only count jobs that satisfy either one of the following scenarios:

- the user submitted a job while passing their default bank on the command line with `--setattr`, or
- the user did not specify any bank when submitting their job

The sharness test `t1011-job-archive-interface.t` is also improved by adding a secondary bank for the test user who submits jobs in the test. In the sharness test, additional jobs are submitted under this new secondary bank and checked to ensure the usage value _only under the secondary bank_ is updated. 

Fixes #218

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -13,6 +13,7 @@ import time
 import pwd
 import csv
 import math
+import json
 
 from flux.resource import ResourceSet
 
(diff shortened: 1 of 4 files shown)


cmoussa1:
> Maybe I'm missing something, but I'm not quite clear on how the bank_specific_jobs function is avoiding adding jobs that don't specify a bank ("bank" not in jobspec["attributes"]["system"]) to the job_records for all banks rather than just the default bank. It looks to me like if a user has jobs that are submitted with no bank, those get added to the jobs list in bank_specific_jobs regardless of whether the bank given as an argument to bank_specific_jobs is the default bank.

Yeah, this is my fault, as I was sacrificing readability for fewer lines of code, which in hindsight made things unnecessarily unclear, so I'm sorry about that. I think what I can do to make this clearer is have two separate functions, named `def_bank_jobs()` and `sec_bank_jobs()`. Each function will be responsible for counting jobs based on what bank the `update_usage()` function is on (either a user's default bank or one of their secondary banks), and the two functions can be called based on the following logic:

```python
if bank != default_bank:
    jobs = sec_bank_jobs(rows, bank) # we are only counting jobs for one of a user's secondary banks
else:
    jobs = def_bank_jobs(rows, default_bank) # we are only counting jobs for a user's default bank
``` 

In this case, I hope it's more clear which function is being called based on what bank we are currently counting jobs for, instead of mixing and mashing the two together and having some unnecessary complicated logic. Let me know what you think of this change.

> Similarly, does the test actually test whether usage when no bank is specified only gets added to the default bank, or just that the usage gets added to the default bank?

Yeah, I think the test doesn't do a great job of making it clear exactly what it's looking for, so I extended the test to `grep` for multiple strings on the same line. I can't `grep` the entire line itself because the "user" that submits jobs in GitHub actions has a different username than the user who submits jobs in my local Docker container, but I can `grep` for pretty much everything else; in this case, the bank the usage value is associated with, the usage value itself, and the fairshare value all on the same line:

```console
grep -E "account1" post_update1.test | grep "4" | grep "0.25"
```

Let me know if this makes the test any clearer.

ryanday36 (approved):
Looks good! thanks for fixing this up @cmoussa1.

cmoussa1:
Thanks @ryanday36! I'll set MWP on this then.

PR #220, opened 2022-03-16
[WIP] priority plugin: add configurable plugin factor weights for priority calculation

### LINKED ISSUE #128: priority plugin: implement custom configuration for factor weights
PR #122 adds a new multi-factor priority plugin that takes each factor and multiplies it by some weight, where the larger the weight, the more important the factor. Currently, the factors in the priority plugin are a set weight inside the calculation function and cannot be changed.

It would probably make sense to be able to define/configure these weights somewhere else where the calculation function could read from, similar to `slurm.conf`, where weights for different factors can be customized.

cmoussa1:
~~_This is a PR built on top of #207, a PR proposing adding a queue factor to priority calculation in the plugin._~~

#### Background

As more and more factors get added to the multi-factor priority plugin, there comes a need for configuring the weight of each factor to be more or less important in the overall calculation of a job's priority. For example, a scheduler operator may want to emphasize the importance of a user/bank's fairshare over what queue a job is submitted in, or vice versa.

Currently, the priority plugin defines the weights for each factor in the `priority_calculation()` function, which means the weights of both the **fairshare** and **queue** factors cannot be changed by a scheduler operator or sys admin.

---

This ~~[WIP]~~ PR looks to add configurable weights to both the flux-accounting DB and the priority plugin by adding a new table to the flux-accounting DB called `plugin_factor_table`. It will store all plugin factors and their associated weights, which are created when the database is first created. Commands to both view and edit these commands are also added, called `view-plugin-factor` and `edit-plugin-factor`, respectively. 

These factors and their weights are included in the bulk update script, where it is then unpacked by the plugin into a map, with the name of the factor as the key, and its associated weight as the value. Its weight is then fetched in `priority_calculation()` when calculating the priority of a user/bank combo's job. 

Both unit tests and sharness tests are added for the new commands and their effects on priority calculation in `t1013-plugin-weights.t`, which adds a user/bank combo who submits jobs under a number of different weight configurations for the various plugin factors.

Fixes #128

### DIFF
--- a/src/bindings/python/fluxacct/accounting/Makefile.am
+++ b/src/bindings/python/fluxacct/accounting/Makefile.am
@@ -17,7 +17,8 @@ TESTSCRIPTS = \
 	test/test_example.py \
 	test/test_job_archive_interface.py \
 	test/test_user_subcommands.py \
-	test/test_queue_subcommands.py
+	test/test_queue_subcommands.py \
+	test/test_plugin_factor_subcommands.py
 
 dist_check_SCRIPTS = \
 	$(TESTSCRIPTS)
--- a/src/bindings/python/fluxacct/accounting/Makefile.am
+++ b/src/bindings/python/fluxacct/accounting/Makefile.am
@@ -3,6 +3,7 @@ acctpy_PYTHON = \
 	user_subcommands.py \
 	bank_subcommands.py \
 	queue_subcommands.py \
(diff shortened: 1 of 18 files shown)


cmoussa1:
OK, I think I'm getting closer, so I'll mark this as ready for some early initial review.

Issue #221, opened 2022-03-18
plugin: is it possible to recover user/bank information on a plugin reload?


cmoussa1:
In an offline conversation with @ryanday36, he mentioned that when the priority plugin is re-loaded after a `systemctl restart flux`, `flux account-priority-update` had to be run again in order to re-populate the plugin with the user/bank information from the flux-accounting database. A suggestion was made to see if the data sent to the plugin could survive a restart so that the `flux account-priority-update` command does not have to be run again.

A couple of thoughts came to mind as I was thinking about this:

1) Is it possible to run flux commands in conjunction with loading a plugin? i.e, is it possible to automatically call `flux account-priority-update` after the plugin is loaded from a TOML file?

2) Is there a way to save the state of the plugin before Flux restarts? In this case, I don't think there would be a requirement to resend database information since it would already have user/bank information before the instance restarted.

3) When a `cron` job is set up, does the command set to be run by `cron` run instantly? I know our [Admin Guide](https://flux-framework.readthedocs.io/en/latest/adminguide.html) lists the set of flux-accounting commands to be run periodically through `cron`, so maybe I need to add a note in the guide saying that these flux-accounting commands should be set up just after loading the multi-factor priority plugin?

Regardless, maybe I should add some clearer documentation stating that the plugin won't allow jobs to run if it doesn't have any user/bank information from the database.

grondo:
It occurs to me a nice side-effect of storing some state in the KVS is that sysadmins and @ryanday36 could easily peek at the current JSON object and perhaps get some idea of the current state of the priority plugin.

cmoussa1:
Ah, nice, thanks @grondo for taking the time to answer these thoughts! I wasn't really sure what approach to take so I was just thinking of random ideas here, and your response to each helped provide a lot of insight.

> Another potential idea here would be to switch `flux account-priority-update` to use the KVS instead of a direct RPC to the priority plugin. The command would write or update a JSON object in the KVS at a well known location. The priority plugin could then use a KVS watch instead of an RPC to grab the updated information.
>
> The benefit is that the latest data is always in the KVS, and the priority plugin only needs one method to read the user/bank information. A drawback is that, unless you did something clever, the entire set of user/bank information would need to be written to the KVS on each update, which might be less efficient that what you are doing now (if you are only sending changes in information as updates).

[...]

PR #223, opened 2022-03-21
flux-accounting: add `etc/` directory, `rc1` script

cmoussa1:
_I'm posting this as an early work-in-progress, as I'm not sure I have everything required to add an `rc1` script to flux-accounting just yet, but I figured I would post it now._

#### Problem

As noted in #221, when the multi-factor priority plugin is re-loaded after a `systemctl restart flux`, `flux account-priority-update` has to be run again in order to re-populate the plugin with the user/bank information from the flux-accounting database. A suggestion was made to see if the data sent to the plugin could be initiated automatically so that the `flux account-priority-update` command does not have to be manually run again.

---

This PR adds an `etc/` directory to flux-accounting and adds an `rc1` script named `01-flux-account-priority-update`, which will attempt to populate a loaded multi-factor priority plugin with user/bank information from the flux-accounting DB on a system start or restart.

In the `Makefile.am`, `01-flux-account-priority-update` gets added to `dist_fluxrc1_SCRIPTS`, that when installed, is located in `/usr/etc/flux/rc1.d/`. I did some dummy testing in an interactive Docker container where I created a flux-accounting DB, loaded the plugin, and then ran `01-flux-account-priority-update` to ensure the information did in fact get sent and I could submit jobs. 

I just had a couple of questions as I am pretty unfamiliar with rc scripts, so I apologize in advance if the following questions are dumb 😅 :

1) I imagine the script could check that the priority plugin is loaded before it calls `flux account-priority-update`? Is there a way to do that in `bash`?

2) The `flux account-priority-update` command in the script doesn't have a path passed to it, so it will attempt to use the default path for the flux-accounting DB, located in `/var/lib/flux/`. Is that okay? 

3) Is it typical to add tests for these scripts? Maybe sharness tests?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,3 +14,6 @@ CODE_COVERAGE_IGNORE_PATTERN = \
     "/usr/lib/*"
 CODE_COVERAGE_LCOV_OPTIONS =
 @CODE_COVERAGE_RULES@
+
+dist_fluxrc1_SCRIPTS = \
+    etc/01-flux-account-priority-update
--- a/etc/01-flux-account-priority-update
+++ b/etc/01-flux-account-priority-update
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+if test $(flux getattr rank) -eq 0 \
+    && flux jobtap list | grep -q mf_priority.so; then
+     flux account-priority-update || true
+fi


cmoussa1:
> You can try `flux jobtap list | grep -q mf_priority.so`.

Ah cool, thanks. Do I capture a return value from `grep`? The `-q` flag suppresses output, right?  

> Is the database configured anywhere (sorry, I'm not remembering off the top of my head). If so, then could the command read this config file? I'm presuming `flux account-priority-update` should run the same way as it would from the cron job?

I think that makes sense. In `configure.ac`, this set of lines exists:

```
adl_RECURSIVE_EVAL([$localstatedir], acct_db_path)
AS_VAR_SET(acct_db_path, $acct_db_path)
AC_SUBST(acct_db_path)
```

And in the Python bindings, the `db_path` is set to `"@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db"`, where, if configured with `./configure --localstatedir=/var/`, sets the DB location to `/var/lib/flux/FluxAccounting.db`. Sorry if that isn't helpful though.

grondo:
I guess my question was whether it is possible for the syadmins to configure an alternate path for the accounting database?

grondo:
> Ah cool, thanks. Do I capture a return value from grep? The -q flag suppresses output, right?

Yeah, `-q` suppresses output. The command will return non-zero if the grep fails to match, so I think you can use this as:
```sh
    if flux jobtap list | grep -q mf_priority.so; then
        # Run flux account-priority-update here
    fi
```

cmoussa1:
> I guess my question was whether it is possible for the syadmins to configure an alternate path for the accounting database?

I'm not sure it's possible to configure an alternate path for the accounting database, but I guess technically an admin could specify a different path when initially creating the database with the `create-db` command, e.g they could say something like:

```console
flux account create-db -p /some/other/path/FluxAccounting.db
```

cmoussa1:
> Sorry, I meant a way for a user/admin to configure at runtime the location of the accounting db.

Oh, okay, then no, I don't believe there is a way to configure at runtime the location of the accounting DB.

grondo:
One suggestion: simply include the rc1 script in the top-level `Makefile.am`, and then you can drop the extra `etc/Makefile.am`. That is:

```diff
diff --git a/Makefile.am b/Makefile.am
index 58c7a36..b9cd2fc 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,3 +14,6 @@ CODE_COVERAGE_IGNORE_PATTERN = \
     "/usr/lib/*"
 CODE_COVERAGE_LCOV_OPTIONS =
 @CODE_COVERAGE_RULES@
+
+dist_fluxrc1_SCRIPTS = \
+       etc/01-flux-account-priority-update
```

Also, I tested this out and the behavior when the plugin is loaded but no database is configured is not pleasant :grin: 
```console
bash-4.4$ FLUX_CONF_DIR=conf flux start -s4 
2022-03-30T17:00:11.630994Z broker.err[0]: rc1.0: Database file does not exist: /usr/var/lib/flux/FluxAccounting.db
2022-03-30T17:00:11.647956Z broker.err[0]: rc1.0: /bin/bash -c /etc/flux/rc1 Exited (rc=1) 0.8s
2022-03-30T17:00:11.688914Z broker.err[1]: rc3.0: flux-module: broker.rmmod resource: Resource temporarily unavailable
2022-03-30T17:00:11.688911Z broker.err[3]: rc3.0: flux-module: broker.rmmod resource: Resource temporarily unavailable
2022-03-30T17:00:11.691374Z broker.err[2]: rc3.0: flux-module: broker.rmmod resource: Resource temporarily unavailable
2022-03-30T17:00:11.706191Z broker.err[3]: rc3.0: flux-module: broker.rmmod job-info: Resource temporarily unavailable
2022-03-30T17:00:11.706681Z broker.err[1]: rc3.0: flux-module: broker.rmmod job-info: Resource temporarily unavailable
2022-03-30T17:00:11.706793Z broker.err[2]: rc3.0: flux-module: broker.rmmod job-info: Resource temporarily unavailable
2022-03-30T17:00:11.721452Z broker.err[3]: rc3.0: flux-module: broker.rmmod job-ingest: Resource temporarily unavailable
2022-03-30T17:00:11.721658Z broker.err[2]: rc3.0: flux-module: broker.rmmod job-ingest: Resource temporarily unavailable
2022-03-30T17:00:11.723018Z broker.err[1]: rc3.0: flux-module: broker.rmmod job-ingest: Resource temporarily unavailable
2022-03-30T17:00:11.737849Z broker.err[2]: rc3.0: flux-module: broker.rmmod barrier: Resource temporarily unavailable
2022-03-30T17:00:11.737843Z broker.err[3]: rc3.0: flux-module: broker.rmmod barrier: Resource temporarily unavailable
2022-03-30T17:00:11.742142Z broker.err[1]: rc3.0: flux-module: broker.rmmod barrier: Resource temporarily unavailable
2022-03-30T17:00:11.752890Z broker.err[3]: rc3.0: flux-module: broker.rmmod kvs-watch: Resource temporarily unavailable
2022-03-30T17:00:11.754431Z broker.err[2]: rc3.0: flux-module: broker.rmmod kvs-watch: Resource temporarily unavailable
2022-03-30T17:00:11.763538Z broker.err[1]: rc3.0: flux-module: broker.rmmod kvs-watch: Resource temporarily unavailable
2022-03-30T17:00:11.770507Z broker.err[2]: rc3.0: flux-module: broker.rmmod kvs: Resource temporarily unavailable
2022-03-30T17:00:11.770611Z broker.err[3]: rc3.0: flux-module: broker.rmmod kvs: Resource temporarily unavailable
2022-03-30T17:00:11.779270Z broker.err[1]: rc3.0: flux-module: broker.rmmod kvs: Resource temporarily unavailable
2022-03-30T17:00:11.780350Z broker.err[3]: rc3.0: flux-content: content.flush: Resource temporarily unavailable
2022-03-30T17:00:11.781012Z broker.err[2]: rc3.0: flux-content: content.flush: Resource temporarily unavailable
2022-03-30T17:00:11.789658Z broker.err[1]: rc3.0: flux-content: content.flush: Resource temporarily unavailable
flux-start: 0 (pid 12764) exited with rc=1
bash-4.4$ 
```

I wonder if this script should be "best effort", i.e. emit an error but otherwise succeed so the instance doesn't fail to start?

Also, this rc1 script should only run on rank 0 right? Currently it is executing on all ranks.

Putting it all together, I suggest:

```diff
diff --git a/etc/01-flux-account-priority-update b/etc/01-flux-account-priority-update
index 631b7ad..f28e0e2 100644
--- a/etc/01-flux-account-priority-update
+++ b/etc/01-flux-account-priority-update
@@ -1,5 +1,6 @@
-#!/bin/bash -e
+#!/bin/bash
 
-if flux jobtap list | grep -q mf_priority.so; then
-    flux account-priority-update
+if test $(flux getattr rank) -eq 0 \
+   && flux jobtap list | grep -q mf_priority.so; then
+    flux account-priority-update || true
 fi
```

grondo:
Oh, meant to paste an example of the end result:
```console
bash-4.4$ FLUX_CONF_DIR=conf flux start -s4 
2022-03-30T17:10:23.074482Z broker.err[0]: rc1.0: Database file does not exist: /usr/var/lib/flux/FluxAccounting.db
bash-4.4$ flux resource list
     STATE NNODES   NCORES    NGPUS NODELIST
      free      4       16        0 asp,asp,asp,asp
 allocated      0        0        0 
      down      0        0        0 
bash-4.4$ flux mini run hostname
flux-mini: ERROR: [Errno 22] user not found in flux-accounting DB
bash-4.4$ 
```

grondo:
> I wonder, is it possible/good practice to check for the existence of a file in that if statement too? i.e can it check to see if the .db file exists before it tries to run flux account-priority-update?

It is a script, so you can basically do anything you want or need. If you want to check explicitly for the database before running `flux account-priority-update`, then you may have to create the script from a `.in` file to substitute the correct path. Other solutions would be to add an option to `flux account-priority-update` to avoid an error if the database file doesn't exist, but I think it is actually better the way it is now, where an error is logged but the script continues.

grondo (approved):
LGTM!

Looks like there is/was an unrelated `black` formatting error in the CI. I imagine we're going to hit this in flux-core CI as well. This SO post has a suggested workaround:

https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click

> As a workaround pin click to last version via pip install --upgrade click==8.0.2

cmoussa1:
Ugh. I noticed the `python format` check failing yesterday; I think it's due to some new version of `click` breaking the version of `black` formatter in https://github.com/psf/black/issues/2964. I should open a small PR to update the GitHub Actions to use the new version of black to fix this.

cmoussa1:
OK, rebased after #228 and it looks like CI is happy again. Thanks for the approval @grondo, I can set MWP!

PR #224, opened 2022-03-21
flux-accounting: miscellaneous fixes, improvements

cmoussa1:
Not a super high priority item, just some random fixes throughout the project I wanted to get posted.

* The `pop-db` command is updated to include the new `max_active_jobs` and `max_running_jobs` columns from the `association_table`. The outdated `QoS` column is replaced with `Queues`.
* The 'deleted' column in the `association_table` won't be able to actually represent a deleted row from the table
since the row is actually removed from the table itself, so just remove the column from the table.
* Some of the lines in `README.md` go way over ~~60~~ 80 characters, so break up those long lines into multiple lines.


grondo:
> Some of the lines in README.md go way over 60 characters, so break up those long lines into multiple lines.

Not a big deal, but did you mean 80 characters here?

cmoussa1:
Thanks for reviewing @ryanday36. I'll set MWP

PR #225, opened 2022-03-23
`edit-user`: change default value of "queues" optional argument

cmoussa1:
#### Background

In an offline conversation with @ryanday36, he reported that when trying to edit a value in a user/bank row in the `association_table`, a `Queue does not exist in the queue_table` message would pop up and prevent the row from being successfully edited. 

```console
[root@machine:~]# flux account edit-user --max-active-jobs 999 user1
Queue specified does not exist in queue_table
```

This is because the "queues" optional argument in the `edit-user` command has a default value of `""` instead of `None`, which means when someone tries to edit a field for a user, the `validate_queues()` function will be triggered because the `edit_user()` function thinks that a queue of name `""` has been passed, which is incorrect.

---

This PR changes the default value of queues to `None` so that when a field other than "queues" is being edited, the `validate_queue()` function does not run.

It also makes some small changes and improvements to the `flux account` tests in `t1007-flux-account.t` and adds a new `edit-user` test to make sure that the `validate_queues()` helper function does not get called when the `queues` column is not being edited in a user/bank row.

### DIFF
--- a/src/cmd/flux-account.py
+++ b/src/cmd/flux-account.py
@@ -142,7 +142,7 @@ def add_edit_user_arg(subparsers):
     subparser_edit_user.add_argument(
         "--queues",
         help="queues the user is allowed to run jobs in",
-        default="",
+        default=None,
         metavar="QUEUES",
     )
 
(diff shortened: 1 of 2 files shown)


ryanday36 (approved):
looks good.

cmoussa1:
Thanks! Setting MWP

Issue #222, opened 2022-03-20
plugin: change behavior in `job.validate` when user/bank combo can't be found


cmoussa1:
Currently, the priority plugin rejects jobs submitted by a user when the user cannot be found in the plugin's user/bank map data structure. This will reject all jobs until the data from the flux-accounting DB is sent to the plugin via `flux account-priority-update`. The behavior in the plugin should be changed to instead hold newly submitted jobs in the PRIORITY state with `flux_jobtap_priority_unavail ()` until the first `flux account-priority-update`.

On an update, it can try to reprioritize all jobs with a call to `flux_jobtap_reprioritize_all ()`, which will in turn invoke the `job.priority.get` callback topic for all pending jobs.

grondo:
As discussed in #221, on jobtap plugin load, `job.new` is currently called on all active jobs, not `job.validate`. On a broker restart, the sequence would likely be `job.new`->`job.state.priority`, and since I think the `mf_priority.so` plugin does not have a `job.new` callback, the bank info will not be available for the job in `job.state.priority` and so a job exception will be raised.

As a start, the job exception here

https://github.com/flux-framework/flux-accounting/blob/c6aacd318372ff2b1cc1660b3fb1f42787a7cc82/src/plugins/mf_priority.cpp#L79-L84

could be relaxed to instead return `flux_jobtap_priority_unavail()` when the priority calculation is done for the `job.state.priority` callback.

Then, when bank information is added via `flux account-prioirty-update`, you could perhaps add the missing bank info to each job still in PRIORITY state and release the jobs by setting a valid priority. Whether this is the right approach or not would be evident after adding some testing.

cmoussa1:
This should be handled by #227, so I'll close this. I can always open a more specific issue if something else arises.

grondo:
I was actually thinking a bit about this today. I'll avoid opening a new issue just yet, but wanted to get something down to think about.

The recent changes fix an issue where jobs are rejected because the bank information for a user hasn't yet been loaded into the plugin by an external source. However, now there is a case where a user can typo a bank name, and this won't be caught as an error, instead the job will sit in PRIORITY state forever.

[...]

PR #227, opened 2022-03-29
plugin: allow user/bank information to be loaded late, add job.new callback

cmoussa1:
#### Background

The plugin currently rejects jobs in `job.validate` when it cannot find the userid of the user/bank combo submitting the job in its internal map data structure. A suggestion was made in #222 to change this behavior to instead hold the job until the correct information was received. Then, upon a data update via `flux account-priority-update`, the plugin could reprioritize all jobs using `flux_jobtap_reprioritize_all ()` with the newly pushed information and release any jobs still waiting on user/bank information.

---

This PR looks to add support for holding jobs when the user/bank information cannot be found in the plugin's internal map. It removes the rejection in `job.validate` and instead creates a special temporary entry in the users map and sets a special `max_run_jobs` value to let the `job.state.priority` callback know to return `flux_jobtap_priority_unavail ()` on this and future submitted jobs from this user until we get their information from the flux-accounting DB.

It also adds a new callback to the plugin for `job.new`, which does much of the same user/bank lookup as `job.validate`, but this callback also includes the accounting for newly submitted jobs by incrementing the user/bank's active jobs count. Like `job.validate`, it will also create a special entry for user/bank combos when it cannot find the user in its internal map, which subsequently holds the submitted job by returning `flux_jobtap_priority_unavail ()`. The `job.new` callback is inserted to handle the following cases (referenced from flux-core's [manpage](https://github.com/flux-framework/flux-core/blob/master/doc/man7/flux-jobtap-plugins.rst) on jobtap plugins):

- on job submission
- when the job manager is restarted and has reloaded a job from the KVS
- when a new plugin is loaded

A new service is also added to the priority plugin, which, when called, will call `flux_jobtap_reprioritize_all ()` on all jobs. This will attempt to assign a priority to jobs that are currently held due to missing user/bank information.

Finally, this PR adds a second lookup on held jobs due to missing user/bank information in `job.state.priority`. If no user/bank information is found, `flux_jobtap_priority_unavail ()` is returned and the job is held in PRIORITY state. If updated information is found, it will update both the current job's `bank_info` struct so that a valid priority can be calculated for the job as well as update the `cur_active_jobs` and `cur_run_jobs` counters in the internal users map since the job will not transition back to `job.state.depend` or `job.new`.

Basic tests are added in a new sharness test file `t1012-mf-priority-load`. This test file goes through the following scenario:

1. the plugin is loaded but no user/bank information has been sent from the database to the plugin yet
2. a user/bank combo submits a job to the plugin; it gets held since no information about the user/bank was found in the plugin yet
3. information for the user gets pushed to the plugin and the trigger callback gets activated, which runs `flux_jobtap_reprioritize_all ()`
4. the previously held job transitions to RUN state
5. more jobs are submitted to check that the user/bank's `max_run_jobs` is correctly enforced

The test for a successful job rejection in `t1001-mf-priority-basic.t` has also been removed since that is not the behavior of the plugin anymore.

###### edit commit message for Python black formatter commit ✅

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -157,6 +157,24 @@ error:
 }
 
 
+static void reprior_cb (flux_t *h,
+                        flux_msg_handler_t *mh,
+                        const flux_msg_t *msg,
+                        void *arg)
+{
+    flux_plugin_t *p = (flux_plugin_t*) arg;
+
+    if (flux_jobtap_reprioritize_all (p) < 0)
+        goto error;
+    if (flux_respond (h, msg, NULL) < 0)
+        flux_log_error (h, "flux_respond");
+    return;
+error:
+    flux_respond_error (h, msg, errno, flux_msg_last_error (msg));
+
+}
+
+
 /*
  * Unpack the urgency and userid from a submitted job and call
  * priority_calculation (), which will return a new job priority to be packed.
@@ -542,7 +560,8 @@ static const struct flux_plugin_handler tab[] = {
 extern "C" int flux_plugin_init (flux_plugin_t *p)
 {
     if (flux_plugin_register (p, "mf_priority", tab) < 0
-        || flux_jobtap_service_register (p, "rec_update", rec_update_cb, p) < 0)
+        || flux_jobtap_service_register (p, "rec_update", rec_update_cb, p)
+        || flux_jobtap_service_register (p, "reprioritize", reprior_cb, p) < 0)
         return -1;
     return 0;
 }
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -313,6 +313,118 @@ static int validate_cb (flux_plugin_t *p,
 }
 
 
+static int new_cb (flux_plugin_t *p,
+                        const char *topic,
+                        flux_plugin_arg_t *args,
+                        void *data)
+{
+    int userid;
+    char *bank = NULL;
+    int max_run_jobs, cur_active_jobs, max_active_jobs = 0;
+    double fairshare = 0.0;
+    struct bank_info *b;
+
+    std::map<int, std::map<std::string, struct bank_info>>::iterator it;
+    std::map<std::string, struct bank_info>::iterator bank_it;
+
+    flux_t *h = flux_jobtap_get_flux (p);
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:i, s{s{s{s?s}}}}",
+                                "userid", &userid,
+                                "jobspec", "attributes", "system",
+                                "bank", &bank) < 0) {
+        return flux_jobtap_reject_job (p, args, "unable to unpack bank arg");
+    }
+
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b != NULL) {
+        max_run_jobs = b->max_run_jobs;
+        fairshare = b->fairshare;
+        cur_active_jobs = b->cur_active_jobs;
+        max_active_jobs = b->max_active_jobs;
+    } else {
+        // make sure user belongs to flux-accounting DB
+        it = users.find (userid);
+        if (it == users.end ()) {
+            // user does not exist in internal map yet, so create a bank_info
+            // struct that signifies it's going to be held in PRIORITY
+            add_missing_bank_info (p, h, userid);
+            return 0;
+        }
+
+        // make sure user belongs to bank they specified; if no bank was passed
+        // in, look up their default bank
+        if (bank != NULL) {
+            bank_it = it->second.find (std::string (bank));
+            if (bank_it == it->second.end ()) {
+                flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                             "mf_priority", 0,
+                                             "job.new: not a member of %s",
+                                             bank);
+                return -1;
+            }
+        } else {
+            bank = const_cast<char*> (users_def_bank[userid].c_str ());
+            bank_it = it->second.find (std::string (bank));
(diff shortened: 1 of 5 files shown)


cmoussa1:
OK, I think I'll mark this is as ready for some very initial review whenever convenient (_BTW, I don't expect this to land with the release of flux-accounting `v0.15.0` since we are at the end of our monthly release cycle_).

grondo (commented):
@cmoussa1 - just did a very quick pass on this one, still haven't tested on my test system though. Thought I'd share current comments thus far.

src/plugins/mf_priority.cpp
```diff
@@ -307,6 +307,128 @@ static int validate_cb (flux_plugin_t *p,
...
+
+    // if a user's fairshare value is 0, that means they shouldn't be able
+    // to run jobs on a system
+    if (fairshare == 0)
+        return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
+
+    // if a user/bank has reached their max_active_jobs limit, subsequently
+    // submitted jobs will be rejected
+    if (max_active_jobs > 0 && cur_active_jobs >= max_active_jobs)
+        return flux_jobtap_reject_job (p, args, "user has max active jobs");
```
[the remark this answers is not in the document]
cmoussa1: Ah, I see. Yeah, I think that, for the most part, I just duplicated what I had in `job.validate`, which is why the `flux_jobtap_reject_job ()` calls are in here, so I don't think these lines are covered by tests so far, sorry about that. 🤦‍♂️

> Maybe we should discuss if we should address that issue in flux-core before merging this change, if it makes your work here more straightforward?

I think that sounds like a good plan. I wonder if it would be okay to instead raise a job exception here in `job.new` instead of rejecting the job?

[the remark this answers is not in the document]
grondo: Yeah, that might be reasonable for now.

cmoussa1:
Thanks for the early initial review @grondo - just squashed and pushed up changes based on the above suggestions.

src/plugins/mf_priority.cpp
```diff
@@ -184,6 +204,71 @@ static int priority_cb (flux_plugin_t *p,
...
+            // passed in, look up their default bank
+            if (bank != NULL) {
+                bank_it = it->second.find (std::string (bank));
+                if (bank_it == it->second.end ())
+                    flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                                 "mf_priority",
+                                                 0,
+                                                 "job.state.priority: user " \
+                                                 "does not belong to " \
+                                                 "specified bank");
```
grondo, line 239: The error emitted by `flux_jobtap_raise_exception()` is directed at the user. It may be confusing to include the plugin topic string in the exception note, so I'd just drop that. Also, it may be useful to include the specified bank name, e.g. (untested):

```suggestion
                    flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
                                                 "mf_priority",
                                                 0,
                                                 "not a member of bank %s",
                                                 bank);
```

src/plugins/mf_priority.cpp
```diff
@@ -184,6 +204,71 @@ static int priority_cb (flux_plugin_t *p,
...
+            } else {
+                bank = const_cast<char*> (users_def_bank[userid].c_str ());
+                bank_it = it->second.find (std::string (bank));
+                if (bank_it == it->second.end ())
+                    flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                                 "mf_priority",
+                                                 0,
+                                                 "job.state.priority: " \
+                                                 "user/default bank entry " \
+                                                 "does not exist");
```
grondo, line 249: similar comments to above: no need for line continuation and generalize error message.

src/plugins/mf_priority.cpp
```diff
@@ -298,11 +477,31 @@ static int validate_cb (flux_plugin_t *p,
...
+    b->max_active_jobs = 0;
+    b->cur_active_jobs = 0;
+    b->held_jobs = std::vector<long int>();
+
+    if (flux_jobtap_job_aux_set (p,
+                                 FLUX_JOBTAP_CURRENT_JOB,
+                                 "mf_priority:bank_info",
+                                 b,
+                                 NULL) < 0)
+        flux_log_error (h, "flux_jobtap_job_aux_set");
```
grondo, line 504: I think I saw this `no_user_info` block twice? Maybe this could/should be a function?

src/plugins/mf_priority.cpp
cmoussa1: Thanks for this suggestion! I just squashed and pushed a change onto [d454f3f](https://github.com/flux-framework/flux-accounting/pull/227/commits/d454f3f16b64293a80ab0b3013d76ec360e28fc3) that includes the bank name in the exception message. I also added a test in `t1012-mf-priority-load.t` that checks for this exception message when a user submits a job using an incorrect bank name before the user/bank information is sent to the plugin.

src/plugins/mf_priority.cpp
```diff
@@ -184,6 +204,71 @@ static int priority_cb (flux_plugin_t *p,
...
 
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.state.priority: bank info is " \
+                                     "missing");
```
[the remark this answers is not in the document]
cmoussa1: Ah, okay, cool. I didn't know you could omit the `\`. I just squashed and pushed a change to remove those forward slashes and change the exception message.

grondo (approved):
Did some basic testing and didn't see any issues. :+1:

Nice work @cmoussa1!

cmoussa1:
Awesome, thanks @grondo! I guess this could make the `v0.15.0` release after all! I'll set MWP shortly

grondo:
Yeah, it will be good to get this tested on fluke

PR #228, opened 2022-03-30
ci: pin click package to 8.0.2

cmoussa1:
#### Problem

A recent release of `click` breaks the version of `black` used in flux-accounting's GitHub Actions as noted by this issue thread:

https://github.com/psf/black/issues/2964

---

Pin the `click` package to 8.0.2 to prevent the break (reference: https://stackoverflow.com/questions/71673404)

I think since this PR modifies GitHub actions, it might need a manual merge?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
         pip install -r requirements.txt
     - name: Lint with black
       run: |
-        pip install black==20.8b1
+        pip install --upgrade black==20.8b1 click==8.0.2
         # return the status of running black formatter
         black . --check
 


.github/workflows/main.yml
```diff
@@ -34,6 +34,7 @@ jobs:
     - name: Lint with black
       run: |
         pip install black==20.8b1
+        pip install --upgrade click==8.0.2
```
grondo, line 37: I think you can put this on the same install line as `black`, e.g.
```
  pip install --upgrade black==20.8b1 click==8.0.2
```
?

grondo (approved):
LGTM, but one suggestion inline. Let me know when you're ready and I'll manually merge.

grondo:
Nice, I'll press the button once CI completes.

cmoussa1:
Thanks @grondo! Was just about to let you know I was ready. :-)

PR #230, opened 2022-04-15
plugin: change behavior for users who have extremely low/zero fairshare

### LINKED ISSUE #141: priority plugin: should jobs be rejected when a user's fairshare value is 0?
A good question that was brought up in #131 was handling submitted jobs when a user's fairshare value is 0. Currently, the behavior of the plugin is to reject the job with a message explaining that the user's fairshare value is 0. But perhaps the behavior of the plugin should instead be to hold the job until:

1) a user's fairshare value is updated to be > 0, and/or
2) the user has not exceeded their max jobs limit

I _believe_ this would include removing the rejection in `validate_cb ()` and adding a check in `priority_cb ()` to compare the fairshare value to 0; if so, we could return `FLUX_JOB_URGENCY_HOLD` or `FLUX_JOB_PRIORITY_MIN`.

I think it might be good to get @ryanday36's thoughts on how the plugin should behave as well. Does SLURM currently **hold** or **reject** jobs that are submitted under a user whose fairshare value is 0?

cmoussa1:
#### Problem 

The multi-factor priority plugin currently rejects jobs from a user who has a fairshare value of exactly 0.0. As discussed in #141, the plugin should instead assign the lowest job priority for the job submitted.

---

This PR removes the job rejection from the plugin when a user has a fairshare value lower than 0.00000 and instead returns `FLUX_JOB_PRIORITY_MIN` when calculating the priority of a job. 

When a fairshare update occurs and updated information is sent to the plugin and jobs are reprioritized, there is potential to update the existing priority for a job and raise its priority to higher than `FLUX_JOB_PRIORITY_MIN`.

The test that checks for a successful job rejection in `t1001-mf-priority-basic.t` is also updated to instead check that the assigned priority is `FLUX_JOB_PRIORITY_MIN`. An updated fairshare value is then sent to the plugin and jobs are reprioritized, bringing the job out of SCHED state and into RUN.

Fixes #141

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -554,11 +554,6 @@ static int validate_cb (flux_plugin_t *p,
     cur_active_jobs = bank_it->second.cur_active_jobs;
     max_active_jobs = bank_it->second.max_active_jobs;
 
-    // if a user's fairshare value is 0, that means they shouldn't be able
-    // to run jobs on a system
-    if (fairshare == 0)
-        return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
-
     // if a user/bank has reached their max_active_jobs limit, subsequently
     // submitted jobs will be rejected
     if (max_active_jobs > 0 && cur_active_jobs >= max_active_jobs)
@@ -651,11 +646,6 @@ static int new_cb (flux_plugin_t *p,
 
         b = &bank_it->second;
     }
-
-    // if a user's fairshare value is 0, that means they shouldn't be able
-    // to run jobs on a system
-    if (fairshare == 0)
-        return flux_jobtap_reject_job (p, args, "user fairshare value is 0");
 
     // if a user/bank has reached their max_active_jobs limit, subsequently
     // submitted jobs will be rejected
--- a/t/t1001-mf-priority-basic.t
+++ b/t/t1001-mf-priority-basic.t
@@ -151,35 +151,6 @@ test_expect_success 'reject job when invalid bank format is passed in' '
 	grep "unable to unpack bank arg" invalid_fmt.out
 '
 
-test_expect_success 'create a fake payload with a 0 fairshare key-value pair' '
-	cat <<-EOF >empty_fairshare.py
-	import flux
-	import pwd
-	import getpass
-	import json
-
-	username = getpass.getuser()
-	userid = pwd.getpwnam(username).pw_uid
-	# create an array of JSON payloads
-	bulk_update_data = {
-		"data" : [
-			{"userid": userid, "bank": "account4", "def_bank": "account3", "fairshare": 0.0, "max_running_jobs": 10, "max_active_jobs": 12}
-		]
-	}
-	flux.Flux().rpc("job-manager.mf_priority.rec_update", json.dumps(bulk_update_data)).get()
-	EOF
-'
-
-test_expect_success 'update plugin with sample test data' '
-	flux python empty_fairshare.py
-'
-
-test_expect_success 'submit a job with new bank and 0 fairshare should result in a job rejection' '
-	test_must_fail flux mini submit --setattr=system.bank=account4 hostname > zero_fairshare.out 2>&1 &&
-	test_debug "zero_fairshare.out" &&
-	grep "user fairshare value is 0" zero_fairshare.out
-'
-
 test_expect_success 'pass special key to user/bank struct to nullify information' '
 	cat <<-EOF >null_struct.json
 	{
--- a/t/t1001-mf-priority-basic.t
+++ b/t/t1001-mf-priority-basic.t
@@ -90,7 +90,8 @@ test_expect_success 'submit a job with default urgency' '
 	cat <<-EOF >job1.expected &&
 	45321
 	EOF
-	test_cmp job1.expected job1.test
+	test_cmp job1.expected job1.test &&
+	flux job cancel $jobid
 '
 
 test_expect_success 'submit a job with custom urgency' '
@@ -99,7 +100,8 @@ test_expect_success 'submit a job with custom urgency' '
 	cat <<-EOF >job2.expected &&
 	45320
 	EOF
-	test_cmp job2.expected job2.test
+	test_cmp job2.expected job2.test &&
+	flux job cancel $jobid
 '
 
 test_expect_success 'submit a job with urgency of 0' '
@@ -118,7 +120,8 @@ test_expect_success 'submit a job with urgency of 31' '
 	cat <<-EOF >job4.expected &&
 	4294967295
 	EOF
-	test_cmp job4.expected job4.test
+	test_cmp job4.expected job4.test &&
+	flux job cancel $jobid
 '
 
 test_expect_success 'submit a job with other bank' '
@@ -127,7 +130,8 @@ test_expect_success 'submit a job with other bank' '
 	cat <<-EOF >job5.expected &&
 	11345
 	EOF
-	test_cmp job5.expected job5.test
+	test_cmp job5.expected job5.test &&
+	flux job cancel $jobid
 '
 
 test_expect_success 'submit a job using default bank' '
@@ -136,7 +140,8 @@ test_expect_success 'submit a job using default bank' '
 	cat <<-EOF >job6.expected &&
 	45321
 	EOF
-	test_cmp job6.expected job6.test
+	test_cmp job6.expected job6.test &&
+	flux job cancel $jobid
 '
 
 test_expect_success 'submit a job using a bank the user does not belong to' '
@@ -201,7 +206,8 @@ test_expect_success 'resend user/bank information with valid data and successful
 	cat <<-EOF >job2.expected &&
 	45321
 	EOF
-	test_cmp job2.expected job2.test
+	test_cmp job2.expected job2.test &&
+	flux job cancel $jobid2
 '
 
 test_done


src/plugins/mf_priority.cpp
```diff
@@ -108,6 +108,13 @@ int64_t priority_calculation (flux_plugin_t *p,
     fshare_factor = b->fairshare;
     queue_factor = b->queue_factor;
 
+    priority = round ((fshare_weight * fshare_factor) +
+                      (queue_weight * queue_factor) +
+                      (urgency - 16));
+
+    if (fshare_factor < 0.00000)
```
grondo, line 115: Question: The commit message states:

>  The multi-factor priority plugin currently rejects jobs from a user who has a fairshare value of 0.0.

Is this supposed to be "rejects jobs from a user who has a fairshare value of _less than_ 0.0" ?

Also, could this test be moved above the calculation of `priority` since it immediately returns? (Thus the `priority` is done needlessly) Apologies if I'm misreading the code in diff format.

src/plugins/mf_priority.cpp
```diff
@@ -108,6 +108,13 @@ int64_t priority_calculation (flux_plugin_t *p,
...
+            return FLUX_JOB_PRIORITY_MIN;
+
     priority = round ((fshare_weight * fshare_factor) +
```
grondo, line 118: Maybe I'm reading the diff wrong, but is `priority` re-calculated here? Perhaps the additional one above wasn't meant to be added to the file.

src/plugins/mf_priority.cpp
grondo, line 115: Also, I don't _think_ `0.00000` is any different than `0.0` or even `0.`.

t/t1001-mf-priority-basic.t
```diff
@@ -90,7 +90,8 @@ test_expect_success 'submit a job with default urgency' '
 	cat <<-EOF >job1.expected &&
 	45321
 	EOF
-	test_cmp job1.expected job1.test
+	test_cmp job1.expected job1.test &&
```
grondo, line 93: Suggestion: add a commit body to this commit that explains the problem the commit is trying to solve (i.e. why the missing `flux job cancel` is problematic), then how that problem is solved by the commit.

grondo (commented):
I was a little confused by the changes in `priority_calculation()` and thought perhaps there was a merge error or something that duplicated the block that calculates `priority`.

Also, you have
```c    
    if (fshare_factor < 0.00000)
            return FLUX_JOB_PRIORITY_MIN;
```

which makes me wonder why `0.00000` was used instead of `0.`. I wonder if you maybe meant a very small value above zero, instead of strictly `0`,  like `1e-5`?

I don't really know how the fairshare factor is calculated, so this confusion may be due to that lack of knowledge!

src/plugins/mf_priority.cpp
cmoussa1: Yup, you're completely right - I must have made a mistake during my rebase and obviously things got a little haywire 🤦‍♂️ There should only be one calculation, and the test for `fshare_factor` is in fact before the priority calculation, as you mentioned above. I'll push up a fix, I'm sorry about that!

src/plugins/mf_priority.cpp
cmoussa1: Yeah, in fact maybe I could use some clarification here too. I always thought that a double value of `0.0001` was treated as larger than `0.00000`. The flux-accounting DB stores and the bulk update script sends fairshare values to the 5th decimal place, so that's why I included the test for `< 0.00000`, and my assumption was that a user/bank combo who had a fairshare value of `0.00001` or larger would be allowed to run jobs (although that is really really low). But now that I'm typing this out, I'm thinking it would be better to instead just compare to exactly `0`, as you suggested above? As that's what the test should really be looking for, for a value of exactly `0`.

grondo: Yes, `0.001` is larger than `0.00000`, but like I said, `0.00000` and `0.0` (and `0.`) are all the same as far as I know.

Note that I think you have to be careful with calculated floating point numbers due to rounding errors, It is very difficult to get an "exact" number with floating point calculations (and I didn't see if the fairshare value is rounded). So you could calculate a fairshare value that is _very_ close to zero (e.g. `1.0e-10`) and which it may be convenient to treat as a literal zero. However, I'm not sure that a real problem here since even `1.0e-10` is greater than zero. Just thought I'd give something to think about (I'm not sure if there is potential for a rounding error to cause a fairshare value to be a very very small negative number which should instead be treated as `0`.)

src/plugins/mf_priority.cpp
grondo: Ah, that is what I figured :+1:

src/plugins/mf_priority.cpp
cmoussa1: Yeah, I think I see what you're saying. Thanks for this note. I don't _think_ it should be an issue since fairshare values are not rounded in the priority calculation, and only the final priority is rounded. So I think I'll push a fix to just compare `fshare` to `0`.

cmoussa1:
OK, just squashed and pushed a couple of changes based on the good suggestions made above:

- the comparison of the value of `fshare_factor` is made to just `0`
- the body of the commit message for adding `flux job cancel` to `t1001-mf-priority-basic.t` was also added, explaining the reasoning for the addition

src/plugins/mf_priority.cpp
```diff
@@ -554,11 +557,6 @@ static int validate_cb (flux_plugin_t *p,
     cur_active_jobs = bank_it->second.cur_active_jobs;
     max_active_jobs = bank_it->second.max_active_jobs;
 
-    // if a user's fairshare value is 0, that means they shouldn't be able
```
grondo, line 557: The commit message still has me a little bit confused, though I'm sure it is just me. However, for my own edification, I'd like to understand. The commit message says:

>  The multi-factor priority plugin currently rejects jobs from a user
> who has a fairshare value of 0.0. To align this behavior more closely with
> the current implementation on our systems, the plugin should instead assign
> the lowest job priority for the job submitted.

However, the change as I see it in the code is that the minimum priority is now returned if the user's fairshare is _less than_ zero, not equal to zero. Jobs with `0` fairshare are now no longer rejected, but they don't get assigned `FLUX_JOB_PRIORITY_MIN` - instead they appear to get whatever priority is calculated in `priority_calculation()` taking into account the fairshare value of `0.`.

Is that the intended operation? If so maybe rewording the commit message would help.

cmoussa1: Ohhh, no it's not you at all, you are absolutely right. Right now the change I proposed is checking that a user's fairshare is _less than_ zero, not _less than or equal to 0_; the latter is the behavior I am trying to add in. I suppose I should change this comparison to

```c++
if (fshare_factor <= 0)
    return FLUX_JOB_PRIORITY_MIN;
```

So now if a user's fairshare is `0.0`, it should return a minimum priority. I should have noticed that myself, sorry about that. I can push up a fix.

grondo: Ok, that makes more sense! Thanks

cmoussa1: I'm sorry, I meant to ping you that I pushed up a fix. I was waiting for Codecov to publish the report so that I could check that the lines were actually covered, but it's slower than I thought. :/

grondo: @dongahn, I knew that testing for equality was not safe but here the code is testing `<= 0.` which I had considered "safe". Please advise if my assumption is incorrect. @cmoussa1, I guess one thing to consider is if you want to treat a very small value _epsilon_ as equivalent to zero, then you might want to use

```c
    if (fshare_factor < std::numeric_limits<double>::epsilon ())
        return FLUX_JOB_PRIORITY_MIN;
```

I'm not 100% sure if that is necessary in this case.

dongahn: >Oh, nice, thanks for pointing this out! In the context of this PR, I am comparing a double to an int instead of comparing two doubles. I'm guessing the precautions you pasted above also apply to this scenario? 

I guess so. The integer will be promoted to `double` for comparison...

dongahn: Can `fshare_factor` become a negative?

cmoussa1: Not to my knowledge, no. `fshare_factor` should be in the range `0.0 <= fshare_factor <= 1.0`

dongahn: So is this comparison just a sanity check that you won't have a negative`fshare_factor`? Or are you trying to set some threshold on a very small factor value?

grondo: Ah, I apologize, some of my comments may have been confusing then since I assumed `fshare_factor` < 0 was a possibility.

cmoussa1: The comparison `fshare_factor <= 0` was put in place in the case that `fshare_factor` was `0.0`, a specific job priority `FLUX_JOB_PRIORITY_MIN` should be returned. I hadn't used `==` for comparison since I had a little knowledge (but definitely not extensive) of the dangers of comparing double values using `==`, so that's why I had used `<=`. 

So maybe yours and @grondo's suggestion of using `std::numeric_limits<double>::epsilon ()` should be used instead?

dongahn: @cmoussa1: From the code comment:

```
 * fairshare: the ratio between the amount of resources allocated vs. resources
 *     consumed.
 ```

Can `fshare_factor` really become zero? Probably just closer and closer to zero but never become zero since this is a ratio with a positive nominator? 

In any case, what is the criteria in determining the threshold to return `FLUX_JOB_PRIORITY_MIN` with? Why should that be `std::numeric_limits<double>::epsilon ()`?

Comment (cmoussa1) [reply src/plugins/mf_priority.cpp:557]: Ohhh, now that I'm typing this out, I _do_ briefly remember an offline conversation you and I had that `fshare_factor` could never really become zero, it could just get closer and closer to zero; I think I had an experimental branch where I confirmed it too. I'm sorry I didn't remember that!

Comment (cmoussa1) [reply src/plugins/mf_priority.cpp:557]: > In any case, what is the criteria in determining the threshold to return `FLUX_JOB_PRIORITY_MIN` with? Why should that be `std::numeric_limits<double>::epsilon ()`?

If I am understanding your question correctly, the criteria for returning `FLUX_JOB_PRIORITY_MIN` was if `fshare_factor` was truly `0.0`, but, if `fshare_factor` can never really become zero, then perhaps it is not necessary to check for this in the first place. Sorry if I misunderstood though.

Comment (dongahn) [reply src/plugins/mf_priority.cpp:557]: @cmoussa1: I was just trying to understand your algorithm a bit. What is the purpose of `FLUX_JOB_PRIORITY_MIN`? Do you have to signal the upper layer about a certain condition? If that condition never happens, it seems that the check is defeating the purpose. But if this is a condition that can happen, I thought I could be a sounding board for you. 

BTW I also see

```c
    priority = round ((fshare_weight * fshare_factor) +
                      (queue_weight * queue_factor) +
                      (urgency - 16));

    if (priority < 0)
        return FLUX_JOB_PRIORITY_MIN;
```

Can `priority` become zero here as well?

I don't think I understand the entire picture. Sorry if this is confusing you.

cmoussa1: > What is the purpose of FLUX_JOB_PRIORITY_MIN?

This is an integer priority (0) which signals that the job should not exit out of PRIORITY state and, as a result, will be held until it receives an integer priority > 0. This was supposed to be the priority returned for a job submitted by a user whose fairshare value was `0.0`, but I think we've since determined such condition is not possible, since fairshare can only get really really close to `0.0`.

> Can `priority` become zero here as well?

Yes, I think technically `priority` can become zero or less than zero. For example, if the weight for both factors (`fshare_weight` and `queue_weight`) are both 0, and the `urgency` for that job was < 16, then `priority` would be negative. But since we don't want to return a negative priority, we just return `FLUX_JOB_PRIORITY_MIN`. I think this is a very unlikely scenario however, as by default we would want the weights for both factors (`fshare_weight` and `queue_weight`) to be large positive integers, and I think only a sys admin would have the privilege to set an urgency < 16 for a job. I hope this makes sense.

dongahn: All make sense. I think you could either just not do that floating point comparison or have a configurable (or set) threshold, `t` such that if `fshare_factor` becomes less than that value, we will deem `FLUX_JOB_PROIRTY_MIN`. It isn't clear why that threshold to be an epsilon, though. But since this is just a policy choice, I wouldn't object to the use of the machine eepsilon as your pollicy for now, though.

cmoussa1: I think I'll just drop the floating point comparison for now. That seems like a reasonable choice. Thanks @dongahn for working through this with me and helping me reason it out, it was really helpful.

cmoussa1:
OK, based on the good feedback I got above, I squashed and pushed a change that just drops the floating point comparison in `priority_calculation ()`. 

I've also updated the original PR description and just dropped the tests in `t1001-mf-priority-basic.t` that deal with a user fairshare value of exactly 0 since we've discussed that the fairshare value for a user/bank entry can only get closer and closer to 0, but cannot be exactly 0.

cmoussa1:
Thanks again for your review and input on this PR @dongahn. If you get a chance, could you take a quick glance at this PR and approve it if it looks good to you? If possible, I'd like to include this in the `v0.16.0` release of flux-accounting this week.

grondo (approved):
@cmoussa1, just taking another look at this PR. Now that you've just dropped the comparison altogether, how does this result in a PRIORITY_MIN being returned for the job? Is this because in `priority_calculation()` you have the following?

```c    
    priority = round ((fshare_weight * fshare_factor) +
                      (queue_weight * queue_factor) +
                      (urgency - 16));

    if (priority < 0)
        return FLUX_JOB_PRIORITY_MIN;

```

It might be more clear in the commit message to allude to this fact, e.g.

"Remove the job rejection from the plugin when a user has a fairshare value
lower than 0.0. This then sresult in the priority being set to FLUX_JOB_PRIORITY_MIN
in priority_calculation() which is the desired result."

Or something :shrug: 

Otherwise, this one LGTM.

cmoussa1:
OK, rebased to catch up after #236 and edited the commit message to hopefully be a little clearer. Thanks again for approving, so I'll go ahead and set MWP

Issue #231, opened 2022-04-20
system instance broker crash after `flux account-priority-update`


grondo:
On fluke, there was a user job stuck in PRIORITY state since the user was not currently in the job accounting databse.

I added the user with
```console
# flux account add-user --username <username> --bank=lc
```

Then manually ran `flux account-priority-update` to update the user in the `mf_priority.so` plugin.

Shortly thereafter, the rank 0 broker crashed here:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007ffff62e7f35 in malloc_consolidate () from /usr/lib64/libc.so.6
[Current thread is 1 (Thread 0x7fffbfb8f700 (LWP 1537986))]
Missing separate debuginfos, use: yum debuginfo-install flux-core-0.38.0-1.t4.x86_64
(gdb) where
#0  0x00007ffff62e7f35 in malloc_consolidate () from /usr/lib64/libc.so.6
flux-framework/flux-core#1  0x00007ffff62ea368 in _int_malloc () from /usr/lib64/libc.so.6
flux-framework/flux-core#2  0x00007ffff62ec8d6 in calloc () from /usr/lib64/libc.so.6
flux-framework/flux-core#3  0x00007ffff7ba5183 in safe_malloc (file=0x7ffff7bb112b "zhash.c", line=79, 
    size=2040) at czmq_internal.h:43
flux-framework/flux-core#4  fzhash_new () at zhash.c:79
flux-framework/flux-core#5  0x00007ffff7b6d55c in composite_future_create () at composite_future.c:41
flux-framework/flux-core#6  future_create_composite (wait_any=wait_any@entry=0)
    at composite_future.c:166
flux-framework/flux-core#7  0x00007ffff7b6d8eb in flux_future_wait_all_create ()
    at composite_future.c:182
flux-framework/flux-core#8  0x00007fffbfba4bd7 in lookup_keys (l=0x7fffb40117b0) at lookup.c:309
flux-framework/flux-core#9  lookup_cb (h=0x7fffb40061d0, mh=<optimized out>, msg=0x7fffb401b240, 
    arg=0x7fffb400f590) at lookup.c:309
flux-framework/flux-core#10 0x00007ffff7b64267 in call_handler (mh=0x7fffb400f610, 
    msg=msg@entry=0x7fffb401b240) at msg_handler.c:344
[...]

PR #233, opened 2022-04-21
plugin: fix memory corruption due to use-after-free of "DNE" bank entry

cmoussa1:
#### Problem 

Mentioned in #231, the `DNE` entry for a user/bank combo gets deleted when an active job under that user/bank transitions to `job.state.inactive`, but any other active job under the same user/bank still referencing that deleted `DNE` entry will cause a seg fault and the plugin to break.

---

This PR removes the deletion of the DNE entry from the users map for a user/bank entry in the internal `users` map. 

I also added a sharness test (`t1014-mf-priority-dne.t`) that attempts to imitate the steps outlined in the reproducer shown in #231, but it might be worth to try this branch against that reproducer itself to make sure the same error does not occur.

### DIFF
--- a/t/t1014-mf-priority-dne.t
+++ b/t/t1014-mf-priority-dne.t
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+test_description='Test cancelling active jobs with a late user/bank info load'
+
+. `dirname $0`/sharness.sh
+MULTI_FACTOR_PRIORITY=${FLUX_BUILD_DIR}/src/plugins/.libs/mf_priority.so
+
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 1 job
+
+flux setattr log-stderr-level 1
+
+test_expect_success 'load multi-factor priority plugin' '
+	flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
+'
+
+test_expect_success 'check that mf_priority plugin is loaded' '
+	flux jobtap list | grep mf_priority
+'
+
+test_expect_success 'submit job #1 with no user/bank info loaded to plugin' '
+	jobid1=$(flux mini submit --wait-event=depend hostname)
+'
+
+test_expect_success 'check that job #1 is in state PRIORITY' '
+	test $(flux jobs -no {state} ${jobid1}) = PRIORITY
+'
+
+test_expect_success 'send the user/bank information to the plugin without reprioritizing all active jobs' '
+	cat <<-EOF >fake_payload.py
+	import flux
+	import pwd
+	import getpass
+	import json
+
+	username = getpass.getuser()
+	userid = pwd.getpwnam(username).pw_uid
+	# create an array of JSON payloads
+	bulk_update_data = {
+		"data" : [
+			{
+				"userid": userid,
+				"bank": "account3",
+				"def_bank": "account3",
+				"fairshare": 0.45321,
+				"max_running_jobs": 10,
+				"max_active_jobs": 12,
+				"queues": "standby,special"
+			}
+		]
+	}
+	flux.Flux().rpc("job-manager.mf_priority.rec_update", json.dumps(bulk_update_data)).get()
+	bulk_queue_data = {
+		"data" : [
+			{
+				"queue": "default",
+				"priority": 0,
+				"min_nodes_per_job": 0,
+				"max_nodes_per_job": 5,
+				"max_time_per_job": 64000
+			}
+		]
+	}
+	flux.Flux().rpc("job-manager.mf_priority.rec_q_update", json.dumps(bulk_queue_data)).get()
+	EOF
+	flux python fake_payload.py
+'
+
+test_expect_success 'submit job #2 which should run before job #1' '
+	jobid2=$(flux mini submit hostname) &&
+	flux job wait-event -t 15 $jobid2 clean
+'
+
+test_expect_success 'cancel job #1' '
+	flux job cancel $jobid1
+'
+
+test_done
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -790,11 +790,6 @@ static int inactive_cb (flux_plugin_t *p,
         b->held_jobs.erase (b->held_jobs.begin ());
     }
 
-    // delete user's "DNE" entry in internal map (if it exists)
-    it = users.find (userid);
-    if (it != users.end ())
-        it->second.erase ("DNE");
-
     return 0;
 }
 
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -14,7 +14,8 @@ TESTSCRIPTS = \
 	t1010-update-usage.t \
 	t1011-job-archive-interface.t \
 	t1012-mf-priority-load.t \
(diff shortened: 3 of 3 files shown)


t/t1014-mf-priority-dne.t
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+test_description='Test multi-factor priority plugin with a single user'
```
grondo, line 3: minor: Update `test_description`?

grondo: An interesting test would be to run this sharness test with `FLUX_TEST_VALGRIND=t` both on this PR branch and current master to see if all valgrind reported errors go away here (and if errors are reported for you on master)

t/t1014-mf-priority-dne.t
```diff
@@ -0,0 +1,73 @@
...
+'
+
+test_expect_success 'add a user to the DB' '
+	username=$(whoami) &&
+	uid=$(id -u) &&
+	flux account -p ${DB_PATH} add-user --username=$username --userid=$uid --bank=account1 --shares=1
+'
+
+test_expect_success 'send the user/bank information to the plugin and reprioritize pending jobs' '
+	flux account-priority-update -p $(pwd)/FluxAccountingTest.db
```
grondo, line 61: To simulate the original reproducer, don't we need to send an update _without_ a reprioritization? Or does the issue still reproduce with the reprioritize (I guess you can check that). Sorry I can't remember where we landed in our discussion about that.

t/t1014-mf-priority-dne.t
```diff
@@ -0,0 +1,73 @@
...
+'
+
+test_expect_success 'check that job #2 transitions to RUN and job #1 is still pending' '
+	test $(flux jobs -no {state} ${jobid2}) = RUN &&
+	test $(flux jobs -no {state} ${jobid1}) = SCHED
```
grondo, line 66: Testing for the job state _might_ lead to a test race condition. Instead I suggest using `flux job wait-event` and waiting for the `start` event. _If_ the first job is supposed to stay in `SCHED` state, then this test is fine as is, but I'm confused if that should actually be true. (I likely just don't understand the test.)

grondo (commented):
I'd suggest splitting this PR into 2 separate commits:

 * plugin: fix potential use-after-free of DNE bank entry
 * testsuite: add test to exercise DNE error path (or something)

t/t1014-mf-priority-dne.t
cmoussa1: Oh, right. Sorry for leaving that test description old there, I copied one of the other sharness test files and left that over. I just pushed up a fix that should have updated the test description there. 

> An interesting test would be to run this sharness test with `FLUX_TEST_VALGRIND=t` both on this PR branch and current master to see if all valgrind reported errors go away here (and if errors are reported for you on master)

I ran this sharness test with `FLUX_TEST_VALGRIND=t` both on this PR branch and on current master and confirmed that no errors were reported in my local Docker container. That makes me think - is it possible to add a builder to CI that runs all of the sharness tests with `FLUX_TEST_VALGRIND=t`? If so, I should submit another PR in the near future (probably before the next flux-accounting release) that adds that type of builder.

t/t1014-mf-priority-dne.t
cmoussa1: Thanks for pointing this out. I just pushed a change to this test to instead `flux job wait-event $jobid2 start` instead of testing for the job state.

t/t1014-mf-priority-dne.t
grondo: Yeah, this would be simple by setting `FLUX_TEST_VALGRIND="t"` in the `env` dict of a new `matrix.add_build()` to add a 'valgrind' builder.

We don't do this in flux-core since it would take way too long, but perhaps here it might be a good idea.

> I ran this sharness test with FLUX_TEST_VALGRIND=t both on this PR branch and on current master and confirmed that no errors were reported in my local Docker container.

I would have expected master to reproduce some of the valgrind errors, so this makes me think the test isn't reproducing the right conditions to hit the failure?

cmoussa1: Oh yeah, you're right... that's weird. I will jump back to `master` and investigate

cmoussa1: OK, after further investigation, it looks like `master` _was_ correctly reproducing the use-after-free errors reported in #231, but the sharness test was cancelling the first job before the second job completed all of `job.state.inactive`. 

I just force pushed a change to the sharness test that waits for the `clean` event instead of `start` before cancelling the first job. Now, valgrind reports no errors when running under this PR branch, but reports errors in `master`, which I think is expected. 👍

t/t1014-mf-priority-dne.t
```diff
@@ -0,0 +1,78 @@
...
+		]
+	}
+	flux.Flux().rpc("job-manager.mf_priority.rec_q_update", json.dumps(bulk_queue_data)).get()
+	EOF
+	flux python fake_payload.py
+'
+
+test_expect_success 'submit job #2 which should run before job #1' '
+	jobid2=$(flux mini submit hostname) &&
+	flux job wait-event $jobid2 clean
```
grondo, line 71: In case something goes wrong and `jobid2` doesn't get a `clean` event (e.g. a bug is introduced and the job is blocked from running somehow), it is useful to provide the `-t, --timeout` option to `flux job wait-event` in tests:
```suggestion
	jobid2=$(flux mini submit hostname) &&
	flux job wait-event -t 15 $jobid2 clean
```

This will cause the test to fail instead of hanging indefinitely.

t/t1014-mf-priority-dne.t
```diff
@@ -0,0 +1,78 @@
...
+	flux jobtap list | grep mf_priority
+'
+
+test_expect_success 'submit job #1 with no user/bank info loaded to plugin' '
+	jobid1=$(flux mini submit hostname)
+'
+
+test_expect_success 'check that job #1 is in state PRIORITY' '
+	test $(flux jobs -no {state} ${jobid1}) = PRIORITY
+'
```
grondo, line 27: There _could_ be an unlikely race here between these two tests. If the first job is submitted, but the `depend` event isn't emitted before the next test runs, then this second test will fail, because the job isn't quite in PRIORITY state yet.

The fix would be to wait for the `depend` event in one of the two tests

```suggestion
test_expect_success 'submit job #1 with no user/bank info loaded to plugin' '
	jobid1=$(flux mini submit --wait-event=depend hostname)
'

test_expect_success 'check that job #1 is in state PRIORITY' '
	test $(flux jobs -no {state} ${jobid1}) = PRIORITY
'
```

This would be a very rare race for this test since `depend` is emitted immediately by the job manager if there are no `dependency-add` events, however, it is perhaps instructive on writing tests that are race condition free (plus if you run a test under valgrind, race conditions tend to be much more likely)

grondo (approved):
This LGTM now!

I had a couple instructive comments in the introduced test, but they are non-critical.

cmoussa1:
Thanks for reviewing and approving @grondo! And thanks for pointing out the instructive benefits for the sharness test. That `--timeout` option looks really really useful. I've just force pushed the change based on your suggestions. I can set MWP shortly

PR #235, opened 2022-05-10
association_table: add `max_nodes` column

cmoussa1:
#### Background

Described in #121, there is a need for supporting a `max_nodes` limit on user/bank associations. More specifically, this limit will enforce a maximum number of nodes a user/bank association can have across all of their running jobs at any given time.

---

This is a small PR that simply adds the `max_nodes` column to the `association_table` in the flux-accounting database. It also adds `max_nodes` as an optional argument to the `add-user` and `edit-user` commands.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -87,18 +87,19 @@ def create_db(
     conn.execute(
         """
             CREATE TABLE IF NOT EXISTS association_table (
-                creation_time    bigint(20)                NOT NULL,
-                mod_time         bigint(20)  DEFAULT 0     NOT NULL,
-                username         tinytext                  NOT NULL,
-                userid           int(11)     DEFAULT 65534 NOT NULL,
-                bank             tinytext                  NOT NULL,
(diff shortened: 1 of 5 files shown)


chu11 (approved):
LGTM!

cmoussa1:
Thanks @chu11! Before I merge this PR, I should probably get some input from @ryanday36 to see if he has any suggestions on a better default value for the `max_nodes` limit - I chose 5 somewhat randomly, so I'm not sure if there's a better value to set based on what defaults are used currently on our systems. Do you have any feedback or suggestions @ryanday36? Would you prefer a different value for this limit?

ryanday36:
Would `unlimited` (or just the max possible value) be an option for the default? The `max_nodes` limit tends to be based on the size of the cluster, so it's hard to come up with a general default value.

cmoussa1:
Yeah, I can just make the default a super large int (2147483647 looks like the max int for c/c++ according to the [Microsoft docs](https://docs.microsoft.com/en-us/cpp/c-language/cpp-integer-limits?view=msvc-170)) for the default. Thanks for clarifying that the limit is different depending on the cluster, I didn't know that!

PR #236, opened 2022-05-10
t: add sharness test for multiple user jobs with different urgencies

### LINKED ISSUE #134: priority plugin: add tests for multiple users with different urgency values

cmoussa1:
#### Background

This was opened in an issue a little while ago (#134), but it was noted that it might be useful to have a sharness test that simulated multiple users submitting jobs with different `urgency` values.

---

This is a [WIP] PR that adds a sharness test which has multiple fake users submit jobs with different `urgency` values. For the purpose of checking ordering of the job queue, I set the same fairshare value for all users and had them submit jobs to the same `default` queue, which would leave `urgency` as the one factor to affect a user's job priority.

Since I am using the `submit_as.py` script that @grondo added a while ago for this test, I think I noticed that the `urgency` option would not be set in the actual job submit command even if it was passed on the command line, so I made a small change in the script to check explicitly for an optional `urgency` argument, and if passed, add it to `submitcmd`. @grondo: I don't know if this is the best approach or change to make to your submit script, so if you have a better idea or solution for your script, I am completely open to any feedback you might have. Let me know if there is a better workaround for my proposed solution. :-)

Fixes #134

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -15,7 +15,8 @@ TESTSCRIPTS = \
 	t1011-job-archive-interface.t \
 	t1012-mf-priority-load.t \
 	t1013-mf-priority-queues.t \
-	t1014-mf-priority-dne.t
+	t1014-mf-priority-dne.t \
+	t1015-mf-priority-urgency.t
 
 dist_check_SCRIPTS = \
 	$(TESTSCRIPTS) \
--- a/t/t1015-mf-priority-urgency.t
+++ b/t/t1015-mf-priority-urgency.t
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+test_description='Test multi-factor priority plugin order with no ties'
+
+. `dirname $0`/sharness.sh
+MULTI_FACTOR_PRIORITY=${FLUX_BUILD_DIR}/src/plugins/.libs/mf_priority.so
+SUBMIT_AS=${SHARNESS_TEST_SRCDIR}/scripts/submit_as.py
+
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 1 job
+
+flux setattr log-stderr-level 1
+
+test_expect_success 'load multi-factor priority plugin' '
+	flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
+'
+
+test_expect_success 'check that mf_priority plugin is loaded' '
+	flux jobtap list | grep mf_priority
+'
+
+test_expect_success 'create a group of users with the same fairshare value' '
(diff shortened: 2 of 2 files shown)


t/scripts/submit_as.py
```diff
@@ -34,6 +34,8 @@ def main():
     userid = int(sys.argv[1])
     os.environ["FLUX_HANDLE_USERID"] = str(userid)
 
+    urgency = sys.argv[2] if "urgency" in sys.argv[2] else None
```
grondo, line 37: Since `submit-as.py` takes any `flux-mini run` option, it won't be guaranteed that `--urgency` will be the second option. Therefore, I think it might be better (i.e. less surprising) if `sys.argv[2:]` was iterated and any option matching `--urgency` was passed down to `flux job submit`.

Also, the code could be simplified if you just append to the `submitcmd` list if needed.

e.g. something like (untested):


```diff
diff --git a/t/scripts/submit_as.py b/t/scripts/submit_as.py
index 4b22c92..6bd45b9 100644
--- a/t/scripts/submit_as.py
+++ b/t/scripts/submit_as.py
@@ -34,7 +34,10 @@ def main():
     userid = int(sys.argv[1])
     os.environ["FLUX_HANDLE_USERID"] = str(userid)
 
-    urgency = sys.argv[2] if "urgency" in sys.argv[2] else None
+    submitcmd = [ "flux", "job", "submit", "--flags=signed" ]
+    for arg in sys.argv[2:]:
+        if "urgency" in arg:
+            submitcmd.append(arg)
 
     minicmd = [
         "flux",
@@ -48,10 +51,6 @@ def main():
 
     signedJ = SecurityContext().sign_wrap_as(userid, jobspec, mech_type="none")
 
-    if urgency is not None:
-        submitcmd = ["flux", "job", "submit", urgency, "--flags=signed"]
-    else:
-        submitcmd = ["flux", "job", "submit", "--flags=signed"]
     jobid = run_process(submitcmd, input=signedJ)
     print(jobid)
```
Good catch by the way!

grondo (commented):
This LGTM! I made some notes on the `submit_as.py` script inline.

t/scripts/submit_as.py
cmoussa1: Thanks @grondo for reviewing and for the suggested improvement! I didn't know you could just append the `urgency` argument to the end of `submitcmd`, so thank you for pointing that out. This solution is much cleaner and did result in passing tests, so I just force pushed a change that implemented this suggestion and rebased to catch up after some of the PR's that have landed before this one. I've also taken it out of [WIP].

grondo (approved):
LGTM!

cmoussa1:
Thanks! Setting MWP

PR #238, opened 2022-05-19
plugin: disable queue validation, `default` queue requirement

### LINKED ISSUE #237: pugin: disable requirement for `default` queue
Currently, the priority plugin rejects jobs when there is no `default` queue information that exists in the plugin (this information is sent from the DB to the plugin with `flux account-priority-update`). This should be disabled until we have a more refined design for queue and multi-queue support within Flux, so we should not require `default` queue information to exist in the plugin for now.

cmoussa1:
#### Problem 

The multi-queue design is not fully designed and implemented throughout the various Flux components yet, but the plugin requires that a `default` queue exist in order to submit jobs with the plugin loaded.

---

Disable the validation of queues as well as its use of an associated integer priority with queues passed in from the flux-accounting DB so that a `default` queue does not need to be present in order for users to submit jobs with the multi-factor priority plugin loaded. I just commented out the sections of code that perform this validation, but if we think it would be cleaner to just remove the code altogether, I am OK with that as well.

Disable the tests that check for queue validation error messages and that check for specific job priorities affected by queue priorities. I just commented out the tests that perform these checks and job priority comparisons, but if we think it would be cleaner to just remove the code altogether, I am also OK with that!

Fixes #237

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -812,6 +812,18 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
         || flux_jobtap_service_register (p, "reprioritize", reprior_cb, p)
         || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p) < 0)
         return -1;
+
+    struct queue_info *q;
+    q = &queues["default"];
+
+    // min_nodes_per_job, max_nodes_per_job, and max_time_per_job are not
+    // currently used or enforced in this plugin, so their values have no
+    // effect in queue limit enforcement.
+    q->min_nodes_per_job = 0;
+    q->max_nodes_per_job = 1024;
+    q->max_time_per_job = 64000;
+    q->priority = 1000;
+
     return 0;
 }
 
--- a/t/t1013-mf-priority-queues.t
+++ b/t/t1013-mf-priority-queues.t
@@ -58,11 +58,11 @@ test_expect_success 'stop the queue' '
 	flux queue stop
 '
 
-test_expect_success 'trying to submit a job without specifying a queue should attempt to use default queue' '
-	test_must_fail flux python ${SUBMIT_AS} 5011 -n1 hostname > no_default_queue.out 2>&1 &&
-	test_debug "no_default_queue.out" &&
-	grep "No default queue exists" no_default_queue.out
-'
+# test_expect_success 'trying to submit a job without specifying a queue should attempt to use default queue' '
+# 	test_must_fail flux python ${SUBMIT_AS} 5011 -n1 hostname > no_default_queue.out 2>&1 &&
+# 	test_debug "no_default_queue.out" &&
+# 	grep "No default queue exists" no_default_queue.out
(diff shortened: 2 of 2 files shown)


cmoussa1:
Ah yikes, looks like coverage dropped a significant amount because of the commented out code no longer triggering the helper functions. Maybe it's worth it for me to save a copy of this plugin with the queue support and just remove the portions of currently-disabled code completely for now!

cmoussa1:
Oh, that's a good point, thanks @grondo. Yeah, in that case I'll just add an internal `queue` default in the map and that way I won't need to comment a bunch of stuff out.

cmoussa1:
Ah, that coverage is better! I also tested this in my Docker container by starting flux and using `flux account-priority-update` without adding a queue via `flux account add-queue default --priority=1000`:

```console
$ flux start
$ flux account create-db
$ flux account add-bank A 1
$ flux account add-user --username=moussa1 --bank=A
$ flux jobtap load mf_priority.so
$ flux account-priority-update
$ flux mini submit -n 1 hostname
```

and confirmed that the job completed with `flux jobs -a`:


I also only needed to disable one sharness test (the one that checks for a `No default queue exists` error message). I think I'll take this out of [WIP] and mark it ready for review whenever convenient.

src/plugins/mf_priority.cpp
```diff
@@ -812,6 +812,15 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
...
         || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p) < 0)
         return -1;
+
+    struct queue_info *q;
+    q = &queues["default"];
+
+    q->min_nodes_per_job = 0;
+    q->max_nodes_per_job = 1024;
+    q->max_time_per_job = 64000;
+    q->priority = 1000;
```
grondo, line 822: Question: since queues are not validated, these numbers don't matter right?

If they do, I'd suggest the equivalent of unlimited for `max_time_per_job` and `max_nodes_per_job`.

grondo (commented):
I had one question about the internal values chosen for the builtin default queue.

Other than that this change is very clean, nice work! :+1:

src/plugins/mf_priority.cpp
cmoussa1: Good question! `min_nodes_per_job`, `max_nodes_per_job`, and `max_time_per_job` are not used or enforced anywhere throughout the plugin, so those values should not matter. `priority` is the only factor used in the priority calculation for jobs.

grondo: I'd add a comment here to that effect then, to avoid confusion for someone browsing the code.

cmoussa1: Sounds good. I just force-pushed a fix to the first commit that adds a comment to make note that those three values aren't currently used or enforced in the plugin.

grondo (approved):
Great! LGTM!

cmoussa1:
Thanks! I'll set MWP

Issue #232, opened 2022-04-20
plugin: need a way to "disable" user/bank entries in internal map


cmoussa1:
Mentioned in #231, we need a method for "disabling" a user/bank entry in the plugin's internal map when a user gets removed from a bank in the flux-accounting DB. Since entries don't get deleted in the internal map, a user can technically still run jobs under a bank they no longer belong to.

PR #239, opened 2022-05-25
flux-accounting DB: add `active` column

cmoussa1:
#### Problem

Mentioned in #232, we need a method for "disabling" a user/bank entry in the multi-factor priority plugin's internal map when a user gets removed from a bank in the flux-accounting DB. Since entries don't get deleted in the internal map, a user can technically still run jobs under a bank they no longer belong to.

I think a potential solution to this would be to signal to the plugin when a user/bank association gets deleted in the flux-accounting DB. A `1` could represent a "deleted" user/bank row or a "deleted" bank row. Since this involves a little rework to the DB, its commands, **and** the priority plugin, I think I should break up this work into at least two separate PR's.

---

This [WIP] PR adds a new column to the `association_table` and `bank_table` in the flux-accounting DB called `deleted`, an integer column that represents when a row has been "deleted." When a user/bank row or a bank row gets "deleted," this column is updated to `1`. 

In the weighted tree library, when the data from the flux-accounting DB is loaded, if a row's `deleted` column is 1, that row is skipped and not added to the tree. 

A couple of updates to the tests that use the `delete-user` and `delete-bank` commands are also updated to accommodate this new functionality. 

A couple of new tests are also added to `t1000-print-hierarchy.t` to make sure that when a user/bank row or a bank row is "deleted," they do not show up in the output of the `flux account-shares` command (the command that prints the DB hierarchy).


cmoussa1:
I think this is getting a little closer, so I'll take this out of [WIP] and mark it as ready for some initial review whenever convenient.

cmoussa1:
Rebased to catch up after #241

chu11:
an initial high level comment, would a field called `active` (or similar) be more desirable?  can a user be re-added to a bank?

chu11:
> I could see using a field called active being easier to look at and keep track of, so thanks for this suggestion. I'm assuming we could use similar integer values to represent a row if it is "active" or not, 1 for active and 0 for inactive. If this makes more sense and is what you were thinking, I can change the direction of this PR to instead add an active column instead of a deleted column.

well to me "active" or perhaps "enabled" seems more binary, like user could be enabled or not, where as "deleted" sounds more permanent.  I haven't reviewed the PR yet so there could be a subtlety I'm missing.

Perhaps @grondo or @ryanday36 can give a vote as well.  It's obviously just a semantics thing and I wouldn't want to cause you too much work if it's just me.

cmoussa1:
Thanks for the input @chu11 and @ryanday36, all good points. Why don't I take some time to rework this PR to instead use `active` to distinguish a user/bank row in the flux-accounting database. I can convert this back to [WIP] and open it up for review once I've made the appropriate changes.

cmoussa1:
Sorry @chu11 that it took me a couple days to get back to this and to your suggestions - got sidetracked with other bug fixes in the meantime. :-) I've adjusted this PR to instead add an `active` column instead of a `deleted` column.

One thing I noticed while working on the change was that the branch name is no longer a good fit (`add.deleted.col`). I can change the name of the branch, but in doing so I _think_ it will close this PR and open a new one. If that sounds like an OK plan to you, I'll just close this PR and re-open with a better branch name.

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -158,6 +158,24 @@ def delete_user(conn, username, bank):
...
+    if default_bank == bank:
+        # get first bank from other potential existing rows from user
+        select_stmt = """SELECT bank FROM association_table WHERE active=1 AND username=?
+                         ORDER BY creation_time"""
+        cur.execute(select_stmt, (username,))
+        rows = cur.fetchall()
+        # update user rows to have a new default bank (the next earliest user/bank row created)
+        if len(rows) > 0:
+            new_default_bank = rows[0][0]
+            edit_user(conn, username, default_bank=new_default_bank)
```
chu11, line 198: what if the user is not a member of any other bank?  should default_bank be set to `None` or something like that?

and would test in follow on test be needed?

src/fairness/reader/data_reader_db.cpp
```diff
@@ -245,33 +245,36 @@ std::shared_ptr<weighted_tree_node_t> data_reader_db_t::get_sub_banks (
             std::string usage = reinterpret_cast<char const *> (
                 sqlite3_column_text (c_assoc, 3));
             double fshare = sqlite3_column_double (c_assoc, 4);
+            int active = sqlite3_column_int (c_assoc, 5);
 
-            try {
-                // add association as a child of the node
-                if (add_assoc (username, shrs, usage, fshare, node) < 0) {
-                    m_err_msg += "Failed to add association\n";
+            if (active == 1) {
```
chu11, line 250: nit: doesn't have to be == 1, just `active`?

src/bindings/python/fluxacct/accounting/test/test_bank_subcommands.py
```diff
@@ -72,12 +72,13 @@ def test_04_add_subaccounts(self):
         rows = cur.fetchall()
         self.assertEqual(len(rows), 1)
 
-    # removing a bank currently in the bank_table
+    # remove a bank currently in the bank_table
```
chu11, line 75: remove -> disable?

cmoussa1:
Thanks for the great feedback @chu11! I went ahead and force-pushed some changes to this branch based on your suggestions:

- added a (hopefully) clear comment in `delete_user()` that talks about the behavior when a user who only belongs to one bank is disabled
- changed the check for `active == 1` to just `active` in `data_reader_db.cpp`
- changed comments that say `delete` to `disable`

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: I'm sorry, I think I am just confused by your question and am probably not answering what you are actually asking. If a user only belongs to one bank (thus, their default bank is the same as the bank they belong to), and the user gets disabled, then their default bank does not need to be updated at all. It will keep the same default bank it had before the user got disabled because the user does not belong to any other banks. Does that make sense? :P

chu11: ohh ok, that makes sense :-)  I'll let ya decide if a comment is warranted, maybe I'm just a dummy :-)

chu11 (approved):
just the minor nits I mentioned above, I think its good to go

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: Haha no you're not - I probably just didn't do a good job explaining it. I'll add a detailed comment just above explaining this case so that if/when other people look at it they won't raise similar questions.

cmoussa1:
Just pushed up a change to address @chu11's good suggestions above and added some additional comments to hopefully help clarify the expected behavior when a user gets disabled while only belonging to one bank. I'll set MWP shortly

PR #241, opened 2022-06-06
t: organize expected files into subdirectories

cmoussa1:
#### Background

While the `t/` directory's expected files are mostly organized into subdirectories for their respective sharness tests, not all of them are. Those which do not belong to a subdirectory yet should be further organized into an appropriate subdirectory.

---

This is a tiny PR that just moves some of the expected files in the `t/` directory into a subdirectory to maintain consistency with the other expected files.


grondo (approved):
This LGTM!

If anything, I'd suggest squashing all three of these commits together since they all accomplish one thing: moving expected output files into subdirectories.

You could then describe the _Problem_ being solved and the solution for more context when people look through commit history.

However, I'd call that change optional, so only do it if you think it is a good idea as well!

cmoussa1:
Thanks @grondo! I also think that's a good idea. I went ahead and squashed those commits into just one and added a _Problem_ statement to the commit message. I'll go ahead and add MWP shortly. Thanks again!

PR #242, opened 2022-06-06
[WIP] plugin: disable user/bank entry in priority plugin when row gets deleted in DB

cmoussa1:
#### Problem

As mentioned in #232, we need a method for "disabling" a user/bank entry in the plugin's internal map when a user gets removed from a bank in the flux-accounting DB. Since entries don't get deleted in the internal map, a user can technically still run jobs under a bank they no longer belong to.

----

This is a small [WIP] PR built on top of #239 that adds the `deleted` column to the group of user/bank information sent to the priority plugin, where it is then checked in `validate_cb ()` when a user/bank association submits a job. If the `deleted` column shows `1`, then that means the user/bank association is "deleted" from the flux-accounting DB, and thus disabled from running jobs.

A small sharness test is also added to exercise this check, where a user is added to two different banks in the flux-accounting DB and can successfully submit jobs under each bank. When the user gets removed from one bank and tries to submit a job under that bank, the job is rejected with an appropriate message saying that the association has been removed from the DB.


grondo:
I'm just starting to look at this (and wondering if @ryanday36 should chime in).
A couple questions came up at first:

 1. So is the plan that deleted users will stay in the database forever? What happens if a user is deleted and then the instance is restarted or the priority plugin is reloaded? (I didn't look, I was just curious)
 2. Would it be easier to have a `deleted = 1` entry just delete the user/bank from the internal plugin map, then a deleted user will be treated the same as a user not in that bank without any extra code? (Sorry if this is a naive question) I'm wondering why a deleted user has to be treated differently than a user not in a bank

grondo:
>  I chose this approach because I was thinking about a case where a user submits a job, but then their user/bank row is deleted from the database while their job is still active.

Ah, very good catch, I hadn't considered that. We should ask @ryanday36 what the expected behavior is in this case. I wonder if any active job referencing a deleted entry should be terminated?

In any case, a different approach might be to use a reference count on user/bank entries. Then an entry that is deleted, but which active jobs are referencing, will be automatically removed when the last job referencing them exits.

I'm actually unsure of the correct approach here, but I am a little wary of a design in which deleted users have to be kept in the database forever. However, I don't know enough about accounting to say if my apprehension is warranted. For example, maybe we want to keep all users in the database anyway for historical reasons?

cmoussa1:
> I'm actually unsure of the correct approach here, but I am a little wary of a design in which deleted users have to be kept in the database forever. However, I don't know enough about accounting to say if my apprehension is warranted. For example, maybe we want to keep all users in the database anyway for historical reasons?

That could very well be the case, good point. I'm not 100% sure either, so some further discussion is probably needed. :-) I'll leave this as [WIP] until we have a clearer idea of what is expected from both the plugin (and probably the database too) in terms of removing users and banks.

PR #243, opened 2022-06-14
flux-account: add DB export command

### LINKED ISSUE #215: support migrating data from one flux-accounting DB to another
In an offline conversation with @ryanday36, he mentioned that eventually there will be a need for supporting the ability to update the flux-accounting database when a new version of the database is released so that the existing DB does not have to be completely re-created and have all values re-inserted. This will be especially important when new columns are added to an already existing table or a new table is added.

In SQLite, this can be done with a query:

```sql
ATTACH DATABASE file_name AS new_db;
CREATE TABLE new_db.table_name(table_definition);
INSERT INTO new_db.table_name SELECT * FROM old_db.table_name;
INSERT INTO new_db.table_name(col1, col2) SELECT col1, col2 FROM old_db.table_name; <-- if columns don't match up
```

But it would probably be a good idea to support this through a `flux account` command that takes care of all of these SQLite queries.

cmoussa1:
#### Background

As mentioned in #215, there will be a need for supporting the ability to update the flux-accounting database when a new version of the database is released so that the existing DB does not have to be completely re-created and have all values re-inserted. This will be especially important when new columns are added to an already existing table or a new table is added.

---

This PR adds a new command to flux-accounting: `export-db`, which allows an admin to export information from the `association_table` and `bank_table` in the flux-accounting DB into two `.csv` files. This can be used to then import user and bank information into a new flux-accounting DB. Custom paths and filenames can be specified for these two files, or, if left default, will just create the two `.csv` files in the current working directory named `banks.csv` and `users.csv`.

The exported DB information will only contain the absolutely necessary data for each row: from the `association_table`:

**username, userid, banks, shares, max_running_jobs, max_active_jobs, max_nodes, queues**. 

From the `bank_table`: 

**bank, parent bank, shares**. 

This will hopefully ensure that when a database needs to be updated, we avoid "missing column" errors and such.

A sharness test is also added to exercise this command in conjunction with the `pop-db` command to migrate database information from one flux-accounting DB to another.

I also added a tiny commit at the beginning of this PR to fix the default values for `max_nodes` in the `add-user` command and `association_table`.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -128,4 +128,5 @@ dist_fluxcmd_SCRIPTS = \
 	cmd/flux-account-update-fshare \
 	cmd/flux-account-priority-update.py \
 	cmd/flux-account-shares \
-	cmd/flux-account-pop-db.py
+	cmd/flux-account-pop-db.py \
+	cmd/flux-account-export-db.py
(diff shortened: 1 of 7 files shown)


cmoussa1:
OK, I think I'll mark this as ready for review whenever convenient. :-)

PR #244, opened 2022-06-21
flux-account: add new `update-db` command

### LINKED ISSUE #215: support migrating data from one flux-accounting DB to another

cmoussa1:
#### Background

Noted in #215, there is a need for supporting the ability to update the flux-accounting database when a new version of the database is released so that the existing DB does not have to be completely re-created and have all values re-inserted. This will be especially important when new columns are added to an already existing table or a new table is added. There currently exists both `export-db` and `pop-db` commands to export and import database information into `.csv` files, but these commands also involve creation of a new DB to migrate this information. It would be nice to have support to just update the database that currently resides on a system with any new potential tables or columns after a new release of flux-accounting.

---

This is a [WIP] PR that adds a new command to flux-accounting: `update-db`, which will look at the DB schema of the database currently residing on a system and update it to add any missing tables or add any missing columns within each of the tables. It does so by creating a temporary database to compare schema; it will then compare tables and then iterate through each table and compare columns. After the iteration is done, the temporary database is removed. For testing purposes, there is an optional argument for the `update-db` command to specify a path for the temporary database; if not specified (which should be the norm and is noted in the description of the command), the temporary database is created and removed from the current working directory.

Two helper scripts are added to the `t/` directory to help with testing:

**modify_accounting_db.py** is a helper script that adds two columns to the association_table and a new table to the flux-accounting DB for testing the functionality of the flux account-update-db command.

**check_db_info.py** is a helper script that checks the names of the tables in a flux-accounting DB or the names of the columns within a table in a flux-accounting DB.

The sharness test file `t1017-export-db.t` uses these two helper scripts to modify a flux-accounting DB to simulate a newer DB being released with an additional table and two additional columns added to one of the existing tables. The test runs the `update-db` command and makes sure the missing table and columns are added to the flux-accounting DB successfully.

Fixes #215

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -129,4 +129,5 @@ dist_fluxcmd_SCRIPTS = \
 	cmd/flux-account-priority-update.py \
 	cmd/flux-account-shares \
 	cmd/flux-account-pop-db.py \
-	cmd/flux-account-export-db.py
+	cmd/flux-account-export-db.py \
+	cmd/flux-account-update-db.py
--- a/src/cmd/flux-account-update-db.py
+++ b/src/cmd/flux-account-update-db.py
@@ -1,0 +1,193 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2022 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import argparse
+import os
+import sqlite3
+import sys
+
+from argparse import RawDescriptionHelpFormatter
+
+import fluxacct.accounting
+from fluxacct.accounting import create_db as c
+
+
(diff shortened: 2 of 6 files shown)


ryanday36 (approved):
This looks good! Thanks @cmoussa1.

cmoussa1:
Thanks @ryanday36! Looking forward to seeing this tool in action. Setting MWP shortly

PR #251, opened 2022-07-13
plugin: add check before deleting held job ID

### LINKED ISSUE #249: plugin: job.state.inactive callback attempts to remove item from list even if list is empty
In the priority plugin, the `job.state.inactive` callback looks to see if it should release any held jobs to transition to RUN state by looking at a list of held job ID's. If it can release the job, it removes the dependency from the job and removes the job ID from the `held_jobs` list. It does not first check the list to make sure that there are ID's present in the list before calling `erase ()`, however. This can cause unexpected behavior if a job ID is already removed by the time the callback calls `erase ()`. 

A check should be added just before this call to ensure that there is in fact a job ID to remove from the list before calling `erase ()`.

cmoussa1:
#### Problem

The `job.state.inactive` callback attempts to remove a job ID from the list of held jobs regardless if there is actually a job ID to be removed, which can cause unexpected behavior if there are no job IDs in the list.

---

This PR adds a check just before the `erase ()` call to make sure that the ID exists in the list before attempting to remove it. If the list is empty, don't attempt to remove a job ID.

Fixes #249

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -777,7 +777,9 @@ static int inactive_cb (flux_plugin_t *p,
             flux_jobtap_raise_exception (p, jobid, "mf_priority",
                                          0, "failed to remove job dependency");
 
-        b->held_jobs.erase (b->held_jobs.begin ());
+        // check that there is in fact a job ID to be removed from the list
+        if (b->held_jobs.begin () != b->held_jobs.end ())
+            b->held_jobs.erase (b->held_jobs.begin ());
     }
 
     return 0;


cmoussa1:
I'll mark this as ready for review whenever convenient. Since this fix is proposed alongside the work done in flux-framework/flux-core#4406, I wasn't exactly sure if/how to add tests for trying to remove a dependency on a job ID that no longer has a dependency on it, but if there are any suggestions I'd be more than willing to add some. :-)

src/plugins/mf_priority.cpp
```diff
@@ -777,7 +777,9 @@ static int inactive_cb (flux_plugin_t *p,
             flux_jobtap_raise_exception (p, jobid, "mf_priority",
                                          0, "failed to remove job dependency");
 
-        b->held_jobs.erase (b->held_jobs.begin ());
+        // check that there is in fact a job ID to be removed from the list
+        if (b->held_jobs.begin () != b->held_jobs.end ())
```
grondo, line 781: I'm not an expert on C++, but could `if (!b->held_jobs.empty ())` work here and be more readable? I assume `.begin () != .end ()` is just a way to test for a non empty vector? Apologies if I'm mistaken.

I actually wonder if this was the problem though. I don't see how this check could ever be false since a nonzero size for `b->held_jobs` was checked before entering this block. Do you have any ideas?

cmoussa1: That's a good point. Before flux-framework/flux-core#4406 landed, valgrind was reporting an `invalid read of size 8` at the line where the ID was erased from the list. So that's what made me think that for some reason the list was empty by the time it got to the `erase ()` call, and an additional check was needed.

I don't know for sure if there was some timing that was happening in the scenario where two jobs under the same user referenced the list of held jobs and were canceled one right after the other, so one job transitioned to `job.state.inactive`, and removed the dependency while the other job was also in `job.state.inactive` but before the `erase ()` call. I don't know if that makes any sense, but that's what I had initially thought was the issue.

cmoussa1:
I think this can be closed now because of #325 landing.

PR #252, opened 2022-07-16
`update-db`: fix command, improve error message/execption clarity

### LINKED ISSUE #248: flux-account-update-db suppresses error from sqlite3.connect
If there is a problem connecting to the db in `flux-account-update-db.py`, the script prints a generic error:
```
Unable to open temporary database file: <file.db>
```

It would be useful to capture the exception thrown by sqlite and print the exception data, so that the real reason for the failure is printed to the user, e.g.

```diff
--- /usr/libexec/flux/cmd/flux-account-update-db.py	2022-06-23 09:40:44.000000000 -0700
+++ flux-account-update-db.py	2022-07-13 08:38:31.550533501 -0700
@@ -158,8 +158,8 @@
         new_conn.close()
 
         os.remove(new_db)
-    except sqlite3.OperationalError:
-        print(f"Unable to open temporary database file: %s" % new_db)
+    except sqlite3.OperationalError as exc:
+        print(f"Unable to open temporary database file: {new_db}: {exc}")
         sys.exit(1)
 

```

### LINKED ISSUE #250: flux-account-update-db fails in v0.17.0
On fluke, when attempting to update the accounting db from v0.16.0 to v0.17.0, the following error was seen

```console
# flux account-update-db -p /var/lib/flux/FluxAccounting.db
checking for new tables...
checking for new columns to add in tables...
new column found in association_table: username
Unable to open temporary database file: tmpFluxAccounting.db: duplicate column name: username
```

The error was from a patched version of `flux-account-update-db.py` as described in #248

cmoussa1:
#### Problem

As described in #248 and in #250, there are a couple of issues regarding the `update-db` command:

- If there is a problem connecting to the DB in `flux-account-update-db.py`, the script prints a generic error that often times is not representable of the issue at hand
- On fluke (or probably any other machine that contains an installed and configured flux-accounting DB), when attempting to update the accounting DB, the command fails when upgrading to a new version where the new table does not contain columns that existed in previous versions, which resulted in a `duplicate column` error message

---

This [WIP] PR looks to improve the message and exception clarity if/when an error occurs in the `flux account-update-db` command. In addition to printing a `Could not connect to DB` message along with the DB path, it will also print a more specific error about where the command encountered an error.

The other major component to this PR adds support for updating versions of the DB when columns are deleted from tables from version to version. `ALTER TABLE DROP COLUMN column_name` support does not exist until SQLite version 3.35 and later, so I've sort of implemented my own version. In the `flux account-update-db` command, the script now has a number of helper functions that create a new table with an updated column list and transfers existing table data to the new table. A new test is also added to ensure that deleting columns from a table and then updating the DB works as expected.

In addition to adding the new tests, a couple of very minor improvements are made to the helper scripts used in `t1017-update-db.t`. In `check_db_info.py`, the printing of table names and column info are changed to a better format for the sharness tests. In `modify_accounting_db.py`, a number of SQLite statements are added that simulate dropping the `queue_table` and re-adding it with one less column than before, indicating a deleted column from the table in a newer flux-accounting version.

Fixes #248
Fixes #250

### DIFF
--- a/src/cmd/flux-account-update-db.py
+++ b/src/cmd/flux-account-update-db.py
@@ -158,8 +158,9 @@ def update_db(path, new_db):
         new_conn.close()
 
         os.remove(new_db)
-    except sqlite3.OperationalError:
-        print(f"Unable to open temporary database file: %s" % new_db)
+    except sqlite3.OperationalError as exc:
+        print(f"Unable to open temporary database file: {new_db}")
+        print(f"Exception: {exc}")
         sys.exit(1)
 
 
--- a/src/cmd/flux-account-update-db.py
+++ b/src/cmd/flux-account-update-db.py
@@ -44,6 +44,77 @@ def est_sqlite_conn(path):
     return conn
 
 
+# gets a final list of columns for the new version of a table, which
+# include any columns added in a newer version or removed from an older version
+def get_cols_list(old_columns, new_columns):
+    cols = []
+
+    for column in new_columns:
+        if column in old_columns and column in new_columns:
+            cols.append(column)
+        elif column in new_columns:
+            cols.append(column)
+
+    return cols
+
+
+# adds a new version of the table to the DB
+def add_tmp_table_to_db(old_cur, table, cols):
+    add_stmt = "CREATE TABLE IF NOT EXISTS " + table[0] + "_tmp" + " ("
+
+    for column in cols:
+        column_name = column[1]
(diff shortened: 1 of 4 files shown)


ryanday36 (approved):
looks good @cmoussa1!

cmoussa1:
Thanks @ryanday36! Will set MWP shortly

PR #254, opened 2022-07-21
plugin: disable user/bank entry in priority plugin when row gets deleted in DB

### LINKED ISSUE #232: plugin: need a way to "disable" user/bank entries in internal map

cmoussa1:
#### Problem

As mentioned in #232, we need a method for "disabling" a user/bank entry in the plugin's internal map when a user gets removed from a bank in the flux-accounting DB. Since entries don't get deleted in the internal map, a user can technically still run jobs under a bank they no longer belong to.

----

This is a small PR built on top of #239 that adds the `active` column to the group of user/bank information sent to the priority plugin, where it is then checked in `validate_cb ()` when a user/bank association submits a job. If the `active` column shows `0`, then that means the user/bank association is disabled from the flux-accounting DB, and thus disabled from running jobs.

A small sharness test is also added to exercise this check, where a user is added to two different banks in the flux-accounting DB and can successfully submit jobs under each bank. When the user gets removed from one bank and tries to submit a job under that bank, the job is rejected with an appropriate message saying that the association has been removed from the DB. Note that if a user is disabled (and the information is sent from the DB to the plugin) while they still have active jobs, the active jobs are still allowed to finish. Subsequently submitted jobs will be rejected, however.

Fixes #232

### DIFF
--- a/t/t1018-mf-priority-disable-entry.t
+++ b/t/t1018-mf-priority-disable-entry.t
@@ -1,0 +1,114 @@
+#!/bin/bash
+
+test_description='Test rejecting jobs from a user who has been disabled from the flux-accounting DB'
+
+. `dirname $0`/sharness.sh
+MULTI_FACTOR_PRIORITY=${FLUX_BUILD_DIR}/src/plugins/.libs/mf_priority.so
+DB_PATH=$(pwd)/FluxAccountingTest.db
+
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 1 job
+
+flux setattr log-stderr-level 1
+
+test_expect_success 'load multi-factor priority plugin' '
+	flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
+'
+
+test_expect_success 'check that mf_priority plugin is loaded' '
+	flux jobtap list | grep mf_priority
+'
+
+test_expect_success 'create flux-accounting DB' '
+	flux account -p $(pwd)/FluxAccountingTest.db create-db
+'
+
+test_expect_success 'add some banks to the DB' '
+	flux account -p ${DB_PATH} add-bank root 1 &&
+	flux account -p ${DB_PATH} add-bank --parent-bank=root account1 1 &&
+	flux account -p ${DB_PATH} add-bank --parent-bank=root account2 1
+'
+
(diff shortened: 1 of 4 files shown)


t/t1018-mf-priority-disable-entry.t
```diff
@@ -0,0 +1,114 @@
...
+	test $(flux jobs -no {state} ${jobid5}) = RUN &&
+	flux job cancel $jobid5
+'
+
+test_expect_success 'trying to submit a job now should result in a job rejection' '
+	test_must_fail flux mini submit --setattr=system.bank=account2 -n1 hostname > deleted_entry2.out 2>&1 &&
+	test_debug "cat deleted_entry2.out" &&
+	grep "user/bank entry has been disabled from flux-accounting DB" deleted_entry2.out
+'
+
```
grondo, line 113: Maybe add a test that re-activates the user/bank and ensures jobs can be submitted again?

grondo (approved):
This looks good to me. Only thought I had was to add a few more tests (comment inline)

t/t1018-mf-priority-disable-entry.t
cmoussa1: I believe I have a couple of tests that go through this scenario earlier in the test file:

```
test_expect_success 're-add second user/bank entry and update-plugin' '
	flux account -p ${DB_PATH} add-user --username=$username --userid=$uid --bank=account2 --shares=1 &&
	flux account-priority-update -p $(pwd)/FluxAccountingTest.db
'

test_expect_success 'submit a job successfully under second bank' '
	jobid3=$(flux mini submit --setattr=system.bank=account2 -n1 hostname) &&
	flux job wait-event -f json $jobid3 priority | jq '.context.priority' > job3.test &&
	cat <<-EOF >job3.expected &&
	10050000
	EOF
	test_cmp job3.expected job3.test
'
```

Sorry if you meant something else though (or in addition to what I had before 😅)!

grondo: Ah, sorry I missed that! LGTM then!

cmoussa1:
Thanks for reviewing @grondo! I'll set MWP shortly :-)

Issue #247, opened 2022-07-11
add "project" support


cmoussa1:
I will probably add more specific sub-issues once I get more information on what is required and how WCKey support is supposed to be implemented, but I wanted to just make a note for now that flux-accounting will need to support workload characterization keys (WCKeys for short) for users in a flux-accounting database. My notes so far consist of the following:

WCKeys allow users from different banks to all work on the same project instead of creating a new bank and adding a bunch of users to it. This WCKey is specified on job submission with, but if not specified, it will use the user’s default WCKey for the cluster. On the job record for a user’s job, the record will be appended with an asterisk (*) to show that the user used their default WCKey.

garlick:
Here's the slurm documentation for wckey:  https://slurm.schedmd.com/wckey.html

If I understand correctly, resource utilization reports can now be generated per user or per bank, and this adds another way to group users arbitrarily and report the group's resource utilization?

Would it be more natural to call this something like a "project" or "team" or even a "group"?

ryanday36:
It sounds like you all are on it. The idea is to be able to group jobs / users for reporting purposes in a way that is independent of their bank / priority. So, it shouldn't be included in or affect fairshare. I see it as more of a way to be able to set a bit of metadata in the accounting db that will be attached to all of a user's jobs unless they specify a different value for that metadata when they submit. Also, yes, calling them 'projects' is perfectly reasonable and a lot less clunky than 'workload characterization keys'.

ryanday36:
> * a user can be a member of multiple projects

True. A user can be a member of multiple projects, with one of them being the user's default project.

> * jobs submitted with an unknown project are rejected

False. A user may run on a cluster even if they are not a member of a project.

> * jobs submitted without a project are not charged to a project

True. Although WCI does tend to want to assign users to a project after the fact if the user has a lot of usage in a given reporting period.

ryanday36:
> I guess I was trying to nail down what happens at submit time if I specify an invalid project name (as opposed to leaving the project out entirely). Shouldn't it be rejected in case the user just typed it wrong?
> 
> A follow on question would be what happens if I specify a valid project that I'm not a member of?
> 
> I was assuming that the accounting db would track all valid project names and their user membership, but maybe this is intended to be looser than that.

Gotcha. I misunderstood. Yes, jobs where a user specifies an invalid project or a project that they're not a member of should probably be rejected.

garlick:
Linking flux-framework/flux-core#3257

[...]

PR #255, opened 2022-07-25
database: add `project_table`, `projects` column to association_table

### LINKED ISSUE #253: flux-accounting DB: add "project" column, "project" table
Probably the first sub-issue to open as a result of #247 is to add support in the flux-accounting database to store "projects" for each user/bank row in the `association_table`, as well as a "projects" table to store names of projects. A quick schema comes to mind for both the row and the table itself:

```sql
project    tinytext    DEFAULT '*'    NOT NULL
```

```sql
CREATE TABLE IF NOT EXISTS projects_table (
    project_id    integer    PRIMARY KEY    AUTOINCREMENT,
    project       text                      NOT NULL
);
```

We can then add support in the `add-user` and `edit-user` commands to specify projects when adding a user/bank row for the first time as well as editing eligible projects for a user/bank row after they've been added.

Separate commands should probably also be added to add, remove, view, and edit projects in the `projects_table`. (side note: the fixes introduced in PR #252 should hopefully alleviate the issues we've run into in the past with updating our currently-installed DB's on fluke, so adding this will be a good test to ensure the `update-db` command is working as expected :P).

cmoussa1:
As discussed in #247 and in #253, there is need to support projects within flux-accounting, or the ability for multiple users from different banks to all work on the same project. One of the first requirements is to be able to add projects to a table to keep track of as well as list projects that a user belongs to in the `association_table`.

---

This is a [WIP] PR that adds a new table to the flux-accounting database: `project_table`, a table storing project names. Its schema consists of the following:

```sql
CREATE TABLE IF NOT EXISTS project_table (
        project_id          integer    PRIMARY KEY AUTOINCREMENT,
        project             tinytext               NOT NULL
);
```

A new column is also added to the `association_table`: `projects`, which stores all of the projects that a user is allowed to specify when running jobs. A default value (`*`) is added as a default project for all users when they are added to the flux-accounting DB:

```sql
projects         tinytext    DEFAULT '*'            NOT NULL    ON CONFLICT REPLACE DEFAULT '*'
```

A couple of new commands to interact with the `project_table` are also added that allows an admin to add, view, and remove project information from the table. A helper function is also added to the `add-user` and `edit-user` commands to validate the projects specified for a user/bank row in the `association_table` so that a user does not get added to a project that does not exist in the `project_table`.

When removing a project from the `project_table`, a quick check is performed in the `association_table` to see if any user/bank rows still reference this project; if so, a warning message is printed notifying the admin that the project being deleted is still being referenced, and thus those user rows should be edited to account for the deletion. 

Both unit tests and sharness tests are added to test both the functions that interact with the `project_table` and the `add-project`, `view-project`, and `delete-project` commands.

Since I am still a little unfamiliar with our current use of projects (currently known as WCKeys), I'll pose a couple questions here that might help add some more useful functionality or change my current PR to be better:

- are deleting projects common on our current machines? When projects get deleted, do any users who still reference that project also get edited automatically to no longer reference that project, or are admins also responsible for changing those user rows?
- do projects have a `job_usage` value? i.e is it common on our current machines to calculate a usage value for all of the jobs that run under a specific project? I know @ryanday36 mentioned in #247 that it does not affect and is not included in fairshare, but I figured I could get a little more clarification in case I should add another column to the `project_table` to include a usage value.

Fixes #253

### DIFF
--- a/src/bindings/python/fluxacct/accounting/Makefile.am
+++ b/src/bindings/python/fluxacct/accounting/Makefile.am
@@ -17,7 +17,8 @@ TESTSCRIPTS = \
 	test/test_example.py \
 	test/test_job_archive_interface.py \
 	test/test_user_subcommands.py \
-	test/test_queue_subcommands.py
+	test/test_queue_subcommands.py \
+	test/test_project_subcommands.py
 
 dist_check_SCRIPTS = \
 	$(TESTSCRIPTS)
--- a/src/bindings/python/fluxacct/accounting/Makefile.am
+++ b/src/bindings/python/fluxacct/accounting/Makefile.am
@@ -3,6 +3,7 @@ acctpy_PYTHON = \
 	user_subcommands.py \
 	bank_subcommands.py \
 	queue_subcommands.py \
+	project_subcommands.py \
 	job_archive_interface.py \
 	create_db.py
 
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -101,6 +101,8 @@ def create_db(
                 max_active_jobs  int(11)     DEFAULT 7              NOT NULL    ON CONFLICT REPLACE DEFAULT 7,
                 max_nodes        int(11)     DEFAULT 2147483647     NOT NULL    ON CONFLICT REPLACE DEFAULT 2147483647,
                 queues           tinytext    DEFAULT ''             NOT NULL    ON CONFLICT REPLACE DEFAULT '',
+                projects         tinytext    DEFAULT '*'            NOT NULL    ON CONFLICT REPLACE DEFAULT '*',
+                default_project  tinytext    DEFAULT '*'            NOT NULL    ON CONFLICT REPLACE DEFAULT '*',
                 PRIMARY KEY   (username, bank)
         );"""
     )
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -177,4 +177,18 @@ def create_db(
             );"""
     )
 
+    # Projects Table
+    # stores projects
+    logging.info("Creating project_table in DB...")
+    conn.execute(
+        """
+            CREATE TABLE IF NOT EXISTS project_table (
+                project_id          integer    PRIMARY KEY AUTOINCREMENT,
+                project             tinytext               NOT NULL,
+                usage               real       DEFAULT 0.0 NOT NULL
+            );"""
+    )
+    conn.execute("INSERT INTO project_table (project) VALUES ('*')")
+    conn.commit()
+
     conn.close()
--- a/src/bindings/python/fluxacct/accounting/project_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/project_subcommands.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2022 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import sqlite3
+
+
+def view_project(conn, project):
+    cur = conn.cursor()
+    try:
+        # get the information pertaining to a project in the DB
+        cur.execute("SELECT * FROM project_table where project=?", (project,))
+        rows = cur.fetchall()
+        headers = [description[0] for description in cur.description]
+        if not rows:
+            print("Project not found in project_table")
+        else:
+            # print column names of project_table
+            for header in headers:
+                print(header.ljust(18), end=" ")
+            print()
+            for row in rows:
+                for col in list(row):
+                    print(str(col).ljust(18), end=" ")
+                print()
+    except sqlite3.OperationalError as e_database_error:
+        print(e_database_error)
+
+
+def add_project(conn, project):
+    try:
+        insert_stmt = "INSERT INTO project_table (project) VALUES (?)"
+        conn.execute(
+            insert_stmt,
+            (project,),
+        )
+
+        conn.commit()
+    # make sure entry is unique
+    except sqlite3.IntegrityError as integrity_error:
+        print(integrity_error)
+
+
+def delete_project(conn, project):
+    cursor = conn.cursor()
+
+    # look for any rows in the association_table that reference this project
+    select_stmt = "SELECT * FROM association_table WHERE projects LIKE ?"
+    cursor.execute(select_stmt, ("%" + project + "%",))
+    rows = cursor.fetchall()
+    if len(rows) > 0:
+        print(
+            """WARNING: user(s) in the assocation_table still reference this project.
+                 Make sure to edit user rows to account for this deleted project."""
+        )
+
+    delete_stmt = "DELETE FROM project_table WHERE project=?"
+    cursor.execute(delete_stmt, (project,))
+
+    conn.commit()
--- a/src/bindings/python/fluxacct/accounting/test/test_create_db.py
+++ b/src/bindings/python/fluxacct/accounting/test/test_create_db.py
@@ -54,6 +54,7 @@ class TestDB(unittest.TestCase):
             "job_usage_factor_table",
             "t_half_life_period_table",
             "queue_table",
+            "project_table",
         ]
         self.assertEqual(list_of_tables, expected)
 
--- a/src/bindings/python/fluxacct/accounting/test/test_project_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/test/test_project_subcommands.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2022 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import unittest
+import os
+import sqlite3
+
+from fluxacct.accounting import create_db as c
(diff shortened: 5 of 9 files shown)


ryanday36:
regarding deleting projects: I think it's okay to require the admin to remove projects from a user before deleting them. That said, an easy tool to do this would no doubt be welcome.

regarding usage: We will want to be able to calculate the total usage for all of the jobs that use a given project in a given time frame, but that's different than the usage that is used for fairshare (which has the half-life decay applied to it). As long as jobs are tagged with a project, we can calculate the usage when needed.

Does this PR include a means of specifying a default project for a user. e.g. if a user has multiple projects, can one be specified as the one to be used if the user doesn't specify a project when they submit their job?

cmoussa1:
> regarding deleting projects: I think it's okay to require the admin to remove projects from a user before deleting them. That said, an easy tool to do this would no doubt be welcome.

Sounds good. Perhaps if it's okay with you, and to keep this PR relatively simple, maybe I can add such a tool in a subsequent PR shortly after this one lands?

> regarding usage: We will want to be able to calculate the total usage for all of the jobs that use a given project in a given time frame, but that's different than the usage that is used for fairshare (which has the half-life decay applied to it). As long as jobs are tagged with a project, we can calculate the usage when needed.

OK, this sounds good. So maybe I should add in fact a `usage` column to the `project_table`; the difference between a project's job usage and a user's job usage is that the project's usage does not have a half-life decay applied to it?  

This means, in the job usage calculation script in this repository (and in the `flux account update-usage` command), perhaps an additional optional argument should be added that allows an admin to query all jobs that ran under a specific project. I think I can accomplish this now if a user sets a project similar to how they currently specify a bank (with `--setattr=system.project=my_project`). Then, we can probably query jobs from the job-archive database in flux-core by only appending jobs that contain a specific project in jobspec. This functionality would probably also be best suited as a subsequent PR.

> Does this PR include a means of specifying a default project for a user. e.g. if a user has multiple projects, can one be specified as the one to be used if the user doesn't specify a project when they submit their job?

Yes, this PR does add a default project for every user who is added to the flux-accounting database - the default project name for a user is just an asterisk (`*`), which I _think_ is what we currently use on our systems? Please let me know if this is incorrect (or if you have a suggestion for a better default name). So a user should not have to specify a project when they run jobs, and if they belong to multiple projects, not specifying a project name will result in running under a default project (`*`).

ryanday36:
> > regarding deleting projects: I think it's okay to require the admin to remove projects from a user before deleting them. That said, an easy tool to do this would no doubt be welcome.
> 
> Sounds good. Perhaps if it's okay with you, and to keep this PR relatively simple, maybe I can add such a tool in a subsequent PR shortly after this one lands?
> 

Yes. That sounds good.

> > regarding usage: We will want to be able to calculate the total usage for all of the jobs that use a given project in a given time frame, but that's different than the usage that is used for fairshare (which has the half-life decay applied to it). As long as jobs are tagged with a project, we can calculate the usage when needed.
> 
> OK, this sounds good. So maybe I should add in fact a `usage` column to the `project_table`; the difference between a project's job usage and a user's job usage is that the project's usage does not have a half-life decay applied to it?
> 
> This means, in the job usage calculation script in this repository (and in the `flux account update-usage` command), perhaps an additional optional argument should be added that allows an admin to query all jobs that ran under a specific project. I think I can accomplish this now if a user sets a project similar to how they currently specify a bank (with `--setattr=system.project=my_project`). Then, we can probably query jobs from the job-archive database in flux-core by only appending jobs that contain a specific project in jobspec. This functionality would probably also be best suited as a subsequent PR.
> 

It sounds like I wasn't quite clear on this. The usage that we report for projects will have arbitrary start and end dates, so isn't really suited to keeping in the association table. We'll be calculating it from archived jobs or something along those lines. The SlurmDB does have tables where it keeps aggregated usages for associations by hour / day / month, and we may want to do something like that eventually, but it's not a near term concern for me.

> > Does this PR include a means of specifying a default project for a user. e.g. if a user has multiple projects, can one be specified as the one to be used if the user doesn't specify a project when they submit their job?
> 
> Yes, this PR does add a default project for every user who is added to the flux-accounting database - the default project name for a user is just an asterisk (`*`), which I _think_ is what we currently use on our systems? Please let me know if this is incorrect (or if you have a suggestion for a better default name). So a user should not have to specify a project when they run jobs, and if they belong to multiple projects, not specifying a project name will result in running under a default project (`*`).

If I'm understanding this correctly, it's not quite what I was looking for. What I'd like would be more like if a user belonged to 'projectA' and 'projectB', a way to say that if the user doesn't specify a project name when they submit a job, the job should get 'projectB'.

cmoussa1:
> It sounds like I wasn't quite clear on this. The usage that we report for projects will have arbitrary start and end dates, so isn't really suited to keeping in the association table. We'll be calculating it from archived jobs or something along those lines. The SlurmDB does have tables where it keeps aggregated usages for associations by hour / day / month, and we may want to do something like that eventually, but it's not a near term concern for me.

Sorry about the confusion from my end; I think I was proposing keeping usage by project in the `project_table`, not in the `association_table`. The usage calculated for a project would be calculated by fetching jobs from flux-core's job-archive DB (where old job records are stored) and could have time filters applied to it as well (e.g "I want all of the jobs that under projectA before time X"). Does that sound more like what you were looking for?

> If I'm understanding this correctly, it's not quite what I was looking for. What I'd like would be more like if a user belonged to 'projectA' and 'projectB', a way to say that if the user doesn't specify a project name when they submit a job, the job should get 'projectB'.

Ah, okay - sorry about that. I misunderstood then! So were you thinking that a user could have a default project name, kind of like how users currently belong to banks and have a default bank?

ryanday36:
> > It sounds like I wasn't quite clear on this. The usage that we report for projects will have arbitrary start and end dates, so isn't really suited to keeping in the association table. We'll be calculating it from archived jobs or something along those lines. The SlurmDB does have tables where it keeps aggregated usages for associations by hour / day / month, and we may want to do something like that eventually, but it's not a near term concern for me.
> 
> Sorry about the confusion from my end; I think I was proposing keeping usage by project in the `project_table`, not in the `association_table`. The usage calculated for a project would be calculated by fetching jobs from flux-core's job-archive DB (where old job records are stored) and could have time filters applied to it as well (e.g "I want all of the jobs that under projectA before time X"). Does that sound more like what you were looking for?
> 

kind of. The trick is that I don't necessarily know ahead of time what filters we would want. I think that it's usually the usage for the past quarter, but sometimes it might be the usage for two quarters ago, and others it might be just the past month. I can talk to Greg T about whether there are any common things that would be worth keeping in the project table, but I don't see that as urgent. It's more important that jobs in the job archive DB have an appropriate project associated with them.

> > If I'm understanding this correctly, it's not quite what I was looking for. What I'd like would be more like if a user belonged to 'projectA' and 'projectB', a way to say that if the user doesn't specify a project name when they submit a job, the job should get 'projectB'.
> 
> Ah, okay - sorry about that. I misunderstood then! So were you thinking that a user could have a default project name, kind of like how users currently belong to banks and have a default bank?

Yes. That would be perfect.

cmoussa1:
> kind of. The trick is that I don't necessarily know ahead of time what filters we would want. I think that it's usually the usage for the past quarter, but sometimes it might be the usage for two quarters ago, and others it might be just the past month. I can talk to Greg T about whether there are any common things that would be worth keeping in the project table, but I don't see that as urgent. It's more important that jobs in the job archive DB have an appropriate project associated with them.

OK. I think that makes sense. Thanks for clarifying that with me. In the case of this PR, I think I will just hold off on adding the capability of associating jobs with projects (just for now) until a future PR. I think it will keep the direction of this one relatively straightforward, and in a future PR, I can start looking at how we want to add associating jobs with projects. 

>>> If I'm understanding this correctly, it's not quite what I was looking for. What I'd like would be more like if a user belonged to 'projectA' and 'projectB', a way to say that if the user doesn't specify a project name when they submit a job, the job should get 'projectB'.

>> Ah, okay - sorry about that. I misunderstood then! So were you thinking that a user could have a default project name, kind of like how users currently belong to banks and have a default bank?

> Yes. That would be perfect.

OK, cool. So I'll go ahead and edit this PR to instead treat projects (and default projects) more like banks, where users can have a default project when none is specified. Does this mean that at least one project _has_ to be specified when a user is added to the `association_table`? That would also be similar to having to specify a bank when a user is added to the `association_table` as well.

ryanday36:
> OK, cool. So I'll go ahead and edit this PR to instead treat projects (and default projects) more like banks, where users can have a default project when none is specified. Does this mean that at least one project _has_ to be specified when a user is added to the `association_table`? That would also be similar to having to specify a bank when a user is added to the `association_table` as well.

No, we want both kinds of defaults :) If no project is specified when the user is created, giving them '*' for their project and default project is what we want.

cmoussa1:
> No, we want both kinds of defaults :) If no project is specified when the user is created, giving them '*' for their project and default project is what we want.

Ah, okay, I think I understand. Thanks! I'll go back and rework this PR slightly to account for this behavior.

cmoussa1:
OK, I've pushed the following changes to this PR in regards to handling default project names:

A user can be added to the `association_table` with specified projects, like so:

```console
$ flux account add-user --username=user1234 --bank=bank_A --projects="project_A,project_B,project_C"
```

`project_A` will become the user's default project (since it was the first project listed), and if no project is specified when running jobs, jobs will fall under `project_A`. A user's default project can also be edited with the `edit-user` command:

```console
$ flux account edit-user user1234 --default_project=project_B
```

If a user is added to the `association_table` without specifying any projects, a default project (`*`) is added for the user automatically, and jobs submitted without specifying a project will fall under `*`.

Every user who is added to the `association_table` belongs to the `*` project, but will only run jobs under `*` project if they do not already have another default project.

@ryanday36 - whenever you get the chance, let me know if this is the functionality you had in mind, or if there's something I missed in your explanation.

ryanday36 (commented):
Looks good. Nice work @cmoussa1!

cmoussa1:
Thanks again for reviewing @ryanday36! I'll set MWP on this

cmoussa1:
Ah, sorry @ryanday36, I didn't realize until just now - I think you might need to explicitly "approve" this PR in order for it to be merged.

ryanday36 (approved):

ryanday36:
Sorry. Forgot to click the radio button. Should be good now.

cmoussa1:
Thanks!!

PR #258, opened 2022-08-04
`update-db`: improve helper functions, add tests for previous versions of flux-accounting DB

### LINKED ISSUE #257: flux account-priority-update and account-update-db fail with v0.18.0
On fluke after an update to flux-core v0.42.0 and flux-accounting v0.18.0, all jobs were getting stuck in PRIORITY.

Looks like `flux account-priority-update` was failing with the error:

```console
$ flux account-priority-update
  File "/usr/libexec/flux/cmd/flux-account-priority-update.py", line 119, in <module>
    main()
  File "/usr/libexec/flux/cmd/flux-account-priority-update.py", line 115, in main
    bulk_update(path)
  File "/usr/libexec/flux/cmd/flux-account-priority-update.py", line 59, in bulk_update
    queues, active FROM association_table"""
sqlite3.OperationalError: no such column: active
```

I assumed that a `flux account-update-db` was required, so I tried that and got the error:
```console
$ flux account-update-db 
checking for new tables...
checking for new columns to add in tables...
  File "/usr/libexec/flux/cmd/flux-account-update-db.py", line 263, in <module>
    main()
  File "/usr/libexec/flux/cmd/flux-account-update-db.py", line 259, in main
    update_db(old_db, args.new_db)
  File "/usr/libexec/flux/cmd/flux-account-update-db.py", line 223, in update_db
    update_columns(old_conn, old_cur, new_cur)
  File "/usr/libexec/flux/cmd/flux-account-update-db.py", line 197, in update_columns
    move_existing_rows(old_cur, cols, old_columns, table)
  File "/usr/libexec/flux/cmd/flux-account-update-db.py", line 104, in move_existing_rows
    old_cur.execute(final_stmt)
sqlite3.IntegrityError: NOT NULL constraint failed: bank_table_tmp.shares
```

Though even after these errors `flux account-priority-update` worked and jobs can now run. :shrug: 

Besides tracking down and adding tests for the errors above, we should perhaps open issues for the following:

 * If the database scheme is going to be updated frequently, we'll should try to make the upgrade of existing database more transparent. Maybe the upgrade-db command could be made part of the cron job that is installed by the flux-accounting RPM, so that admins don't have to remember to do it?
 * Need much better testing for the upgrade-db command soon. Before the end of the month this will be in use on  production clusters with real users. We don't want to extend downtimes by having to fix small issues by hand.

cmoussa1:
#### Problem

Reported in #257, an exception message that reported that a `NOT NULL` constraint was not satisfied because the data from the `shares` column in the old version of the `bank_table` was not transported to the new version of the `bank_table`. 

In `move_existing_rows()` and `get_cols_list()`, the list of columns from the old version of a table are compared with the list of columns from the new version of a table so that all of the old data can be moved to a new version of the table. The two lists are compared with each other, but instead of comparing just the names of the columns, the entire tuple is compared, which results in false negatives because the index of the tuples could be different (although the rest of the data is identical). An example:

```python
(1, 'some text', True) != (2, 'some text', True)
```

---

This PR modifies these two helper functions to use lists of columns that contain just the names of the columns instead of the tuples containing all of the column information, which should prevent these false negatives from occurring when columns are added or removed from one version to the next.

Fixes #257

##### TODO

- If possible, a new CI check should be added in this PR that creates a flux-accounting DB (`flux account create-db`) from a previous tag and runs `flux account-update-db` to make sure that the command runs successfully and no exceptions are raised.


cmoussa1:
OK, I've pushed up some additional changes to the `update-db` command to add some more exception handling and message output if an error were to come up while updating a flux-accounting database in response to the `sqlite3.IntegrityError` message that was reported in #257.

I've also added additional tests to `t1017-update-db.t` that runs the `update-db` command on every version of a flux-accounting DB since `v0.10.0` (I don't think (and hope) there exists any installation of flux-accounting on our systems that run an earlier version). Each sample database has two bank entries and one user entry, and the `update-db` command is run on each one to ensure that no exceptions are raised.

Now that I've added some additional tests, I'l go ahead and mark this as ready for initial review whenever convenient.

t/t1017-update-db.t
```diff
@@ -87,4 +96,70 @@ test_expect_success 'get all the columns from the queue_table and make sure the
 	test_cmp queue_table_columns.expected queue_table_columns.test
 '
 
+test_expect_success 'successfully update flux-accounting DB from v0.10.0' '
```
grondo, line 99: You could make use of a for loop here to avoid using a variable per db version. Then, all you would need to do is drop a new database into `expected/test_dbs` for it to be checked in this test file.

For an example, see the end of flux-core `t/t2219-job-manager-restart.t`:

https://github.com/flux-framework/flux-core/blob/3c367a3a15e888857cab2e8dbf2704e97861a494/t/t2219-job-manager-restart.t#L57-L66

grondo (approved):
This looks good to me.

I noted a possible improvement in the testsuite, but that could also be an enhancement saved for later if you'd like.

t/t1017-update-db.t
cmoussa1: Ah, nice, thanks for this suggestion. If it's okay with you, I'd like to open an issue on this and work on this in the near future, just for the purposes of landing this PR by today.

cmoussa1:
Thanks for reviewing and approving @grondo! I've opened a new issue on the suggestion you've made above, so I'll go ahead and set MWP on this :-)

PR #264, opened 2022-08-17
plugin: add new `plugin.query` callback

### LINKED ISSUE #263: plugin: need a callback that will return plugin's internal data
As a result of trying to debug #262, perhaps a useful utility for the multi-factor priority plugin to have would be a callback that could return all data that the plugin has at a given moment, which would include all:

- users and bank information
- users' current job limits
- users' default banks

This data should be returned in a readable format and should give us as much helpful information regarding currently running jobs and any held jobs as possible.

cmoussa1:
#### Background

As described in #263, it would be useful to have a utility to fetch the internal state of the multi-factor priority plugin to look at the current data it has on users, banks, and their current running and active jobs. This would be helpful when looking at jobs currently held because of a jobs limit or to see what users or banks are currently in the plugin.

---

This PR adds a new `plugin.query` callback that returns all the internal data that the plugin has about its users, banks, and their running and active job counts at the time of the query in a JSON object. It includes data such as:

- all the users/banks it has received from the flux-accounting DB via `flux account-priority-update`
- current fair-share values, `max_run_jobs` and `max_active_jobs` counts for each user/bank
- any held job IDs for each user/bank

Fixes #263

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -198,11 +198,187 @@ int check_queue_factor (flux_plugin_t *p,
 }
 
 
+/*
+ * Add held job IDs to a JSON array to be added to a bank_info JSON object.
+ */
+static json_t *add_held_jobs (
+                            const std::pair<std::string, struct bank_info> &b)
+{
+    json_t *held_jobs = NULL;
+
+    // add any held jobs to a JSON array
+    if (!(held_jobs = json_array ()))
+        goto error;
+
+    for (auto const &j : b.second.held_jobs) {
+        json_t *jobid = json_integer (j);
+
+        if (!jobid)
+            goto error;
+
+        if (json_array_append_new (held_jobs, jobid) < 0) {
+            json_decref (jobid);
+            goto error;
+        }
+    }
+
+    return held_jobs;
+error:
+    json_decref (held_jobs);
+    return NULL;
+}
+
+
+/*
+ * Create a JSON object for a bank that a user belongs to.
+ */
+static json_t *pack_bank_info_object (
+                            const std::pair<std::string, struct bank_info> &b)
+{
+    json_t *bank_info, *held_jobs = NULL;
+
+    held_jobs = add_held_jobs (b);
+    if (held_jobs == NULL)
+        goto error;
+
+    if (!(bank_info = json_pack ("{s:s, s:f, s:i, s:i, s:i, s:i, s:o, s:i}",
+                                 "bank", b.first.c_str (),
+                                 "fairshare", b.second.fairshare,
+                                 "max_run_jobs", b.second.max_run_jobs,
+                                 "cur_run_jobs", b.second.cur_run_jobs,
+                                 "max_active_jobs", b.second.max_active_jobs,
+                                 "cur_active_jobs", b.second.cur_active_jobs,
+                                 "held_jobs", held_jobs,
+                                 "active", b.second.active))) {
+            goto error;
+    }
+
+    return bank_info;
+error:
+    json_decref (held_jobs);
+    return NULL;
+}
+
+
+/*
+ * For each bank that a user belongs to, create a JSON object for each bank and
+ * add it to the user JSON object.
+ */
+static json_t *banks_to_json (
+                    flux_plugin_t *p,
+                    std::pair<int, std::map<std::string, struct bank_info>> &u)
+{
+    json_t *bank_info, *banks = NULL;
+
+    banks = json_array (); // array of banks that user belongs to
+    if (!banks)
+        goto error;
+
+    for (auto const &b : u.second) {
+        bank_info = pack_bank_info_object (b); // JSON object for one bank
+        if (bank_info == NULL)
+            goto error;
+
+        if (json_array_append_new (banks, bank_info) < 0) {
+            json_decref (bank_info);
+            goto error;
+        }
+    }
+
+    return banks;
+error:
+    json_decref (banks);
+    return NULL;
+}
+
+
+/*
+ * Iterate thrpugh each user in users map and create a JSON object for each
+ * user.
+ */
+static json_t *user_to_json (
+                    flux_plugin_t *p,
+                    std::pair<int, std::map<std::string, struct bank_info>> u)
+{
+    json_t *user = json_object (); // JSON object for one user
+    json_t *userid, *banks = NULL;
+
+    if (!user)
+        return NULL;
+
+    userid = json_integer (u.first);
+    if (!userid)
+        goto error;
+
+    if (json_object_set_new (user, "userid", userid) < 0) {
+        json_decref (userid);
+        goto error;
+    }
+
+    banks = banks_to_json (p, u);
+    if (banks == NULL)
+        goto error;
+
+    if (json_object_set_new (user, "banks", banks) < 0) {
+        json_decref (banks);
+        goto error;
+    }
+
+    return user;
+error:
+    json_decref (user);
+    return NULL;
+}
+
+
 /******************************************************************************
  *                                                                            *
  *                               Callbacks                                    *
  *                                                                            *
  *****************************************************************************/
+
+/*
+ * Get state of all user and bank information from plugin
+ */
+static int query_cb (flux_plugin_t *p,
+                     const char *topic,
+                     flux_plugin_arg_t *args,
+                     void *data)
+{
+    flux_t *h = flux_jobtap_get_flux (p);
+    json_t *all_users = json_array (); // array of user/bank combos
+
+    if (!all_users)
+        return -1;
+
+    for (auto const &u : users) {
+        json_t *user = user_to_json (p, u);
+        if (user == NULL) {
+            json_decref (all_users);
+            return -1;
+        }
+
+        if (json_array_append_new (all_users, user) < 0) {
+            json_decref (user);
+            json_decref (all_users);
+            return -1;
+        }
+    }
+
+    if (flux_plugin_arg_pack (args,
+                              FLUX_PLUGIN_ARG_OUT,
+                              "{s:O}",
+                              "mf_priority_map",
+                              all_users) < 0)
+        flux_log_error (flux_jobtap_get_flux (p),
+                        "mf_priority: query_cb: flux_plugin_arg_pack: %s",
+                        flux_plugin_arg_strerror (args));
+
+    json_decref (all_users);
+
+    return 0;
+}
+
 
 /*
  * Unpack a payload from an external bulk update service and place it in the
@@ -801,6 +977,7 @@ static const struct flux_plugin_handler tab[] = {
     { "job.priority.get", priority_cb, NULL },
     { "job.state.inactive", inactive_cb, NULL },
     { "job.state.depend", depend_cb, NULL },
+    { "plugin.query", query_cb, NULL},
     { 0 },
 };
 
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -19,7 +19,8 @@ TESTSCRIPTS = \
 	t1015-mf-priority-urgency.t \
 	t1016-export-db.t \
 	t1017-update-db.t \
-	t1018-mf-priority-disable-entry.t
+	t1018-mf-priority-disable-entry.t \
+	t1019-mf-priority-info-fetch.t
 
 dist_check_SCRIPTS = \
 	$(TESTSCRIPTS) \
@@ -71,7 +72,9 @@ EXTRA_DIST= \
 	expected/pop_db/db_hierarchy_new_users.expected \
 	expected/job_usage/pre_update.expected \
 	expected/job_usage/post_update1.expected \
-	expected/job_usage/post_update2.expected
+	expected/job_usage/post_update2.expected \
+	expected/plugin_state/internal_state_1.expected \
+	expected/plugin_state/internal_state_3.expected
 
 clean-local:
 	rm -fr trash-directory.* test-results .prove *.broker.log */*.broker.log *.output
--- a/t/expected/plugin_state/internal_state_1.expected
+++ b/t/expected/plugin_state/internal_state_1.expected
@@ -1,0 +1,27 @@
+[
+  {
+    "userid": 1001,
+    "banks": [
+      {
+        "bank": "account1",
+        "fairshare": 0.5,
+        "max_run_jobs": 2,
+        "cur_run_jobs": 0,
+        "max_active_jobs": 7,
+        "cur_active_jobs": 0,
+        "held_jobs": [],
+        "active": 1
+      },
+      {
+        "bank": "account2",
+        "fairshare": 0.5,
+        "max_run_jobs": 5,
+        "cur_run_jobs": 0,
+        "max_active_jobs": 7,
+        "cur_active_jobs": 0,
+        "held_jobs": [],
+        "active": 1
+      }
+    ]
+  }
+]
--- a/t/expected/plugin_state/internal_state_3.expected
+++ b/t/expected/plugin_state/internal_state_3.expected
@@ -1,0 +1,42 @@
+[
+  {
+    "userid": 1001,
+    "banks": [
+      {
+        "bank": "account1",
+        "fairshare": 0.5,
+        "max_run_jobs": 2,
+        "cur_run_jobs": 0,
+        "max_active_jobs": 7,
+        "cur_active_jobs": 0,
+        "held_jobs": [],
+        "active": 1
+      },
+      {
+        "bank": "account2",
+        "fairshare": 0.5,
+        "max_run_jobs": 5,
+        "cur_run_jobs": 0,
+        "max_active_jobs": 7,
+        "cur_active_jobs": 0,
+        "held_jobs": [],
+        "active": 1
+      }
+    ]
+  },
+  {
+    "userid": 1002,
+    "banks": [
+      {
+        "bank": "account3",
+        "fairshare": 0.5,
+        "max_run_jobs": 5,
+        "cur_run_jobs": 0,
+        "max_active_jobs": 7,
+        "cur_active_jobs": 0,
+        "held_jobs": [],
+        "active": 1
+      }
+    ]
+  }
+]
--- a/t/t1019-mf-priority-info-fetch.t
+++ b/t/t1019-mf-priority-info-fetch.t
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+test_description='Test getting internal state of plugin using flux jobtap query'
+
+. `dirname $0`/sharness.sh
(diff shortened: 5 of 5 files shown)


grondo (commented):
Actually, it would be much better to just emit the JSON object on `stdout`. Then the output can easily be parsed by filters like `jq`. If a user wants to send the output to a file they can easily do so with shell redirection.

This greatly simplifies the "command" since you don't have to support any options processing. (In fact it will basically be one line)

Another note, instead of an RPC that is called `print_map_info` it might be more idiomatic to use `get_map_info` or `dump_map_info` since the RPC is not printing the map info, but just fetching it.

cmoussa1:
Oh! Okay, that does make it easier :-) thanks for this suggestion @grondo! I'll work on changing this to just print to `stdout`.

grondo:
@cmoussa1: @garlick had the great idea to add a new `flux jobtap info [PLUGIN]` command for jobtap and a corresponding jobtap callback from which a plugin can optionally generate detailed info to return to the caller.

Let's pause this PR until we have that support merged in flux-core. (Your work on generating the JSON object representation of the plugin internal state won't be wasted though)

cmoussa1:
Okay! Thanks for letting me know about that. Good idea @garlick :-)

grondo:
@cmoussa1 - check out flux-framework/flux-core#4507 for a proposed way to do this with `flux jobtap query mf_priority.so`. I'll request a review from you.

cmoussa1:
OK, I think I've hopefully made the right adjustments to this PR after flux-framework/flux-core#4507, so I'll take this out of [WIP] and mark it as ready for some initial review whenever convenient.

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +204,71 @@ int check_queue_factor (flux_plugin_t *p,
...
+                                                b.second.max_run_jobs));
+            json_object_set_new (
+                bank_info, "cur_run_jobs", json_integer (
+                                                b.second.cur_run_jobs));
+            json_object_set_new (
+                bank_info, "max_active_jobs", json_integer (
+                                                b.second.max_active_jobs));
+            json_object_set_new (
+                bank_info, "cur_active_jobs", json_integer (
+                                                b.second.cur_active_jobs));
```
grondo, line 243: I was going to suggest that return codes be checked here, but maybe this would be simpler if you created `bank_info` with `json_pack()`? e.g.:

```c
    if (!(bank_info = json_pack ("{s:s s:f s:i s:i s:i s:i}",
                                 "bank", b.first.c_str (),
                                 "fairshare", b.second.fairshare,
                                 etc....    
```

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +204,71 @@ int check_queue_factor (flux_plugin_t *p,
...
+
+    if (flux_plugin_arg_pack (args,
+                              FLUX_PLUGIN_ARG_OUT,
+                              "{s:O}",
+                              "plugin_map_info",
+                              all_users) < 0)
+        flux_log_error (flux_jobtap_get_flux (p),
+                        "mf_priority: query_cb: flux_plugin_arg_pack: %s",
+                        flux_plugin_arg_strerror (args));
+
```
grondo, line 267: Missing `json_decref (all_users)`

grondo (commented):
I've just done a quick pass and noted a couple things.

src/plugins/mf_priority.cpp
cmoussa1: Ah, yeah, that might be simpler. I just pushed up a `fixup` commit that just uses `json_pack ()` as well as adds a `json_decref ()` to the JSON objects that are created. It definitely does reduce the number of lines needed to create the object. Let me know if that's what you had in mind (and/or if I didn't use the `decref ()` call correctly 😬)

src/plugins/mf_priority.cpp
```diff
@@ -204,6 +204,68 @@ int check_queue_factor (flux_plugin_t *p,
...
+                    "{s:s, s:f, s:i, s:i, s:i, s:i, s:o, s:i}",
+                    "bank", b.first.c_str (),
+                    "fairshare", b.second.fairshare,
+                    "max_run_jobs", b.second.max_run_jobs,
+                    "cur_run_jobs", b.second.cur_run_jobs,
+                    "max_active_jobs", b.second.max_active_jobs,
+                    "cur_active_jobs", b.second.cur_active_jobs,
+                    "held_jobs", held_jobs,
+                    "active", b.second.active)) < 0) {
+                json_decref (bank_info);
```
grondo, line 243: Did you mean `json_decref (held__jobs)` here?

grondo (changes requested):
Be sure to check and handle error return from `json_object ()`, `json_array ()`, `json_object_set_new ()`, etc.

cmoussa1:
thanks for the quick reviews @grondo! I've squashed and force-pushed your suggestions to add checks to the `json_* ()` calls in the callback, as well as changed the `json_decref ()` to `held_jobs` instead of `bank_info`. I just returned `-1` if any of the `json_* ()` calls fail.

grondo (commented):
More suggestions inline :smile: Hope this helps!

cmoussa1:
Thanks for all of the reviews and suggestions so far @grondo - can you tell I've not worked with Jannson much? 😄

I just added a `fixup` commit to try and abstract both the construction of a "user" JSON object and any subsequent "bank" JSON object(s) into their own separate functions. Hopefully this clears things up and makes it easier to do error checking.

I think the part where I find myself most confused is when to call `json_decref ()` - do you have to call that on the JSON object in the case of every error check? In the `fixup` commit I've tried to include a `json_decref ()` in every error handling body, but I'm not sure if that is necessary (or if I am `decref`-ing the right object.

grondo:
> I think the part where I find myself most confused is when to call json_decref () - do you have to call that on the JSON object in the case of every error check? I

Yeah, the Jansson `json_t *` object reference count based memory management can get a little confusing. It may be helpful to read through the [Reference Count](https://jansson.readthedocs.io/en/2.10/apiref.html#apiref-reference-count) section of the Jansson docs first to get a basic understanding.

Basically, when you create a new `json_t` object via `json_object ()`, `json_array ()`, `json_integer ()`, and so on, libjansson is allocating some memory for you, and returning a reference to that memory as a `json_t *`. The reference count for the `json_t` starts at 1. If you don't call `json_decref ()`, or pass the reference to another object (e.g. via `json_array_append_new ()` or `json_pack()` with the little `o`), then the memory can't be freed.

Therefore as you accumulate json objects in a function with a reference count, you have to ensure you decref those objects before leaving the function due to an error. If you look at a lot of flux-core code, you will see we use a `goto error:` idiom and call `json_decref ()` on all variables that could possibly still have a reference. This sometimes allows you to reduce the code duplication.

This is also why I suggest encapsulating the creation of each inner JSON object into its own function. This makes error handling easier in that you just have to make sure the components of that one object are freed before returning an error.

You do not have to `json_decref (object)` if the object failed to get allocated, e.g. as you do currently, e.g.

```c
banks = json_array ();
if (!banks)
    json_decref (banks)
```

Since `banks == NULL`, `json_decref (banks)` does nothing and is not needed.

However, later, if `bank_info` fails to be created, you return without calling `json_decref (banks)`. Hopefully you now can see that this is a leak, since `banks` is definitely allocated at this point.

Hopefully this short description and the Jansson documentation helps. Then you can take another crack at it :-)

cmoussa1:
OK, I've squashed and force-pushed a change to hopefully and correctly use `json_decref ()` in the functions that create the JSON objects. Like you mentioned in your suggestions above, I created a `goto error` idiom that `decref`s the `json_t` objects that are created in the two helper functions in the case when an error occurs. Let me know if this is better and/or I need to make additional changes. Thanks for helping me work through this! :P

src/plugins/mf_priority.cpp
```diff
@@ -198,6 +198,71 @@ int check_queue_factor (flux_plugin_t *p,
...
+            goto error;
+        
+        json_decref (jobid);
+    }
+
+    return held_jobs;
+error:
+    json_decref (held_jobs);
+    json_decref (jobid);
+    return NULL;
```
garlick, line 228: This looks correct to me.

I do have one style suggestion: declare `jobid` within the for loop and instead of calling `json_decref (jobid)` at the `error:` label,  use
```c
if (json_array_append_new (held_jobs, jobid) < 0) :
    json_decref (jobid)
    goto error;
}
```
(and skip the `json_decref (jobid)` at the bottom of the for loop).

This is a simple case but I find it gets really confusing in general when you allocate something in a loop and free it outside of the loop.

src/plugins/mf_priority.cpp
```diff
@@ -198,6 +198,71 @@ int check_queue_factor (flux_plugin_t *p,
...
+static json_t *pack_bank_info_object (
+                            const std::pair<std::string, struct bank_info> &b)
+{
+    json_t *bank_info, *held_jobs = NULL;
+
+    bank_info = json_object ();
+    if (!bank_info)
+        goto error;
+
+    held_jobs = add_held_jobs (b);
```
garlick, line 244: Need to handle the case where `add_held_jobs()` returns NULL.

garlick (commented):
Just revewing that last fixup commit since it was the result of some offline discussion.  I think you've nearly got it @cmoussa1.

src/plugins/mf_priority.cpp
cmoussa1: Ah, I like this style suggestion - thanks. This seems more straightforward to me to declare and free the variable within the for-loop, so I'll adopt this!

src/plugins/mf_priority.cpp
cmoussa1: Yes, thanks for pointing this out - I just squashed and pushed a fix to add a check for `held_jobs` after this function call to check for `NULL`.

cmoussa1:
Thanks for the great suggestions @garlick - I just squashed and pushed based on the feedback you gave above. Let me know what you think!

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,181 @@ int check_queue_factor (flux_plugin_t *p,
...
+    for (auto const &j : b.second.held_jobs) {
+        json_t *jobid = json_integer (j);
+
+        if (!jobid)
+            goto error;
+
+        if (json_array_append (held_jobs, jobid) < 0) {
+            json_decref (jobid);
+            goto error;
+        }
```
garlick, line 222: Use `json_array_append_new()` here.  The distinction is that `_new()` doesn't take a reference on `jobid`.  It "steals" the existing reference, so once that call is successful, you no longer have to worry about freeing `jobid`.  Conversely, when it fails you do.

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,181 @@ int check_queue_factor (flux_plugin_t *p,
...
+
+    banks = json_array (); // array of banks that user belongs to
+    if (!banks)
+        goto error;
+
+    for (auto const &b : u.second) {
+        if (!(bank_info = json_object ())) // JSON object for one bank
+            goto error;
+
+        bank_info = pack_bank_info_object (b);
```
garlick, line 286: Need to check for NULL failure here.

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,181 @@ int check_queue_factor (flux_plugin_t *p,
...
+{
+    json_t *bank_info, *banks = NULL;
+
+    banks = json_array (); // array of banks that user belongs to
+    if (!banks)
+        goto error;
+
+    for (auto const &b : u.second) {
+        if (!(bank_info = json_object ())) // JSON object for one bank
+            goto error;
```
garlick, line 284: Like the other function, you _could_ declare `bank_info` inside the loop, then use `json_array_append_new()` as described earlier.

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,181 @@ int check_queue_factor (flux_plugin_t *p,
...
+        return NULL;
+
+    userid = json_integer (u.first);
+    if (!userid)
+        return NULL;
+
+    if (json_object_set_new (user, "userid", userid) < 0)
+        goto error;
+
+    if (json_object_set_new (user, "banks", banks_to_json (p, u)) < 0)
```
garlick, line 323: Need to check for NULL return from `banks_to_json()`.

Also if the `json_object_set_new()` on "banks" fails, now userid is owned by user, so you will get a double free on the error path (userid is freed once when user is freed, then again explicitly).

garlick (commented):
Oops, some of the same comments I made earlier apply to the other functions.  Sorry I made the one comment standalone - I meant to group all these together in one review.

src/plugins/mf_priority.cpp
cmoussa1: I see. Thanks for pointing this out! There are two other calls to `json_array_append ()` in the helper functions (in `banks_to_json ()` and `query_cb ()`, but I left those as is because I'm pretty sure I free the JSON object after the call to `array_append ()`. Let me know if those would be better to also change to `array_append_new ()`.

src/plugins/mf_priority.cpp
cmoussa1: Ah, okay. Maybe I should just declare a separate JSON object to put the result of `banks_to_json ()` into. Then I can check for `NULL` before adding it with `json_object_set_new ()`.

> Also if the `json_object_set_new()` on "banks" fails, now userid is owned by user, so you will get a double free on the error path (userid is freed once when user is freed, then again explicitly).

I think I see what you're saying. So in this case, I do not need to explicitly free `userid` in the error path since it is already owned by `user`?

```diff
error:
    json_decref (user);
-   json_decref (userid);
    return NULL;
```

garlick: Yes but you will need to free it if `json_object_set_new()` on `userid` fails.

cmoussa1:
Thanks for taking another look at this and providing some more feedback - I just squashed and force-pushed based on the above suggestions. 

I think I should also change the following based on some of the feedback above:

```c++
    for (auto const &u : users) {
        if (json_array_append (all_users, user_to_json (p, u)) < 0) {
            json_decref (all_users);
            return -1;
        }
    }
```

`user_to_json (p, u)` should also be checked for NULL before appending it to the `all_users` JSON_array.

cmoussa1:
Just squashed and pushed the change to check the return value for `user_to_json (p, u)` and to change to use `json_array_append_new ()`.

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,195 @@ int check_queue_factor (flux_plugin_t *p,
...
+ * Create a JSON object for a bank that a user belongs to.
+ */
+static json_t *pack_bank_info_object (
+                            const std::pair<std::string, struct bank_info> &b)
+{
+    json_t *bank_info, *held_jobs = NULL;
+
+    bank_info = json_object ();
+    if (!bank_info)
+        goto error;
```
garlick, line 242: Oops I think this was left over from earlier?   The `bank_info` variable is overwritten by `json_pack()` further down, so the reference to the object created here is leaked.

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,195 @@ int check_queue_factor (flux_plugin_t *p,
...
+                                 "cur_active_jobs", b.second.cur_active_jobs,
+                                 "held_jobs", held_jobs,
+                                 "active", b.second.active))) {
+            goto error;
+    }
+
+    return bank_info;
+error:
+    json_decref (held_jobs);
+    json_decref (bank_info);
```
garlick, line 263: bank_info will always be NULL here once the extraneous `json_object()` call is removed,  so this line can be removed.

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,195 @@ int check_queue_factor (flux_plugin_t *p,
...
+    json_t *bank_info, *banks = NULL;
+
+    banks = json_array (); // array of banks that user belongs to
+    if (!banks)
+        goto error;
+
+    for (auto const &b : u.second) {
+        json_t *bank_info = json_object (); // JSON object for one bank
+        if (!bank_info)
+            goto error;
```
garlick, line 285: The reference to the object created here is overwritten by `pack_bank_info_object()` and leaked.  This should be dropped.

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,195 @@ int check_queue_factor (flux_plugin_t *p,
...
+{
+    json_t *user = json_object (); // JSON object for one user
+    json_t *userid, *banks = NULL;
+
+    if (!user)
+        return NULL;
+
+    userid = json_integer (u.first);
+    if (!userid)
+        return NULL;
```
garlick, line 320: This error path leaks `user`.  Use `goto error` here.

garlick (commented):
Getting very close, but there remain a few things to fix.

cmoussa1:
Thanks @garlick! I just squashed and force pushed based on your suggestions above

src/plugins/mf_priority.cpp
```diff
@@ -198,12 +198,186 @@ int check_queue_factor (flux_plugin_t *p,
...
+    flux_t *h = flux_jobtap_get_flux (p);
+    json_t *all_users = json_array (); // array of user/bank combos
+
+    if (!all_users)
+        return -1;
+
+    for (auto const &u : users) {
+        json_t *user = user_to_json (p, u);
+        if (user == NULL)
+            return -1;
```
garlick, line 357: Need to `json_decref (all_users)` before returning -1 here, or it leaks.

garlick (approved):
Just one more memory leak and LGTM!  I'll approve and you can set MWP once you've addressed that.

grondo:
Probably should change the PR title prefix from `cmd:` to `plugin:`?

cmoussa1:
> Probably should change the PR title prefix from `cmd:` to `plugin:`?

Totally, thanks for pointing this out. I forgot to change it after flux-framework/flux-core#4507.

> Just one more memory leak and LGTM! I'll approve and you can set MWP once you've addressed that.

Thanks for taking the time to help review this PR! Really appreciate all the feedback and suggestions - I just pushed up a fix to address the above suggestion, and will set MWP shortly.

cmoussa1:
Ah, perhaps this hasn't landed yet because of your requested changes @grondo? Is it OK with you if I remove you as a reviewer? Or you could take a look and give your approval too. :-)

grondo (approved):
Just tested this out and LGTM! Nice work! Ran the test with `FLUX_TEST_VALGRIND=t` and no leaks or errors detected.

grondo:
Weird, some old comments popped up when I added that approval comment. Those were stale from a previous review so can be ignored.

PR #266, opened 2022-09-01
t1017: condense `update-db` tests with old DBs

### LINKED ISSUE #260: t1017: improve sharness test by using for-loop to iterate through DB versions
@grondo made a good suggestion in #258 to improve the tests introduced in that PR to instead use a `for` loop to avoid using a variable per DB version. Then, we could just drop a new database into `expected/test_dbs` for to be checked in the test file. An example was provided in flux-core's `t/t2219-job-manager-restart.t`. 

https://github.com/flux-framework/flux-core/blob/3c367a3a15e888857cab2e8dbf2704e97861a494/t/t2219-job-manager-restart.t#L57-L66

cmoussa1:
#### Background

`t1017-update-db.t` has separate tests that run the `update-db` command on previous versions of flux-accounting DBs starting with `v0.10.0`. They are all separate tests, but they run the same set of commands and it really lengthens the amount of lines in the test file.

---

This is a small PR that just condenses all of the tests that run the `update-db` command on old versions of flux-accounting databases into a for-loop that iterates through the directory containing the DBs and updates them. I looked to https://github.com/flux-framework/flux-core/blob/3c367a3a15e888857cab2e8dbf2704e97861a494/t/t2219-job-manager-restart.t#L57-L66 for some guidance on how to do this. :-)

Fixes #260

### DIFF
--- a/t/t1017-update-db.t
+++ b/t/t1017-update-db.t
@@ -5,17 +5,9 @@ test_description='Test flux account-update-db command'
 
 DB_PATHv1=$(pwd)/FluxAccountingTestv1.db
 DB_PATHv2=$(pwd)/FluxAccountingTestv2.db
-DB_v0_10_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-10-0.db
-DB_v0_11_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-11-0.db
-DB_v0_12_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-12-0.db
-DB_v0_13_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-13-0.db
-DB_v0_14_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-14-0.db
-DB_v0_15_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-15-0.db
-DB_v0_16_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-16-0.db
-DB_v0_17_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-17-0.db
-DB_v0_18_0=${SHARNESS_TEST_SRCDIR}/expected/test_dbs/FluxAccountingv0-18-0.db
 MODIFY_DB=${SHARNESS_TEST_SRCDIR}/scripts/modify_accounting_db.py
 CHECK_TABLES=${SHARNESS_TEST_SRCDIR}/scripts/check_db_info.py
+DB_INTEGRITY_CHECK=${SHARNESS_TEST_SRCDIR}/scripts/db_integrity_check.py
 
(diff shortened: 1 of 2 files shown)


t/t1017-update-db.t
```diff
@@ -96,70 +87,15 @@ test_expect_success 'get all the columns from the queue_table and make sure the
...
-	rm tmp_v0_16_0.db &&
-	rm tmp_v0_17_0.db &&
-	rm tmp_v0_18_0.db
-'
+# test updating flux-accounting databases from older versions,
+# starting with v0.10.0
+for db in ${SHARNESS_TEST_SRCDIR}/expected/test_dbs/*; do
+	if [[ $db == *FluxAccounting* ]]; then
+		cp $db $(basename $db | cut -d '.' -f 1)_tmp.db
+		tmp_db=$(basename $db | cut -d '.' -f 1)_tmp.db
```
grondo, line 95: Set `tmp_db` first then you can use that variable in the `cp`:
```suggestion
		tmp_db=$(basename $db | cut -d '.' -f 1)_tmp.db
		cp $db $tmp_db
```

t/t1017-update-db.t
```diff
@@ -96,70 +87,15 @@ test_expect_success 'get all the columns from the queue_table and make sure the
...
+		chmod +rw $tmp_db
+		test_expect_success 'update old DB: '$(basename $db) "flux account-update-db -p $tmp_db"
```
grondo, line 97: One thought would be to test the updated db for validity after the update? (This wasn't in the original test so it is optional)

grondo (approved):
LGTM! Just a couple optional comments inline.

t/t1017-update-db.t
cmoussa1: Oh, duh! I don't know why I didn't do this. Thanks for pointing this out, I'll push up a fix.

t/t1017-update-db.t
cmoussa1: Yeah, I think that would be a great next step. I wonder how I might be able to do this - maybe try to add some users and banks and make sure the operations are successful?

cmoussa1: I just squashed and pushed up a fix that runs both `add-bank` and `add-user` commands after updating each DB to make sure the command runs successfully. Let me know if that was along the lines of what you were thinking - I know you already approved but I just want to make sure!

cmoussa1: OK, after some digging, I also found some basic [integrity/validity checks](https://www.sqlite.org/pragma.html#pragma_integrity_check) that SQLite performs on the DB using a pragma. It looks for:

- Table or index entries that are out of sequence
- Misformatted records
- Missing pages
- Missing or surplus index entries
- `UNIQUE`, `CHECK`, and `NOT NULL` constraint errors

Since the test DB's only have a couple of entries in them, the check performs quickly and the result is now checked in the for-loop to make sure that the check is `"ok"`. I've squashed and pushed that up as well.

grondo: That's cool! Nice work.

grondo: Feel free to set MWP when you are ready, already approved :+1:

cmoussa1: Thanks! Will do

cmoussa1:
Thanks again for the review @grondo! Setting MWP

PR #267, opened 2022-09-06
python: add additional exception messages to python commands

cmoussa1:
#### Problem

Often times when looking at the error message of one of the Python commands, the message is unclear or does not tell the whole story of what went wrong.

---

This is a small PR that adds additional exception messages to be printed to `STDOUT` when an exception occurs when running one of the flux-accounting Python commands to provide more clarity on what error actually occurred.


grondo (approved):
This looks like an improvement to me. :+1:

cmoussa1:
Thanks! I'll set MWP on this then. Probably would be good to have for our next release :-)

PR #268, opened 2022-09-06
t: add `v0.19.0` DB to the test DBs directory

cmoussa1:
This is a small PR that adds a sample flux-accounting DB from version `v0.19.0` to the directory of test DBs that are updated in `t1017-update-db.t`.


grondo (approved):
:+1:

cmoussa1:
Thanks for reviewing! I'll set MWP

PR #269, opened 2022-09-07
plugin: change dependency message for running jobs limit

cmoussa1:
#### Problem

The "max-jobs-limit" dependency that is added on active jobs when a user hits their `max_running_jobs` limit is not totally clear on why their submitted job is stuck in PD state.

---

This is a small PR that changes the dependency name to `"max-running-jobs-limit"` to be clear on what limit was hit and enforced on the submitted job in the first place.


garlick (approved):
Ah good idea.

This is good but would it be even clearer if it were `max-running-jobs-user-limit`?  (So that people know it's the number of jobs they have running, not the total jobs on the system?)

cmoussa1:
I like it! Thanks for this suggestion @garlick. I will make the change and force a push.

cmoussa1:
OK, I'll set MWP on this now. Thanks for the quick review! :-)

Issue #262, opened 2022-08-15
all jobs stuck in DEPEND:max-jobs-limit on fluke


grondo:
All jobs for a couple of the `testq*` users are stuck in DEPEND state with `max-jobs-limit`. Since there are no jobs running, this must be a bug:

```
 grondo@fluke108:~$ flux jobs -c 25 -Ao "{id.f58:>16} {username:<8} {state:<8} {dependencies}"
           JOBID USER     STATE    DEPENDENCIES
     ƒ6ATTD8ExHm testqd   DEPEND   max-jobs-limit
     ƒ6ATTJScNuV testqd   DEPEND   max-jobs-limit
     ƒ6ATTPkyoXD testqd   DEPEND   max-jobs-limit
     ƒ6ATTV3sErb testqd   DEPEND   max-jobs-limit
     ƒ6ATTaNEfUK testqd   DEPEND   max-jobs-limit
     ƒ6ATTfde7XM testqd   DEPEND   max-jobs-limit
     ƒ6ATTkvXYrj testqd   DEPEND   max-jobs-limit
     ƒ6ATTrDQzC7 testqd   DEPEND   max-jobs-limit
     ƒ6ATTwWJRXV testqd   DEPEND   max-jobs-limit
     ƒ6ATU2oBrrs testqd   DEPEND   max-jobs-limit
     ƒ6ATU865JCF testqd   DEPEND   max-jobs-limit
     ƒ6ATUDNxjXd testqd   DEPEND   max-jobs-limit
     ƒ6ATUJeNBaf testqd   DEPEND   max-jobs-limit
     ƒ6ATUPwFcv3 testqd   DEPEND   max-jobs-limit
     ƒ6ATUVH72p7 testqd   DEPEND   max-jobs-limit
     ƒ6ATUappLxw testqd   DEPEND   max-jobs-limit
     ƒ6ATUgNXf7m testqd   DEPEND   max-jobs-limit
     ƒ6ATUmsGzhu testqd   DEPEND   max-jobs-limit
     ƒ6Mmwz6sB6w testqc   DEPEND   max-jobs-limit
     ƒ6Mmx5PkcSK testqc   DEPEND   max-jobs-limit
     ƒ6MmxAge3mh testqc   DEPEND   max-jobs-limit
     ƒ6MmxFvZWYP testqc   DEPEND   max-jobs-limit
     ƒ6MmxPG3xRM testqd   DEPEND   max-jobs-limit
[...]

PR #273, opened 2022-09-15
[WIP] plugin: rework `inactive_cb ()`, `priority_cb ()`, and more

cmoussa1:
#### Problem

As described in issue #262, a number of jobs were stuck in `DEPEND` state on fluke due to a `max-running-jobs-limit`: a limit that determines how many _running_ jobs a user can have at any given time. While digging through the plugin and trying to come up with a reproducer, a number of sub-issues popped up that could be a cause of this issue. 

##### sub-issue 1

When a job held in `DEPEND` state transitions to inactive before running, it still decrements the `cur_running_job` count for the user/bank combo, which results in incorrect job counts.

##### sub-issue 2

When a job held in `DEPEND` state transitions to inactive before running and there is more than one held job, it still releases a currently held job even if there are already a max number of running jobs already running.

##### sub-issue 3

When a job held in `DEPEND state` transitions to inactive before running and is not the first held job in the user/bank's list of held jobs, it still releases the first held job even if there are already a max number of running jobs already running.

The changes proposed in this PR look to fix some of the issues reported on fluke about jobs being stuck in `DEPEND` state without having any running jobs to correctly release them.

---

The bulk of this PR re-works the `inactive_cb ()` to check the job ID of the job that transitioned to inactive. It checks that the `held_jobs` list is greater than 0, and if it is, looks for the job ID in the list of held jobs. If it finds itself in the list of held jobs, it only decrements the `cur_active_jobs` count, removes itself from the list of held jobs, and `return`s. Otherwise, it determines that this job transitioned from `RUN` state, so it decrements both `cur_active_jobs` and `cur_run_jobs`.

Another addition to the plugin is checking for both `cur_run_jobs` and `cur_active_jobs` in `priority_cb ()` (when a job enters `job.state.priority`. These checks will enforce limits on jobs that have been submitted by a user/bank combo whose information has not yet been sent from the flux-accounting database to the plugin's internal map. When the user/bank's information has been sent in an update and `reprioritize_all ()` is called on all pending jobs and each job re-enters `job.state.priority`, these limits are checked. If the limit has been hit, it will raise an exception on the job.

Some of the values of a user's "DNE" entry (created when a user/bank combo submits a job before their information has been loaded from the flux-accounting DB to the plugin) of the plugin's internal map have also been adjusted to hopefully better values, especially the `active` field, which should be set to 1 instead of 0.

As a result of some of the plugin rework, one section of the limits sharness test, `t1005-max-jobs-limit.t` has been reworked. One of the sections in the test increases the `max_active_jobs` count of a user and submits a max number of active jobs.
This incorrectly tests that the last submitted job receives a priority from the plugin, when it really should be held in `DEPEND` state. This PR changes the test that checks for an integer priority for the last submitted job to instead check that the job is in `DEPEND` state.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -940,6 +940,8 @@ static int inactive_cb (flux_plugin_t *p,
                         void *data)
 {
     int userid;
+    long int id;
+    long int jobid;
     struct bank_info *b;
     std::map<int, std::map<std::string, struct bank_info>>::iterator it;
     std::map<std::string, struct bank_info>::iterator bank_it;
@@ -947,7 +949,8 @@ static int inactive_cb (flux_plugin_t *p,
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:i}",
+                                "{s:I, s:i}",
+                                "id", &id,
                                 "userid", &userid) < 0) {
         flux_log (h,
                   LOG_ERR,
@@ -969,19 +972,44 @@ static int inactive_cb (flux_plugin_t *p,
         return -1;
     }
 
-    b->cur_run_jobs--;
-    b->cur_active_jobs--;
+    // if job has the same ID exists in held_jobs, only decrement
+    // cur_active_jobs, remove the ID from held_jobs list, and return
+    // since this job was never in RUN state, and therefore should not
+    // remove a dependency on another held job
+    if (b->held_jobs.size () > 0) {
+        if (std::count (b->held_jobs.begin (), b->held_jobs.end (), id)) {
+            // the job ID exists in held_jobs, so we only need to decrement
+            // the active jobs count
+            b->cur_active_jobs--;
+            // remove job ID from held_jobs
+            remove (b->held_jobs.begin (), b->held_jobs.end (), id);
+
+            return 0;
+        }
+        else {
+            // the job ID transitioned from RUN state, so we need to decrement
+            // both the running jobs count and the active jobs count
+            b->cur_run_jobs--;
+            b->cur_active_jobs--;
+        }
+    } else {
+        // the job has transitioned from RUN state, so we need to decrement
+        // both the running jobs count and the active jobs count
+        b->cur_run_jobs--;
+        b->cur_active_jobs--;
+    }
 
     // if the user/bank combo has any currently held jobs and the user is now
     // under their max jobs limit, remove the dependency from first held job
     if ((b->held_jobs.size () > 0) && (b->cur_run_jobs < b->max_run_jobs)) {
-        long int jobid = b->held_jobs.front ();
+        jobid = b->held_jobs.front ();
 
         if (flux_jobtap_dependency_remove (p,
                                            jobid,
-                                           "max-running-jobs-user-limit") < 0)
+                                           "max-running-jobs-user-limit") < 0) {
             flux_jobtap_raise_exception (p, jobid, "mf_priority",
                                          0, "failed to remove job dependency");
+        }
 
         b->held_jobs.erase (b->held_jobs.begin ());
     }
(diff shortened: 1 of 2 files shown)


grondo:
I haven't looked at this yet, but two immediate things come to mind:

1. >  Otherwise, it determines that this job transitioned from RUN state, so it decrements both cur_active_jobs and cur_run_jobs.

    FYI [jobtap plugin arguments](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html#jobtap-plugin-arguments) include the previous job state as `prev_state` - I don't know if that would be useful here as a simpler way to trigger the lookup of a held job, or if it could just be used as a double-check.

2. Now that you know a few scenarios that could cause this behavior, were you able to craft a reproducer and see the same result we were seeing on fluke?

cmoussa1:
> FYI [jobtap plugin arguments](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man7/flux-jobtap-plugins.html#jobtap-plugin-arguments) include the previous job state as prev_state - I don't know if that would be useful here as a simpler way to trigger the lookup of a held job, or if it could just be used as a double-check.

Good to know, I can check this out.

> Now that you know a few scenarios that could cause this behavior, were you able to craft a reproducer and see the same result we were seeing on fluke?

I don't think I was able to craft an exact reproducer of the scenario mentioned in #262, but I _think_ the above scenarios could be classified as contributors to it. I'm not sure if it would be easy to test (both with the old plugin and the changed proposed in this PR) with the workflow that caused this issue in the first place in a local Docker container or something? 

One random thought just came to mind, though, and I'm not sure if this would be easy to test (or if it's even correct), but regardless I thought of the following:

What happens with a Flux system instance shutdown and/or restart with jobs that had were held in `DEPEND` state before the shutdown/restart? Do those jobs just remain in `DEPEND` state? What scenario I'm envisioning goes like this:

A user is running a bunch of jobs (some in `RUN` state, some in `DEPEND` state). The jobs in `DEPEND` state were depending on the jobs in `RUN` state to transition to `job.state.inactive` so that they could be released with `dependency_remove ()`. Before the jobs in `DEPEND` state get released, the system instance is restarted or shut down and the jobs in `DEPEND` state get loaded back in, but they never get released with `dependency_remove ()` because there are no currently running jobs to release them. Thus, they remain in `DEPEND` state forever, but new jobs can be submitted and things work as expected after the plugin gets loaded. 

This might also explain why we've only seen this bug creep up twice I think, both which have been around the time of a Flux upgrade on fluke. Maybe I am just rambling nonsense, but I thought I'd at least put this out there to see what you think (btw, know you're traveling today so no rush on a response!).

grondo:
> Before the jobs in DEPEND state get released, the system instance is restarted or shut down and the jobs in DEPEND state get loaded back in, but they never get released with dependency_remove () because there are no currently running jobs to release them. Thus, they remain in DEPEND state forever, but new jobs can be submitted and things work as expected once the plugin gets loaded.

Ah, you could be right! I thought we had a plan for how to handle jobs in DEPEND state on reload of a jobtap plugin, but I can't think of how precisely that was supposed to work. 

I would actually think that when the Flux instance is shutdown, the running jobs should get killed and proceed through `job.state.inactive`, so all jobs in DEPEND should be released. Perhaps we can test this and see if there is a bug here? Maybe the jobtap plugin callbacks aren't working at shutdown?

cmoussa1:
> I would actually think that when the Flux instance is shutdown, the running jobs should get killed and proceed through `job.state.inactive`, so all jobs in DEPEND should be released. Perhaps we can test this and see if there is a bug here? Maybe the jobtap plugin callbacks aren't working at shutdown?

That sounds like a good idea. Do you know if this can be done in a Docker container? Is there an example somewhere of shutting down or restarting a Flux instance with jobs in DEPEND state?

grondo:
> That sounds like a good idea. Do you know if this can be done in a Docker container? Is there an example somewhere of shutting down or restarting a Flux instance with jobs in DEPEND state?

There are system instance restart tests in flux-core, but I'm not sure if any restart the instance with jobs in DEPEND state.

cmoussa1:
I'm revisiting this PR after last month's release and I think it's gotten a little out of scope of the issue described in #262 and probably a little difficult to review. Nonetheless, I think there are some important fixes that should be considered for the plugin, so I think I'll close this PR, and open subsequent smaller PRs instead.

PR #274, opened 2022-09-20
flux-accounting: add DB schema version to flux-accounting DB

### LINKED ISSUE #270: database: add a schema version to flux-accounting database
#### Problem

When a new version of flux-accounting is released that changes the schema of an already-existing flux-accounting database on a system, many of the commands that interact with this database do not work because of inconsistent schemas between the commands and the already-existing database. Often times it results in a Python exception. Currently, it is required of the admin of the system to update the database with `flux account-update-db`, which will look at the already-existing database and add or remove any necessary tables or columns. Then, the Python commands that interact with this database should work.

It might be worthwhile to tie a schema version to the database, and if the version numbers are out of sync between the commands and the database, it should refuse to execute those commands and instead output a helpful message that the database needs to be updated before interacting with it.

I'll put some links that might be worth exploring:
- https://stackoverflow.com/questions/550662/database-schema-updates

cmoussa1:
This is a small PR that adds a schema version number to the flux-accounting database. This integer version number represents the version of the most current DB schema and will only change when the schema of the DB changes from one version to the next. It is defined with the other `fluxacct` package variables, `db_dir` and `db_path` in the project's `__init__.py.in` file.

A check has also been added to the flux accounting commands that looks at the version number of the database the command is trying to connect to. If the version number is less than the version number defined in `__init__.py.in`, a message will be output saying that the DB is out of date and should be updated with `flux account-update-db`. This command will now do two additional things:

1) if the DB is out of date, it will update the version number of the DB after it has completed updating all existing tables and columns, and/or

2) will add a version number to an existing flux-accounting DB that may not have a schema version number associated with it.

The `flux account-priority-update` command also now has a check to check the schema version of the flux-accounting DB. If it is out of date, it automatically calls flux `account-update-db` to update the database before sending the DB information to the multi-factor priority plugin. If for some reason the update fails, it prints out the exception message(s) from the update and aborts both the DB update and _before_ sending the information to the multi-factor priority plugin. 

Fixes #270

##### TODO

- add tests for checking DB schema version

### DIFF
--- a/src/bindings/python/fluxacct/accounting/__init__.py.in
+++ b/src/bindings/python/fluxacct/accounting/__init__.py.in
@@ -1,4 +1,5 @@
 db_dir = "@X_LOCALSTATEDIR@/lib/flux/"
 db_path = "@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db"
+db_schema_version = 20
 
-__all__ = ["db_dir", "db_path"]
+__all__ = ["db_dir", "db_path", "db_schema_version"]
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -16,6 +16,8 @@ import pathlib
 import math
 import time
 
+import fluxacct.accounting
+
 
 def add_usage_columns_to_table(
     conn, table_name, priority_usage_reset_period=None, priority_decay_half_life=None
@@ -81,6 +83,9 @@ def create_db(
     except sqlite3.OperationalError as exception:
         logging.error(exception)
         sys.exit(1)
+
+    # set version number of database
+    conn.execute("PRAGMA user_version = %d" % (fluxacct.accounting.db_schema_version))
 
     # Association Table
     logging.info("Creating association_table in DB...")
--- a/src/cmd/flux-account-export-db.py
+++ b/src/cmd/flux-account-export-db.py
@@ -42,6 +42,17 @@ def est_sqlite_conn(path):
         print(f"Exception: {exc}")
         sys.exit(1)
 
+    # check version of database; if not up to date, output message
+    # and exit
+    cur = conn.cursor()
+    cur.execute("PRAGMA user_version")
+    db_version = cur.fetchone()[0]
+    if db_version < fluxacct.accounting.db_schema_version:
+        print(
+            "flux-accounting database out of date; please update DB with 'flux account-update-db' before running commands"
+        )
+        sys.exit(1)
+
     return conn
 
 
--- a/src/cmd/flux-account-pop-db.py
+++ b/src/cmd/flux-account-pop-db.py
@@ -44,6 +44,17 @@ def est_sqlite_conn(path):
         print(f"Exception: {exc}")
         sys.exit(1)
 
+    # check version of database; if not up to date, output message
+    # and exit
+    cur = conn.cursor()
+    cur.execute("PRAGMA user_version")
+    db_version = cur.fetchone()[0]
+    if db_version < fluxacct.accounting.db_schema_version:
+        print(
+            "flux-accounting database out of date; please update DB with 'flux account-update-db' before running commands"
+        )
+        sys.exit(1)
+
     return conn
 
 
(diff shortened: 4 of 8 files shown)


garlick:
Quick thought:  If we can just have the script run automatically, that would be easier on the admins.

Ideally the script would work for any prior version (I think maybe you said that already was the case?)

cmoussa1:
> Quick thought: If we can just have the script run automatically, that would be easier on the admins.

Ah, like, it could call `flux account-priority-update` automatically when it detects an out-of-date DB?

garlick:
Well I guess I didn't entirely think that comment through.  It would be nice if, after a software upgrade when Flux is restarted, if the database somehow was automatically upgraded before `mf_priority` starts handling jobs.   Any thoughts on how that could be done?

cmoussa1:
> Well I guess I didn't entirely think that comment through. It would be nice if, after a software upgrade when Flux is restarted, if the database somehow was automatically upgraded before `mf_priority` starts handling jobs. Any thoughts on how that could be done?

Hm, well, the command that sends data from the flux-accounting DB to the `mf_priority` plugin (`flux account-priority-update`) would also check the schema version of the DB _before_ executing the command (i.e sending the database data to the plugin). 

So maybe I can add some functionality to that command (`flux account-priority-update`) which after an upgrade or a restart, checks the version of the DB, and if it determines that the DB is out of date, automatically upgrade the DB (`with flux account-update-db`) before sending the information. This way, an admin does not have to remember to update the DB themselves while at the same time populating the plugin so that users can submit jobs. Would this be something along the lines of what you were thinking? Sorry if I'm off the mark here.

garlick:
Yeah that sounds right to me!  I was forgetting how the various scripts were related.

What would happen if the update failed for some reason?  (What would be the symptom and how would an admin debug?)

cmoussa1:
> What would happen if the update failed for some reason? (What would be the symptom and how would an admin debug?)

If an update were to fail, any upgrades made to the existing DB _do not_ get made to the DB and the operation is aborted, leaving it in the state that it were in before the upgrade. A specific sqlite3 exception message should also be output with an explanation as to why it failed. An admin could then look at the exception message and decide what appropriate action needs to take place, depending on what error occurred.

garlick:
OK.  Just wanted to walk through that.    We should make sure that `flux account-priority-update` fails when the upgrade script it called fails, and that the logs end up somewhere that a sys admin can find them (I presume `flux dmesg` since the script is run by flux's cron?)

Then my presumption is if `flux account-priority-update` fails, jobs would be stuck in PRIORITY which tends to be a good clue that flux accounting is not letting things run.

cmoussa1:
> Then my presumption is if `flux account-priority-update` fails, jobs would be stuck in PRIORITY which tends to be a good clue that flux accounting is not letting things run.

Right. I think that makes sense. Thanks for talking through that with me! I'll go ahead and work on adding some upgrade capability into `flux account-priority-update` to attempt to upgrade the DB when it finds that it's out of date.

cmoussa1:
OK, I think I'll mark this as ready for review now. I went ahead and added a check in `flux-account-priority-update.py` that checks the schema version of the DB, and if it detects that it is out of date, it will call `flux account-update-db` to update the DB before attempting to send data to the multi-factor priority plugin. It calls `flux account-update-db` using Python's [subprocess](https://docs.python.org/3/library/subprocess.html) module. 

If the update were to fail, any in-progress changes to the DB should be aborted and both the update and the priority-update calls should print an exception message and then exit.

ryanday36 (approved):
Looks good. There's one super minor thing that you might want to fix before merging, in your comment for the db version check, you have:
```
# check version of database; if not up to database, output message
```
which I assume should be 'if not up to date, output message'.

cmoussa1:
Whoops, haha, thanks for catching that @ryanday36. I'll force push a fix to correct that comment.

garlick:
I've been a little slow to try this out and may not get to it today.  If you want to merge based on @ryanday36 's review, I'm good with that.  I can open a new issue if I find anything amiss.

cmoussa1:
No problem, thanks for letting me know @garlick. I'll go ahead and remove you as a reviewer for this PR and set MWP shortly. Please feel free to let me know if things look wonky whenever you take a look at things on your system.

Issue #276, opened 2022-09-28
`add-user` does not validate bank passed in when adding a user


cmoussa1:
#### Problem

The `add-user` command does not validate the bank specified when adding a user to the flux-accounting database. So, for example, if the current bank structure looks like the following:

```
root
  A
  B
  C
```

A user an still be added to a bank called `foo`. Some validation should be added to the `add_user()` function to make sure that the bank that the user is being added to exists.

cmoussa1:
Oops, this was never closed as a result of #277 landing. Closing this now.

PR #277, opened 2022-09-28
add-user: add bank validation

cmoussa1:
#### Problem

As noted in #276, the `add-user` command does not validate the bank specified when adding a user to the flux-accounting database. So, a user can be added to a bank that does not exist in the `bank_table`.

---

This is a small PR that adds a validation helper function that looks at the bank being specified when adding a user. If the bank does not exist in the `bank_table`, it raises a ValueError with a message saying that the bank does not exist in the database, and does not add the user.

Both a unit test and sharness test are added to ensure that a non-existent bank in fact does raise a `ValueError` and prevents a user from being added to the flux-accounting database.


ryanday36 (approved):
Looks good. Thanks @cmoussa1!

cmoussa1:
Thanks @ryanday36! I'll set MWP on this

PR #278, opened 2022-09-30
testsuite: adjust queue tests for flux-core changes

garlick:
Problem: flux-accounting's `t1013-mf-priority-queues.t` test was failing in CI for flux-framework/flux-core#4627.

The job manager is made more picky about the queue config by that PR.  It's no longer possible to submit a job for a queue that is not in a `[queues]` configuration,  as the job manager now rejects it at submission.

There was also a failing test which checked that flux-accounting assigned a default queue if one was configured.  This is now handled at ingest by the "frobnicator" so flux-accounting should not need to assign defaults.  I removed the test (and another related one that was already commented out).  I'll open another issue to track changes that might be needed in flux-accounting for these queue changes.

There are a couple of minor cleanups here as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -93,7 +93,6 @@ adl_RECURSIVE_EVAL([$localstatedir], acct_db_path)
 AS_VAR_SET(acct_db_path, $acct_db_path)
 AC_SUBST(acct_db_path)
 
-AC_CONFIG_MACRO_DIR([config])
 X_AC_EXPAND_INSTALL_DIRS
 
 fluxplugin_ldflags="-avoid-version -export-symbols-regex '^flux_plugin_init\$\$' --disable-static -shared -export-dynamic"
--- a/t/sharness.d/flux-sharness.sh
+++ b/t/sharness.d/flux-sharness.sh
@@ -34,6 +34,29 @@ test_size_large() {
 
 
 #
+#  Like test_must_fail(), but additionally allow process to be
+#   terminated by SIGKILL or SIGTERM
+#
+test_must_fail_or_be_terminated() {
+    "$@"
+    exit_code=$?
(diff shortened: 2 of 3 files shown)


garlick:
Oh @cmoussa1 - just realized after your notes in #279 that maybe I misread that "default" test that I removed.  It isn't assigning the jobspec queue attribute (it can't do that can it).  It's just looking up a default priority value since no queue was specified.

If that's correct then I should leave that test in place and tweak the config  so that we can submit a job without a queue in that test.

cmoussa1:
> Oh @cmoussa1 - just realized after your notes in https://github.com/flux-framework/flux-accounting/issues/279 that maybe I misread that "default" test that I removed. It isn't assigning the jobspec queue attribute (it can't do that can it). It's just looking up a default priority value since no queue was specified.

Ah, yes, you are right. That test is just ensuring that if a user does not specify a queue, it uses the `default` queue's associated priority when calculating the priority of the job. I don't believe it assigns a jobspec queue attribute.

garlick:
I had sort of botched this the first time around because I got confused about which version of flux-core I was testing against.  It was working fine because I wast testing against flux-core master, not the proposed changes in flux-framework/flux-core#4627.

It turns out I needed to update `t/sharness.d/flux-sharness.sh` from flux-core so that `flux-start` options can be used on the `test_under_flux` command line.

So I just forced a push with the following
- update `flux-sharness.sh`
- drop the commit removing the "default" test mentioned above
- configure TOML queues _after_ the default test (since it wants to submit a job without a queue)

Apologies for the extra noise!

cmoussa1 (approved):
This LGTM! Thanks for putting this together @garlick.

garlick:
Thanks!  I'll set MWP.

PR #281, opened 2022-10-03
plugin: disable queue validation on an unknown queue, no configured "default" queue

### LINKED ISSUE #280: plugin: disable queue validation, add default priority when queue cannot be found
As a result of some good discussion in #279 and during today's coffee call, I think some cleanup of the multi-factor priority plugin could be had. Specifically, the multi-factor priority plugin should drop the validation of a queue if it is passed in when an association submits a job (i.e it should **not** reject jobs when a queue is specified that does not exist in the `queue_table` in the flux-accounting DB).

Instead, if it finds a queue that it does not know about, it should use a default integer priority when calculating the priority for a job.

---

Some further questions should still be considered (probably best suited in another future issue) about what to do in the case where a user submits a job under a queue to which they do not have access to (e.g somebody trying to run a job in a high priority queue when they are not supposed to!). The multi-factor priority plugin currently **rejects** jobs when a user submits a job to a queue they do not have access to.

Since flux-accounting and the plugin hold the information about which associations belong to which queue(s), it relies on an admin to keep track of who belongs to what queues in the flux-accounting database **as well as** any queue configuration in flux-core. This has the potential to get out of sync and could cause some unwanted behavior, so we should try to consider all possibilities and look for a good solution here.

cmoussa1:
#### Background

The multi-factor priority plugin currently rejects jobs which submit a queue that the plugin does
 not know about and when the plugin does not have a configured "default" queue. The changes proposed in https://github.com/flux-framework/flux-core/pull/4627
 enable the flux-core job-manager to be more aware of queues and instill queue validation on its own.
 As a result of this, the plugin should relax some of its current enforcements
 on queue validation.

---

This PR relaxes some of its queue validation in 2 cases when an association submits a job:

- when it does not recognize the queue specified
- when no queue is specified

In both of these cases, the plugin assign's a default factor of 0 to the overall job priority calculation. It changes a couple of the preprocessor directives the plugin was using to account for these validation changes. However, it is worth noting that the plugin still has the ability to reject jobs when an association submits a job to a queue they do not belong to.

For example, let's say that flux-accounting knows of 3 queues: `bronze`, `silver`, and `gold` and an association has access to both the `bronze` and `silver` queues. If the association submits a job under the `gold` queue, the job will be rejected. If the same association submits a job under any other named queue that flux-accounting does **not** know about (e.g a `special` queue), the plugin will assign a queue factor of 0, not positively or negatively affecting the priority of a job.

It also removes the "default" queue entry in the `queues` map of the plugin, since it no longer needs a "default" queue for jobs to be submitted under.

As a result of the changes made to the plugin, a couple of changes also needed to be made to the sharness tests that deal with job rejection based on queue validation.

Fixes #280

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -29,10 +29,21 @@ extern "C" {
 #include <vector>
 #include <sstream>
 
+// the plugin does not know about the association who submitted a job and will
+// assign default values to the association until it receives information from
+// flux-accounting
 #define BANK_INFO_MISSING -9
-#define NO_SUCH_QUEUE -5
+
+// a queue is specified for a submitted job that flux-accounting does not know
+// about
+#define UNKNOWN_QUEUE 0
+
+// no queue is specified for a submitted job
+#define NO_QUEUE_SPECIFIED 0
+
+// a queue was specified for a submitted job that flux-accounting knows about and
+// the association does not have permission to run jobs under
 #define INVALID_QUEUE -6
-#define NO_DEFAULT_QUEUE -7
 
 std::map<int, std::map<std::string, struct bank_info>> users;
 std::map<std::string, struct queue_info> queues;
@@ -126,13 +137,16 @@ static int get_queue_info (
 {
     std::map<std::string, struct queue_info>::iterator q_it;
 
-    // make sure that if a queue is passed in, it 1) exists, and 2) is a valid
-    // queue for the user to run jobs in
+    // make sure that if a queue is passed in, it is a valid queue for the
+    // user to run jobs in
     if (queue != NULL) {
-        // check #1) the queue passed in exists in the queues map
+        // check #1) the queue passed in exists in the queues map;
+        // if the queue cannot be found, this means that flux-accounting
+        // does not know about the queue, and thus should return a default
+        // factor
         q_it = queues.find (queue);
         if (q_it == queues.end ())
-            return NO_SUCH_QUEUE;
+            return UNKNOWN_QUEUE;
 
         // check #2) the queue passed in is a valid option to pass for user
         std::vector<std::string>::iterator vect_it;
@@ -145,13 +159,8 @@ static int get_queue_info (
             // add priority associated with the passed in queue to bank_info
             return queues[queue].priority;
     } else {
-        // no queue was specified, so use default queue and associated priority
-        q_it = queues.find ("default");
-
-        if (q_it == queues.end ())
-            return NO_DEFAULT_QUEUE;
-        else
-            return queues["default"].priority;
+        // no queue was specified, so just use a default queue factor
+        return NO_QUEUE_SPECIFIED;
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -30,9 +30,9 @@ extern "C" {
 #include <sstream>
 
 #define BANK_INFO_MISSING -9
-#define NO_SUCH_QUEUE -5
+#define UNKNOWN_QUEUE 0
+#define NO_QUEUE_SPECIFIED 0
```
grondo, line 34: Commit message nit: The commit message says:
> Change the preprocessor directive INVALID_QUEUE to UNKNOWN_QUEUE

But it _appears_ that here `INVALID_QUEUE` is still defined.

Also:

> Change the preprocessor directive NO_DEFAULT_QUEUE to UNSPECIFIED_QUEUE,

I don't see a `UNSPECIFIED_QUEUE`, but there is a `NO_QUEUE_SPECIFIED`...

t/t1018-mf-priority-disable-entry.t
```diff
@@ -44,7 +44,7 @@ test_expect_success 'submit a job successfully under default bank' '
 	jobid1=$(flux mini submit -n1 hostname) &&
 	flux job wait-event -f json $jobid1 priority | jq '.context.priority' > job1.test &&
 	cat <<-EOF >job1.expected &&
-	10050000
+	50000
```
grondo, line 47: It might be useful to state in the commit message why the priority changes are a result of the changes in validating queues since, at least to me, that is not an obvious effect. Though please correct me if this should be clear.

grondo (approved):
LGTM besides a minor nit and a question about testing.

src/plugins/mf_priority.cpp
cmoussa1: Whoops, I totally messed that up, didn't I? I meant to update this commit message before posting the PR but completely forgot. I'll update the commit message to use both `UNKNOWN_QUEUE` (instead of `INVALID_QUEUE`) and `NO_QUEUE_SPECIFIED` (instead of `UNSPECIFIED_QUEUE`). Sorry for the confusion!

t/t1018-mf-priority-disable-entry.t
cmoussa1: Yeah, looking back I realize this isn't as obvious as I initially thought, sorry about that.

The reason the priority changes for these jobs is because the queue factor is `UNSPECIFIED_QUEUE`, which is 0. This changes the resulting integer priority of the job from `1005000` to just `50000`, as before, the multi-factor priority plugin had a `"default"` queue entry (and a factor of `1000`) when an association submitted jobs under a default queue (like in the test cases above). I should make note of this in the commit message, and will push up a fix.

src/plugins/mf_priority.cpp
grondo: While you are in there, it might be good to add comments on each of these #defines to say more of what each means. Not critical, but if you are modifying the commit anwyay...

cmoussa1: Yes, good suggestion. I will do that!

cmoussa1:
Thanks for reviewing and approving @grondo! I've force-pushed a couple of fixes based on your suggestions to both fix the commit messages and add some comments in the plugin to describe the preprocessor directives. I'll go ahead and set MWP shortly!

PR #286, opened 2022-10-04
plugin: change default values of "DNE" entry to allow multiple jobs to be submitted

### LINKED ISSUE #283: mf_priority has inconsistent response to unknown user/bank
Problem: when mf_priority encounters a job for which it has no user/bank info, it accepts the first job submission and rejects subsequent submissions, e.g.
```
 garlick@picl0:~$ flux mini submit -q debug hostname
ƒBLUHbdBZks
 garlick@picl0:~$ flux mini submit -q debug hostname
[Errno 22] user/bank entry has been disabled from flux-accounting DB
garlick@picl0:~$ flux jobs --format=deps
       JOBID QUEUE    NAME       URG PRI          STATE    DEPENDENCIES
 ƒBLUHbdBZks debug    hostname   16               PRIORITY 
```

The two job submissions should probably get the same result.

cmoussa1:
#### Problem

As noted in #283, when the multi-factor priority plugin encounters a job for which it has no user/bank info, it allows one job to be submitted, but not subsequently submitted jobs. This is because it **does not** set the `active` field (which has a default value of 0) for the "DNE" entry it creates for the user/bank to during the first job submission, which means that any following submitted job would be met with a rejection message explaining that the user is disabled from the database:

```
[Errno 22] user/bank entry has been disabled from flux-accounting DB
```

---

This PR changes some of the default values of the "DNE" entry created for a user/bank when they submit jobs while their information is not loaded from the flux-accounting database. It sets `active` to 1 to prevent them from having their submitted jobs rejected, and changes their job count limits, `max_running_jobs` and `max_active_jobs` to more reasonable values to allow them to submit multiple jobs (999 and 1000, respectively).

A couple of new tests are also added to `t1014-mf-priority-dne.t` to make sure that a user/bank can submit multiple jobs (in this test case, 3 jobs) without having their information loaded to the plugin. The test makes sure that the jobs can be submitted and that they are held in PRIORITY state.

Fixes #283

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -32,7 +32,7 @@ extern "C" {
 // the plugin does not know about the association who submitted a job and will
 // assign default values to the association until it receives information from
 // flux-accounting
-#define BANK_INFO_MISSING -9
+#define BANK_INFO_MISSING 999
 
 // a queue is specified for a submitted job that flux-accounting does not know
 // about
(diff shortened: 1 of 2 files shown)


garlick (approved):
Tested and now the behavior is consistent (jobs get stuck in PRIORITY) so good!

cmoussa1:
Woohoo! Thanks for testing this and for reviewing @garlick. I'll set MWP on this

PR #288, opened 2022-10-05
`update-db`: use tempfile module for temporary flux-accounting DB during upgrade

### LINKED ISSUE #285: flux account-update-db: ERROR:root:unable to open database file
Problem:  `flux account-update-db` makes a temporary copy of the database in the current working directory, which may fail depending on the credentials the command is run with, and the working directory of the user.
```
garlick@picl0:~$ sudo -u flux flux account-update-db
ERROR:root:unable to open database file
```
This snippet of strace demonstrates what happened when I tried to run the command as the `flux` user from my home directory:
```
openat(AT_FDCWD, "/var/lib/flux/FluxAccounting.db", O_RDWR|O_NOFOLLOW|O_CLOEXEC) = 3
fstat(3, {st_mode=S_IFREG|0644, st_size=49152, ...}) = 0
fstat(3, {st_mode=S_IFREG|0644, st_size=49152, ...}) = 0
newfstatat(AT_FDCWD, "/var/lib/flux/FluxAccounting.db", {st_mode=S_IFREG|0644, st_size=49152, ...}, 0) = 0
pread64(3, "SQLite format 3\0\20\0\1\1\0@  \0\0\0\21\0\0\0\f"..., 100, 0) = 100
mkdirat(AT_FDCWD, ".", 0777)            = -1 EEXIST (File exists)
newfstatat(AT_FDCWD, ".", {st_mode=S_IFDIR|0755, st_size=4096, ...}, 0) = 0
newfstatat(AT_FDCWD, "tmpFluxAccounting.db", 0x7fd7f524e8, AT_SYMLINK_NOFOLLOW) = -1 ENOENT (No such file or directory)
getcwd("/nfshome/garlick", 511)         = 17
getpid()                                = 103233
newfstatat(AT_FDCWD, "/nfshome/garlick/tmpFluxAccounting.db", 0x7fd7f520d8, 0) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/nfshome/garlick/tmpFluxAccounting.db", O_RDWR|O_CREAT|O_NOFOLLOW|O_CLOEXEC, 0644) = -1 EACCES (Permission denied)
openat(AT_FDCWD, "/nfshome/garlick/tmpFluxAccounting.db", O_RDONLY|O_NOFOLLOW|O_CLOEXEC) = -1 ENOENT (No such file or directory)
getpid()                                = 103233
write(2, "ERROR:root:unable to open databa"..., 40ERROR:root:unable to open database file
) = 40
fstat(3, {st_mode=S_IFREG|0644, st_size=49152, ...}) = 0
newfstatat(AT_FDCWD, "/var/lib/flux/FluxAccounting.db", {st_mode=S_IFREG|0644, st_size=49152, ...}, 0) = 0
close(3)                             
```

cmoussa1:
#### Problem

As noted in #285, the `update-db` command fails during an automatic update because it is trying to create a temporary flux-accounting database in the broker's current working directory, which is `/`. The `update-db` command's behavior creates this temporary database in the current working directory, which causes problems if the directory does not have the correct permissions.

---

This PR uses the `tempfile` module to create a temporary directory inside `/tmp/` to store the temporary flux-accounting DB that gets created during an update. After the update is completed, the directory is cleaned up and removed. This should prevent the temporary database from attempting to be created wherever the command is running from, which might not have the necessary permissions.

Fixes #285

### DIFF
--- a/src/cmd/flux-account-update-db.py
+++ b/src/cmd/flux-account-update-db.py
@@ -13,6 +13,8 @@ import argparse
 import os
 import sqlite3
 import sys
+import tempfile
+import shutil
 
 from argparse import RawDescriptionHelpFormatter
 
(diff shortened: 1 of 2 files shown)


garlick:
Would it be possible to add a test that changes the working directory to / (guaranteed not writable) and ensures the command works?

src/cmd/flux-account-update-db.py
```diff
@@ -212,40 +213,41 @@ def update_db(path, new_db):
...
+            # close connections to DB's and remove temporary database
+            old_conn.close()
+            new_conn.close()
     except sqlite3.OperationalError as exc:
         print(f"Unable to open temporary database file: {new_db}")
         print(f"Exception: {exc}")
         sys.exit(1)
     except sqlite3.IntegrityError as exc:
         print(f"Exception: {exc}")
-        os.remove(new_db)
```
garlick, line 248: The temporary directory should also be removed, not just the db file, and preferably on all exit paths.

cmoussa1: Oh, yes, thank you for pointing this out. I just force-pushed a fix to call to remove the directory (and its files, which should just be the temporary database) on a Python Exception. When `update-db` succeeds, the directory and files are cleaned up automatically upon completion (I'm looking at [this](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) explanation on the Python docs).

cmoussa1:
> Would it be possible to add a test that changes the working directory to / (guaranteed not writable) and ensures the command works?

I think so! I just force-pushed an additional commit to add a small test that calls `update-db` from the `/` directory on one of the old databases. Let me know if that was something along the lines of what you were looking for.

garlick (approved):
Looks good, although the first commit message should say "Fixes #285".
The test is exactly what I was thinking, thanks.

cmoussa1:
Ah, yes you are absolutely right. I'll edit the commit message to include "Fixes #285" and force-push, and then set MWP. Thanks!

PR #290, opened 2022-10-11
[WIP] plugin: improve removing dependencies from held jobs, rework decrementing job counts

cmoussa1:
#### Background

While working on #262, I noticed a couple of glaring issues in the way the multi-factor priority plugin handles jobs transitioning to INACTIVE and removing dependencies from held jobs. I think there are three issues with how the plugin currently handles inactive jobs:

1) When a job held in `DEPEND` state transitions to `INACTIVE` before running, it still decrements the `cur_running_job` count for the user/bank combo, which results in incorrect job counts.

2) When a job held in `DEPEND` state transitions to `INACTIVE` before running and there is one or more held jobs, it still releases a currently held job even if there are already a max number of running jobs already running.

3) When a job held in `DEPEND` state transitions to `INACTIVE` before running and is not the first held job in the user/bank's list of held jobs, it still releases a currently held job even if there are already a max number of running jobs already running.

---

This PR slightly re-works the `inactive_cb ()` function to better handle jobs that transition to `INACTIVE` either from `RUN` state or straight from `DEPEND`. The `inactive_cb ()` now checks the job ID of the job that transitioned to `INACTIVE`. It checks that the `held_jobs` list is greater than 0, and if it is, looks for the job ID in the list of held jobs. If it finds itself in the list of held
jobs, it only decrements the `cur_active_jobs` count, removes itself from the list of held jobs, and `return`. Otherwise, we know that the job transitioned from `RUN`, so the plugin needs to decrement both `cur_active_jobs` and `cur_run_jobs`.

The changes made to the plugin also require a tiny rework of a couple of the tests in `t1005-max-jobs-limit.t` that were incorrectly testing that the last submitted job received an integer priority, when it really should be held in `DEPEND`, so a couple of the tests were reworked to make sure that the job is in fact in `DEPEND`.


cmoussa1:
I think this can also be closed now because of #325, which introduced a better version of the changes proposed here.

Issue #293, opened 2022-10-18
multi-factor priority plugin: associate jobs with projects


cmoussa1:
It was mentioned briefly in #247 to add "project" support to both the flux-accounting database and the multi-factor priority plugin. The first part is completed (by PR #255); now it is time to work on the latter. To recap, here is a quick summary of the functionality introduced by #255:

There now exists `projects` and `default_project` column for every association in the `association_table` where projects can be listed for associations to submit jobs under. Below is a quick example:

```console
$ flux account add-user --username=user1234 --bank=bank_A --projects="project_A,project_B,project_C"
```

`projectA` will become the association's default project (since it was the first project listed). A user's default project can also be edited with the `edit-user` command:

```console
$ flux account edit-user user1234 --default_project=project_B
```

If an association is added to the `association_table` without specifying any projects, a default project (`*`) is added for the association automatically. Every association who is added to the `association_table` belongs to the `*` project, but will only have their default project set to `*` if they do not already have another default project.

The latter half of adding basic project support is to be able to communicate this information to the multi-factor priority plugin, where it can be validated and set during job submission.

---

#### The Requirement

It was mentioned some in discussion in #255 about the need to associate jobs with projects that have arbitrary start and end dates. It will be calculated from archived jobs or something along those lines. At the very least, there is a requirement to associate jobs in the job-archive DB with a project name.

#### The Implementation

It should be fairly straightforward to send project data (both the projects that each association belongs to as well as the projects listed in the `project_table`) to the multi-factor priority plugin, where it can be unpacked and stored in an internal data structure. From there, jobs submitted with a project name can have it validated in `job.validate`, and jobs that specify an invalid project name _or_ a project they do not belong to will be rejected with an appropriate rejection message.

Jobs that do successfully associate with a project, on the other hand, will need to keep that data somewhere in the job record. What comes to mind at first is keeping this data in jobspec. If a project is specified on the command line, for example, then it can be found in the `attributes.system` section of the jobspec:

```json
"atttributes": {
  "project": "projectA"
}
```

But if a project is not specified on the command line (e.g an association is using their default project), then I think the plugin should be able to annotate the jobspec with the project name. Similar to the idea brought up in #272, perhaps I can use `jobspec-update` to add the project name to the jobspec. I played with this some yesterday afternoon, and I _believe_ I was able to successfully post an update, as shown from a job's eventlog:

```json
[...]

PR #294, opened 2022-10-21
plugin: add project validation and annotation

cmoussa1:
#### Background

Mentioned in #293, there is a requirement to be able to associate jobs with "projects," which allow multiple users who belong in different banks to submit jobs under one entity. Currently, the flux-accounting database supports adding projects and adding associations to certain projects, but this information is not relayed to the multi-factor priority plugin, so there is no way to associate a job with a project.

---

This PR adds project support in the priority plugin. The two big components that make up this PR are 1) providing some validation for projects if one is specified during `job.validate`, and 2) updating the jobspec with the project name in the PRIORITY event of a job.

Projects (along with the default project) for each association are now sent via `flux account-priority-update` to the plugin's internal map, where they are unpacked and stored in each association's `bank_info` object. The project information for each association is also added to the information returned in the `plugin.query` callback.

When an association specifies a project during job submission with `--setattr=system.project=my_project`, the project name is checked to both exist and be a valid project for the user to submit jobs under. Failure in either of these checks results in a job rejection with an appropriate message as to why the job was rejected.

If a user/bank submits a job under a default project, a `jobspec-update` event is posted under `attributes.system.project` in the PRIORITY event to add the user/bank's default project. The project name can then be found in the job's eventlog:

```json
{"timestamp": 1666106606.9670131,"name":"jobspec-update","context":{"attributes.system.project":"my_default_project"}}
```

### DIFF
--- a/src/cmd/flux-account-priority-update.py
+++ b/src/cmd/flux-account-priority-update.py
@@ -71,12 +71,14 @@ def bulk_update(path):
     data = {}
     bulk_user_data = []
     bulk_q_data = []
+    bulk_p_data = []
 
     # fetch all rows from association_table (will print out tuples)
     for row in cur.execute(
         """SELECT userid, bank, default_bank,
            fairshare, max_running_jobs, max_active_jobs,
-           queues, active FROM association_table"""
+           queues, active, projects, default_project
+           FROM association_table"""
     ):
         # create a JSON payload with the results of the query
         single_user_data = {
@@ -88,6 +90,8 @@ def bulk_update(path):
             "max_active_jobs": int(row[5]),
             "queues": str(row[6]),
             "active": int(row[7]),
+            "projects": str(row[8]),
+            "def_project": str(row[9]),
         }
         bulk_user_data.append(single_user_data)
 
@@ -110,6 +114,18 @@ def bulk_update(path):
     data = {"data": bulk_q_data}
 
     flux.Flux().rpc("job-manager.mf_priority.rec_q_update", json.dumps(data)).get()
+
+    # fetch all rows from project_table
+    for row in cur.execute("SELECT project FROM project_table"):
+        # create a JSON payload with the results of the query
+        single_p_data = {
+            "project": str(row[0]),
+        }
+        bulk_p_data.append(single_p_data)
+
+    data = {"data": bulk_p_data}
+
+    flux.Flux().rpc("job-manager.mf_priority.rec_project_update", data).get()
 
     flux.Flux().rpc("job-manager.mf_priority.reprioritize")
 
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -754,17 +754,19 @@ static int priority_cb (flux_plugin_t *p,
     int urgency, userid;
     char *bank = NULL;
     char *queue = NULL;
+    char *project = NULL;
     int64_t priority;
     struct bank_info *b;
 
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:i, s:i, s{s{s{s?s, s?s}}}}",
+                                "{s:i, s:i, s{s{s{s?s, s?s, s?s}}}}",
                                 "urgency", &urgency,
                                 "userid", &userid,
                                 "jobspec", "attributes", "system",
-                                "bank", &bank, "queue", &queue) < 0) {
+                                "bank", &bank, "queue", &queue,
+                                "project", &project) < 0) {
         flux_log (h,
                   LOG_ERR,
                   "flux_plugin_arg_unpack: %s",
@@ -848,6 +850,21 @@ static int priority_cb (flux_plugin_t *p,
                                              "failed to update jobspec "
                                              "with bank name");
                 return -1;
+            }
+
+            // if the user/bank is submitting their job under a default
+            // project, then we need to update that too
+            if (project == NULL) {
+                if (update_jobspec_project (p, userid, bank) < 0) {
+                    flux_jobtap_raise_exception (p,
+                                                 FLUX_JOBTAP_CURRENT_JOB,
+                                                 "mf_priority",
+                                                 0,
+                                                 "failed to update jobspec "
+                                                 "with project name");
+
+                    return -1;
+                }
             }
         }
     }
@@ -1023,6 +1040,7 @@ static int new_cb (flux_plugin_t *p,
     int userid;
     char *bank = NULL;
     char *queue = NULL;
+    char *project = NULL;
     int max_run_jobs, cur_active_jobs, max_active_jobs = 0;
     double fairshare = 0.0;
     struct bank_info *b;
@@ -1033,10 +1051,11 @@ static int new_cb (flux_plugin_t *p,
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:i, s{s{s{s?s, s?s}}}}",
+                                "{s:i, s{s{s{s?s, s?s, s?s}}}}",
                                 "userid", &userid,
                                 "jobspec", "attributes", "system",
-                                "bank", &bank, "queue", &queue) < 0) {
+                                "bank", &bank, "queue", &queue,
+                                "project", &project) < 0) {
         return flux_jobtap_reject_job (p, args, "unable to unpack bank arg");
     }
 
@@ -1128,6 +1147,21 @@ static int new_cb (flux_plugin_t *p,
                                  b,
                                  NULL) < 0)
         flux_log_error (h, "flux_jobtap_job_aux_set");
+
+    if (project == NULL) {
+        // using a default project, so add it to main eventlog
+        // via jobspec-update
+        if (update_jobspec_project (p, userid, bank) < 0) {
+            flux_jobtap_raise_exception (p,
+                                         FLUX_JOBTAP_CURRENT_JOB,
+                                         "mf_priority",
+                                         0,
+                                         "failed to update "
+                                         "jobspec with project name");
+
+                return -1;
+        }
+    }
 
     b->cur_active_jobs++;
 
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -183,15 +183,17 @@ static int get_queue_info (
 }
 
 
-static void split_string (char *queues, struct bank_info *b)
+static void split_string_and_push_back (char *list,
+                                        std::vector<std::string> *vec)
 {
     std::stringstream s_stream;
 
-    s_stream << queues; // create string stream from string
+    s_stream << list; // create string stream from string
+
     while (s_stream.good ()) {
         std::string substr;
         getline (s_stream, substr, ','); // get string delimited by comma
-        b->queues.push_back (substr);
+        vec->push_back (substr);
     }
 }
 
@@ -531,7 +533,7 @@ static void rec_update_cb (flux_t *h,
 
         // split queues comma-delimited string and add it to b->queues vector
         b->queues.clear ();
-        split_string (queues, b);
+        split_string_and_push_back (queues, &b->queues);
 
         users_def_bank[uid] = def_bank;
     }
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -462,6 +462,36 @@ static bank_info_result get_bank_info (int userid, char *bank)
 }
 
 
+static int update_jobspec_project (flux_plugin_t *p, int userid, char *bank)
+{
+    std::string project;
+
+    // look up user/bank info based on unpacked information
+    bank_info_result lookup_result = get_bank_info (userid, bank);
+
+    if (lookup_result.first == BANK_USER_NOT_FOUND
+        || lookup_result.first == BANK_INVALID
+        || lookup_result.first == BANK_NO_DEFAULT)
+        return -1;
+
+    // set bank_info to the struct that holds the user/bank information
+    // and get user's default project
+    auto bank_info = lookup_result.second->second;
+    project = bank_info.def_project;
+
+    if (!project.empty ()) {
+        // post jobspec-update event
+        if (flux_jobtap_jobspec_update_pack (p,
+                                             "{s:s}",
+                                             "attributes.system.project",
+                                             project.c_str ()) < 0)
+            return -1;
+    }
+
+    return 0;
+}
+
+
 /******************************************************************************
  *                                                                            *
  *                               Callbacks                                    *
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -45,6 +45,14 @@ extern "C" {
 // the association does not have permission to run jobs under
 #define INVALID_QUEUE -6
 
+// a project was specified for a submitted job that flux-accounting knows about
+// and the association does not have permission to submit jobs under
+#define INVALID_PROJECT -6
+
+// a project is specified for a submitted job that flux-accounting does not
+// know about
+#define UNKNOWN_PROJECT -7
+
 // different codes to return as a result of looking up user/bank information:
 //
 // BANK_SUCCESS: we found an entry for the passed-in user/bank
@@ -213,6 +221,30 @@ int check_queue_factor (flux_plugin_t *p,
                                      prefix, queue);
         return -1;
     }
+
+    return 0;
+}
+
+
+int validate_project (std::vector<std::string> user_projects, char *project)
+{
+    std::vector<std::string>::iterator vect_it;
+
+    // make sure that if a project is passed in, it exists in flux-accounting's
+    // "projects" vector
+    vect_it = std::find (projects.begin (), projects.end (), project);
+
+    if (vect_it == projects.end ())
+        return UNKNOWN_PROJECT;
+
+    // make sure that if a project is passed in, it is a valid project for the
+    // user to submit jobs under
+    vect_it = std::find (user_projects.begin (),
+                         user_projects.end (),
+                         project);
+
+    if (vect_it == user_projects.end ())
+        return INVALID_PROJECT;
 
     return 0;
 }
@@ -847,6 +879,7 @@ static int validate_cb (flux_plugin_t *p,
     char *queue = NULL;
     flux_job_state_t state;
     int max_run_jobs, cur_active_jobs, max_active_jobs, queue_factor = 0;
+    char *project = NULL;
     double fairshare = 0.0;
     bool only_dne_data;
 
@@ -857,11 +890,12 @@ static int validate_cb (flux_plugin_t *p,
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:i, s:i, s{s{s{s?s, s?s}}}}",
+                                "{s:i, s:i, s{s{s{s?s, s?s, s?s}}}}",
                                 "userid", &userid,
                                 "state", &state,
                                 "jobspec", "attributes", "system",
-                                "bank", &bank, "queue", &queue) < 0) {
+                                "bank", &bank, "queue", &queue,
+                                "project", &project) < 0) {
         return flux_jobtap_reject_job (p, args, "unable to unpack bank arg");
     }
 
@@ -916,6 +950,22 @@ static int validate_cb (flux_plugin_t *p,
     if (queue_factor == INVALID_QUEUE)
         return flux_jobtap_reject_job (p, args, "Queue not valid for user: %s",
                                        queue);
+
+    // validate project passed in (if any); a job can be rejected if the
+    // following conditions are met:
+    //   - an association submits a job under a project that does not exist
+    //   - an association submits a job under a project they do not belong to
+    if (project != NULL) {
+        int result = validate_project (bank_it->second.projects, project);
+
+        if (result == UNKNOWN_PROJECT)
+            return flux_jobtap_reject_job (p, args, "project does not exist: "
+                                           "%s", project);
+
+        if (result == INVALID_PROJECT)
+            return flux_jobtap_reject_job (p, args, "project not valid for "
+                                           "user: %s", project);
+    }
 
     max_run_jobs = bank_it->second.max_run_jobs;
     fairshare = bank_it->second.fairshare;
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -62,6 +62,7 @@ typedef std::pair<bank_info_codes, std::map<std::string, struct bank_info>::iter
 
 std::map<int, std::map<std::string, struct bank_info>> users;
 std::map<std::string, struct queue_info> queues;
+std::vector<std::string> projects;
 std::map<int, std::string> users_def_bank;
 
 struct bank_info {
@@ -74,6 +75,8 @@ struct bank_info {
     std::vector<std::string> queues;
     int queue_factor;
     int active;
+    std::vector<std::string> projects;
+    std::string def_project;
 };
 
 // min_nodes_per_job, max_nodes_per_job, and max_time_per_job are not
@@ -485,7 +488,7 @@ static void rec_update_cb (flux_t *h,
                            const flux_msg_t *msg,
                            void *arg)
 {
-    char *bank, *def_bank, *queues = NULL;
+    char *bank, *def_bank, *queues, *projects, *def_project = NULL;
     int uid, max_running_jobs, max_active_jobs = 0;
     double fshare = 0.0;
     json_t *data, *jtemp = NULL;
@@ -499,9 +502,6 @@ static void rec_update_cb (flux_t *h,
         goto error;
     }
 
-    if (flux_respond (h, msg, NULL) < 0)
-        flux_log_error (h, "flux_respond");
-
     if (!data || !json_is_array (data)) {
         flux_log (h, LOG_ERR, "mf_priority: invalid bulk_update payload");
         goto error;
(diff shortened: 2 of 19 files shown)


cmoussa1:
So that was odd! My `python-lint` check was failing with an error message saying it could not locate Python `3.6`. From some quick and rough Google searches it looks like it _might_ have something to do with the latest version of Ubuntu being used (I was using `ubuntu-latest`): https://github.com/actions/setup-python/issues/162. I updated the version number to match flux-core's Ubuntu version for the `python-lint` check, and now it works as excepted, so I think I'll move the most recent commit on this PR to another branch and post a separate, small fix PR.

garlick:
Just skimming a bit, but should the `jobspec-update` event only be posted when the project is not present in the jobspec already?

cmoussa1:
@garlick, ah, you're absolutely right. If a user specifies a project on the command line the project will be present in the jobspec already, so there won't be a need to add it manually via `jobspec-update`. It should only be added if a user doesn't specify one. Perhaps that can be done in `priority_cb ()` instead of in `validate_cb ()`. I'm about to push up some changes to this PR to only add a project via `jobspec-update` _only if_ if a user doesn't specify one on the command line.

cmoussa1:
OK, I've pushed up some changes to this PR to only post a `jobspec-update` if a user/bank submits a job under a default project (thanks for pointing that out yesterday!). With that, I'll open it up to some initial review whenever convenient. 😄

cmoussa1:
@grondo if you have some time later this week or next, could you give this PR a look?

src/plugins/mf_priority.cpp
```diff
@@ -774,6 +776,31 @@ static int priority_cb (flux_plugin_t *p,
...
+            }
+        }
+    }
+
+    // if using a default project, add it to jobspec
+    if (project == NULL) {
+        if (update_jobspec_project (p, userid, bank, project) < 0) {
+            flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                         "mf_priority", 0, "failed to update "
+                                         "jobspec with project name");
```
grondo, line 843: is there any way to combine this duplicated code? I might need help understanding why `update_jobspec_project()` needs to be called two different places like this here, sorry!

grondo (commented):
@cmoussa1 - I made a first pass at this PR. Sorry it took so long!

A few comments inline. I also had some general questions:

 - Is the use of projects now a requirement? What happens at a site where no projects are configured at all
 - Do we need to add `attributes.system.project` to the official list of system attributes in RFC 14?
 - Does job-list need to support `project` as a job attribute it can return (and so job-archive can store this attribute)

 Sorry if all this has been discussed before, I may need a refresher on the use case for job "project" attribute!

grondo:
> Perhaps that can be done in priority_cb () instead of in validate_cb (). I'm about to push up some changes to this PR to only add a project via jobspec-update only if if a user doesn't specify one on the command line.

A job could enter the PRIORITY state multiple times, so would this mean the `jobspec-update` could unnecessarily occur multiple times? Also, if you delay the `jobspec-update` until PRIORITY, then the job would be without an assigned project until it passes the dependency state, which might be unexpected.

We may want to think about how to issue the `jobspec-update` event during `job.validate` as it was before, perhaps after the job has been fully validated?

grondo:
Hm, I don't why the pr-validator thinks all your commits have empty bodies, but I'll look into it and get that action fixed.

src/cmd/flux-account-priority-update.py
```diff
@@ -111,6 +115,18 @@ def bulk_update(path):
...
+    for row in cur.execute("SELECT project FROM project_table"):
+        # create a JSON payload with the results of the query
+        single_p_data = {
+            "project": str(row[0]),
+        }
+        bulk_p_data.append(single_p_data)
+
+    data = {"data": bulk_p_data}
+
+    flux.Flux().rpc("job-manager.mf_priority.rec_p_update", json.dumps(data)).get()
```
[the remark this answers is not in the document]
cmoussa1: Ah, I did not know this, thanks for pointing this out! There are other parts in this script that also wrap `data` with `json.dumps()`. I should open a separate PR later that drops it.

src/plugins/mf_priority.cpp
```diff
@@ -168,15 +168,24 @@ static int get_queue_info (
 }
 
 
-static void split_string (char *queues, struct bank_info *b)
+static void split_string (char flag, char *list, struct bank_info *b)
```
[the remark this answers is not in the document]
cmoussa1: I think I see what you're saying! Adjusting the parameters of the function to just take a string to split on `,` and the vector to push the results cleans the code up quite considerably (no having to pass funky single-character flags); instead it can just look like this:

```
split_string_and_push_back (projects, &b->projects);
```

and the function only needs one `while` loop. Thanks for the suggestion!

I can make this change as well as add the callers of this function to the same commit (sorry about that!!).

src/plugins/mf_priority.cpp
```diff
@@ -1046,7 +1104,8 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
     if (flux_plugin_register (p, "mf_priority", tab) < 0
         || flux_jobtap_service_register (p, "rec_update", rec_update_cb, p)
         || flux_jobtap_service_register (p, "reprioritize", reprior_cb, p)
-        || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p) < 0)
+        || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p)
+        || flux_jobtap_service_register (p, "rec_p_update", rec_p_cb, p) < 0)
```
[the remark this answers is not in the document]
cmoussa1: Yeah absolutely, sorry for the confusion! I was trying to stay under the 80 character limit here but I'll it can be hard for me to look at too, so I should just break it up 👍 will fix this

src/cmd/flux-account-priority-update.py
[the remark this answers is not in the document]
grondo: Yeah, it doesn't cause a problem to call `json.dumps()` first, but it would probably be good to clean up in the future.

src/plugins/mf_priority.cpp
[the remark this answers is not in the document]
grondo: Ah! I can sympathize with that! However, no problem breaking the function across multiple lines!

t/t1015-mf-priority-urgency.t
```diff
@@ -46,12 +41,8 @@ test_expect_success 'create a group of users with the same fairshare value' '
 			}
 		]
 	}
-	flux.Flux().rpc("job-manager.mf_priority.rec_update", json.dumps(bulk_user_data)).get()
 	flux.Flux().rpc("job-manager.mf_priority.rec_q_update", json.dumps(bulk_queue_data)).get()
 	EOF
-'
-
-test_expect_success 'send the user information to the plugin' '
```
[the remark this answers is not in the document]
grondo: Ah, ok, thanks for the clarification. I figured everything was good since tests were passing, but just thought I'd check!

cmoussa1:
Oh great, thanks for letting me know! Once I address your suggestions in this PR I'll push them up and see if the check passes. I thought it could have been my fault as a result of landing #369 yesterday 😅 lol

src/plugins/mf_priority.cpp
cmoussa1: Ah, you're so right! Thanks for pointing this out. I don't know why I have this code duplicated; it doesn't need to be. I think I must have thought that if a job is submitted and held in PRIORITY because a user is waiting for their information to be added to the plugin, it required a special case in `priority_cb ()` to add the project name, but that's not the case. Sorry for the confusion!

I'll remove this duplicate code and add a test which shows that the project name gets updated in the case where a user's job is held in PRIORITY as the plugin waits for (and then receives) information from the flux-accounting DB.

cmoussa1:
> Is the use of projects now a requirement? What happens at a site where no projects are configured at all

The use of projects I _believe_ is a requirement for our site. If no projects are configured at all, then all jobs should just run under a default project of `*`.

> Do we need to add `attributes.system.project` to the official list of system attributes in RFC 14? Does job-list need to support project as a job attribute it can return (and so job-archive can store this attribute)
>
> Sorry if all this has been discussed before, I may need a refresher on the use case for job "project" attribute!

That's okay! I think @ryanday36's input would be useful for these too.

I'm not sure if it is necessary to add it to an official list of attributes, but I don't know how often users are added to a project and need to submit jobs under it; if it is very common, maybe it would be helpful to add it.

To my knowledge the usage reporting requirement for jobs under a specific project is not known ahead of time. Here's a comment from @ryanday36 I remembered from a while back on the PR where projects were first added to the flux-accounting DB:

> The trick is that I don't necessarily know ahead of time what filters we would want. I think that it's usually the usage for the past quarter, but sometimes it might be the usage for two quarters ago, and others it might be just the past month. I can talk to Greg T about whether there are any common things that would be worth keeping in the project table, but I don't see that as urgent. It's more important that jobs in the job archive DB have an appropriate project associated with them.

So, maybe not entirely necessary to also have it be a filter for `job-list`. If anything, perhaps it could be a filter for flux-accounting's job-archive (whenever that gets added)?

grondo:
> the update to the eventlog for the project name was moved to the end of the callback for job.state.run.

This means that jobs that never run (e.g. are canceled before they run) won't ever have a project assigned. This could be surprising since you've stated above that a project is required. 

> And in some offline conversation, it was mentioned that calling event_post_pack () while the job is in PRIORITY state could invalidate variables that are needed shortly thereafter to calculate the job's priority. My opinion is that it will be much clearer if the jobspec is updated at the end of `job.validate` or possibly  in `job.new` (which is only called for valid jobs)

To clarify, it is the `jobspec-update` event that could invalidate local variables _if_ they were assigned by unpacking directly from the `jobspec` member vi `flux_plugin_arg_unpack()`. This is because `const char *` variables point directly into the `jobspec_redacted` member of the job structure in the job-manager, so if that is updated in place, then the variables can end up pointing to garbage. The fix would be to either copy the variables or only updated the jobspec at the very end of the function.

I'm kind of wondering if the job-manager should make this easier on plugin authors and perhaps make an immutable copy of the `jobspec` while callbacks are running -- or somehow delay processing of `jobspec-update` events while a given plugin callback is running. I'll take a look and see if this is possible.

> Moving the jobspec-update to the job.state.run callback also allows for use of the suggested implementation for flux_jobtap_event_post_pack () in the review above.

Hm, that method of calling `post_pack()` should be usable from any of the callbacks. What was the reason it wasn't working before `job.state.run`?

src/plugins/mf_priority.cpp
```diff
@@ -343,6 +343,61 @@ static json_t *user_to_json (
...
+        return -1;
+    }
+
+    // post jobspec-update event
+    if (flux_jobtap_event_post_pack (p, FLUX_JOBTAP_CURRENT_JOB,
+                                         "jobspec-update", "O",
+                                         project_name) < 0) {
+        json_decref (project_name);
+        return -1;
+    }
```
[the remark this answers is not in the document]
grondo: @cmoussa1: D'oh! I gave bad advice here by not paying attention and forgetting how the `jobpsec-update` event works. Sorry! You can ignore this comment. My suggestion should have been:

```c
// post jobspec-update event
    if (flux_jobtap_event_post_pack (p,
                                     FLUX_JOBTAP_CURRENT_JOB,
                                     "jobspec-update",
                                     "{s:s}",
                                     "attributes.system.project", project) < 0) {
        return -1;
    }   
```

I apologize for the confusion :facepalm:

cmoussa1:
OK, I've just force-pushed up a fix after the new suggested format for `flux_jobtap_event_post_pack ()` which fixed the weird workaround I had proposed before (basically calling `jobspec-update` in the callback for `job.state.run`). In this most recent push, the project name gets added to the main eventlog in two cases:

1) A user/bank submits their job with a default project before any information is loaded to the plugin, in which case their job gets held in PRIORITY state. When the information is loaded and jobs are reprioritized, add the project name in the section of `job.state.priority` where the previously held job gets updated with the user/bank information.

2) A user/bank submits their job with a default project where user/bank information is already loaded to the plugin. Here, just add the project name once the job has been fully validated, i.e at the end of the callback for `job.validate`.

grondo:
@cmoussa1 FYI: A new function `flux_jobtap_jobspec_update_pack(3)` was just merged into flux-core if you'd like to update to use that. This would mean flux-accounting would require the latest flux-core, but that may be ok in this instance.

cmoussa1:
@grondo, yup, I just force-pushed to use the new `flux_jobtap_jobspec_update_pack ()` function after flux-framework/flux-core#5386!

grondo:
> @grondo, yup, I just force-pushed to use the new flux_jobtap_jobspec_update_pack ()

Oh, oops I missed that! :+1: Sorry!

cmoussa1:
No problem! It was good timing 😅 let me know if I am not using that function correctly if you get the chance.

grondo:
Looks correct to me! I'll give this PR another review now.

cmoussa1:
Made a slight change to this PR - I moved the `jobspec-update` of the project name to `job.new` instead of `job.validate`, since in #378, we've established that a job can now re-enter `job.validate` multiple times, and we don't want to add multiple updates of the same project name if a job gets updated.

src/plugins/mf_priority.cpp
```diff
@@ -266,27 +266,68 @@ static json_t *add_held_jobs (
...
+static json_t *add_projects (const std::pair<std::string, struct bank_info> &b)
+{
+    json_t *projects = NULL;
+
+    // add any projects to a JSON array
+    if (!(projects = json_array ()))
+        goto error;
+
+    for (auto const &j : b.second.projects) {
+    }
```
grondo, line 281: Should this empty loop be deleted?

src/plugins/mf_priority.cpp
```diff
@@ -266,27 +266,68 @@ static json_t *add_held_jobs (
...
                                  "max_run_jobs", b.second.max_run_jobs,
                                  "cur_run_jobs", b.second.cur_run_jobs,
                                  "max_active_jobs", b.second.max_active_jobs,
                                  "cur_active_jobs", b.second.cur_active_jobs,
                                  "held_jobs", held_jobs,
-                                 "active", b.second.active))) {
+                                 "active", b.second.active,
+                                 "projects", projects,
+                                 "def_project",
+                                 b.second.def_project.c_str ()))) {
```
grondo, line 330: It appears that `projects` is leaked on error here. (missing `json_decref (projects)` in `error:` label I think.

grondo (commented):
Just made another pass with some minor comments (and a couple comments that were sitting in the queue for week, sorry about that!)

Also, it appears there is still a case where the prioirty plugin updates the jobspec in the `job.priority` callback? Do you remember why this is so (I'm sorry if I suggested it and can't remember why). If there is a good reason, then a comment in the code there might be helpful.

src/plugins/mf_priority.cpp
```diff
@@ -437,10 +439,15 @@ static void rec_update_cb (flux_t *h,
         b->max_run_jobs = max_running_jobs;
         b->max_active_jobs = max_active_jobs;
         b->active = active;
+        b->def_project = def_project;
```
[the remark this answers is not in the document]
cmoussa1: Ah, yes, sorry for making that confusing. I should just split this into two commits with the format you suggested: one to refactor `split_string ()` and a second to use the new function to parse a `projects` string. Will fix this 👍

src/plugins/mf_priority.cpp
cmoussa1: yup, it should definitely be deleted 🤦 thanks for catching that, will fix

cmoussa1:
OK, I've just pushed some fixes based on the most recent set of suggestions - thanks! I've split the commits that deal with expanding `split_string ()` and making use of the helper function into two separate commits.

> Also, it appears there is still a case where the prioirty plugin updates the jobspec in the job.priority callback? Do you remember why this is so (I'm sorry if I suggested it and can't remember why). If there is a good reason, then a comment in the code there might be helpful.

Yeah, this was mistakenly left over because I failed to realize that even when a job is submitted under a default project _before_ any user/bank information is loaded to the plugin, the job will still enter `job.new` and have its jobspec updated with the project name, so I don't think there's any need to have that additional update. I've updated that commit to no longer have that update in `job.state.priority` as well as updated the commit message.

grondo:
>  I failed to realize that even when a job is submitted under a default project before any user/bank information is loaded to the plugin, the job will still enter job.new and have its jobspec updated with the project name

Sorry, I don't recall at the moment how this works, but how will the jobspec be updated with the project name in `job.new` if the priority plugin does not have any bank information yet? Maybe you _do_ also need the jobspec-update in `job.priority` to handle this case?

I could be wrong though, are there tests that exercise this case (jobs submitted with default bank/projects but the first upload to the priority plugin has not yet occurred?)

cmoussa1:
Yes, the first set of tests in `t1028-mf-priority-projects.t` should demonstrate this case:

```shell
test_expect_success 'submit a job under a user before plugin gets updated' '
	job0=$(flux python ${SUBMIT_AS} 1003 hostname) &&
	test $(flux jobs -no {state} ${job0}) = PRIORITY
'

test_expect_success 'add user to flux-accounting DB and to plugin; job transitions to RUN' '
	flux account -p ${DB_PATH} add-user --username=user1003 --userid=1003 --bank=account1 &&
	flux account-priority-update -p ${DB_PATH} &&
	test $(flux jobs -no {state} ${job0}) = RUN
'

test_expect_success 'check that project gets updated for submitted job' '
	flux job info $job0 eventlog > eventlog.out &&
	grep "{\"attributes.system.project\":\"\*\"}" eventlog.out &&
	flux job cancel $job0
'
```

In the above, a job gets submitted before any data is loaded to the plugin. The user is then added to the flux-accounting DB and the DB info is sent to the plugin, the job transitions to RUN, and the project name is checked in the job's main eventlog to look for the default project name.

t/t1028-mf-priority-projects.t
```diff
@@ -0,0 +1,140 @@
...
+	test_cmp job1.expected job1.test &&
+	flux job info $job1 jobspec > jobspec.out &&
+	grep "projectA" jobspec.out &&
+	flux job cancel $job1
+'
+
+test_expect_success 'submit a job under a project that does not exist' '
+	test_must_fail flux python ${SUBMIT_AS} 1001 --setattr=system.project=projectFOO \
+		hostname > project_dne.out 2>&1 &&
+	test_debug "project_dne.out" &&
```
grondo, line 83: Just noticed this while trying to run this test with `-d -v` -- `test_debug()` takes a command to run, not a file. Since there's no command `project_dne.out`, this line causes a failure. Change to 
```
test_debug "cat project_dne.out" &&
```
here and in the other use of `test_debug` in this sharness test. (Also perhaps audit the rest of the testsuite)

t/t1028-mf-priority-projects.t
```diff
@@ -0,0 +1,140 @@
...
+
+test_expect_success 'submit a job under a user before plugin gets updated' '
+	job0=$(flux python ${SUBMIT_AS} 1003 hostname) &&
+	test $(flux jobs -no {state} ${job0}) = PRIORITY
+'
+
+test_expect_success 'add user to flux-accounting DB and to plugin; job transitions to RUN' '
+	flux account -p ${DB_PATH} add-user --username=user1003 --userid=1003 --bank=account1 &&
+	flux account-priority-update -p ${DB_PATH} &&
+	test $(flux jobs -no {state} ${job0}) = RUN
```
grondo, line 51: There's a slight race condition here (and actually above in the check for PRIORITY) because of the eventual consistency properties of `job-list`. It might be safer to check the eventlog and ensure the expected _event_ has been posted.

E.g. instead of checking that the job is in the PRIORITY state above, check that the last event in the eventlog is `depend`.

and instead of checking for the `RUN` state here, wait for an `alloc` event...

grondo:
> The user is then added to the flux-accounting DB and the DB info is sent to the plugin, the job transitions to RUN, and the project name is checked in the job's main eventlog to look for the default project name.

Sorry, still getting clarification here!

In this test though the project name of `*` (is this and empty project name or something?) is added immediately when the job is submitted. When the accounting database is updated, the user still has no project information, so when the database info is sent to the priority plugin it doesn't need to update anything. What about users that _do_ have a default project, doesn't the plugin have to assign these default projects to pending jobs in PRIOIRITY after the first update?

I apologize if I'm still not getting it and I appreciate your patience!

cmoussa1:
No need to apologize, I think I am doing a poor job explaining the requirements here, so I'm sorry! Here is the requirement for how each job needs to be associated with a project, taken from conversation with @ryanday36 in #255 (feel free to correct me I am misunderstanding this Ryan):

---

If a user is added to the DB without specifying any projects, a default project `*` is added for the user automatically, and jobs submitted without specifying a project will fall under `*`; every user who is added to the DB belongs to the `*` project, but will only run jobs under `*`  if they do not already have another default project.

If a user is added to the DB with a specified project name, any job submitted without specifying a project name will fall under that project name.

---

Does that explanation make sense? So, I believe you are absolutely right, in the event where a user _does_ have a default project, the plugin has to assign this project to pending jobs while the plugin waits for flux-accounting information. I was able to confirm that this was not yet supported with the most current push of this branch by submitting a job under a user/bank before the plugin gets any data, adding the user with a specific default project, updating the plugin, and seeing that the project name for the user/bank _does not_ get updated to the correct default project.

To fix this, I can place a check in `job.state.priority` when the information for the job is being updated to reflect the loaded flux-accounting information to update the project name to the user/bank's default _if_ it is not `*`.

I also wrote a set of tests that submits two jobs from two different users before the plugin gets any information (so both jobs are pending). One user gets added with no projects specified, and the other user gets added with one project specified (which becomes their default). Once the plugin gets loaded with the flux-accounting DB information, the jobs get updated. The job from the user/bank with no projects specified keeps its project as `*`, but the job from the user/bank _with_ a specified default project gets updated to use the newly found default project name. I can add those tests to the sharness test file that's created in this PR as well as include a comment describing how the project assignment works in either case.

Let me know if this explanation does or does not make sense. I'll push up these changes (as well as the changes based on your feedback above) shortly.

grondo:
Ah, ok.  I guess my naive question would be why not just not set a project when no project is set instead of using `*`?

cmoussa1:
If I am understanding your question correctly, `*` is the name of the project that gets set when no project is set.

grondo:
Yeah, sorry I meant just don't set `attributes.system.project` when no project is set. This might reduce the number of necessary `jobspec-update` events.

cmoussa1:
Oh, whoops. I misunderstood your question, sorry. My understanding is that we want a default whether a user/bank has a default project or not. Here is a comment from @ryanday36 in #255 when projects were first being added to the flux-accounting DB:

> No, we want both kinds of defaults :) If no project is specified when the user is created, giving them '*' for their project and default project is what we want.

grondo:
Ok, thanks! I didn't read that issue carefully enough! :facepalm:

src/plugins/mf_priority.cpp
```diff
@@ -183,15 +194,17 @@ static int get_queue_info (
...
     std::stringstream s_stream;
 
-    s_stream << queues; // create string stream from string
+    s_stream << list; // create string stream from string
+
     while (s_stream.good ()) {
         std::string substr;
         getline (s_stream, substr, ','); // get string delimited by comma
-        b->queues.push_back (substr);
+            vec->push_back (substr);
```
grondo, line 207: Is there a stray indent here?

src/plugins/mf_priority.cpp
```diff
@@ -760,6 +948,8 @@ static void add_missing_bank_info (flux_plugin_t *p, flux_t *h, int userid)
     b->cur_active_jobs = 0;
     b->active = 1;
     b->held_jobs = std::vector<long int>();
+    b->projects = { "*" };
+    b->def_project = "*";
```
grondo, line 952: When the flux-accounting info is missing, shouldn't the project be treated like the bank info, i.e. be somehow recorded as unknown? That is, if the bank info is missing, we don't know if the default project should be `"*"` since the real default should be set in the accounting database. In fact, should the default project be something encoded in the database and left our of the plugin itself (that way other sites could make their own decision what a default project should be, or if there should even be a default)?

grondo (commented):
Just a couple other comments I had while looking at this again. Not a full review.

src/plugins/mf_priority.cpp
cmoussa1: That's a great question. I can change this assignment to be set to indicate that the project for the user/bank's job is unknown until it gets its information - just re-ran the tests after changing these values to something like `"DNE"`, which is the same as the bank information, and I think everything still works as expected. 👍 

In the user table in the flux-accounting DB, the default project for any user/bank when being added is `*` (if one is not specified at user creation time; the default project can be changed later if desired):

```python
"""
projects         tinytext    DEFAULT '*'            NOT NULL    ON CONFLICT REPLACE DEFAULT '*',
default_project  tinytext    DEFAULT '*'            NOT NULL    ON CONFLICT REPLACE DEFAULT '*',
"""
```

and `*` is added as a valid project name to the DB when the DB is first created:

```python
    conn.execute(
        """
            CREATE TABLE IF NOT EXISTS project_table (
                project_id          integer    PRIMARY KEY AUTOINCREMENT,
                project             tinytext               NOT NULL,
                usage               real       DEFAULT 0.0 NOT NULL
            );"""
    )
    conn.execute("INSERT INTO project_table (project) VALUES ('*')")
```

Since this is the case, maybe we don't even need to assign values for `project` and `def_project` here? Removing these assignments should also be fine (at least, the the testsuite passes). What do you think? Sorry if I am misunderstanding you.

grondo: Yes, that would look more correct to me. If the default is already set in the database, then the code should use that default and not a copy, since that might be redundant and also could cause the values to diverge if someone updates the database and forget to do the same in the plugin. This is why it is usually bad practice to copy a default value into more than one place. 

Of course, I'm not all that familiar with the code here, so maybe it made more sense in this situation, so I was just posing the question.

cmoussa1: just pushed up a change (really just dropped this commit) that added this assignment of a project/default project in the case where a user/bank submits a job before any information is loaded to the plugin.

cmoussa1:
I've been able to make some decent progress on adding an external class for user/bank (i.e accounting) information that I think would really clean up this PR, so I think I'll just close this and re-propose something else at a later date.

Issue #291, opened 2022-10-12
multi-factor priority plugin: add new "age" factor


cmoussa1:
Another factor that could be added to the multi-factor priority plugin is an "age" factor, or a factor that represents the amount of time that a job has been sitting in a queue waiting to run. The priority for a job, in general, should _increase_ the longer it sits in a queue eligible to run.

A first thought: Perhaps the `t_submit` field can be extracted when the job enters `job.validate` and be set along with the other necessary job information. Then, it can be subtracted from the current time when the priority is being calculated to determine an age, resulting in an age factor that could increase the priority of a job. The age factor should also probably be less than the fair-share factor. 

I think this would also work during a [`reprioritize`](https://github.com/flux-framework/flux-core/blob/1dcbcd97d974ba686b8514f714934093eb7c2162/doc/man7/flux-jobtap-plugins.rst), which calls `job.priority.get` for all jobs (by "all" jobs, I'm assuming all jobs that are not currently in `RUN` state). The priorities for these jobs would be updated to reflect how long they've been scheduled to run.

cmoussa1:
After playing around with this some yesterday and this morning, I think I am getting closer to a working solution. Since jobtap plugins allow `t_submit` to be extracted from a job, I can pull this out as a `double`. During the job priority calculation, `t_submit` can be subtracted from the current time to get an "age," which can be used to further increase the priority of a job, i.e the longer a job sits in a queue, the greater its "age" becomes, subsequently increasing its integer priority:

```
priority = round ((fshare_weight * fshare_factor) +
                  (queue_weight * queue_factor) +
                  (age_weight * age_factor) +
                  (urgency - 16));
```

cmoussa1:
As I think about how to test this, I am wondering how it might be possible to test for consistent priority values, which is what I have been doing currently. I guess I could change the sharness tests to just check that the job receives a priority value (and not compare to a specific value), and that's it. The age factor is sure to change the integer priorities that I currently test for, and I'm not sure I can guarantee it stay the same in between runs.

grondo:
Maybe you'll want to add some configuration that controls whether to include or exclude the `age` factor in priority. This would allow you to test this factor separately from the others. I'm not sure if the priority plugin has any configuration at the moment, so that might be a bigger project than it at first seems.

As a stopgap, you could add a test-only RPC to the `mf_priority` module that disables the age in the priority calculation for tests that currently assume non-time-based priorities (this would probably be the easiest solution).

[...]

PR #295, opened 2022-11-07
[WIP] plugin: add configurable priority factors via TOML `conf.update`

cmoussa1:
#### Background

While working on #291, I found it hard to add a new factor to the priority plugin without having to change all of the sharness tests that checked for a specific priority number. A great suggestion was made that perhaps the plugin could make use of the `conf.update` callback topic, which would allow for configuration of values in a TOML file that can be extracted in the plugin and stored for future use. This would be especially useful for the integer weights of the various factors used in the priority calculation of a job for a couple of reasons:

- it would make adding new factors easier so that existing tests do not have to be adjusted (since the new factor can be set to 0 in these tests).
- it would allow an admin to define and set the weights for the factors used in job priority calculation, adjusting for whatever is most appropriate for the center during use, instead of using hard-coded weights defined in the plugin.

---

This is a [WIP] PR that adds TOML configuration support for the priority factors to the multi-factor priority plugin. The `conf_update_cb ()` callback that is added to the plugin looks for a `[priority_factors]` section in the config file where the various priority factors are defined along with their associated integer weights. It extracts these values and places them in a map, where the key is the name of the factor and the value is its integer weight. These weights are then used in the `priority_calculation ()` helper function where an integer priority is calculated for a job. Previously, this helper function was using hard-coded values for the weights of each factor.

As part of this PR, a config update is added to the existing sharness tests that interact with the multi-factor priority plugin and check for job priorities. The weights used in the sample TOML files in the tests are the same weights that used to be defined in the plugin, so no changes were required in the integer priorities checked for in the tests.

Finally, a new sharness test, `t1020-mf-priority-config.t` is added, where a couple of tests are added to check that the priority of a job can be adjusted depending on the `[priority_factors]` configuration in the TOML file, as well as ensuring that no configuration of the priority factors (as well as using the default `urgency`) results in a job being held in the `PRIORITY` state.


cmoussa1:
I've continued to make progress on adding the "age" factor to the priority plugin, and I think it would be helpful to have this PR reviewed first as the "age" factor stuff is built on top of these commits. :-) @grondo, if you have a chance later this week or next, could you give this a review please?

src/plugins/mf_priority.cpp
```diff
@@ -170,6 +171,38 @@ static void add_special_association (flux_plugin_t *p, flux_t *h, int userid)
...
+                           void *data)
+{
+    int fshare_weight = -1, queue_weight = -1;
+    flux_t *h = flux_jobtap_get_flux (p);
+
+    // unpack the various factors to be used in job priority calculation
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s?{s?{s?i, s?i}}}",
+                                "conf", "priority_factors",
```
grondo, line 195: Suggestion: put all flux-accounting values under one heading, maybe `[accounting]`? 

Will all configurable priority factors actually be weights? If so, then maybe the sub-table should
just be `factor-weights` and the individual factor weights like `fairshare`, `queue`, `age`, etc.

So in the config file it would be
```toml
[accounting.factor-weights]
fairshare = 1
age = 1
```
and so on?

grondo (commented):
This looks pretty good! I just had a suggestion on the layout of the configuration to more closely match how other Flux subsystems are organized. Let me know what you think. Definitely open for discussion!

A `flux-config-accounting(5)` man page should probably be added as well documenting the new configuration parameters and expanded when new configuration values are added.

src/plugins/mf_priority.cpp
cmoussa1: That's a good suggestion, thanks! I've just pushed up my branch that makes this change to use a sub-table `[accounting.factor-weights]`.

cmoussa1:
Thanks for the review @grondo!

> A `flux-config-accounting(5)` man page should probably be added as well documenting the new configuration parameters and expanded when new configuration values are added.

Sure thing, I'd be happy to document this. Should I add that manpage to the `doc` folder in flux-accounting in this PR?

grondo:
> Sure thing, I'd be happy to document this. Should I add that manpage to the doc folder in flux-accounting in this PR?

Yeah, are there no manpages yet for flux-accounting?

We should start with this one. It should probably go in `/doc/man5/flux-config-accounting.rst`. You could probably borrow the setup from `flux-sched` which I think has the same structure.

cmoussa1:
Not yet, but I'll go ahead and take a crack at adding one for the config parameters proposed in this PR. 👍

cmoussa1:
I just pushed up a commit to add a new manpage to the `doc/` folder - let me know what you think and if you'd like to see anything else added!

grondo:
Great! Unfortunately I think some infrastructure is still needed to build and install the man page. I think you could check out flux-sched for the requirements, but I can try to help later

cmoussa1:
Thanks @grondo! I've started trying to copy over the infrastructure from flux-core and flux-sched to enable building and installing the man pages (I've pushed up another commit that you'll probably notice is failing 😉 ). I'm going to try and keep hacking away at this to see if I can get it to build successfully

cmoussa1:
OK, after a lot of tinkering I think I was finally able to get the man pages infrastructure set up. 😆 The last two commits should have all of the required files added and modified to be able to build and install the documentation for the priority plugin configuration, which can be viewed with `man flux-config-accounting`.

doc/man5/flux-config-accounting.rst
```diff
@@ -0,0 +1,36 @@
...
+=========================
+
+
+DESCRIPTION
+===========
+
+The flux-accounting priority plugin can be configured to assign different
+weights to the different factors used when calculating a job's priority.
+Assigning a higher weight to a factor will result in it having more
+influence in the calculated priority for a job.
```
grondo, line 12: Maybe this should mention the default weight for factors that are not set in the config file?

cmoussa1: Good suggestion, thanks. I've just force-pushed up a fix that adds a table which lists the default weights for factors if they are not configured.

grondo (approved):
Great! LGTM!

cmoussa1:
Thanks! I'll set MWP here then

PR #297, opened 2022-11-28
[WIP] plugin: add "age" factor to priority calculation

cmoussa1:
#### Background

As mentioned in #291, there is potential to add another factor to the priority calculation of a job that consists of a job's "age," or how long a job has been submitted and waiting to run. This will help further refine the priority of a job and can help increase a low priority job's position in the queue on a reprioritization of all jobs.

---

This is a [WIP] PR built on top of #295 that adds this new "age" factor to the priority calculation of a job. It unpacks `t_submit` from the job and subtracts it from the current time to calculate a job's age. This factor is then added to the other two factors that are currently used in the multi-factor priority calculation.

With the changes proposed in #295, it is possible to "disable" this factor (or any factor for that matter) by setting its weight to 0 in the TOML config file:

```toml
[priority_factors]
fshare_weight = 1000
queue_weight = 100
age_weight = 0
```

which is what I have done in the existing set of sharness tests, as to avoid having to adjust every test that checks for a specific priority number.

##### TODO

- add tests for age factor

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -28,6 +28,7 @@ extern "C" {
 #include <cinttypes>
 #include <vector>
 #include <sstream>
+#include <chrono>
 
 // the plugin does not know about the association who submitted a job and will
 // assign default values to the association until it receives information from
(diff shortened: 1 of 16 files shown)


cmoussa1:
I don't expect to get back to this until higher priority stuff has eventually landed in flux-accounting, but I wanted to jot some notes down from an offline discussion so that I don't forget about it when I circle back to this.

Something to keep in mind when considering the **age** factor is to calculated it from when a job is _released_, not when it is _submitted_, or in other words, consider only the time that the job was available to be scheduled but failed due to another constraint (like a resource constraint) that could be satisfied if it was the next job up in the queue.

This is to prevent users from purposely submitting held jobs for a long time and then getting a very large priority bump when they release their held job.

cmoussa1:
Going to close this PR and re-open a hopefully cleaner implementation. :-)

PR #299, opened 2022-11-30
edit-bank: add ability to edit parent bank

### LINKED ISSUE #298: edit-bank doesn't change parent
I was just asked to move an existing bank on corona under a new parent bank. It doesn't look like the `edit-bank` command can change anything other than shares, so I deleted the bank and re-created it under the new parent. I also had to delete and re-create all of the users of the bank. Can the `edit-bank` command be extended to moving banks to a new parent?

cmoussa1:
#### Problem

As described in #298, the parent bank of a bank cannot be edited, so in order to move a bank and its users, it must be deleted and re-created under the new parent bank. The `edit-bank` command should be extended to allow the parent bank to be edited.

---

This is a small PR that extends the `edit-bank` command to allow edits of a parent bank. This would remove the need to delete and then re-add a bank and all of its users under the new parent bank. 

A couple of unit test and sharness tests are also added to make sure it works as expected. The unit tests make sure that a bank's parent bank can be changed, and trying to edit a bank's parent bank to one that does not exist raises a `ValueError`. The sharness test adds a bank and some users before editing the parent bank to confirm that the users are also moved to the new parent bank in the hierarchy displayed by the `flux account-shares` command.

Fixes #298

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -199,22 +199,36 @@ def delete_bank(conn, bank):
     return 0
 
 
-def edit_bank(conn, bank, shares):
-    print(shares)
-    # if user tries to edit a shares value <= 0,
-    # raise an exception
-    if int(shares) <= 0:
-        raise ValueError("New shares amount must be >= 0")
-    try:
-        # edit value in bank_table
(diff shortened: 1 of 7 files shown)


ryanday36 (approved):
Looks good @cmoussa1. Thanks for the fast work!

PR #300, opened 2022-12-06
github: change Ubuntu version for python-lint check

cmoussa1:
#### Problem

Using ubuntu-latest in the python-lint causes the check to fail with an error message saying that Python `3.6` cannot be found.

---

This PR just switches the version of Ubuntu to match flux-core's Ubuntu version in their `python-lint` check: `20.04`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,7 @@ jobs:
 
   python-lint:
     name: python lint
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

cmoussa1:
Thanks @grondo! Will set MWP on this :-)

PR #301, opened 2022-12-09
plugin: record bank name to jobspec when user submits under default bank

### LINKED ISSUE #272: idea: record bank name in priority event
Problem: the job eventlog doesn't record the bank a job used when its priority was assigned, but this may be useful information, e.g. see flux-framework/flux-core#4569

Perhaps we could let the multifactor priority plugin record accounting-specific info, like the bank, in the [priority](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html#priority-event) event?

cmoussa1:
#### Problem

Described in #272, currently, when an association submits a job under their default bank, the bank name is not recorded anywhere in the jobspec. Only when an association specifies the bank they want to submit jobs under (i.e `flux mini submit --setattr=system.bank=my_bank`) does the bank name get recorded. It was proposed in that issue to record the bank name in the PRIORITY event of a job even when a user does not specify a bank when they submit a job (i.e they use their default bank).

---

This PR records the bank name in the jobspec during the PRIORITY event when an association submits a job under their default bank. It makes use of the `jobspec-update` event to post the `bank` key-value pair. This event takes place just before the integer priority for the job is calculated. This event should take place if the association was not known to the flux-accounting DB/priority plugin at the time of job submission; in the latter case, when all pending jobs are reprioritized after a plugin update, the default bank is looked up and then added to the jobspec. 

In the case where the default bank is known at the time of job submission, it adds the bank name via `jobspec-update` under `attributes.system.bank` during `job.new`.

Some sharness tests are also added to a new test file, `t1029-mf-priority-bank.t`, that test this functionality, making sure the bank name can be found in the eventlog of a job when jobs are submitted under a default bank as well as when an association does not yet exist in the flux-accounting DB.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -350,6 +350,34 @@ static bool check_map_for_dne_only ()
 }
 
 
+/*
+ * Update the jobspec with the default bank the association used to
+ * submit their job under.
+ */
+static int update_jobspec_bank (flux_plugin_t *p, int userid)
+{
+    char *bank = NULL;
+    std::map<int, std::map<std::string, struct bank_info>>::iterator it;
+
+    it = users.find (userid);
+    if (it == users.end ()) {
+        return -1;
+    }
+
+    // look up default bank
+    bank = const_cast<char*> (users_def_bank[userid].c_str ());
+
+    // post jobspec-update event
+    if (flux_jobtap_jobspec_update_pack (p,
+                                         "{s:s}",
+                                         "attributes.system.bank",
+                                         bank) < 0)
+        return -1;
+
+    return 0;
+}
+
+
 /******************************************************************************
  *                                                                            *
  *                               Callbacks                                    *
@@ -639,6 +667,17 @@ static int priority_cb (flux_plugin_t *p,
                                          &bank_it->second,
                                          NULL) < 0)
                 flux_log_error (h, "flux_jobtap_job_aux_set");
+
+            // now that we know the user/bank info associated with this job,
+            // we need to update jobspec with the default bank used to
+            // submit this job under
+            if (update_jobspec_bank (p, userid) < 0) {
+                flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                             "mf_priority", 0,
+                                             "failed to update jobspec "
+                                             "with bank name");
+                return -1;
+            }
         }
     }
 
@@ -851,6 +890,14 @@ static int new_cb (flux_plugin_t *p,
                                              "mf_priority", 0,
                                              "job.new: user/default bank "
                                              "entry does not exist");
+                return -1;
+            }
+            // update jobspec with default bank
+            if (update_jobspec_bank (p, userid) < 0) {
+                flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                            "mf_priority", 0,
+                                            "failed to update jobspec "
+                                            "with bank name");
                 return -1;
             }
         }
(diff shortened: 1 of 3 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -331,6 +331,46 @@ static json_t *user_to_json (
...
+        return -1;
+    }
+
+    // post jobspec-update event
+    if (flux_jobtap_event_post_pack (p, FLUX_JOBTAP_CURRENT_JOB,
+                                     "jobspec-update", "O",
+                                     bank_name) < 0) {
+        json_decref (bank_name);
+        return -1;
+    }
```
chu11, line 367: I think you can use the new flux_jobtap_jobspec_update_pack() now instead.

https://github.com/flux-framework/flux-core/pull/5386

t/t1028-mf-priority-bank.t
```diff
@@ -0,0 +1,97 @@
...
+'
+
+test_expect_success 'add a user to the DB' '
+	flux account -p ${DB_PATH} add-user --username=user1001 --userid=1001 --bank=account1 &&
+	flux account -p ${DB_PATH} add-user --username=user1001 --userid=1001 --bank=account2
+'
+
+test_expect_success 'submit a job before plugin has any flux-accounting information' '
+	job4=$(flux python ${SUBMIT_AS} 1002 hostname) &&
+	test $(flux jobs -no {state} ${job4}) = PRIORITY
```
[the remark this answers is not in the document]
cmoussa1: d'oh 🤦 you're right; i re-ordered these and didn't re-order them. Thanks for pointing this out! I just went ahead and removed the numbers altogether from the `jobid` variable.

t/t1029-mf-priority-default-bank.t
```diff
@@ -0,0 +1,98 @@
...
+'
+
+test_expect_success 'add a user to the DB' '
+	flux account -p ${DB_PATH} add-user --username=user1001 --userid=1001 --bank=account1 &&
+	flux account -p ${DB_PATH} add-user --username=user1001 --userid=1001 --bank=account2
+'
+
+test_expect_success 'submit a job before plugin has any flux-accounting information' '
+	jobid=$(flux python ${SUBMIT_AS} 1002 hostname) &&
+	flux job info $jobid eventlog > eventlog.out &&
```
grondo, line 45: I think there is a race condition here because the job may not have reached the DEPEND state by the time `flux job info` is run. Maybe `flux job wait-event -vt 60 $jobid depend` would be better here? (Sorry if I originally had suggested `flux job info`!)

I'm not actually sure how to definitively test that a job _stays_ in DEPEND state though.. so maybe right after the wait-event you can dump the eventlog and check again like you have here now?

grondo (commented):
This LGTM. One note about a race condition in the tests. Note: I discovered it by running under `valgrind` which made the race reproducible actually.

t/t1029-mf-priority-default-bank.t
cmoussa1: Thanks for this suggestion! Sorry that I was using the wrong method to check for the job state here. I just re-ran this set of tests (both under and not under valgrind) and using `flux job wait-event` seemed to work. I can then check the eventlog afterwards and look for the "depend" event. I've just pushed up this change. Let me know if it looks better!

On a side note, I should probably step through the test suite tomorrow (like you mentioned in #294) and change the tests that test for job states to use the suggestion above. I'll open a separate issue on that.

grondo (approved):
Once the test is fixed LGTM!

cmoussa1:
Thanks! I believe I've fixed the test as suggested in the review above. I'll set MWP on this 👍

PR #302, opened 2022-12-10
testsuite: do not assume queues default to stopped

chu11:
Problem: In t1013-mf-priority-queues.t, newly created job queues
are assumed to be stopped due to a call to `flux queue stop` earlier
in the tests.  A queue defaulting to stopped cannot be assumed.

Add an additional call to flux queue stop after the job queues
are created.

See investigation in https://github.com/flux-framework/flux-core/pull/4776


garlick (approved):
LGTM!

PR #303, opened 2022-12-10
testsuite: stop all queues with --all option

chu11:
Problem: flux-core PR 4776 added the ability for queues to be individually started and stopped.  As an update, starting/stopping all queues with the `flux queue` command now requires the --all option to be specified, otherwise a warning is output.

To avoid the warning, update all callers to specify --all when multiple queues are started or stopped.

This is similar to https://github.com/flux-framework/flux-sched/pull/992

***This should not be merged until https://github.com/flux-framework/flux-core/pull/4776 is merged***

PR #304, opened 2022-12-12
NEWS: add release notes for `v0.21.0`

cmoussa1:
After the added feature in #299 and the fixes proposed by @chu11 in #302 and #303, I think it's definitely worth tagging and releasing a new version of flux-accounting. This PR adds release notes for `v0.21.0`.

This PR should not be merged until after #303 has been merged.

Once this lands, I'll create an annotated tag with the following command:

```console
git tag -a v0.21.0 -m "Tag v0.21.0" && git push upstream v0.21.0
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,16 @@
+flux-accounting version 0.21.0 - 2022-12-12
+-------------------------------------------
+
+#### Features
+
+* Add ability to edit the parent bank of a bank (#299)
+
+#### Testsuite
+
+* Do not assume queues default to stopped (#302)
+
+* Stop all queues with `--all` option (#303)
+
 flux-accounting version 0.20.1 - 2022-10-05
 -------------------------------------------
 


NEWS.md
```diff
@@ -1,3 +1,14 @@
+flux-accounting version 0.21.0 - 2022-12-12
+-------------------------------------------
+
+#### Fixes
+
+* Specify `--all` option to `flux queue stop` in rc1 and sharness tests (#303)
```
grondo, line 6: It might be better to just use the verbatim title from the PR:

>  * testsuite: stop all queues with --all option (#303)

Also an entry is missing for #302 
>  * testsuite: do not assume queues default to stopped (#302)

Since these are both testsuite-only fixes, you may want to list them under `#### Testsuite`, and likely after new Features since the new Features are most important (I leave that up to you though)

cmoussa1: Thanks, I really like that suggestion. Sorry about missing the other PR! I just force-pushed a fix to add the other fix for #302, as well as move both of these to a separate `#### Testsuite` section under the `#### Features` section.

grondo (approved):
LGTM!

NEWS.md
grondo: We have a script in `flux-core` that can help you generate release notes, e.g.

```
 grondo@asp:~/git/flux-accounting$ ../flux-core.git/src/test/relnotes.sh 
Changes since v0.20.1:
 * edit-bank: add ability to edit parent bank (#299)
 * testsuite: do not assume queues default to stopped (#302)
 * github: change Ubuntu version for python-lint check (#300)
 * add-user: add bank validation (#277)
```

In case it ever comes in useful!

cmoussa1: Oh awesome! I had no idea. I'll definitely have to steal this for future releases ;)

Issue #186, opened 2021-12-15
database: allow remote access to accounting database


ryanday36:
users should have some means of querying bank information from the login node when the database resides on a management node. It would also be useful for staff/admins to be able to add users to banks remotely.

cmoussa1:
This was brought up in our project meeting today, so I'll begin reviving this to think about it more.

I have to apologize in advance that I'm not very familiar with how remote access works on our clusters, so I don't know for sure how to go about this, at least right now. I know that SQLite is not a network database, but I don't _think_ that should be an issue when a user/admin/somebody is already on the cluster.

I also know that the job-archive DB in flux-core is a SQLite database. Can that database be accessed from any node on a cluster? If so, is there some special setup that happens in flux-core that enables such functionality?

grondo:
@garlick had a good idea that the remote access interface to the accounting database could be implemented as a systemd-managed python script which connects to the system instance and registers a service using [`handle.service_register()`](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/python/autogenerated/flux.core.handle.html#flux.core.handle.Flux.service_register).

There are multiple benefits to this approach:
 - Use of the accounting database service can occur from any node in the Flux instance
 - No need for separate host:port etc configuration for the database (the service would be available on rank 0)
 - Security and user authentication handled by Flux ([RFC 12](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_12.html))
 - Allows the use of Python for the implementation, rapid development
 - Separates the db server from the address space of the broker

The main drawback is that there is potentially extra development here since client tools will have to be written in terms of Flux RPCs, the Python server would need to be implemented. If using a database which allows remote access then perhaps the utilities could work as-is once they are connected to the database.

grondo:
@cmoussa1, as promised on the :coffee: call today, here's a skeleton example of a python script that registers a service with the broker then handles a couple of rpc endpoints under that service name. Given that we're proposing to run this script under systemd, the script also tries to handle `SIGTERM` and gracefully exit.

```python
#!/usr/bin/env python3
import signal
import sys
import errno

import flux
from flux.constants import FLUX_MSGTYPE_REQUEST


class TestServer:
    def __init__(self, flux_handle):

        self.fh = flux_handle

        #  Register service with broker. Allows this
        self.fh.service_register("test-server").get()
        print("registered test-server service", file=sys.stderr)

        #  Register signal watcher for SIGTERM to initiate shutdown
[...]

PR #307, opened 2023-01-13
python: redesign functions in `Accounting` class

cmoussa1:
#### Background

The Python functions in the `Accounting` class directly emit output to `STDOUT` instead of `return`-ing information to the caller of the function. To prepare for the transition to using a systemd-managed service that accepts an RPC to call the Python functions, this behavior should be changed to `return` information to the caller of the function.

---

This PR redesigns a number of the functions in the `Accounting` class to `return` any information that would otherwise be emitted to `STDOUT` using `print()` statements. The front-end Python script that calls these functions has also been adjusted to assign the returned value to a variable, where it is then `print()`-ed out.

A couple of the existing sharness tests required slight changes to expected output as a result of the redesign described above. A couple of the unit tests also had to be adjusted slightly to account for some of the changes to the functions in the `Accounting` class. What was being tested in these unit and sharness tests, though, did not change.

This work is a prerequisite to the refactoring being discussed in #186 to eventually transition the calling of the functions in the `Accounting` class from directly calling these functions to issuing RPCs to a service that calls these functions.

##### TODO

- [x] the commit that redesigns `user_subcommands.py` could probably be split into two smaller commits: one that adds the `return` functionality and one that fixes some of the misspelling and poor variable naming
- [x] the commit that redesigns `bank_subcommands.py` could probably be split into two smaller commits: one that adds the `return` functionality and one that removes the helper function for the `-t` flag, which is removed in this PR

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -21,25 +21,28 @@ from fluxacct.accounting import user_subcommands as u
 
 # helper function to print user information in a table format
 def print_user_rows(cur, rows, bank):
-    print("\nUsers Under Bank {bank_name}:\n".format(bank_name=bank))
+    user_str = "\nUsers Under Bank {bank_name}:\n\n".format(bank_name=bank)
     user_headers = [description[0] for description in cur.description]
     # print column names of association_table
     for header in user_headers:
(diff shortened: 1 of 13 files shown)


ryanday36 (approved):
Looks good @cmoussa1. Glad to see you diving into this!

Issue #311, opened 2023-02-17
bindings: improve clarity and context to error messages


cmoussa1:
A lot of the exception handling in the Python bindings simply print out the error message associated with the caught `Exception`, which can be confusing to decipher (e.g is it a result of `sqlite3`? Something else?).

The error handling for the bindings should be improved to provide more context to the error that occurred.

cmoussa1:
This work should also be completed by #327, #328, and #329, so I'll close this one too.

PR #315, opened 2023-02-28
flux-accounting: add systemd unit file to repository

### LINKED ISSUE #314: repo: add systemd unit file to repository
Now that #308 has landed, the next step is to add a systemd unit file to the repository. I'm not super familiar with what this would look like. I looked around in flux-core and I _think_ I found something similar in the `etc/` directory but I'm not 100% if those are the right files. Perhaps flux-accounting needs something similar to [flux.service.in](https://github.com/flux-framework/flux-core/blob/master/etc/flux.service.in)? And then add it to a `Makefile.am`  like it does in flux-core:

```
#if HAVE_SYSTEMD
systemdsystemunit_DATA = flux.service
#endif
```

cmoussa1:
#### Background

The flux-accounting commands that interact with the database can be started as a service now, but cannot be controlled with systemd. It currently has to be started manually. 

---

This PR adds a systemd unit file to flux-accounting so that it can be controlled with systemd. The unit file is located in the `etc/` directory and, if configured the same as flux-core, will be installed to the same directory as flux-core's `flux.service` file.

This enables the flux-accounting service to be started with `systemctl` and run in conjunction with a Flux system instance:

```console
$ systemctl enable /path/to/flux-accounting.service
$ systemctl start flux-accounting
$ flux account add-bank root 1
```

Per the systemd unit file, a requirement for the flux-accounting service to be run with systemd is that flux-core's service is currently up and running. On a Flux system instance restart or complete shutdown, flux-accounting will also be restarted or shut  down. It will then come back up shortly after the Flux system instance is started and running.

#### Documentation

As a result of these changes and additions to installation and setup for flux-accounting, the setup guide for flux-accounting in flux-framework/flux-docs should be updated to reflect that. I plan on updating the documentation as soon as this PR is landed. I should probably also update some documentation here in this repository as well. 

#### Testing

This was tested manually in a RHEL virtual machine with flux-core installed and running as a system instance. I'm not sure if the flux-core equivalent of this is currently tested in CI anywhere, but I figured I'd just post this as is, and if testing should be added, I'd be happy to do so. 😄 

Fixes #314


grondo:
Great! I'll give this a try in the fluxorama container and see how it goes.

cmoussa1:
Sounds great! Thanks for taking the time to test it out in your container. Let me know if I can be of any help.

etc/Makefile.am
```diff
@@ -0,0 +1,3 @@
+#if HAVE_SYSTEMD
```
grondo, line 1: Commit message typo: "interacted with systemd" :arrow_right: "controlled with systemd" or "started with systemd"?

grondo (approved):
LGTM! Just one typo in a commit message.

cmoussa1:
Nice! Glad to see that it is working.

> One issue I did hit just playing with commands:

Ah, yeah that error message is ugly looking. Sorry about that. That should be cleaned up as part of the work outlined in #311. I think it might be an issue with the way I am passing arguments to that Python binding when I converted using the flux-accounting service. I'll open an issue on this.

etc/Makefile.am
cmoussa1: Thanks for catching this! Just force-pushed a fix for this commit message.

cmoussa1:
Thanks for taking the time to review and test this PR @grondo! I'll go ahead and set MWP on this shortly. Looking forward to releasing and testing this on a machine this month!

PR #316, opened 2023-02-28
view-job-records: fix arguments passed in via `flux-account-service.py`

cmoussa1:
#### Background

The call to `output_job_records()` in **flux-account-service.py** currently fails due to a mismatch in arguments to the function.

```console
$ sudo flux account view-job-records
  File "/usr/libexec/flux/cmd/flux-account.py", line 679, in <module>
    main()
  File "/usr/libexec/flux/cmd/flux-account.py", line 675, in main
    select_accounting_function(args, output_file, parser)
  File "/usr/libexec/flux/cmd/flux-account.py", line 554, in select_accounting_function
    return_val = flux.Flux().rpc("accounting.view_job_records", data).get()
  File "/usr/lib64/flux/python3.6/flux/rpc.py", line 77, in get
    resp_str = self.get_str()
  File "/usr/lib64/flux/python3.6/flux/util.py", line 90, in func_wrapper
    retval = func(future, *args, **kwargs)
  File "/usr/lib64/flux/python3.6/flux/rpc.py", line 71, in get_str
    self.pimpl.flux_rpc_get(payload_str)
  File "/usr/lib64/flux/python3.6/flux/util.py", line 66, in func_wrapper
    raise EnvironmentError(error.errno, errmsg.decode("utf-8")) from None
OSError: [Errno 22] a non-OSError exception was caught: output_job_records() takes 2 positional arguments but 6 were given
```

This is because the parameters in the call to `output_job_records()` are not labeled.

The path passed to the `output_job_records()` function is incorrect. The path should be to the flux-core job-archive DB, not the flux-accounting DB.

---

This PR makes a couple of small fixes to the `view-job-records` command in **flux-account-service.py** and adjusts the helper functions for that command slightly to account for the use of the new service, most notably in its output. A couple of tests are also added to **t1011-job-archive-interface.t** that call this command, both with and without a passed output file.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -65,25 +65,29 @@ def write_records_to_file(job_records, output_file):
 
 
 def fetch_job_records(job_records):
-    job_record_str = ""
-    job_record_str += "{:<10} {:<10} {:<10} {:<15} {:<15} {:<15} {:<10}".format(
-        "UserID",
-        "Username",
-        "JobID",
-        "T_Submit",
-        "T_Run",
-        "T_Inactive",
-        "Nodes",
+    job_record_str = []
+    job_record_str.append(
+        "{:<10} {:<10} {:<20} {:<20} {:<20} {:<20} {:<10}".format(
+            "UserID",
+            "Username",
+            "JobID",
+            "T_Submit",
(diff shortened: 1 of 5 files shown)


ryanday36 (approved):
looks good @cmoussa1

cmoussa1:
Thanks @ryanday36! I'll set MWP on this

PR #319, opened 2023-03-07
edit-user: make "userid" an editable field

### LINKED ISSUE #318: can't change uid for users in flux accounting db
If a user gets created in the flux accounting db before they're in the passwd file, they get a default UID of 65534. Once they get added to the passwd file, their UID needs to get changed to the correct value. There isn't an option to do this with `flux account edit-user`. Deleting and re-adding the user doesn't fix it either:
```
[day36@corona81:~]$ flux account view-user sochat1
creation_time    mod_time         active           username         userid           bank             default_bank     shares           job_usage        fairshare        max_running_jobs max_active_jobs  max_nodes        queues           projects         default_project  
1661527388       1661527388       1                sochat1          65534            lc               lc               1                0.0              0.528487         100              1000             2147483647                        *                *                
[day36@corona81:~]$ flux account delete-user sochat1 lc
[day36@corona81:~]$ flux account add-user --username sochat1 --bank lc --shares 1 --max-running-jobs 100 --max-active-jobs 1000 --userid 34633
[day36@corona81:~]$ flux account view-user sochat1
creation_time    mod_time         active           username         userid           bank             default_bank     shares           job_usage        fairshare        max_running_jobs max_active_jobs  max_nodes        queues           projects         default_project  
1661527388       1661527388       1                sochat1          65534            lc               lc               1                0.0              0.528487         100              1000             2147483647                        *                *                
[day36@corona81:~]$ grep sochat1 /etc/passwd
sochat1:x:34633:34633:Vanessa Elizabeth Sochat, +19254243702:/g/g0/sochat1:/bin/bash
[day36@corona81:~]$
```

cmoussa1:
#### Problem

As reported in #318, the "userid" field for a user is not editable in the `association_table`, so when a user is created in the DB before they are in the passwd file, they get a default UID of 65534 with no way to edit it unless they go into a SQLite shell and manually edit the column themselves.

---

This PR adds "userid" as an editable field in the `edit-user` command. It also adds a unit test and a sharness test that confirms that the userid can be edited.

Fixes #318

### DIFF
--- a/src/bindings/python/fluxacct/accounting/test/test_user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/test/test_user_subcommands.py
@@ -248,6 +248,18 @@ class TestAccountingCLI(unittest.TestCase):
 
         self.assertRaises(ValueError)
 
+    # edit a user's userid
+    def test_14_edit_user_userid(self):
+        cur = acct_conn.cursor()
+        u.add_user(acct_conn, username="test_user5", bank="A")
+
+        cur.execute("SELECT userid FROM association_table WHERE username='test_user5'")
+        self.assertEqual(cur.fetchone()[0], 65534)
+
+        u.edit_user(acct_conn, username="test_user5", userid="12345")
+        cur.execute("SELECT userid FROM association_table WHERE username='test_user5'")
+        self.assertEqual(cur.fetchone()[0], 12345)
+
     # remove database and log file
     @classmethod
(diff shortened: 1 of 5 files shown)


ryanday36 (approved):
looks good. Thanks @cmoussa1!

cmoussa1:
Thanks @ryanday36! Will set MWP on this

PR #320, opened 2023-03-08
`view-*` commands: raise `ValueError` when item cannot be found in flux-accounting DB

### LINKED ISSUE #309: view-user: command succeeds even when user cannot be found
The `view-user` command succeeds even when a user cannot be found in the flux-accounting database, e.g:

```console
bash-4.4$ flux account view-user grondo
User not found in association_table
bash-4.4$ echo $?
0
```

This behavior should be changed to instead return a non-zero exit code.

### LINKED ISSUE #310: bindings: `view-*` commands succeed even when item cannot be found
Related to #309, all of the `view-*` commands "succeed" even when the item cannot be found. In the case of an unknown user, bank, queue, or a project, the corresponding function simply returns a `"<item> was not found in <item_table>"`. These should be changed to instead return a non-zero exit code.

cmoussa1:
#### Background

As mentioned in #309 and #310, the `view-*` commands for flux-accounting still "succeed" even when the item being searched cannot be found in the flux-accounting DB. It's current behavior is just to return a string saying "Item not found in *_table." This behavior should be changed to instead raise an exception.

---

This PR changes the functions called by the various `view-*` commands to instead raise a `ValueError` when an item cannot be found in the flux-accounting database. The error is caught by the flux-accounting service and a message is sent saying, for example:

```console
$ flux account view-user foo
error in view-user: User not found in association_table
```

The respective unit and sharness tests are also slightly changed to account for this new raised exception when a user cannot be found, i.e instead of just checking for string output, checking for a raised `ValueError` in the unit tests and checking that the failed `view-*` commands _do_ fail when an item cannot be found.

Fixes #309
Fixes #310

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -114,7 +114,7 @@ def view_bank(conn, bank, users=False):
         if rows:
             bank_str = get_bank_rows(cur, rows, bank)
         else:
-            return "Bank not found in bank_table"
+            raise ValueError(f"Bank {bank} not found in bank_table")
 
         # if users is passed in, print out all potential users under
         # the passed in bank
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -191,7 +191,7 @@ def view_user(conn, user):
         rows = cur.fetchall()
         headers = [description[0] for description in cur.description]  # column names
         if not rows:
-            return "User not found in association_table"
+            raise ValueError(f"User {user} not found in association_table")
 
         user_str = get_user_rows(headers, rows)
 
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -133,6 +133,8 @@ class AccountingService:
             handle.respond(msg, payload)
         except KeyError as exc:
             handle.respond_error(msg, 0, f"missing key in payload: {exc}")
+        except ValueError as val_err:
+            handle.respond_error(msg, 0, f"error in view-user: {val_err}")
         except Exception as exc:
             # fall through to a non-OSError exception
             handle.respond_error(
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -420,6 +420,8 @@ class AccountingService:
             handle.respond(msg, payload)
         except KeyError as exc:
             handle.respond_error(msg, 0, f"missing key in payload: {exc}")
+        except ValueError as val_err:
+            handle.respond_error(msg, 0, f"error in view-project: {val_err}")
         except Exception as exc:
             handle.respond_error(
                 msg, 0, f"a non-OSError exception was caught: {str(exc)}"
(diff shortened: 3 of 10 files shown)


src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -191,7 +191,7 @@ def view_user(conn, user):
         rows = cur.fetchall()
         headers = [description[0] for description in cur.description]  # column names
         if not rows:
-            return "User not found in association_table"
+            raise ValueError("User not found in association_table")
```
grondo, line 194: Suggestion: Not a requirement, but it might be nice to include the user name (or is `user` a userid) in the error string:

```python
raise ValueError(f"User {user} not found in association table")
```

src/cmd/flux-account-service.py
```diff
@@ -133,6 +133,8 @@ def view_user(self, handle, watcher, msg, arg):
             handle.respond(msg, payload)
         except KeyError as exc:
             handle.respond_error(msg, 0, f"missing key in payload: {exc}")
+        except ValueError as val_err:
+            handle.respond_error(msg, 0, f"error in view-user: {str(val_err)}")
```
grondo, line 137: I don't think `str(val_err}` is required here, just `{val_err}` should be sufficient

src/bindings/python/fluxacct/accounting/bank_subcommands.py
```diff
@@ -114,7 +114,7 @@ def view_bank(conn, bank, users=False):
         if rows:
             bank_str = get_bank_rows(cur, rows, bank)
         else:
-            return "Bank not found in bank_table"
+            raise ValueError("Bank not found in bank_table")
```
grondo, line 117: Similar comment as before, it might be nice to include the failing bank name in the error message.

src/cmd/flux-account-service.py
```diff
@@ -220,6 +220,8 @@ def view_bank(self, handle, watcher, msg, arg):
             handle.respond(msg, payload)
         except KeyError as exc:
             handle.respond_error(msg, 0, f"missing key in payload: {exc}")
+        except ValueError as val_err:
+            handle.respond_error(msg, 0, f"error in view-bank: {str(val_err)}")
```
grondo, line 224: Again: I'm not sure `str(val_err)` is needed here. I do notice you use that other places, so I apologize if I"m wrong!

grondo (commented):
Nice!

I had a couple suggestions that apply to all code commits. I'll let you respond to those before reviewing further!

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: That's a great suggestion! I like that. I just force-pushed up a fix to include the username in the error string, along with all of the other respective names (bank, queue, project) in their error strings.

src/cmd/flux-account-service.py
cmoussa1: You are right, good catch, thank you! `{val_err}` is sufficient here. I just force-pushed up a fix to just use `{val_err}`

grondo (approved):
LGTM!

cmoussa1:
Thanks! Will set MWP on this

PR #325, opened 2023-03-14
plugin: rework increment/decrement of running job counts

cmoussa1:
#### Problem

As noted in #262, the plugin currently incorrectly handles calculating running job counts for associations (user/bank combos). It increments the running job count in `job.state.priority`, which could result in a double increment if the job re-enters PRIORITY state after a reprioritization of jobs via `flux account-priority-update`.

The plugin also incorrectly handles decrementing running and active job counts when a job enters `job.state.inactive`. Even if a job did not run, the plugin still decrements the association's current running jobs count and looks to release any held jobs from a max running jobs dependency, which could result in incorrect running job counts.

---

This PR looks to rework the way the plugin increments **and** decrements running and active job counts for an association. It adds a new callback for when a job enters `job.state.run` and increments the current running jobs count there.

It also reworks the way the plugin decrements active and running job counts in `inactive_cb ()`. It checks if the job received an `alloc` event to determine if it actually ran. If True, it decrements both the running and the active job counts, and proceeds to check if there are any held jobs to release. If False, then the plugin determines that the job never actually ran, so it only decrements the active job count and `return`s from `inactive_cb ()`.

A new sharness test is also added that tests some of the problematic scenarios brought up in #262. Specifically, the test submits one job to run and two jobs that remain in SCHED state. It then simulates a couple of different cases and checks that the active and running job counts are expected. The scenarios include:

- updating the plugin with flux-accounting DB information while jobs are running/scheduled via `flux account-priority-update`
- changing the priority of a scheduled job while jobs are running/scheduled
- canceling one of the scheduled jobs
- canceling both running/scheduled jobs

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -977,8 +977,14 @@ static int inactive_cb (flux_plugin_t *p,
         return -1;
     }
 
-    b->cur_run_jobs--;
     b->cur_active_jobs--;
+    // nothing more to do if this job was never running
+    if (!flux_jobtap_job_event_posted (p, FLUX_JOBTAP_CURRENT_JOB, "alloc"))
+        return 0;
+
+    // this job was running, so decrement the current running jobs count
+    // and look to see if any held jobs can be released
+    b->cur_run_jobs--;
 
     // if the user/bank combo has any currently held jobs and the user is now
     // under their max jobs limit, remove the dependency from first held job
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -605,10 +605,9 @@ static int priority_cb (flux_plugin_t *p,
                 return -1;
 
             // if we get here, the bank was unknown when this job was first
-            // accepted, and therefore the active and run job counts for this
+            // accepted, and therefore the active job counts for this
             // job need to be incremented here
             bank_it->second.cur_active_jobs++;
-            bank_it->second.cur_run_jobs++;
 
             // update current job with user/bank information
             if (flux_jobtap_job_aux_set (p,
@@ -633,8 +632,6 @@ static int priority_cb (flux_plugin_t *p,
                   flux_plugin_arg_strerror (args));
         return -1;
     }
-
-    b->cur_run_jobs++;
 
     return 0;
 }
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -914,6 +914,34 @@ static int depend_cb (flux_plugin_t *p,
 }
 
 
+static int run_cb (flux_plugin_t *p,
+                   const char *topic,
+                   flux_plugin_arg_t *args,
+                   void *data)
+{
+    int userid;
+    struct bank_info *b;
+
+    b = static_cast<bank_info *>
+        (flux_jobtap_job_aux_get (p,
+                                  FLUX_JOBTAP_CURRENT_JOB,
+                                  "mf_priority:bank_info"));
+
+    if (b == NULL) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.state.run: bank info is " \
+                                     "missing");
+
+        return -1;
+    }
+
+    // increment the user's current running jobs count
+    b->cur_run_jobs++;
+
+    return 0;
+}
+
+
 static int inactive_cb (flux_plugin_t *p,
                         const char *topic,
                         flux_plugin_arg_t *args,
@@ -977,6 +1005,7 @@ static const struct flux_plugin_handler tab[] = {
     { "job.priority.get", priority_cb, NULL },
     { "job.state.inactive", inactive_cb, NULL },
     { "job.state.depend", depend_cb, NULL },
+    { "job.state.run", run_cb, NULL},
     { "plugin.query", query_cb, NULL},
     { 0 },
 };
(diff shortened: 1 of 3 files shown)


grondo:
>  It adds a new handler for a job's alloc event by subscribing to all events for the job in job.new. Once the job receives the alloc event, the current running jobs counter for the association is incremented. The addition of this new handler removes the need to increment the running jobs count in job.state.priority.

Did you first try adding a callback for `job.state.run` and increment the running jobs counter there? (This is the approach noted in [this comment](https://github.com/flux-framework/flux-accounting/issues/262#issuecomment-1464716028)) That might be more lightweight than subscribing to all job events.

cmoussa1:
> Did you first try adding a callback for job.state.run and increment the running jobs counter there? (This is the approach noted in https://github.com/flux-framework/flux-accounting/issues/262#issuecomment-1464716028) That might be more lightweight than subscribing to all job events.

Oh, whoops. I didn't even listen to my own comment in that thread. 🤦 I'm sorry about that. Yeah, looks like adding it to `job.state.run` works, so I'll update the commit and PR description right now and force up a push. Thanks for reminding me!

cmoussa1:
OK, I think I've updated both the commit and PR description now to use `job.state.run` instead of adding a new event handler to the plugin. I think I'll mark this as ready for review whenever convenient.

src/plugins/mf_priority.cpp
```diff
@@ -914,6 +914,36 @@ static int depend_cb (flux_plugin_t *p,
...
 
+static int run_cb (flux_plugin_t *p,
+                      const char *topic,
+                      flux_plugin_arg_t *args,
+                      void *data)
+{
+    int userid;
+    struct bank_info *b;
+
+    flux_t *h = flux_jobtap_get_flux (p);
```
grondo, line 925: Is `h` unused here?

src/plugins/mf_priority.cpp
```diff
@@ -914,6 +914,36 @@ static int depend_cb (flux_plugin_t *p,
 }
 
 
+static int run_cb (flux_plugin_t *p,
+                      const char *topic,
+                      flux_plugin_arg_t *args,
+                      void *data)
```
grondo, line 920: looks like some whitespace errors here?

src/plugins/mf_priority.cpp
```diff
@@ -979,8 +979,18 @@ static int inactive_cb (flux_plugin_t *p,
...
+    // current active jobs count; if not, then we only need to decrement the
+    // current active jobs count and exit this callback.
+    if (flux_jobtap_job_event_posted (p, FLUX_JOBTAP_CURRENT_JOB, "alloc")) {
+        b->cur_run_jobs--;
+        b->cur_active_jobs--;
+    } else {
+        b->cur_active_jobs--;
+
+        return 0;
+    }
```
grondo, line 993: IMO, this logic would be clearer by

 1. first decrement active jobs count, this makes it clear that count is always decremented.
 2. check if no `alloc` event posted, then nothing more to do, so return
 3. proceed to decrement `cur_run_jobs` and check for any held jobs that need to be released

i.e.:
```c
    b->cur_active_jobs--;
    // Nothing more to do if this job was never running
    if (!flux_jobtap_job_event_posted (p, FLUX_JOBTAP_CURRENT_JOB, "alloc"))
        return 0;

    // This job was running, decrement running jobs count and check if a job can be released
    b->cur_run_jobs--;

```

t/t1020-mf-priority-issue262.t
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+
+test_description='Test comparing job counts when submitting jobs that take up all resources'
```
grondo, line 3: Nicely done. Did you verify that before the changes in this PR this test fails? (just curious)

grondo (commented):
Nice! :+1: Just some initial comments inline. Mostly very minor stuff.

t/t1020-mf-priority-issue262.t
cmoussa1: Yes I did - here are the results of running this test with the plugin before any changes:

#### query_1.json

##### expected
cur_running_jobs: 1
cur_active_jobs: 3

##### failed
cur_running_jobs: 3
cur_active_jobs: 3

---

#### query_2.json

##### expected
cur_running_jobs: 1
cur_active_jobs: 3

##### failed
cur_running_jobs: 5
cur_active_jobs: 3

---

#### query_3.json

##### expected
cur_running_jobs: 1
cur_active_jobs: 3

##### failed
cur_running_jobs: 8
cur_active_jobs: 3

---

#### query_4.json

##### expected
cur_running_jobs: 1
cur_active_jobs: 2

##### failed
cur_running_jobs: 7
cur_active_jobs: 2

---

#### query_5.json

##### expected
cur_running_jobs: 0
cur_active_jobs: 0

##### failed
cur_running_jobs: 5
cur_active_jobs: 0

src/plugins/mf_priority.cpp
cmoussa1: Ah, yes, thanks. I always forget to adjust the whitespace for these function headers when I copy/paste the callbacks. 🤦 Just pushed up a change to fix this.

cmoussa1:
Thanks for the review @grondo. Just force-pushed some fixes to this PR based on your feedback above, including the whitespace adjustments and re-work to the logic in `inactive_cb ()`.

grondo (approved):
LGTM!

cmoussa1:
Thanks! Will set MWP on this

Issue #313, opened 2023-02-21
bindings: change behavior of bindings to raise error to caller


cmoussa1:
The Python bindings for flux-accounting currently do a lot of/all of the error handling within the function itself instead of raising the error to the caller of the function. It handles the error and then returns a string. This behavior should be changed to instead raise the error and let the caller of the function handle the error.

cmoussa1:
This should be completed now by the work in #327, #328, and #329, so I'll close this now. I can re-open in case there are any stragglers that I find in the future.

PR #327, opened 2023-03-16
`user-*` commands: change error handling to raise error to caller

cmoussa1:
#### Background

The functions called by all of the `user-*` commands handle any exceptions raised during execution and just return a string containing the error message. This behavior should be changed to be more Pythonic and raise errors to the caller of the function.

As noted in #313, this behavior currently exists in all of the Python bindings, but instead of creating one large PR that changes the behavior for all of the bindings at once, it might be a little cleaner to post subsequent, smaller PRs that change the bindings one file at a time. This one starts with `user_subcommands.py`.

---

This PR changes the behavior of a number of the functions in `user_subcommands.py` to raise any error to the caller of the function with a clear message explaining what the error is. For `sqlite3.OperationalError`, a new comment is added to provide more context on what the error means and add a link to official docs containing more information about `OperationalError`.

The unit/sharness tests for the `user-*` commands are also adjusted to account for the new behavior of the commands that raise an exception on certain errors, and a couple new unit and sharness tests are also added to further check for expected raised errors.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/test/test_project_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/test/test_project_subcommands.py
@@ -105,8 +105,8 @@ class TestAccountingCLI(unittest.TestCase):
 
     # editing a user's project list with a bad project name should raise a ValueError
     def test_07_edit_projects_list_bad_name(self):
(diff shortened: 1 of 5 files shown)


ryanday36 (approved):
Nice. Error messages look pretty good to me.

PR #328, opened 2023-03-21
`bank-*` commands: change error handling to raise error to caller

cmoussa1:
This PR continues the work brought on by #327 and changes the behavior of a number of the functions in `bank_subcommands.py` to raise any error to the caller of the function with a clear message explaining what the error is.

The unit/sharness tests for the `bank-*` commands are also adjusted to account for the new behavior of the commands that raise an exception on certain errors, and a couple new unit and sharness tests are also added to further check for expected raised errors.


ryanday36 (approved):
Looks good to me.

cmoussa1:
Thanks! Setting MWP on this

PR #329, opened 2023-03-23
`*-queue`/`*-project` commands: change error handling to raise error to caller

cmoussa1:
This PR completes the work brought on by https://github.com/flux-framework/flux-accounting/pull/327 and changes the behavior of a number of the functions in `queue_subcommands.py` and `project_subcommands.py` to raise any error to the caller of the function with a clear message explaining what the error is.

The unit/sharness tests for the `queue-*` and `*-project` commands are also adjusted to account for the new behavior of the commands that raise an exception on certain errors, and a couple new unit and sharness tests are also added to further check for expected raised errors.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/project_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/project_subcommands.py
@@ -12,6 +12,34 @@
 import sqlite3
 
 
+###############################################################
+#                                                             #
+#                      Helper Functions                       #
+#                                                             #
+###############################################################
+
+
+# check if project already exists and is active in project_table;
+# if so, return True
+def project_is_active(cur, project):
+    cur.execute(
+        "SELECT * FROM project_table WHERE project=?",
+        (project,),
+    )
+    project_exists = cur.fetchall()
+    if len(project_exists) > 0:
+        return True
+
+    return False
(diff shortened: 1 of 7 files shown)


src/bindings/python/fluxacct/accounting/project_subcommands.py
```diff
@@ -12,6 +12,34 @@
...
 
 
+###############################################################
+#                                                             #
+#                      Helper Functions                       #
+#                                                             #
+###############################################################
+
+
+# check if bank already exists and is active in bank_table;
```
ryanday36, line 22: From the department of trivial things, this comment should probably refer to 'project' and 'project_table' rather than 'bank' and 'bank_table'.

ryanday36 (approved):
looks good save the one comment on a comment.

src/bindings/python/fluxacct/accounting/project_subcommands.py
cmoussa1: Whoops, you're totally right. 🤦 Thanks for catching this. Just force-pushed a fix to this comment.

cmoussa1:
Thanks for approving @ryanday36!

PR #330, opened 2023-03-26
flux-accounting service: make certain commands accessible to all users

cmoussa1:
#### Problem

All of the message handlers in the flux-accounting service are only available to the instance owner, but some of them should be open to all users. Commands that simply read from the flux-accounting DB, like `view-user`, `view-bank`, etc. should be able to be run by any user in on the machine, while others that write to the DB, like `add-user`, `add-bank`, should only be able to be run by a privileged group of administrators.

---

This PR breaks up the `endpoints` list into two smaller lists, one that is "general" and can be run by any user, and one that is "privileged" and is only available to the instance owner. Specifically, from the "general" list, the watchers set a rolemask to `FLUX_ROLE_USER` when creating the watcher.

A couple of tests are also added to `t1007-flux-account.t` that make sure that certain commands fail for a normal user but others succeed.

### DIFF
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -15,6 +15,7 @@ import argparse
 import logging
 
 import flux
+import flux.constants
 import fluxacct.accounting
 
 from flux.constants import FLUX_MSGTYPE_REQUEST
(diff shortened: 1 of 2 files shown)


cmoussa1:
Thanks @ryanday36!

> We're going to want users to be able to view their own jobs, but if the job records contain things like paths or names of executables, there can be problems with allowing users to see other users job records. If they just have things like jobids, times, nodes, cores, etc., it's not a problem.

Here's what `view-job-records` contains for each job record:

```
"UserID",
"Username",
"JobID",
"T_Submit",
"T_Run",
"T_Inactive",
"Nodes",
"R",
```

I forget if "R" contains things like paths or names of executables. Perhaps @grondo or somebody with better knowledge of R than me should comment here with their expertise.

cmoussa1:
@ryanday36 just wanted to update you about the `view-job-records` command; spoke offline with @chu11 about this yesterday and confirmed that the information returned by this command should be okay to be visible by other users since this information is also returned by `flux jobs`, so I think this command would be okay to be made accessible to all users.

cmoussa1:
@grondo, if you have some time next week, could you give this PR a review? @ryanday36 already gave his approval but I want to make sure I'm not messing anything up with the rolemask stuff in `flux-account-service.py`.

grondo (approved):
This looks good to me.

The only suggestion I have is that it would seem to be easy to add a list of endpoints that should be restricted and ensure every one of them is instance-owner only. Since it is quite important to ensure none of those RPCs is accidentally exposed to normal users, that seems like it would be valuable in the testing instead of only testing a sample endpoint.

cmoussa1:
Thanks for the suggestion @grondo! I went ahead and force-pushed a fix to add tests for all of the privileged commands to make sure that they cannot be run by just any user. They all follow pretty much the same format.

I'll go ahead and set MWP on this shortly

Issue #333, opened 2023-03-27
users still prevented from submitting to queue after it has been removed


ryanday36:
Possibly related to #332, if I delete a queue with `flux account delete-queue ...`, then run `flux account-priority-update`, users who were not given access to it continue to be unable to submit to it.

cmoussa1:
Sorry, I think I could probably use a little more clarification here. 😅 This sounds like expected behavior? If a queue is removed from the flux-accounting database, shouldn't users **not** be able to submit jobs under that queue, since it does not exist?

ryanday36:
The queue still exists on the cluster, just not in the DB. I expected it to revert to the behavior before the queue had been added where the plugin skips the check for whether or not the user has the queue.

cmoussa1:
This should also be resolved by #334, so I will close this too.

Issue #332, opened 2023-03-27
user access to queues persists after being removed in DB


ryanday36:
If a user who has been added to a queue via `flux account edit-user --queue=_queue_name user_name_` is removed from that queue with `flux account edit-user --queue= _user_name_`, their access to the queue persists after running an `account-priority-update`. E.g. on the management node:
```
[root@fluke1:~]# flux account view-user day36
creation_time   mod_time        active          username        userid          bank            default_bank    shares          job_usage       fairshare       max_running_jobsmax_active_jobs max_nodes       queues          projects        default_project 
1648069813      1679958340      1               day36           36075           dev             lc              1               72530.383804999990.99375         100             7               5               debug           *               *               
1648069811      1679958340      1               day36           36075           guests          lc              1               72530.383804999990.025           1               7               5               debug           *               *               
1648069812      1679958340      1               day36           36075           lc              lc              1               526456.357985   0.26875         100             7               5               debug           *               *               

[root@fluke1:~]# flux account edit-user --queues= day36
[root@fluke1:~]# flux account view-user day36
creation_time   mod_time        active          username        userid          bank            default_bank    shares          job_usage       fairshare       max_running_jobsmax_active_jobs max_nodes       queues          projects        default_project 
1648069813      1679958403      1               day36           36075           dev             lc              1               72530.383804999990.99375         100             7               5                               *               *               
1648069811      1679958403      1               day36           36075           guests          lc              1               72530.383804999990.025           1               7               5                               *               *               
1648069812      1679958403      1               day36           36075           lc              lc              1               526456.357985   0.26875         100             7               5                               *               *               

[root@fluke1:~]# /bin/flux account-priority-update
[root@fluke1:~]# date
Mon Mar 27 16:15:59 PDT 2023
[root@fluke1:~]#
```
but I can still submit to debug:
```
[day36@fluke108:~]$ date && flux run -n1 --queue debug hostname
Mon Mar 27 16:18:09 PDT 2023
fluke96
[day36@fluke108:~]$
```

cmoussa1:
OK, I had a chance to write a reproducer in my Docker environment and the bug is as I expected; the `queues` list for a user in the plugin does not get cleared when it is cleared in the flux-accounting database. So, above, the user's `queues` list is empty in the DB, as confirmed by the output in `flux account view-user`, but the user's `queues` list still contains `debug` in its `bank_info->queues` list, which is why jobs still successfully submit to the queue.

[...]

PR #334, opened 2023-03-29
plugin: clear queues on flux-accounting DB update

cmoussa1:
#### Problem

As reported in #332 and #333, there are a couple of issues with how the plugin handles unpacking queues on a flux-accounting update.

1) If a user who has been added to a queue via `flux account edit-user --queue=_queue_name user_name_` is removed from that queue with `flux account edit-user --queue= _user_name_`, their access to the queue persists after running an `account-priority-update`.
2) If a queue is deleted with `flux account delete-queue ...`, and then `flux account-priority-update` is run, users who were not given access to it continue to be unable to submit to it.

This is because both the `queues` field for a user entry in the plugin and the `queues` map itself are not cleared before an update, so if queues are removed for a user or if queues are removed in the `queues_table` in the flux-accounting DB, they still remain in the plugin's internal data.

---

This PR adds a small helper function to the `edit-user` command that allows a `queues` list to be cleared for user by just passing `--queues=` to the command.

It also adds `.clear ()` calls to both of the helper functions that unpack `queues` attributes for each user/bank entry in the plugin and the `queues` map before receiving the newest update via `flux account-priority-update`. This should ensure consistency with the database even when queues are removed.

A new sharness test is also added that reproduces the steps outlined in issues #332 and #333. It creates a number of queues and a user who has access to all of these queues. It clears the queues for the user, sends an update to the plugin, and ensures that a user can no longer submit jobs to those queues after an update. It then removes the queues from the flux-accounting database, sends an update to the plugin, and ensures that users can submit jobs without being prevented by the plugin.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -195,6 +195,26 @@ def update_mod_time(conn, username, bank):
     conn.execute(update_stmt, mod_time_tup)
 
 
+def clear_queues(conn, username, bank=None):
+    if bank is None:
+        conn.execute(
+            "UPDATE association_table SET queues='' WHERE username=?", (username,)
+        )
+    else:
+        conn.execute(
+            "UPDATE association_table SET queues='' WHERE username=? AND bank=?",
+            (
+                username,
+                bank,
+            ),
+        )
+        update_mod_time(conn, username, bank)
+
+    conn.commit()
+
+    return 0
+
+
 ###############################################################
 #                                                             #
 #                   Subcommand Functions                      #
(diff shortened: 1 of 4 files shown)


cmoussa1:
@ryanday36 let me know your thoughts on this PR and if this fulfills the issues you reported in #332 and #333 whenever it's convenient for you! Let me know if I got the steps right in the sharness test added in this PR: `t1021-mf-priority-issue332.t` and if that's the behavior you expected. 🙂

ryanday36 (approved):
Looks good. Thanks for the quick work on this @cmoussa1

cmoussa1:
Thanks @ryanday36! Will set MWP on this

PR #338, opened 2023-04-07
flux-accounting service: change `Requires` to `BindTo`

### LINKED ISSUE #337: systemd: use `BindTo=flux.service` instead of `Requires=flux.service`
When upgrading a system to the latest flux-core, we noticed that flux was unexpectedly restarted by systemd, even after shutdown caused the broker to exit with the special exit code 42 defined in `RestartPreventExitStatus=42` (See flux-framework/flux-core#5072).

Through some debugging, @garlick was able to determine that the reason flux was restarted by systemd was because it is mentioned as a `Requires` by `flux-accounting`, which has `Restart=always`.

We found that if `Requires` is replaced with `BindsTo`, then things work more as expected, i.e. if flux is shutdown, then flux-accounting is also, and similarly when flux is started,  the flux-accounting service is started.

Let's make this change in the `flux-accounting.service` and get it into the next TOSS 4 release, since auto-restarting flux when the admins don't expect it will be frustrating.

If you do not want to generate a whole release for this small change, we can also apply a patch in the RPM specfile. Up to you @cmoussa1.

cmoussa1:
#### Problem

Flux can be restarted by systemd because it is mentioned as a `Requires` by flux-accounting, which has `Restart=Always`. We don't want Flux to be auto-restarted when the admins don't expect it.

---

This PR changes `Requires=flux.service` to `BindTo=flux.service`. This should allow things to behave more as expected, i.e if Flux is shut down, then flux-accounting is also, and similarly when Flux is started, the flux-accounting service is started.

Fixes #337

### DIFF
--- a/etc/flux-accounting.service.in
+++ b/etc/flux-accounting.service.in
@@ -1,6 +1,6 @@
 [Unit]
 Description=Flux accounting service
-Requires=flux.service
+BindTo=flux.service
 After=flux.service
 
 [Service]


garlick:
One thing I didn't try in my testing is seeing what happens when flux-accounting is installed on rank > 0, which I guess is required so that users can get to the tools that now work remotely.

Presumably if the flux-accounting unit is not enabled, then everything will be fine.  I'll try it though and report back ASAP

garlick:
No problem on rank > 0.  As expected when the unit is not enabled, it doesn't get auto-started when flux starts.  Which in retrospect is a "duh" :-)

grondo (approved):
Looks great! Thanks for the extra testing @garlick!

jameshcorbett:
@cmoussa1 isn't it `BindsTo` and not `BindTo`? https://www.freedesktop.org/software/systemd/man/systemd.unit.html#BindsTo=

cmoussa1:
Oh no, did I mess it up? I used `BindTo` since that was what was in the header of the issue in #337, but I see now that in the description of the issue it uses `BindsTo`. @grondo, should I post a fixup PR and create a new release for flux-accounting?

cmoussa1:
@garlick mentioned that he tested it but maybe he edited the `systemd` unit file himself and used the correct `BindsTo`.

grondo:
Hm, maybe `BindTo` is a synonym because it seems to be working. Though something else went wrong on fluke that I'm looking into now, I was able to start the flux-accounting service with the current unit file. :shrug: Probably best to fix the unit file, but generating a release doesn't seem to be necessary.

```
[root@fluke1:~]# systemctl show flux-accounting | grep Bind
BindsTo=flux.service
[root@fluke1:~]# systemctl status flux-accounting
● flux-accounting.service - Flux accounting service
   Loaded: loaded (/usr/lib/systemd/system/flux-accounting.service; disabled; v>
   Active: active (running) since Fri 2023-04-14 09:09:39 PDT; 4min 47s ago
 Main PID: 2025109 (python3)
    Tasks: 1 (limit: 99830)
   Memory: 13.2M
   CGroup: /system.slice/flux-accounting.service
           └─2025109 /usr/bin/python3 /usr/libexec/flux/cmd/flux-account-servic>

Apr 14 09:09:39 fluke1 systemd[1]: Started Flux accounting service.
```

grondo:
No problem, at least half my fault for missing it in review :facepalm:

I did notice that the flux-accounting service did not restart after a broker crash on fluke. We'll have to look into that and it may require some further adjustments of the unit file, fyi.

cmoussa1:
OK, that sounds good. I'll keep an eye out.

PR #341, opened 2023-04-14
flux-accounting service: change `BindTo` to `BindsTo`

cmoussa1:
#### Problem

The flux-accounting systemd unit file uses `BindTo` instead of `BindsTo`. @jameshcorbett  pointed out that `BindsTo` is the correct variable to use according to: https://www.freedesktop.org/software/systemd/man/systemd.unit.html#BindsTo=

---

This PR changes `BindTo` to `BindsTo` in the flux-accounting systemd unit file. Note that the flux-accounting service was still able to be started on fluke according to @grondo, so a new release doesn't seem to be necessary at the moment.

### DIFF
--- a/etc/flux-accounting.service.in
+++ b/etc/flux-accounting.service.in
@@ -1,6 +1,6 @@
 [Unit]
 Description=Flux accounting service
-BindTo=flux.service
+BindsTo=flux.service
 After=flux.service
 
 [Service]


jameshcorbett (approved):
LGTM!

cmoussa1:
Thanks! Setting MWP shortly

Issue #340, opened 2023-04-13
`view-user`: formatting of command output needs improvement


cmoussa1:
When the `view-user` command was first created, there were only a few columns to display for each user, so it made sense to print the information out for a user from the `association_table` horizontally. Now that the table has many more columns, each of them pretty useful to display, the output of the command on a terminal is really pretty ugly because it wraps around the width of the terminal (unless you don't mind looking at your terminal in, like, 6pt font 😆).

The output of this command should be reformatted and improved. Perhaps the output should be switched to print information vertically instead of horizontally. This might make output longer if a user belongs to multiple banks, but I think it would still be better then multiple user rows being wrapped around and jumbling output on a terminal.

cmoussa1:
This should be fixed by #342 now, so I'll close this.

PR #342, opened 2023-04-14
`view-user`: improve formatting of output of command

cmoussa1:
#### Problem

Mentioned in #340, the output of the `view-user` command is quite ugly because of how many columns there are in the `association_table`. Often times the output wraps around the terminal which makes it hard to look at.

---

This PR changes the format of the output of `view-user` to print user information vertically instead of horizontally. This way the output of the command looks cleaner and easier to look at.

It also slightly changes the tests that look for output of this command to use `grep -w` instead of calling `grep` multiple times. Passing the `-w` flag allows `grep` to look for multiple strings in the output, which was the original intention of the tests.

@ryanday36 let me know if this new output of the `view-user` command looks okay to you, or if you had another format in mind, I'd be happy to adjust it.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -84,8 +84,21 @@ def set_default_project(projects):
     return "*"
 
 
-def get_user_rows(headers, rows):
+def get_user_rows(headers, rows, parseable):
     user_str = ""
+
+    if parseable is True:
+        # find length of longest column name
+        col_width = len(sorted(headers, key=len)[-1])
+
+        for header in headers:
+            user_str += header.ljust(col_width)
+        user_str += "\n"
+        for row in rows:
+            for col in list(row):
+                user_str += str(col).ljust(col_width)
+
+        return user_str
 
     for row in rows:
         # iterate through column names of association_table and
@@ -217,7 +230,7 @@ def clear_queues(conn, username, bank=None):
 #                   Subcommand Functions                      #
 #                                                             #
 ###############################################################
-def view_user(conn, user):
+def view_user(conn, user, parseable=False):
     cur = conn.cursor()
     try:
         # get the information pertaining to a user in the DB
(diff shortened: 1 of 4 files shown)


ryanday36 (approved):
This looks good and should improve readability. The one thing that I have some reservations about is that parsing the output with another script is generally a bit cleaner when everything is on one line. Maybe a flag (e.g. --parseable) could be added that would toggle the formatting of the output back to putting all of the information on one line.

grondo:
@ryanday36: would a `--json` option be useful for these commands, or would that not work for your use case? Seems like that is usually a good alternative to `--parseable` or `--machine-readable` options these days.

cmoussa1:
OK! I can do that. 🙂 Let me add a `--json` option to this command too.

cmoussa1:
OK, I think I've successfully added a `--json` option (thanks for the suggestion @grondo!) to the `view-user` command. Let me know what you think @ryanday36, the format looks like this (note that this list will have multiple entries if a user belongs to more than one bank):

```json
[
    {
        "creation_time": 1681750380,
        "mod_time": 1681750380,
        "active": 1,
        "username": "user5014",
        "userid": 5014,
        "bank": "C",
        "default_bank": "C",
        "shares": 1,
        "job_usage": 0.0,
        "fairshare": 0.5,
        "max_running_jobs": 5,
        "max_active_jobs": 7,
        "max_nodes": 2147483647,
        "queues": "",
        "projects": "*",
        "default_project": "*"
    }
]
```

grondo:
Yes, although @ryanday36 should comment if that format is too complex for parsing from a script in his use case.

However, the more structured output allows for easier expansion in the future, and reduces duplication of items that are the same for every bank (username, userid, default_bank, etc)

ryanday36:
This looks good to me.

cmoussa1:
Thanks @ryanday36! Setting MWP on this shortly

cmoussa1:
Just rebased to catch up after #330. Setting MWP

PR #343, opened 2023-04-17
`update-db`: fix statement when updating a table with no primary key

cmoussa1:
#### Problem

The `add_stmt` used to add tables to an old version of a flux-accounting DB breaks when a table does not have a primary key. The generated statement used to update the table ends in a comma with nothing else after it, which results in a `sqlite3.OperationalError`.

---

This PR changes the construction of the `add_stmt` to only add a comma if the column is not the last one in the table. It moves the outer parentheses and semi-colon to the very end of the `add_stmt` construction instead of inside the for-loop where a PRIMARY KEY is generated. It also includes some minor cleanup that removes some unused variables from a couple function definitions.


ryanday36 (approved):
looks good.

cmoussa1:
Thanks! Setting MWP

PR #344, opened 2023-04-27
test: replace "flux mini" usage

cmoussa1:
#### Problem

`flux mini CMD` is deprecated in favor of `flux CMD`.

---

This PR replaces the old usage in the test suite.


cmoussa1:
Thanks for the quick review @chu11!

PR #358, opened 2023-06-27
t: add a Python test directory to `t/`

cmoussa1:
#### Background

flux-accounting doesn't have a directory to run Python tests in `t/`. Having one would be useful to run unit tests on some of the front-end Python scripts.

---

This PR copies the `t/python/` directory from flux-core and adds it to flux-accounting's `t/` directory. It also adds an example unit test to the directory to make sure it works.

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -23,7 +23,8 @@ TESTSCRIPTS = \
 	t1019-mf-priority-info-fetch.t \
 	t1020-mf-priority-issue262.t \
 	t1021-mf-priority-issue332.t \
-	t1022-mf-priority-issue346.t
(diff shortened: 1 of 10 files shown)


chu11:
> Add an example unit test to the directory to make sure it works.

IIWM I'd take the t1000 example and the Makefile change and make that one new commit together.  So `t/python` is one commit, then the example is another commit.

cmoussa1:
@chu11, good point, thanks. Could you give my most recent push another look to make sure it looks like what you suggested?

chu11 (approved):
LGTM

cmoussa1:
thanks! setting MWP

PR #359, opened 2023-06-29
`view-bank`: improve `-t` option

cmoussa1:
#### Problem

There is no way to see an overarching view of a flux-accounting hierarchy without running `flux account-shares`, which includes job usage values for banks and users all in one place. It would be nice if you could see this view with a Python command.

---

This PR edits the `-t` option of the `view-bank` command to more closely match the output of the `flux account-shares` command in Python, which includes job usage values for banks and fair share values for users all in one place. The new format of the `-t` option is the following:

```console
$ flux account view-bank -t root
bank_id        bank           active         parent_bank    shares         job_usage      
1              root           1                             1              0.0            

Bank                            Username           RawShares            RawUsage           Fairshare
root                                                       1                 0.0
 A                                                         1                 0.0
  A                             user5011                   1                 0.0                 0.5
  A                             user5012                   1                 0.0                 0.5
 B                                                         1                 0.0
  B                             user5013                   1                 0.0                 0.5
 C                                                         1                 0.0
  C                             user5014                   1                 0.0                 0.5
 D                                                         1                 0.0
  E                                                        1                 0.0
  F                                                        1                 0.0


```

This option is slightly more extensible than `flux account-shares` because it can generate a hierarchy from any bank in the flux-accounting database, not just the root bank. So, if you only want to see a sub-tree in the overall hierarchy, you can just pass that bank in the `view-bank -t` command.

It also adds a new column to the `bank_table` which stores job usage; this gets updated when all of the job usage values for all users also get updated.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -129,6 +129,51 @@ def reactivate_bank(conn, cur, bank, parent_bank):
     conn.commit()
 
 
+def print_hierarchy(cur, bank, hierarchy_str, indent=""):
+    # look for all sub banks under this parent bank
+    select_stmt = "SELECT bank,shares,job_usage FROM bank_table WHERE parent_bank=?"
+    cur.execute(select_stmt, (bank,))
+    sub_banks = cur.fetchall()
+
+    if len(sub_banks) == 0:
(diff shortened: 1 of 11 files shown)


ryanday36 (approved):
This looks like what I want. Thanks @cmoussa1!

cmoussa1:
cool cool! i'll set MWP on this

PR #361, opened 2023-06-30
flux-accounting DB: update schema version

### LINKED ISSUE #360: database: update DB schema version after #359
this is just a reminder to myself to open a PR that updates the DB schema version after #359 has landed since the `bank_table` has been updated

cmoussa1:
#### Background

The schema for the flux-accounting database has been changed as a result of #359. The schema version number thus needs to be updated.

---

This PR updates the schema version number for the database in `__init__.py.in` and in `flux-account-service.py`. This means that any existing flux-accounting database will need to be updated with `flux account-update-db` before running new commands. The periodically-run update script, `flux account-priority-update`, will call this update-db script if it hasn't already been run.

A version of the flux-accounting DB before #359 landed has also been added to the test suite so that it can be run by `t1017-update-db.t` during `make check`.

Fixes #360


chu11 (approved):
LGTM

cmoussa1:
thanks! setting MWP

PR #363, opened 2023-07-13
build: add `make deb` target for test packaging

garlick:
Problem: We have `make deb` for testing flux-core and flux-sched, but not flux-accounting.

Copy and the scripts and metadata added by Mark Grondona to flux-core in  flux-framework/flux-core@a0b0bad271309579a9f7882d20e64e566e29a4f8 and customize them for this project.

I used this to install flux-accounting on my test cluster at home and it seems OK.


cmoussa1 (approved):
LGTM! Thanks @garlick! Just noticed one nit in the commit message that maybe you could drop the "and" in the commit message, i.e.

`"Copy and the scripts..."` --> `"Copy the scripts..."`

garlick:
Oops, thanks!  Will fix.

I forgot to add python3-devel to the dependencies. I didn't spot any other dependent python packages in the autoconf.ac.  Are there any?

cmoussa1:
> I didn't spot any other dependent python packages in the autoconf.ac. Are there any?

Here is the list of `BuildRequires` packages listed in flux-accounting's `.spec` file (sorry if this isn't related to your question, figured I'd be safe than sorry):

```
pkgconfig{jannson} >= 2.6
python36
python3-devel
python3-cffi
python3-six
python3-yaml
python3-jsonschema
sqlite-devel
flux-core
```

flux-accounting also requires `sqlite >= 3.6.0`

garlick:
Perfect, thanks!

There is some logic missing in configure.ac for these packages.  I'll open a separate issue.

grondo:
```
python3-devel
python3-cffi
python3-six
```

Are these really requirements? I don't see `import six` or `ffi` anywhere? And I don't see a python devel check in configure (after a quick look though)

cmoussa1:
To be honest I'm not sure if they are actually requirements. I think when I originally created the `.spec` file I just copied what flux-core had, and I don't think flux-accounting uses `six` or `ffi` anywhere.

grondo:
Yeah, not even flux-core requires `six` anymore since that is just for Python 2/3 compatibility and we don't support Python 2 anymore. If you don't use cffi or `Python.h` anywhere then probably the other 2 requirements can be dropped as well ...

garlick:
Is sqlite3 required for the command line `sqlite3` program?  It looks like python support for sqlite3 is built-in to python3.

cmoussa1:
You are right that `sqlite3` is built-in to Python, but yes, I believe sqlite3 is required for the `sqlite3` program and CLI, which is useful in some cases for the flux-accounting database, so if it is okay, I would like to keep that as a requirement.

garlick:
I didn't see python3-yaml or python3-jsonschema in use either.  Just pushed an update which lists the following
```
Build-Depends:
  debhelper (>= 10),
  flux-core,
  libjansson-dev,
  libsqlite3-dev,
  sqlite3
```
Is that getting closer to reality?

Should python3 be in that list?  I didn't see it in flux-core's list.

cmoussa1:
If python3 doesn't appear in flux-core's list, than I am assuming flux-accounting won't need it either. I have to admit I am very unfamiliar with package requirements, so I'm sorry for misdirecting you with false requirements!

garlick:
This is only for test packages so it's not super critical that we get everything perfect here.  However it's probably good to get the RPM requirements right if they need tweaking, and add any missing checks to configure.ac too.  I'll open an issue so that can get fixed whenever it is convenient.

cmoussa1:
OK, that sounds like a good plan, thanks 🙂

grondo:
Hm, I wonder if we should have python3 in flux-core requirements. I guess it is implied by other python3 packages like python3-yaml, python3-cffi -- and I guess flux-accounting gets python3 transitively from that. Also python3 may be part of build-essential so maybe it doesn't _need_ to be listed..

garlick:
I'm OK with just deferring that issue for someday when we submit upstream packages (which we definitely should do at some point!).  Maybe some linter will tell us everything we did wrong at that point anyway.

grondo:
Ok, well if that's the case then I wouldn't worry about the other requirements either since they'll be satisfied by requiring flux-core.

garlick:
OK, well maybe this can go in then.  The dependencies are a pretty good guess anyway (barring the things that are likely to be part of a baseline, like you said).

Thanks!

garlick:
@Mergifyio refresh

Issue #365, opened 2023-07-13
config.h include is missing from source code


garlick:
Problem: just happened to notice that `config.h` is not included in many of the cpp source modules.  It's a good practice in autotools projects to include that right at the top of all (non-vendored) `.c` and  `.cpp` source files, immediately following the copyright header, e.g.

```c
#if HAVE_CONFIG_H
#include "config.h"
#endif
```

One reason is that sometimes configure defines macros that affect how other header files are interpreted.  That is why it needs to go first.

cmoussa1:
thanks for pointing this out! i noticed i do have something like this already in the priority plugin, but nowhere else:

```cpp
extern "C" {
#if HAVE_CONFIG_H
#include "config.h"
#endif
}
```

I'm assuming this is accomplishing the same thing?

garlick:
Yep, although the `extern "C" {}` part isn't doing anything AFAIK and should probably be dropped.

garlick:
Oh sorry, double `include "config.h"` is correct not `include <config.h>`.  Oops!

cmoussa1:
Closing this now that #366 has landed.

trws:
The extern C part is necessary if it can ever include any function declaration or include of any other header.  If that's missing, it should come back.

chu11:
Adding comment I put in https://github.com/flux-framework/flux-sched/issues/1050

> I think an audit that #include <flux/core.h> (and other flux stuff, and maybe also jansson.h) is also within the extern "C" might be worthwhile while we're at it. A cursory glance is we're doing it correctly in flux-sched, but I saw atleast one file in accounting that was missing it.

src/plugins/mf_priority.cpp doesn't seem to have `extern C` surround the config.h and flux headers.

cmoussa1:
After some offline conversation with @chu11 and following the thread in https://github.com/flux-framework/flux-sched/issues/1050, it looks like I should wrap `config.h` in `extern C` as a just-in-case, as well as the flux-core headers that are referenced in the multi-factor priority plugin. I can open a PR shortly with these changes!

PR #366, opened 2023-07-14
.cpp: add config.h include to source code

cmoussa1:
#### Problem

As mentioned in #365, `config.h` is not included in many of the `.cpp` source modules. Sometimes configure defines macros that affect how other header files are interpreted.

---

This PR adds `config.h` include statements in the `.cpp` source files that do not already have them.

mf_priority.cpp already had this include, but it also had an `extern "C"` preface, which is not needed, so this PR also drops that.

### DIFF
--- a/src/fairness/account/account.cpp
+++ b/src/fairness/account/account.cpp
@@ -7,6 +7,9 @@
  *
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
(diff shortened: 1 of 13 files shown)


garlick (approved):
LGTM!

cmoussa1:
thanks! setting MWP

PR #367, opened 2023-07-17
t: reorganize `t1007-flux-account.t` into multiple sharness tests

cmoussa1:
#### Problem

`t1007-flux-account.t` has grown very large because it includes tests for almost every `flux account` command. As a result, it is unorganized and hard to read (my fault!!).

---

This PR reorganizes the sharness test file and creates multiple test files according to the category of `flux account` commands it deals with. So, `t1007-flux-account.t` becomes the test file that's responsible for testing all user-related commands; it is now `t1007-flux-account-users.t`. New test files are created for the `bank`, `queue`, and `project` commands. There is also a new file that tests the permissions for the flux-accounting commands that should only be available to administrators.

This reorganization allows for an easier time to add/edit tests for the suite of `flux account` commands if necessary.


t/t1007-flux-account-users.t
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
+
+test_description='Test flux-account commands'
```
chu11, line 3: user commands?

t/t1023-flux-account-banks.t
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+test_description='Test flux-account commands'
```
chu11, line 3: bank? seems needs update in all new files :-)

t/t1025-flux-account-projects.t
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
+
```
chu11, line 2: commit message, why `mv` and not `move` like other messages?  NBD, thought maybe a typo.

t/t1007-flux-account-users.t
```diff
@@ -0,0 +1,114 @@
...
+	flux account-service -p ${DB_PATH} -t
+'
+
+test_expect_success 'add some banks to the DB' '
+	flux account add-bank root 1 &&
+	flux account add-bank --parent-bank=root A 1 &&
+	flux account add-bank --parent-bank=root B 1 &&
+	flux account add-bank --parent-bank=root C 1 &&
+	flux account add-bank --parent-bank=root D 1 &&
+	flux account add-bank --parent-bank=D E 1
```
ryanday36, line 29: should there be a '&&' at the end of this line?

chu11 (approved):
Other than some nits, nothing huge.

I thought I'd comment that the way that you split up the commits does have a minor side effect.  Basically after the first commit, technically you have "lost tests" in the commit history.  i.e. for a small moment in commit history, these tests disappeared.  Then in the middle few commits you added new test files but didn't add them to the Makefile.  So for a small moment in history the re-added tests probably won't run when running `make check`.

If you were hunting down some bug that was introduced awhile back, this could be a nuisance.  i.e. `make check` would not run tests that were there before.

I might suggest a commit history like:

`t: split bank tests out of t1007....` (add this new file to Makefile.am)
`t: split queue tests out of t1007...` (and add new file to Makefile.am)
...
`t: rename t1007 to ...`

that way the history is never lost momentarily and `make check` works on each commit.

It's not a huge deal, so I leave it to you if you'd like to do.

t/t1023-flux-account-banks.t
```diff
@@ -0,0 +1,128 @@
...
+	flux account-service -p ${DB_PATH} -t
+'
+
+test_expect_success 'add some banks' '
+	flux account add-bank root 1 &&
+	flux account add-bank --parent-bank=root A 1 &&
+	flux account add-bank --parent-bank=root B 1 &&
+	flux account add-bank --parent-bank=root C 1 &&
+	flux account add-bank --parent-bank=root D 1 &&
+	flux account add-bank --parent-bank=D E 1
```
ryanday36, line 29: same as last file, should this line end in '&&'?

t/t1024-flux-account-queues.t
```diff
@@ -0,0 +1,121 @@
...
+	flux account-service -p ${DB_PATH} -t
+'
+
+test_expect_success 'add some banks to the DB' '
+	flux account add-bank root 1 &&
+	flux account add-bank --parent-bank=root A 1 &&
+	flux account add-bank --parent-bank=root B 1 &&
+	flux account add-bank --parent-bank=root C 1 &&
+	flux account add-bank --parent-bank=root D 1 &&
+	flux account add-bank --parent-bank=D E 1
```
ryanday36, line 29: &&

t/t1024-flux-account-queues.t
```diff
@@ -0,0 +1,121 @@
...
+'
+
+test_expect_success 'remove a queue' '
+	flux account delete-queue special &&
+	test_must_fail flux account view-queue special > deleted_queue.out 2>&1 &&
+	grep "queue special not found in queue_table" deleted_queue.out
+'
+
+test_expect_success 'add a queue with no optional args to the queue_table' '
+	flux account add-queue queue_1
```
ryanday36, line 79: &&

t/t1025-flux-account-projects.t
```diff
@@ -0,0 +1,136 @@
...
+	flux account-service -p ${DB_PATH} -t
+'
+
+test_expect_success 'add some banks to the DB' '
+	flux account add-bank root 1 &&
+	flux account add-bank --parent-bank=root A 1 &&
+	flux account add-bank --parent-bank=root B 1 &&
+	flux account add-bank --parent-bank=root C 1 &&
+	flux account add-bank --parent-bank=root D 1 &&
+	flux account add-bank --parent-bank=D E 1
```
ryanday36, line 29: &&

ryanday36 (approved):
Looks good to me overall. There were a few lines that I *think* should have ended in '&&', but maybe not. Feel free to ignore those comments if I'm wrong.

My only other comment would be to update the 'test_description' line in the various files to describe the specific tests in the file (e.g. 'Test flux-account bank commands' rather than 'Test flux-account commands' in 'f1023-flux-account-banks.t').

t/t1007-flux-account-users.t
cmoussa1: Actually, my understanding of the `&&` is to signal that there are additional commands to be run on the next line, so there isn't a need to put the `&&` at the end of the last command in the test.

t/t1007-flux-account-users.t
cmoussa1: Yeah, all of these test descriptions should be updated/made clearer. Thanks for pointing that out!

t/t1025-flux-account-projects.t
cmoussa1: I used `mv` here to fit the commit title in the (is it 50?) character limit for the title. I can update the titles for these commands so that they are all the same and fit in the character limit. Sorry if that caused any confusion!

chu11: ahh the 50 char limit, it's fine then ... i just thought maybe you typoed it

PR #368, opened 2023-08-02
`.cpp`: wrap `"config.h"`, C headers in `extern "C"`

cmoussa1:
#### Problem

There are C++ files in this repo that include headers from C files, but they are not wrapped in "extern C", which could cause some unwanted mucky behavior and make it tricky to diagnose link issues.

---

Wrap the `.cpp` files in this repository that include `config.h` in `extern "C"`. In `src/plugins/mf_priority.cpp`, also wrap `flux/core.h`, `flux/joptap.h`, and `jansson.h` in `extern "C"`.

### DIFF
--- a/src/fairness/account/account.cpp
+++ b/src/fairness/account/account.cpp
@@ -7,9 +7,11 @@
  *
  * SPDX-License-Identifier: LGPL-3.0
 \************************************************************/
+extern "C" {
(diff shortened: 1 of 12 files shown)


chu11 (approved):
LGTM

cmoussa1:
Thanks! Setting MWP

PR #369, opened 2023-08-08
docker: transition `bionic` container to `jammy`

cmoussa1:
### background

Like flux-sched, flux-accounting also uses a `bionic` container as a part of its CI checks, but the Ubuntu release is no longer under LTS standard support.

---

This PR transitions the `bionic` image to `jammy` in flux-accounting and removes references of `bionic` in the `docker/` directory's README.md.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -5,7 +5,7 @@ queue_rules:
       - status-success="validate commits"
       - status-success="python format"
       - status-success="python lint"
-      - status-success="bionic - py3.6"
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
LGTM!

cmoussa1:
thanks!! setting MWP

cmoussa1:
Oh, I think I just remembered that this probably requires a manual merge since it makes changes to `.mergify.yml`. I can press the merge button

PR #371, opened 2023-08-30
[experimental] plugin: add bank info lookup helper function

cmoussa1:
#### Problem

There are multiple spots in the plugin code where we need to look up user/bank information, and the code isn't super easy to read because a map iterator needs to be declared and used to look up the information in a map. IMHO, It might be easier to read and understand what is going on if this lookup was moved to a helper function.

---

This PR is just a proposal to see if others would agree that this might be an improvement - if others feel it looks/reads better as before, then I can just close this. 🙂 It creates a helper function, `get_bank_info ()`, that accepts a `bank_info` map struct iterator, userid, and a bank parameter and performs this lookup in the map. It returns a pair: a return code as well as the iterator to the user/bank map.

In the spots where this new helper function is called, the same logic is retained that handles validating the user/bank information passed in, i.e rejecting the job or raising a job exception because the user/bank information passed in cannot be validated.

It might also be helpful to have a general validation function like this as a result of the conversation in flux-framework/flux-core#5409 since there's talk of potentially having a callback for job updates; it could be helpful to just call this function. 🤷‍♂️

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -45,6 +45,19 @@ extern "C" {
 // the association does not have permission to run jobs under
 #define INVALID_QUEUE -6
 
+// different codes to return as a result of looking up user/bank information:
+//
+// BANK_SUCCESS: we found an entry for the passed-in user/bank
+// INVALID_BANK: the user specified a bank they don't belong to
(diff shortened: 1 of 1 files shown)


cmoussa1:
Think I'm gonna close this PR; I'm finding the helper function I wrote pretty useful (with some very slight adjustments), so gonna paste it as a commit to another PR later.

PR #373, opened 2023-08-31
t: add valgrind folder to flux-accounting

cmoussa1:
#### Problem

flux-accounting currently has no way to run tests under valgrind, but it is useful to run tests in this mode, especially when working with the multi-factor priority plugin. I've been using it while developing on my local machine but never added it officially to the repository. 

---

This PR just copies the `t/valgrind/` folder from flux-core and adds it to flux-accounting.


chu11:
don't forgot to copy over t5000-valgrind.t as that is what actual runs it.  Some tweaking to the file may be necessary.

cmoussa1:
> don't forgot to copy over t5000-valgrind.t as that is what actual runs it. Some tweaking to the file may be necessary.
>
> and possibly, checks need to be done to see if valgrind is installed. Can't remember how that's done off the top of my head.

I could definitely be wrong but including this folder has been all I _think_ I've needed to be able to run any of the flux-accounting tests under valgrind. The tests in `t5000-valgrind.t` get skipped when I include it in the flux-accounting test suite because there is no `valgrind.h` file (and I can't seem to find one to copy in flux-core using the GitHub search :/ ).

chu11:
You probably need to add a check to find valgrind.h in your configure script.  Check out `AX_VALGRIND_H` in flux-core.  That should find the header for you and then t5000 should work.

cmoussa1:
Ah, thanks for that clarification. I think I see what you mean. I've edited this PR to add that check for `valgrind.h`; I basically just copied what flux-core has in their repo. I've added `t5000-valgrind.t`, and I've also edited the `workload.d` folder to have one basic flux-accounting-specific test that just loads the multi-factor priority plugin. I believe that valgrind test runs now (screenshot from GitHub actions):

image

t/valgrind/workload.d/job
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+NJOBS=${NJOBS:-10}
+
+flux submit -n 1 hostname
```
chu11, line 5: it seems you dropped the NJOBS carbon copy when you copied the file over here.  Perhaps testing/playing around and forget to put that back?

Alternate idea.  Instead of adding `mf-priority-plugin-load` in the next commit, just update this test to load the mf-priority-plugin, so now `workload.d/job` is flux-accounting specific.   Also the `--wait` option should be kept so that the job completes before returning.

cmoussa1: Oh okay! I like that suggestion. Will do that

cmoussa1: OK, just force-pushed a change to this PR to just have one commit and updates the test to load the priority plugin (I also removed the `job-cancel` test because that didn't have anything flux-accounting specific in it), and added back the `--wait` option.

chu11 (approved):
LGTM!  Although I would tweak the commit message header and the PR header to be a tad more generic, like "add valgrind testing" or whatever.  Since it's not just copying over a folder.

cmoussa1:
Thanks @chu11! I modified the commit message a bit; setting MWP

PR #375, opened 2023-09-05
ci: update github actions `main.yml` file

cmoussa1:
#### Problem

flux-accounting's GitHub actions uses a number of near-deprecated/deprecated actions that show warnings when running CI.

---

This PR just updates the workflow file for CI for flux-accounting to use newer versions of a number of actions. I've mostly tried to just copy what flux-core has in their workflow file.

One point of clarification I think I might need is on this proposed change:

```diff
- run: echo "::set-output name=tarball::$(echo flux-accounting*.tar.gz)"
+ run: echo "$(echo flux-accounting*.tar.gz)"
```

Is the new change correct? I couldn't find the flux-core equivalent via quick GitHub search.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
     name: validate commits
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 0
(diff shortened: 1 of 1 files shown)


grondo:
I'd actually update the release to `softprops/action-gh-release@v1` which we're using in flux-core. Using this action is much simpler than what we had before, and doesn't require setting the tarball in an output variable. For reference, here's the action usage in flux-core:
```yaml
    - name: create release
      id: create_release
      if: |
        success()
        && matrix.create_release
        && github.repository == 'flux-framework/flux-core'
      env: ${{matrix.env}}
      uses: softprops/action-gh-release@v1
      with:
        tag_name: ${{ matrix.tag }}
        name: flux-core ${{ matrix.tag }}
        prerelease: true
        files: flux-core*.tar.gz
        body: |
          View [Release Notes](https://github.com/${{ github.repository }}/blob/${{ matrix.tag }}/NEWS.md) for flux-core ${{ matrix.tag }}
```

cmoussa1:
Oh, thanks! That's helpful. I've force-pushed a change to this PR to use the suggested `create-release` action you've pasted, adapted for flux-accounting.

grondo (approved):
A few minor comments inline, but looks good to me.

If you wanted you could probably collapse all these changes into one commit `ci: update github actions`, but I leave that up to you.

cmoussa1:
Looks like things are working as expected, so I'll set MWP here!

PR #378, opened 2023-09-06
plugin: check for `FLUX_JOB_STATE_NEW` in `validate_cb()`

cmoussa1:
#### Problem

The multi-factor priority plugin checks a user/bank's active jobs limit when a job enters `job.validate` and will reject a job if the user/bank is currently at their max active jobs limit. Now that Flux supports job updates, previously validated jobs can re-enter `job.validate`, which breaks the plugin's assumptions that `job.validate` is only for newly submitted jobs, particularly with this check for a user/bank's active jobs count.

---

This PR adds an unpack for the **state** of a job in `job.validate` and check this state for `FLUX_JOB_STATE_NEW` *before* checking for a user/bank's active jobs count to protect it from jobs that were already validated by the priority plugin.

It also adds a small number of tests that validates the priority plugin can support job updates for job duration while maintaining the integrity of a user/bank's active job count.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -678,6 +678,7 @@ static int validate_cb (flux_plugin_t *p,
     int userid;
     char *bank = NULL;
     char *queue = NULL;
+    flux_job_state_t state;
     int max_run_jobs, cur_active_jobs, max_active_jobs = 0;
     double fairshare = 0.0;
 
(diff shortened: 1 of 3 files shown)


grondo (approved):
Nicely done! LGTM!

cmoussa1:
Thanks! Setting MWP on this

grondo:
I wonder if we'll want to get a release of flux-accounting tagged to go out with the upcoming version of flux-core. While the error this PR fixes would be rare, the error message will be confusing and users and admins may find the behavior a bit annoying.

I'll give #294 a review and see if we can get that in also?

cmoussa1:
It's up to you - not sure how urgent #294 is at the moment but I agree that landing this and then creating a release for flux-accounting would be worth it!

PR #383, opened 2023-09-10
testsuite: load content module in rc scripts

garlick:
Problem: flux-core is migrating the content service from the broker to a broker module.

Load the content module in test rc scripts.

Ignore any errors from 'flux module load' so that flux-accounting continues to work with older flux-core versions.

I would hold off merging until it looks like flux-framework/flux-core#5435 is going to be accepted.

### DIFF
--- a/t/rc/rc1-job
+++ b/t/rc/rc1-job
@@ -28,6 +28,7 @@ modload() {
 }
 
 
+modload all content || :
 modload 0 content-sqlite
 modload all kvs
 modload all kvs-watch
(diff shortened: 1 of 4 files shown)


cmoussa1 (approved):
LGTM! Thanks @garlick. Looks like flux-framework/flux-core#5435 was approved too; did you want this to be merged first (so that the flux-accounting check on the flux-core PR passes)?

garlick:
As discussed on :coffee: I did try dropping the `job` personality from accounting sharness tests and that did not go well.  I'd say let's just go with the proposed change.  Since you already approved, I'll just set MWP.

Thanks!

PR #386, opened 2023-09-13
plugin: improve check of internal user/bank map in `job.validate`

### LINKED ISSUE #385: plugin: subsequently-submitted jobs get rejected even when plugin has only "DNE" data
I discovered a bug last night in the multi-factor priority plugin where even if it only has the special does-not-exist (`"DNE"`) entries for at least one user/bank, subsequently submitted jobs get rejected with the `"no bank found for user"` error message because the plugin thinks that it has actual data.

```c++
if (users.empty ())
```

I think this can be fixed with an expanded check in `validate_cb ()`. This check should look to see if the plugin's internal user/bank data map is empty _or_ if the map only contains the special `"DNE"` entries. In this case, multiple users can submit jobs to the priority plugin when it has no actual flux-accounting data and they will all be held in state `PRIORITY` while it waits for data:

```c++
if (users.empty () || only_dne_data)
```

I think I was able to confirm this with a set of tests.

```shell
test_expect_success 'submit a job with no user/bank info loaded to plugin' '
	jobid1=$(flux python ${SUBMIT_AS} 1001 --wait-event=depend sleep 60) &&
	test $(flux jobs -no {state} ${jobid1}) = PRIORITY
'

test_expect_success 'submit a job as another user, check that it is also in state PRIORITY' '
	jobid2=$(flux python ${SUBMIT_AS} 1002 --wait-event=depend sleep 60) &&
	test $(flux jobs -no {state} ${jobid2}) = PRIORITY
'
```

Then, when the plugin gets loaded with actual flux-accounting data, the jobs transition to state `RUN`. I'll plan on posting a PR today with this fix.

cmoussa1:
#### Problem

As noted in #385, in `validate_cb ()`, the multi-factor priority plugin has a bug where even if it only has the special does-not-exist (`"DNE"`) entries for at least one user/bank (indicating that the plugin has not received any flux-accounting data yet), subsequently submitted jobs by other users get rejected with the `"no bank found for user"` error message, which is the rejection message typically output when the plugin has valid user/bank data.

---

This PR adds a helper function to the priority plugin that scans the users map and checks the default bank for each user. As soon as it finds a user/bank with a default bank that is not the special `"DNE"` entry, it returns `false`. This helper function is called in the `job.validate` callback to ensure that the plugin's internal user/bank map contains valid user/bank data before
rejecting a user's job because they don't have a valid entry in the flux-accounting database.

It also adds a small set of tests that exercise this functionality of submitting multiple jobs from multiple users and checking that they are held in state `PRIORITY` before any data is loaded to the plugin; then, valid data is sent to the plugin and the jobs are checked to confirm that they transition to `RUN`. It also adds a test to make sure that a user that **does not** have a valid entry in the flux-accounting DB cannot submit jobs to the plugin.

Fixes #385

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -331,6 +331,25 @@ error:
 }
 
 
+// Scan the users map and look at each user's default bank to see if any one
+// of them have a valid bank (i.e one that is not "DNE"; if any of the users do
+// do have a valid bank, it will return false)
+static bool check_map_for_dne_only ()
+{
+    // the users map iterated through in this for-loop, along with the
+    // users_def_bank map used to look up a user's default bank, are
+    // both global variables
+    for (const auto& entry : users) {
+        auto def_bank_it = users_def_bank.find(entry.first);
+        if (def_bank_it != users_def_bank.end() &&
+                def_bank_it->second != "DNE")
+            return false;
+    }
+
+    return true;
+}
+
+
 /******************************************************************************
  *                                                                            *
  *                               Callbacks                                    *
@@ -681,6 +700,7 @@ static int validate_cb (flux_plugin_t *p,
     flux_job_state_t state;
     int max_run_jobs, cur_active_jobs, max_active_jobs = 0;
     double fairshare = 0.0;
+    bool only_dne_data;
 
     std::map<int, std::map<std::string, struct bank_info>>::iterator it;
     std::map<std::string, struct bank_info>::iterator bank_it;
@@ -708,7 +728,10 @@ static int validate_cb (flux_plugin_t *p,
     // the plugin.
     it = users.find (userid);
     if (it == users.end ()) {
-        if (users.empty ()) {
+        // check if the map only contains DNE entries
+        bool only_dne_data = check_map_for_dne_only ();
+
+        if (users.empty () || only_dne_data) {
             add_missing_bank_info (p, h, userid);
             return 0;
         } else {
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -29,6 +29,7 @@ TESTSCRIPTS = \
 	t1025-flux-account-projects.t \
 	t1026-flux-account-perms.t \
 	t1027-mf-priority-issue376.t \
+	t1028-mf-priority-issue385.t \
 	t5000-valgrind.t \
 	python/t1000-example.py
 
--- a/t/t1028-mf-priority-issue385.t
+++ b/t/t1028-mf-priority-issue385.t
@@ -1,0 +1,84 @@
+#!/bin/bash
+
+test_description='Test submitting jobs to queues after queue access is changed'
+
+. `dirname $0`/sharness.sh
+
+mkdir -p conf.d
+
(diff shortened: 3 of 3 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -331,6 +331,25 @@ static json_t *user_to_json (
 }
 
 
+// Scan the users map and look at each user's default bank to see if any one
+// of them have a valid bank (i.e one that is not "DNE"; if any of the users do
+// do have a valid bank, it will return false)
+static bool check_map_for_dne_only ()
+{
+    // check if the map only contains DNE entries
+    bool only_dne_data = true;
```
grondo, line 340: This variable appears to be unused

src/plugins/mf_priority.cpp
```diff
@@ -331,6 +331,25 @@ static json_t *user_to_json (
 }
 
 
+// Scan the users map and look at each user's default bank to see if any one
+// of them have a valid bank (i.e one that is not "DNE"; if any of the users do
+// do have a valid bank, it will return false)
+static bool check_map_for_dne_only ()
```
grondo, line 337: A really quick thought here before a full review.

If this function instead returned the default bank for a user or DNE if the user has no default bank, would it potentially be useful elsewhere? Also, it might be more clear if the users map was passed as a parameter to the function...

grondo (commented):
Just a quick couple early comments/questions. Haven't fully reviewed yet.

src/plugins/mf_priority.cpp
cmoussa1: Whoops, thanks for catching this. I must have missed that when I was moving this block of code to a helper function. Will fix this

src/plugins/mf_priority.cpp
cmoussa1: I think there are a couple of other places where the default bank for a user is queried, but I had actually posted an experimental PR (#371) where the bank lookup was moved to its own helper function. I had originally thought that it would be useful to have functionality like the one proposed in this PR (just scanning the users map for the first non-DNE bank it sees) kept separate, but perhaps it would be better to combine the two? Off the top of my head, maybe in the check where it looks for a default bank, it could add something like:

```c++
else {
        bank = const_cast<char*> (users_def_bank[userid].c_str ());
        bank_it = it->second.find (std::string (bank));
+       if (it->second == "DNE")
+           return {DNE_DEFAULT_BANK, bank_it};
        else if (bank_it == it->second.end ())
            return {NO_DEFAULT_BANK, bank_it};
    }
```

> Also, it might be more clear if the users map was passed as a parameter to the function...

The `users` map is a global variable in the plugin. Would it be sufficient to add a comment in this function to mention this?

grondo: Yeah, a comment would definitely help!

cmoussa1: just force-pushed up a change to add a comment about both the `users` and `users_def_bank` maps since they are both global variables.

grondo (approved):
Ok, this looks good and we should get it merged since it fixes an issue.

Issue #377, opened 2023-09-06
mf_priority: handle job updates


grondo:
As noted in #376, the mf_priority plugin should probably add a `job.update` callback so that it can handle updates to jobspec for pending jobs.

The `job.update` callback is issued after any `jobspec-update` event is processed, so the jobspec passed in the plugin args will have the updated values. An `updates` key is also passed in to the plugin args so this object could be scanned to see if any keys of interest were updated before reprocessing the jobspec, but that is probably optional for now.

The `job.update` callback is probably going to duplicate much of the code currently in the `job.new` callback, so possible a good solution here would be to split off the common code into a function that can be called by both `job.new` and `job.update`.

grondo:
PR flux-framework/flux-core#5424 is ready to merge once the accounting priority plugin is capable of handling the queue update. I wonder if we'll want to prepare a WIP PR here, then merge the flux-core changes so that the implementation here can be tested...

cmoussa1:
Sure, I can try to work on this this week and see if I can get something posted soon! Is there a way to for me to work in an interactive docker container with that specific flux-core branch? Not sure if I've ever done that before...

grondo:
Oh, good question. What I would do is create the docker container then build and install that branch with something like:
```console
$ ./configure    --prefix=/usr  --sysconfdir=/etc  --with-systemdsystemunitdir=/etc/systemd/system --localstatedir=/var  --with-flux-security
$ sudo make -j 8 install
```

cmoussa1:
Sorry for leaving you hanging on this - I definitely forgot to acknowledge your suggestion yesterday and didn't try this until just now - think I was able to get it working though.

(notes for myself for the future): I cloned your branch, launched an interactive Docker container, built and installed with the above commands, cloned the appropriate **flux-accounting** branch, built that, and was able to use the custom flux-core to do testing. 👍 thanks!!

cmoussa1:
Just doing some super rudimentary testing this morning and wanted to report my findings so far.

This is specific to queue updates, but at least in my interactive Docker container, there was not a need to add a `job.update` callback to the priority plugin to be able to validate updated queues; I _think_ this is because as we mentioned in #378, after a job update, the job will re-enter `job.validate`, which _does_ check queue values passed in for a user/bank. The following set of tests I've been running attempt a workflow of submitting a pending job and updating its queue:

```shell
test_expect_success 'submit job for testing' '
[...]

PR #388, opened 2023-09-14
plugin: move queue priority assignment to `job.new` callback

cmoussa1:
#### Problem

The multi-factor priority plugin assigns the priority associated with a given queue to the `bank_info` struct associated with each job in `job.validate`. Now that Flux can accept job updates and jobs can re-enter `job.validate` multiple times, this assignment should be moved to `job.new`, and `job.validate` should only be used for validating a queue if one is passed in.

---

This PR removes the assignment of the priority associated with a passed in queue in `job.validate`.

In `job.new`, instead of only assigning a queue's priority in the case where the `bank_info` struct unpacked is `NULL`, it moves the assignment to later in the callback so that it is executed in all cases.

This will result in the `bank_info` struct associated with each job to have an associated queue priority by the end of `job.new` instead of by the end of `job.validate`.

This work is part of some of the changes to the plugin described in #377, but I'm trying to avoid posting a big PR with a lot of different changes, so I split this off and am posting this as kind of a "prerequisite" PR to that work in #377.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -698,7 +698,7 @@ static int validate_cb (flux_plugin_t *p,
     char *bank = NULL;
     char *queue = NULL;
     flux_job_state_t state;
-    int max_run_jobs, cur_active_jobs, max_active_jobs = 0;
+    int max_run_jobs, cur_active_jobs, max_active_jobs, queue_factor = 0;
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

cmoussa1:
thanks! setting MWP

PR #389, opened 2023-09-15
plugin: add queue update validation

cmoussa1:
#### Background

As noted in #377, Flux can now handle job updates to pending jobs, particularly with queues. Since flux-accounting's priority plugin is involved in queue validation (i.e making sure that the user/bank has access to the queue they want to submit their job to), it should be expanded to handle job updates.

---

This PR adds a new callback to the priority plugin for `job.update`, which will unpack user, bank, and queue information and assign the new priority associated with the updated queue to the `bank_info` struct of a job so that when the job re-enters `job.state.priority`, a new overall job priority can be calculated.

A new helper function is added to the plugin called `get_bank_info ()`, which accepts a userid and an optional bank value (if one is passed during job submission) that performs a lookup in the plugin's internal map to retrieve the necessary information to associate with the submitted job. It returns a pair: a return code as well as the iterator to the user/bank map that is associated with the job. The `job.update` callback makes use of this to retrieve the user/bank information, and then retrieves the new priority associated with the updated queue and assigns it to the `bank_info` struct associated with a job, which could affect its priority.

This new helper function can most likely be adapted in other places in the plugin code, but I think its best to probably leave it for a future PR to refactor some of the callbacks to make use of it.

Finally, a new set of tests are added to the testsuite which test updating the queue for a job in which the user/bank **does** have access to and updating the queue for a job in which the user/bank **does not** have access to. In the former, the job is updated and the priority is increased, and in the latter, an error is raised when the job re-enters `job.validate` and the priority remains the same as it was before the attempted update. Note that these tests are commented out until flux-framework/flux-core#5424 has landed; otherwise, I think these tests will fail in the meantime.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1074,6 +1074,70 @@ static int run_cb (flux_plugin_t *p,
 }
 
 
+static int job_updated (flux_plugin_t *p,
+                        const char *topic,
+                        flux_plugin_arg_t *args,
+                        void *data)
+{
+    std::map<std::string, struct bank_info>::iterator bank_it;
+    int userid;
+    char *bank = NULL;
+    char *queue = NULL;
+    struct bank_info *b;
+
+    if (flux_plugin_arg_unpack (args,
+                                FLUX_PLUGIN_ARG_IN,
+                                "{s:i, s{s{s{s?s, s?s}}}}",
+                                "userid", &userid,
+                                "jobspec", "attributes", "system",
+                                "bank", &bank, "queue", &queue) < 0) {
+        return flux_jobtap_reject_job (p, args, "unable to unpack bank arg");
+    }
+
+    // grab bank_info struct for user/bank (if any)
+    b = static_cast<bank_info *> (flux_jobtap_job_aux_get (
+                                                    p,
+                                                    FLUX_JOBTAP_CURRENT_JOB,
+                                                    "mf_priority:bank_info"));
+
+    if (b == NULL) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.update: bank info is missing");
+
+        return -1;
+    }
+
+    // look up user/bank info based on unpacked information
+    bank_info_result lookup_result = get_bank_info (userid, bank);
+
+    if (lookup_result.first == BANK_USER_NOT_FOUND) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                     "mf_priority", 0,
+                                     "job.update: cannot find info for user: ",
+                                     userid);
+    } else if (lookup_result.first == BANK_INVALID) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                     "mf_priority", 0,
+                                     "job.update: not a member of %s",
+                                     bank);
+    } else if (lookup_result.first == BANK_NO_DEFAULT) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                     "mf_priority", 0,
+                                     "job.update: user/default bank "
+                                     "entry does not exist");
+    } else if (lookup_result.first == BANK_SUCCESS) {
+        bank_it = lookup_result.second;
+    }
+
+    // fetch the priority of the validated queue; assign it to the bank_info
+    // struct associated with the job
+    int queue_factor = get_queue_info (queue, bank_it);
+    b->queue_factor = queue_factor;
+
+    return 0;
+}
+
+
 static int inactive_cb (flux_plugin_t *p,
                         const char *topic,
                         flux_plugin_arg_t *args,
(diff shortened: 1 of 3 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -45,6 +45,21 @@ extern "C" {
...
+// USER_NOT_FOUND: the user could not be found in the plugin map
+// BANK_SUCCESS: we found an entry for the passed-in user/bank
+// INVALID_BANK: the user specified a bank they don't belong to
+// NO_DEFAULT_BANK: the user does not have a default bank in the plugin map
+enum bank_info_codes {
+    USER_NOT_FOUND,
+    BANK_SUCCESS,
+    INVALID_BANK,
+    NO_DEFAULT_BANK
+};
```
grondo, line 59: Suggestion: it is sometimes good practice to name enumerations with a common prefix, e.g. if it works here prefix all these values with `BANK_`: `BANK_USER_NOT_FOUND`, `BANK_NO_DEFAULT`, `BANK_INVALID` etc. This just helps when reading and writing code.

Also, just another suggestion, but is more typical to see the first enumeration value (i.e. `0`) be SUCCESS. Since that is more common practice, I'd suggest moving `BANK_SUCCES` to the first position of the enumeration.

t/t1029-mf-priority-update-queue.t
```diff
@@ -0,0 +1,87 @@
...
+test_expect_success 'submit job for testing' '
+	jobid1=$(flux python ${SUBMIT_AS} 5001 --queue=bronze sleep 30)
+'
+
+# test_expect_success 'update of queue of pending job works' '
+# 	flux update $jobid1 queue=silver &&
+# 	flux job eventlog $jobid1 > eventlog.out &&
+# 	grep "attributes.system.queue=\"silver\"" eventlog.out
+# '
+
```
grondo, line 72: Is there a way to check that the update had the intended effect, i.e. check for a new priority (or maybe query the mf_priority plugin and ensure the queue factor has been updated?

grondo (commented):
Looks good! Just a quick first pass with a couple suggestions, but otherwise I think this looks good!

src/plugins/mf_priority.cpp
cmoussa1: Thanks for these suggestions! I was unaware of the practices here (I think it's my first time using enumeration values? can't remember if i ever did this), so I'll make these changes to use a common `BANK` prefix as well as using the SUCCESS value as the first position. 👍

t/t1029-mf-priority-update-queue.t
cmoussa1: Yes there is. I forgot to add those to the official test; was just printing out the priority values before, but I think I can check priority values with `flux job wait-event -f json $jobid priority | jq '.context.priority'`. I'll add these to the testsuite. Thanks!

grondo:
flux-framework/flux-core#5424 has been merged, so tests could be enabled here now.

cmoussa1:
OK, I've uncommented out the tests that were dependent on flux-framework/flux-core#5424 and rebased; I'll mark it as ready for some initial review.

src/plugins/mf_priority.cpp
```diff
@@ -1074,6 +1074,71 @@ static int run_cb (flux_plugin_t *p,
...
+                        flux_plugin_arg_t *args,
+                        void *data)
+{
+    std::map<std::string, struct bank_info>::iterator bank_it;
+    int userid;
+    char *bank = NULL;
+    char *queue = NULL;
+    struct bank_info *b;
+
+    flux_t *h = flux_jobtap_get_flux (p);
```
grondo, line 1088: I could have missed it, but it appears `h` is assigned here and never used?

t/Makefile.am
```diff
@@ -31,6 +31,7 @@ TESTSCRIPTS = \
 	t1027-mf-priority-issue376.t \
 	t1028-mf-priority-issue385.t \
 	t1029-mf-priority-default-bank.t \
+	t1030-mf-priority-update-queue.t \
```
grondo, line 34: be sure to remove the `note:` from the commit message here.

t/t1030-mf-priority-update-queue.t
```diff
@@ -0,0 +1,92 @@
...
+	flux update $jobid1 queue=silver &&
+	flux job wait-event -f json $jobid1 priority &&
+	flux job eventlog $jobid1 > eventlog.out &&
+	grep "attributes.system.queue=\"silver\"" eventlog.out &&
+	grep 2050000 eventlog.out
+'
+
+test_expect_success 'updating a job using a queue the user does not belong to fails' '
+	test_must_fail flux update $jobid1 queue=gold > unavail_queue.out 2>&1 &&
+	test_debug "unavail_queue.out" &&
```
grondo, line 80: This should be `test_debug "cat unavail_queue.out"`

grondo (approved):
Looks good! Just a couple minor comments.

cmoussa1:
Thanks for the review @grondo! I've pushed up the fixes you mentioned in your comments. Since you've approved, I'll set MWP on this. Thanks again!!

PR #392, opened 2023-09-18
testsuite: allow sharness tests to be run by hand

grondo:
Problem: sharness tests in the testsuite can't be run by hand because `PYTHONPATH` is not set to the in-tree accounting python path.

In `t/sharness.d/flux-accounting.sh` add a function to prepend to a colon-separated path variable if the argument is not already the first entry. Use this to prepend the in-tree python path to `PYTHONPATH`h if necessary. Also use it for `FLUX_EXEC_PATH_PREPEND` so we don't prepend the same path multiple times.
 
```
$ ./t1000-print-hierarchy.t  -d -v
```

or 
```
$ FLUX_TEST_VALGRIND=t ./t1000-print-hierarchy.t  -d -v
```


cmoussa1 (approved):
LGTM! thanks @grondo

Issue #390, opened 2023-09-18
testsuite: fix tests that look at job state


cmoussa1:
Mentioned in #294 and #301, there might be some tests in the flux-accounting test suite that might produce a race condition, especially when run under valgrind. Some of the tests look at specific job states, but there is a chance the job may not have reached a certain state, like DEPEND, by the time `flux job info` is run. It might be better to use `flux job wait-event` to look for `depend` or `alloc` events instead of looking at DEPEND or RUN states of jobs.

I should step through the test suite in flux-accounting and make any necessary changes.

cmoussa1:
This should be fixed by #393, so I can close this now.

PR #393, opened 2023-09-20
testsuite: change check for specific job states

cmoussa1:
#### Problem

As mentioned in #390, there are some tests in the flux-accounting testsuite that might produce a race condition when run under `valgrind`. Some of the tests look at specific job states, but there is a chance the job may not have reached a certain state, like DEPEND, by the time `flux job info` is run. It might be better to use `flux job wait-event` to look for `depend` or `alloc` events instead of looking at DEPEND or RUN states of jobs.

---

This PR makes a number of changes to the testsuite so that flux-accounting tests can be run under `valgrind`. The first commit changes the checks for certain job states to use `flux job wait-event` and `flux job info $jobid eventlog` to check for certain events and states in the tests.

The next commit removes a couple of checks for specific job usage values because running this set of tests in `t1011-job-archive-interface.t` under `valgrind` might cause the job usage values as a result of the submitted job to be different than what is checked.

The third commit changes the order of job cancellations of the submitted jobs in `t1019-mf-priority-info-fetch.t` because we don't want the last submitted job to enter the `alloc` event while the first two jobs are being canceled, since in this test we're checking for certain `running_jobs` and `active_jobs` values in the tests.

The last commit just converts `t1020-mf-priority-issue262.t` to use tabs instead of spaces, which would make it consistent with the rest of the tests in the flux-accounting testsuite.

If desired, perhaps I could add a `valgrind` builder to the GitHub actions for flux-accounting in this PR? Is that possible? I can't remember if flux-core has one; I tried doing a quick GitHub search but couldn't find anything.

### DIFF
--- a/t/t1005-max-jobs-limits.t
+++ b/t/t1005-max-jobs-limits.t
@@ -90,14 +90,16 @@ test_expect_success 'submit max number of jobs' '
 	jobid2=$(flux python ${SUBMIT_AS} 5011 sleep 60)
 '
 
-test_expect_success 'submit job while already having max number of running jobs' '
+test_expect_success 'a submitted job while at max-running-jobs limit will have a dependency added to it' '
 	jobid3=$(flux python ${SUBMIT_AS} 5011 sleep 60) &&
-	test $(flux jobs -no {state} ${jobid3}) = DEPEND
+	flux job wait-event -vt 60 \
+		--match-context=description="max-running-jobs-user-limit" \
+		$jobid3 dependency-add
 '
 
 test_expect_success 'a job transitioning to job.state.inactive should release a held job (if any)' '
 	flux job cancel $jobid1 &&
-	test $(flux jobs -no {state} ${jobid3}) = RUN &&
+	flux job wait-event -vt 60 $jobid3 alloc &&
 	flux job cancel $jobid2 &&
 	flux job cancel $jobid3
 '
@@ -106,9 +108,11 @@ test_expect_success 'submit max number of jobs with other bank' '
 	jobid1=$(flux python ${SUBMIT_AS} 5011 --setattr=system.bank=account2 sleep 60)
 '
 
-test_expect_success 'submit a job while already having max number of running jobs' '
-	jobid2=$(flux python ${SUBMIT_AS} 5011 --setattr=system.bank=account2 sleep 60)
-	test $(flux jobs -no {state} ${jobid2}) = DEPEND &&
+test_expect_success 'a submitted job while at max-running-jobs limit will have a dependency added to it' '
+	jobid2=$(flux python ${SUBMIT_AS} 5011 --setattr=system.bank=account2 sleep 60) &&
+	flux job wait-event -vt 60 \
+		--match-context=description="max-running-jobs-user-limit" \
+		$jobid2 dependency-add &&
 	flux job cancel $jobid1 &&
 	flux job cancel $jobid2
 '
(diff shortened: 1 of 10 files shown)


cmoussa1:
I've just rebased after #399, so I think I'll mark this as ready for review.

t/t1005-max-jobs-limits.t
```diff
@@ -92,12 +92,16 @@ test_expect_success 'submit max number of jobs' '
 
 test_expect_success 'submit job while already having max number of running jobs' '
 	jobid3=$(flux python ${SUBMIT_AS} 5011 sleep 60) &&
-	test $(flux jobs -no {state} ${jobid3}) = DEPEND
+	flux job wait-event -vt 60 $jobid3 dependency-add &&
+	flux job info $jobid3 eventlog > eventlog.out &&
+	grep "dependency-add" eventlog.out
```
grondo, line 97: The `validate` event should transition a job from NEW to DEPEND, so it may be possible to replace the test for `state == DEPEND` with `flux job wait-event JOBID validate`. However, I may be misunderstanding the intent here.

Also, it probably isn't necessary to use `flux job wait-event` and then grep for the event in the eventlog fetched from `flux job info`. In this case, if `flux job wait-event` has reported that the event has appeared, it is guaranteed that the event is posted to the eventlog. Therefore I _think_ you could skip the two extra steps here as redundant.

cmoussa1: Thanks for this pointer! The intent with this test is to ensure that the job has a dependency added to it by the priority plugin. In this case, should I still change the check to instead use your suggestion `flux job wait-event JOBID validate`? I should update the description of the test to make it clearer what it is trying to do, sorry about that.

Thanks also for your suggestion about adding `flux job info` after a `flux job wait-event` - I can just remove those from this test and the others that have it.

grondo: Sorry, I misunderstood the intent. If the intent is to ensure a dependency is added, the `flux job wait-event dependency-add` is the correct approach.

You could also consider using the `flux job wait-event` `-m, --match-context` option to ensure that the context matches the dependency you expect to be adding too. 

Sorry for my confusion!

cmoussa1: Not a problem! I just force-pushed up a change to use `--match-context` with `flux job wait-event` per your suggestion above (I also updated the test description to hopefully be a little bit clearer), as well as removed the redundant `flux job info` calls throughout the changed tests that look at job state.

t/t1005-max-jobs-limits.t
```diff
@@ -90,14 +90,16 @@ test_expect_success 'submit max number of jobs' '
 	jobid2=$(flux python ${SUBMIT_AS} 5011 sleep 60)
 '
 
-test_expect_success 'submit job while already having max number of running jobs' '
+test_expect_success 'a submitted job while at max-running-jobs limit will have a dependency added to it' '
 	jobid3=$(flux python ${SUBMIT_AS} 5011 sleep 60) &&
-	test $(flux jobs -no {state} ${jobid3}) = DEPEND
```
grondo, line 95: Update commit message to remove "and grep-ing for certain events..."

t/t1011-job-archive-interface.t
```diff
@@ -175,8 +175,9 @@ test_expect_success 'run update-usage and update-fshare commands' '
 '
 
 test_expect_success 'check that job usage and fairshare values get updated' '
-	flux account-shares -p $(pwd)/FluxAccountingTest.db > post_update2.test &&
```
grondo, line 178: Looks like perhaps two commits were accidentally squashed here? (Commit message has two Problem statements)

grondo (approved):
LGTM! Noted a couple commit message issues inline.

cmoussa1:
Thanks @grondo! Will set MWP here

PR #395, opened 2023-09-21
`view-bank`: fix `-t` option for a sub bank with users in it

### LINKED ISSUE #394: `view-bank`: `-t` option does show hierarchy for a sub bank with users in it
While working on some other stuff yesterday I discovered a bug in the `-t` option of the `view-bank` where the hierarchy is not printed for a sub bank with users in it. In the reproducer, I had the following hierarchy set up:

```
root
  A
    user1
    user2
  B
    user3
    user4
  C
  D
    E
      user5
      user6
    F
```

And calling `view-bank -t E` just returns the information for bank `E` and does not call the recursive function for printing any hierarchical information. This is because of this check:

```python
            if result:
                # traverse the user/bank hierarchy and add the information from all
                # banks and users to the hiearchy string
                hierarchy_str = print_hierarchy(cur, name, hierarchy_str, "")
                bank_str += "\n" + hierarchy_str
            else:
                bank_str += "\n" + hierarchy_str
```

The `result` variable stores any potential sub banks if the bank being passed has any, but the function should just call `print_hierarchy ()` regardless if `result` has sub banks in it or not, so the fix should be relatively simple: just remove the `if` check and call `print_hierarchy ()` always. I'll open a PR to fix this.

cmoussa1:
#### Problem

As mentioned in #394, the `-t` option of the `view-bank` command does not print out a full hierarchy for a sub bank with users in it because there is an `if` check that looks to see if the bank has any sub banks in it before calling the recursive function that prints out any potential information under it `print_hierarchy()`, but it won't call this recursive function if it doesn't have any sub banks, when it really should be called all the time, whether it has additional sub banks, users, or nothing at all.

---

This PR just removes the `if` check before calling the recursive function `print_hierarchy()` so that it is always called whether a sub bank has any additional sub banks, users, or neither.

I've also added a test that uses the `-t` option for a sub bank with users under it to make sure that it works as intended.

Fixes #394

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -268,17 +268,8 @@ def view_bank(conn, bank, tree=False, users=False):
                 + "\n"
             )
 
-            # get all potential sub banks
-            cur.execute("SELECT * FROM bank_table WHERE parent_bank=?", (name,))
-            result = cur.fetchall()
-
-            if result:
-                # traverse the user/bank hierarchy and add the information from all
-                # banks and users to the hiearchy string
-                hierarchy_str = print_hierarchy(cur, name, hierarchy_str, "")
-                bank_str += "\n" + hierarchy_str
-            else:
-                bank_str += "\n" + hierarchy_str
+            hierarchy_str = print_hierarchy(cur, name, hierarchy_str, "")
+            bank_str += "\n" + hierarchy_str
         # if users is passed in, print out all potential users under
         # the passed in bank
         if users is True:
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
Looks reasonable to me. So with this change, if I try to print information about a sub-bank, it will now print out information about parent banks as well?

cmoussa1:
Actually, with this change, when you print out information from a sub-bank, you'll be able to customize your hierarchy starting from a sub-bank that isn't the root bank, which could be very large. For example, you're in a sub-bank that is, like, 4-or-5 layers deep and you just want to see your information and everybody else who belongs to that sub-bank. Previously, if you used the hierarchy option `-t` and passed the sub-bank, it wouldn't print out your (or any other users in that sub-bank's) information; this was unintended, so now you'll be able to see this information.

cmoussa1:
Thanks for reviewing and approving this @jameshcorbett; I'll go ahead and set MWP here

jameshcorbett:
Oh ok makes sense and sounds good!

PR #397, opened 2023-10-02
plugin: improve `job.validate` callback, clean up unused variables

cmoussa1:
#### Background

While working on some other stuff in the plugin, I noticed there were either unused or unneeded variables in `validate_cb ()` because the things they were originally used for were moved to other callbacks in the priority plugin. For example, the `queue_factor` variable that used to store a result of calling `get_queue_info ()` could be removed in favor of just using the function call directly in its place.

Furthermore, the lookup of the user/bank information can be cleaned up to use the new helper function in the plugin that was added in #389.

---

This is a relatively low-priority PR that just cleans up the `job.validate` callback by removing some unused/unneeded variables as well as makes use of the new user/bank lookup helper function. IMHO, it makes reading the `validate_cb ()` function easier and clearer about what it is checking.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -809,8 +809,8 @@ static int validate_cb (flux_plugin_t *p,
     // if the plugin has NO data about users/banks and the user does not have
     // an entry in the plugin, the job will be held until data is received by
     // the plugin.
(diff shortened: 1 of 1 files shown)


cmoussa1:
Most of the changes here landed in #413, so I'll close this. :-)

PR #398, opened 2023-10-12
feat: developer container environment

vsoch:
Problem: there is no easy way to develop (in a container) and this would be nice if we consider other kinds of bindings. Solution: add a .devcontainer setup.

This is the first step (I think at least) to thinking about Go bindings. I see this is a project in C++ (not C) so I'll look into good ways to do that.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,17 @@
+{
+  "name": "Flux Accounting Developer Environment",
+  "dockerFile": "Dockerfile",
+  "context": "../",
+
+  "customizations": {
+    "vscode": {
+      "settings": {
+        "terminal.integrated.defaultProfile.linux": "bash"
+  },
+      "extensions": [
+         "ms-vscode.cmake-tools"
+      ]
+    }
+  },
+  "postStartCommand": "git config --global --add safe.directory /workspaces/flux-accounting"
+}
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,17 @@
...
+      "settings": {
+        "terminal.integrated.defaultProfile.linux": "bash"
+  },
+      "extensions": [
+         "ms-vscode.cmake-tools"
+      ]
+    }
+  },
+  "postStartCommand": "git config --global --add safe.directory /workspaces/flux-accounting"
+}
```
cmoussa1, line 17: does it matter if this line does or doesn't have a newline at the end of it?

cmoussa1 (approved):
this LGTM! thanks @vsoch! just had one minor question about `devcontainer.json`

.devcontainer/devcontainer.json
vsoch: It does not - it looks like json but actually it breaks many rules of it. I tested this fully and the newline does not seem to make a difference.

vsoch: hey you have a new picture! I like it!

cmoussa1: thanks! my hair is a lot longer now, i don't look much like that graduation photo from 2019 anymore 😆 feel free to set MWP on this whenever you are ready

vsoch: Done! Thanks @cmoussa1

PR #400, opened 2023-10-18
[WIP] plugin: add support for updating bank of a pending job

cmoussa1:
~~note: this is built on top of #399, so I think it might be best to hold off on this PR until that one lands.~~ (#399 has now landed)

#### Problem

flux-accounting currently has no support for a user/bank to update the bank they want their pending job to run under.

---

This is a PR that adds support for updating the bank of a pending job. It adds a new callback to the plugin which looks for the `"job.update.attributes.system.bank"` topic string and validates the new bank that the user is trying to update their job to. If they have access to the bank they want to update their job to _but_ this new bank is currently at its max running jobs limit, the plugin will reject the update. Since `job.state.depend` is the final place where a job can have dependencies added to it (and adding/removing dependencies is how the plugin handles running job limits), I've chosen to just reject the update and keep the job under its current bank.

Once a new bank has been validated, its attributes are then applied to the job's `bank_info` struct in the `job.update` callback. You'll notice that before the new attributes/limits are assigned to the job, the `active_jobs` count of the old bank is decremented. Then, just before the new attributes/limits are assigned to the job, the `active_jobs` count of the new bank is incremented (since this count usually gets incremented when the job is in `job.new`).

A new test file, `t1031-mf-priority-update-bank.t` is also added to the test suite which simulates submitting some jobs under default and non-default banks and updating the bank between each, while also making sure that a job being updated to a bank that is already at their current max running jobs limit gets rejected.

---

#### TODO

- [x] add a check in the test suite to ensure that active/running jobs limits are as expected after an update

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1078,8 +1078,8 @@ static int run_cb (flux_plugin_t *p,
 
 
 /*
- *  apply an update on a job with regard to its queue once it has been
- *  validated.
+ *  apply an update on a job with regard to its queue or associated bank once
+ *  the requested attribute has been validated.
  */
 static int job_updated (flux_plugin_t *p,
                         const char *topic,
@@ -1089,16 +1089,20 @@ static int job_updated (flux_plugin_t *p,
     std::map<std::string, struct bank_info>::iterator bank_it;
     int userid;
     char *bank = NULL;
+    char *new_bank = NULL;
     char *queue = NULL;
     struct bank_info *b;
 
+    flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:i, s{s{s{s?s}}}, s:{s?s}}",
+                                "{s:i, s{s{s{s?s}}}, s:{s?s, s?s}}",
                                 "userid", &userid,
-                                "jobspec", "attributes", "system", "bank", &bank,
+                                "jobspec", "attributes", "system", "bank",
+                                &bank,
                                 "updates",
-                                    "attributes.system.queue", &queue) < 0)
+                                    "attributes.system.queue", &queue,
+                                    "attributes.system.bank", &new_bank) < 0)
         return flux_jobtap_error (p, args, "unable to unpack plugin args");
 
     // grab bank_info struct for user/bank (if any)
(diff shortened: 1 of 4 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -1078,8 +1078,8 @@ static int run_cb (flux_plugin_t *p,
 
 
 /*
- *  apply an update on a job with regard to its queue once it has been
- *  validated.
+ *  apply an update on a job with regard to its queue or associated bank once
```
grondo, line 1081: Commit message: " flux-accounting currently has no support for a user/bank to update the bank they want their pending job to run under." Should this just be "flux-accounting currently has no support for a user to update the bank of their job"?

grondo (commented):
Just did a quick first pass here. I wonder if some tests that update both bank and queue at the same time should be added, both for valid and invalid combinations of updates?

Also, as an aside, while reviewing this I was thinking that it might be good to start thinking about refactoring the bank_info structure in his plugin into a class. The class interface could be pulled out o the plugin itself, and tested with unit tests. I think this would really clean up some of the code for looking up and inserting new bank information for users, and make the code easier to read (I can open an issue on this if you'd like)

src/plugins/mf_priority.cpp
cmoussa1: That's a better message than what I had suggested 😅 thanks! I've just updated this commit message.

cmoussa1:
> Just did a quick first pass here. I wonder if some tests that update both bank and queue at the same time should be added, both for valid and invalid combinations of updates?

That's a good suggestion, thanks. I've added a commit in this PR that adds another test file that tests updating both queue and bank at the same time as well as a couple of invalid combinations (i.e the queue specified or the bank specified was invalid). Let me know if this is what you had in mind!

> Also, as an aside, while reviewing this I was thinking that it might be good to start thinking about refactoring the bank_info structure in his plugin into a class. The class interface could be pulled out o the plugin itself, and tested with unit tests. I think this would really clean up some of the code for looking up and inserting new bank information for users, and make the code easier to read (I can open an issue on this if you'd like)

Oh sure! That could be really helpful. Yes, if you'd like to open an issue I think that would be useful for me. :-)

cmoussa1:
Now that the priority plugin is cleaned up and there are new external functions, I plan to revisit this and update it to reflect the new changes. Will convert this back to [WIP] in the meantime!

cmoussa1:
Going to close this as I now have a (hopefully much cleaner) re-implementation of the changes proposed in this PR.

PR #401, opened 2023-11-17
[experimental] plugn: add new `bank_info` class

cmoussa1:
### Background

The plugin uses its own `bank_info` struct to hold user/bank information and has a number of methods for access and modification of these objects. As the plugin's feature set has grown, so have the requirements for the `bank_info` struct, resulting in a very large and hard-to-parse piece of code.

---

This is more of an experimental PR based on some of the feedback in #400 that suggested considering creating a class for the `bank_info` object. For now, I've just built this on top of #400 with one additional commit that does the following:

A new `.hpp` and `.cpp` file are added to the `plugins/` directory called **bank_info.cpp** and **bank_info.hpp**. This set of files stores a new class called `user_bank_info`, which is made up of the same attributes as the original `bank_info` struct. They are `#include`d to the plugin file and the definition of the struct is thus removed from the plugin and instances of that struct are replaced with the use of the new class.

For now, this class file just has the definition of the `user_bank_info` object, but, if this is the right idea, then I can begin to slowly move some of the related functionality that is currently defined in the plugin over to this new class file.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -17,7 +17,8 @@ noinst_HEADERS = \
 	fairness/reader/data_reader_db.hpp \
 	fairness/writer/data_writer_base.hpp \
 	fairness/writer/data_writer_db.hpp \
-	fairness/writer/data_writer_stdout.hpp
+	fairness/writer/data_writer_stdout.hpp \
+	plugins/bank_info.hpp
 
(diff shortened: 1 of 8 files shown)


cmoussa1:
Think I'm gonna close this PR, as I've just opened #403 that I think does a better job of what I tried to do here. 😆

PR #403, opened 2023-12-04
plugin: add new external class for user/bank information

cmoussa1:
#### Background

The plugin uses its own `bank_info` struct to hold user/bank information and has a number of methods for access and modification of these structs. As the plugin's feature set has grown, so have the requirements for the `bank_info` struct, resulting in a very large and hard-to-parse piece of code. It would benefit the readability of this plugin if some basic features were abstracted out to an external class that the plugin could just `#include`.

---

This PR begins this process of moving some of the definitions and features related to accessing user/bank information from the plugin's internal map to an external class. `bank_info.hpp` and `bank_info.cpp`, in this PR, is proposed to now hold the following:

- the definition of the user/bank class
- a helper function, `get_user_info ()`, which will perform a lookup of a user/bank in the internal map and return its location in the map or `NULL` on failure

References to this new user/bank class (called `user_bank_info`) are now used in place of the old `struct bank_info`.

I was also able to add some basic unit tests for this `get_user_info ()` helper function that get executed with `make check`. This test suite can grow as more and more features are abstracted out of the plugin. :-)

As a result of this new class and helper function, I also adjusted one of the other helper functions that validates a queue for a user/bank by changing one of its function arguments to just the user/bank's list of permissible queues (instead of the map iterator to the old `bank_info` object, which was ugly to look at).

Finally, in the last commit, I give an example of what using these new external `.hpp` and `.cpp` files does to a function in the plugin in terms of cleanup and improved readability. `validate_cb ()` is adjusted to use this new helper function for the user/bank lookup and no longer needs to define and use a map iterator anywhere in its function. I've also added/adjusted some of the outdated comments in this function to help clarify some of the checks in this callback with the intention to help explain what it is doing more clearly.

For the purposes of (hopefully) keeping this PR concise, I've just adjusted `validate_cb ()` to showcase this new use of the external `.cpp` files; in the future, I can open similar, subsequent ones to improve each callback.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -17,7 +17,8 @@ noinst_HEADERS = \
 	fairness/reader/data_reader_db.hpp \
 	fairness/writer/data_writer_base.hpp \
 	fairness/writer/data_writer_db.hpp \
-	fairness/writer/data_writer_stdout.hpp
+	fairness/writer/data_writer_stdout.hpp \
+	plugins/bank_info.hpp
 
 fairness_libweighted_tree_la_SOURCES = \
 	fairness/account/account.cpp \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -47,7 +47,8 @@ fairness_libweighted_tree_la_CXXFLAGS = \
 TESTS = \
 	weighted_tree_test01.t \
 	data_reader_db_test01.t \
-	data_writer_db_test01.t
+	data_writer_db_test01.t \
+	user_bank_info_test01.t
 check_PROGRAMS = $(TESTS)
 
 TEST_EXTENSIONS = .t
@@ -92,6 +93,14 @@ data_writer_db_test01_t_CXXFLAGS = $(AM_CXXFLAGS) -I$(top_srcdir) $(SQLITE_CFLAG
 data_writer_db_test01_t_LDFLAGS = $(SQLITE_LIBS)
 data_writer_db_test01_t_LDADD = \
 	fairness/libweighted_tree.la \
+	common/libtap/libtap.la
+
+user_bank_info_test01_t_SOURCES = \
+	plugins/test/user_bank_info_test01.cpp \
+	plugins/bank_info.cpp \
+	plugins/bank_info.hpp
+user_bank_info_test01_t_CXXFLAGS = $(AM_CXXFLAGS) -I$(top_srcdir)
+user_bank_info_test01_t_LDADD = \
 	common/libtap/libtap.la
 
 noinst_PROGRAMS = \
--- a/src/plugins/Makefile.am
+++ b/src/plugins/Makefile.am
@@ -8,5 +8,6 @@ jobtapdir = \
   $(fluxlibdir)/job-manager/plugins/
 
 jobtap_LTLIBRARIES = mf_priority.la
-mf_priority_la_SOURCES = mf_priority.cpp
+mf_priority_la_SOURCES = mf_priority.cpp bank_info.cpp
+mf_priority_la_CPPFLAGS = -I$(top_srcdir)/src/plugins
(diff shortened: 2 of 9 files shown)


grondo:
@cmoussa1, sorry I keep trying to go through this PR and getting interrupted. I think this is a good start, and definitely seems to clean up some calls in the plugin. However, what I was picturing would go a bit further and make a real `bank_info` class with methods that abstract most the bank information related calls in the plugin.

However, I'm afraid I'd have to study the code a bit more and get familiar with the interfaces such a class would need in order to make any useful suggestions. Also, I'm a novice at C++ at best, and perhaps wouldn't be the best person to make suggestions or review C++ code. But, I'll try to go through the plugin again and maybe make some suggestions along the lines of what I was thinking.

grondo:
Here's a couple examples after a quick look:

 - There's still a few global variables and maps in the plugin. Maybe these could be moved into the class definition, and all functions accessing those maps/variables would become methods?
 - There's a function `banks_to_json()` in `mf_priority.cpp`. IMO, this should be a method on a bank_info or similar object. e,g so you'd call `bank_info.to_json()`. (also, a `flux_plugin_t` appears to be passed to that function, but isn't used)
 - `check_map_for_dne_only()` could also be a method?

Maybe in a comment you could summarize the data structures that are currently being used by the plugin, and the functions they need, and we can then work on the design of an abstraction that could encapsulate all this within one or more classes?

O/w, perhaps we could walk through the code at some point (maybe on a coffee call) and kind of wrap our heads around it that way?

cmoussa1:
@grondo no problem at all, thanks for beginning to take a look at this last week! Just got back from vacation so am just getting back to this now.

I'm sorry I probably didn't make it very clear in the PR description, but the intent with this PR at least was to just begin moving some of the definitions/methods of the `bank_info` class out of the plugin and into a separate class, and to showcase what the new class looks like in just one of the callbacks (`validate_cb ()`) in order to avoid making this PR a complete overhaul of the plugin, which would probably be tedious to review all at once. Then, if this looked okay, I would continue with subsequent PRs that continue to move/abstract more of the methods that directly deal with the `bank_info` object to the new class. Does that make sense?

In any case, I completely agree with you that perhaps a separate comment/issue outlining what methods should be moved out of the plugin and into a separate class should be opened. I can work on it this week and open it up for further discussion where folks who are more familiar with C++ than I am can also chime in with their expertise!

grondo:
Sounds great!

cmoussa1:
note: I've added a couple commits to this PR that move the `check_map_for_dne_only ()` helper function to `bank_info.cpp`, as well as add a couple unit tests for that function.

I've continued working on walking through the plugin and moving things out as I notice room for improvement, but for the most part I've kept them in separate branches (to be proposed as separate, future PRs) to prevent overloading this one into a huge PR that's a pain to manage & review. Progress is being tracked in #403 on what work is located where at the moment 😅

grondo:
Also, it might be useful to pair up on the design phase here. Maybe you could walk through the current design with someone and make a first pass on the design. I don't think I'm the best person for this since I'm not very adept at C++, but perhaps someone else on the team might have the expertise to help (e.g. @jameshcorbett?)

cmoussa1:
Oh, that's a great suggestion @grondo! I think having another set of eyes during design would be super helpful. Yes, I should take a pass on this approach and see where we get. We can propose stuff in an issue/discussion thread or a coffee call as you've mentioned. :-)

jameshcorbett:
> Also, it might be useful to pair up on the design phase here. Maybe you could walk through the current design with someone and make a first pass on the design. I don't think I'm the best person for this since I'm not very adept at C++, but perhaps someone else on the team might have the expertise to help (e.g. @jameshcorbett?)

I'm down!

cmoussa1:
This work is going to be re-proposed in a more concise manner (such as in #411), so I'll close this PR.

PR #405, opened 2024-01-03
[WIP] plugin: move helper functions for `plugin.query` callback to external functions

cmoussa1:
#### Background

Pretty much the same background as in #403, the plugin defines a number of helper functions inside the plugin itself, which over time has resulted in a very large and hard-to-read piece of code. Specifically, there are a number of helper functions for the `plugin.query` callback which converts all of the user/bank data in the plugin's internal map to JSON format. It would significantly clean up the code if this was moved out of the plugin into an external location.

---

This [WIP] PR looks to do exactly that. This PR has the same "base" commits as #403 (creating an external `user_bank_info` class and external files that are `#include`d in the plugin file, but I've added a couple (everything from `"bank_info: add to_json method, function"` onwards) that move the helper functions responsible for creating JSON objects of all of the user/bank data in the plugin to these external files.

In the `user_bank_info` class, I've added a `.to_json ()` method which converts the object to JSON format. I've also added a new function, `map_to_json ()`, which iterates through all users in the plugin's internal map and creates a JSON object containing all of its information.

As a result, the callback for `plugin.query` is also changed to use this new external function, and the previously defined helper functions inside `mf_priority.cpp` are removed since they are no longer needed. You'll notice that the expected output for the tests for this callback have changed slightly, as I've chosen to include pretty much every attribute of each `user_bank_info` object.

##### TODO/might need further feedback or suggestions

- [ ] I noticed that I had to add an `-ljansson` to `src/Makefile.am` in order for the unit test file in `src/plugins/test` to run and find all of the jansson methods. Is this the right approach? Perhaps there is something else I am missing or doing something completely wrong.
- [ ] I did not add unit tests for the `map_to_json ()` function since it is tested in `t1019-mf-priority-info-fetch.t`, but if it's desired, maybe I can look to add a couple
- [ ] I chose not to pass the user/bank map as an argument to the `map_to_json ()` function. Is that okay? Should I change this? Unsure if this is an acceptable practice or not. 🤷


cmoussa1:
Going to close this and re-open a similar PR now that we have a more structured design as a result of #410.

Issue #406, opened 2024-01-08
all pending jobs killed after Flux update


grondo:
After a flux-core update to v0.58.0, all the pending jobs on fluke were killed with an exception from the `mf_priority` plugin:
e.g.
```
[root@fluke108:~]# flux job eventlog -H f2m1t75LN7vo
[Jan05 18:00] submit userid=61494 urgency=16 flags=0 version=1
[  +0.018063] jobspec-update attributes.system.bank="guests"
[  +0.018097] jobspec-update attributes.system.bank="guests"
[  +0.018144] validate
[  +0.034171] depend
[  +0.034206] priority priority=625
[Jan08 08:20] flux-restart
[  +0.000034] exception type="mf_priority" severity=0 note="not a member of guests"
[  +0.000062] priority priority=16
[  +0.000083] clean
```

The user, `testqe`, is in the `guests` bank though:
```console
$ flux account view-bank --users guests | grep testqe
testqe            61494             guests            1                 175899119.293700070.00625           100                                 
```

cmoussa1:
Without trying to construct a reproducer (yet), I believe this type of exception message is raised when a job's user/bank information is being updated in `job.state.priority` and the plugin cannot find a valid user/bank entry for the user that this job is submitted under:

https://github.com/flux-framework/flux-accounting/blob/1a84215ef29b9a5d6acfcef78cf1706cf0affa3d/src/plugins/mf_priority.cpp#L676-L679

Looking at the timestamps of the eventlog, it looks like the job was:

1) submitted successfully (and received a priority)
2) Flux was restarted
3) the priority plugin was loaded
4) jobs were reprioritized _before_ the plugin received any flux-accounting data, so it rejected this job and presumably all pending jobs.

I'll see if I can reproduce this behavior.

cmoussa1:
OK, I actually think I was able to reproduce this without having to restart Flux and instead just unloading/reloading the plugin. If I have a number of jobs in SCHED state (i.e they've received a priority and are waiting to run) and I reload the plugin _without updating it with flux-accounting information_ and call `reprioritize` on all jobs:

 (`flux.Flux().rpc("job-manager.mf_priority.reprioritize")`)

they will have an exception raised on them saying that the plugin cannot find a valid user/bank entry for those previously held jobs.

Is there a process in restarting Flux where jobs could be reprioritized? My thinking is that might have been what caused this.

[...]

PR #407, opened 2024-01-08
plugin: keep jobs in `PRIORITY` after reprioritization

cmoussa1:
#### Problem

As noted in #406, the priority plugin will raise an exception on a job if it is held in `SCHED` state while the plugin is reloaded (or Flux is restarted) and jobs are reprioritized without first loading flux-accounting data to this plugin. This behavior is not graceful and we should instead continue to hold a job in `PRIORITY` while the plugin waits to receive flux-accounting data.

---

This PR adds a check of the plugin's internal map to see if we are still waiting on flux-accounting data to be loaded in; if so, it will continue to hold the job while we wait for data.

I've also added a sharness test that reproduces the issue raised in #406 and ensures that jobs continue to be held after a reprioritization without loading flux-accounting data to the priority plugin.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -666,7 +666,9 @@ static int priority_cb (flux_plugin_t *p,
     if (b->max_run_jobs == BANK_INFO_MISSING) {
         // try to look up user again
         it = users.find (userid);
-        if (it == users.end ()) {
+        if (it == users.end () || check_map_for_dne_only () == true) {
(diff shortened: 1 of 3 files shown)


grondo (approved):
This LGTM @cmoussa1!

I wonder if we should generate a release with this fix to go out with the next TOSS release?

cmoussa1:
Thanks @grondo! Sure, I can generate a release today and have a ticket submitted by COB tomorrow so that it can go in the next TOSS release. Will work on this now 👍 setting MWP here

PR #412, opened 2024-01-30
plugin: add external `Association`  class to be used in plugin

### LINKED ISSUE #411: plugin: create new `Association` class
Siphoned off from #410, I'm going to create separate issues for each component of the improvement process of the priority plugin. This issue will cover the outlined task 1:

A new class should be created called `Association` since that is what we define as the combination of a user who belongs to a certain bank in the flux-accounting DB. This will replace the use of the `bank_info` struct throughout the plugin.

cmoussa1:
#### Background

As mentioned in #410, the plugin uses its own `bank_info` struct to hold user/bank information and has a number of methods for access and modification of these structs. As the plugin's feature set has grown, so have the requirements for the `bank_info` struct, resulting in a very large and hard-to-parse piece of code. It would benefit the readability of this plugin if some basic features were abstracted out to an external class that the plugin could just be `#include`d to the plugin.

---

This PR begins this process of moving some of the definitions and features related to accessing user/bank information from the plugin's internal map to an external class. `accounting.hpp` and `accounting.cpp`, in this PR, are proposed to now hold the following:

- the definition of the `Association` class, an object which holds a set of accounting information of a user who belongs to a bank.

References to this new `Association` class are now used in place of the old `struct bank_info`.

I also added a unit testing framework for `accounting.*`; for now, it just has a test that ensures direct map access works to successfully fetch an `Association`.  As more features are abstracted out of the plugin, this testsuite can grow to ensure things work as expected.

Fixes #411

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -1,0 +1,12 @@
+/************************************************************\
+ * Copyright 2024 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#include "accounting.hpp"
+
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -65,6 +65,7 @@ std::map<std::string, struct queue_info> queues;
 std::map<int, std::string> users_def_bank;
 
 struct bank_info {
+    std::string bank_name;
     double fairshare;
     int max_run_jobs;
     int cur_run_jobs;
@@ -524,6 +525,7 @@ static void rec_update_cb (flux_t *h,
         struct bank_info *b;
         b = &users[uid][bank];
 
+        b->bank_name = bank;
         b->fairshare = fshare;
         b->max_run_jobs = max_running_jobs;
         b->max_active_jobs = max_active_jobs;
@@ -755,6 +757,7 @@ static void add_missing_bank_info (flux_plugin_t *p, flux_t *h, int userid)
     b = &users[userid]["DNE"];
     users_def_bank[userid] = "DNE";
 
+    b->bank_name = "DNE";
     b->fairshare = 0.1;
     b->max_run_jobs = BANK_INFO_MISSING;
     b->cur_run_jobs = 0;
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -29,6 +29,9 @@ extern "C" {
 #include <vector>
 #include <sstream>
 
+// custom bank_info class file
+#include "accounting.hpp"
+
 // the plugin does not know about the association who submitted a job and will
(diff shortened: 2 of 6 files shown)


src/plugins/accounting.cpp
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2024 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+#include "accounting.hpp"
```
jameshcorbett, line 11: What is `accounting.cpp` for if all the code is in `accounting.hpp`? Do you plan to implement methods on the `Association` class in this file later?

src/plugins/test/accounting_test01.cpp
```diff
@@ -0,0 +1,121 @@
...
+    add_user_to_map (users, 1002, "bank_A", user2);
+}
+
+
+static void test_basic_string_comparison ()
+{
+    const std::string string1 = "hello, world!";
+    const std::string string2 = "hello, world!";
+
+    ok (string1 == string2, "i can perform a basic string comparison test");
```
jameshcorbett, line 89: What's this test for?

jameshcorbett (approved):
LGTM, some minor comments below.

src/plugins/accounting.cpp
cmoussa1: Yes, I was planning on adding any methods as well as standalone functions that relate to any `Accounting` objects in this file (I know of at least three at the moment).

src/plugins/mf_priority.cpp
```diff
@@ -479,7 +469,7 @@ static int query_cb (flux_plugin_t *p,
  * Unpack a payload from an external bulk update service and place it in the
  * multimap datastructure.
  */
-static void rec_update_cb (flux_t *h,
+static void rec_update_cb(flux_t *h,
```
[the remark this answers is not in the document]
cmoussa1: Oops! This was an unintended change. Thanks for catching it; will revert

src/plugins/test/accounting_test01.cpp
cmoussa1: Oh, my bad, I had written that test when I was trying to get the unit test file to actually run with `make check`. Sorry about that! I'll just remove this test altogether since it's not related to any flux-accounting stuff.

src/plugins/test/accounting_test01.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+/*
+ * helper function to add a user/bank to the default bank map
+ */
+void add_user_to_def_bank_map (std::map<int, std::string> &users_def_bank,
+                               int userid,
+                               const std::string& bank)
+{
+    users_def_bank[userid] = bank;
+}
```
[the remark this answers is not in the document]
cmoussa1: Not really; now that you mention it, it'd probably just make more sense to write the line of code directly instead of creating a function for it. Will fix this 👍

jameshcorbett (approved):
👍

cmoussa1:

PR #415, opened 2024-02-05
plugin: improve callback for `job.validate`

cmoussa1:
#### Background

There are now helper functions available for the plugin to make use of to help with lookups and access of the internal map for users and banks, but the plugin does not make use of them.

---

This PR adds a new standalone helper function to `association.cpp` called `get_association ()`, which is just a lookup function of accounting information for a user ID and an optional bank. On success, the address of the `Association` object in the plugin's internal map is returned. On failure (if the user ID cannot be found **or** a particular user ID/bank combination cannot be found), `nullptr` is returned.

A couple basic unit tests are also added for `get_association ()`.

I've also added a small commit that changes one of the parameters for a helper function that validates a queue for a given `Association` to expect the list of permissible queues for the `Association` instead of the entire object. Eventually, I think this might be better to move out of the plugin, but for the sake of keeping this PR's scope as narrow as possible, I'll save that for a later date. :-)

Finally, some basic cleanup of `job.validate` is added using this new standalone helper function, as well as some improvements to comment structure within the callback.

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -10,3 +10,28 @@
 
 #include "accounting.hpp"
 
+Association* get_association (int userid,
+                              const char *bank,
+                              std::map<int, std::map<std::string, Association>>
+                                &users,
+                              std::map<int, std::string> &users_def_bank)
+{
+    auto it = users.find (userid);
+    if (it == users.end ())
+        // user could not be found
+        return nullptr;
+
+    std::string b;
+    if (bank != NULL)
+        b = bank;
+    else
+        // get the default bank of this user
+        b = users_def_bank[userid];
+
+    auto bank_it = it->second.find (b);
+    if (bank_it == it->second.end ())
+        // user does not have accounting information under the specified bank
+        return nullptr;
+
+    return &bank_it->second;
+}
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -140,9 +140,8 @@ int64_t priority_calculation (flux_plugin_t *p,
 }
 
 
-static int get_queue_info (
-                      char *queue,
-                      std::map<std::string, Association>::iterator bank_it)
+static int get_queue_info (char *queue,
+                           std::vector<std::string> permissible_queues)
 {
     std::map<std::string, struct queue_info>::iterator q_it;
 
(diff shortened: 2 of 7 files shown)


src/plugins/accounting.cpp
```diff
@@ -10,3 +10,28 @@
 
 #include "accounting.hpp"
 
+Association* get_association (int userid,
+                              const char *bank,
```
jameshcorbett, line 14: Why not take a `std::string` bank and return an `Association&` rather than a pointer? You'd need to throw an exception instead of returning `nullptr` though I think.

src/plugins/mf_priority.cpp
```diff
@@ -798,47 +809,29 @@ static int validate_cb (flux_plugin_t *p,
...
             add_missing_bank_info (p, h, userid);
             return 0;
         } else {
-            return flux_jobtap_reject_job (p, args,
-                                    "no bank found for user: %i", userid);
+            return flux_jobtap_reject_job (p,
+                                           args,
+                                           "cannot find user/bank or "
+                                           "user/default bank entry "
+                                           "for: %i", userid);
```
jameshcorbett, line 829: Maybe add `for uid: %i`? Just to clarify the number that follows is a uid.

jameshcorbett (approved):
Looks reasonable to me!

src/plugins/accounting.cpp
cmoussa1: The reason for using `char *` instead of `std::string` is because the type of `bank` when it is unpacked with `flux_plugin_arg_unpack ()` is of type `char *`, and (to my limited C++ knowledge) there isn't an easy way of converting it to `std::string`.

The reason I chose to `return` a pointer to an `Association` object is because the existence of an `Association` object is not guaranteed (i.e there could be a user or user/bank combo that does not exist in the flux-accounting DB), and the plugin needs to handle this case, which I don't _believe_ is easily handled if I `return` a reference (I could be wrong though).

src/plugins/mf_priority.cpp
cmoussa1: That's a great suggestion, thanks; will fix this.

src/plugins/accounting.cpp
jameshcorbett: Couldn't you do something like pass `std::string{bank}` when you invoke the function, and then take a const reference to it in the function invocation?

And yeah in the case where the object doesn't exist, I don't think you could return a null reference, you'd need to throw something like `std::out_of_range`. 

Anyway, it's fine as it is, it's just my understanding that using raw pointers in C++ is not recommended except in special cases. But I'm not confident in my C++ style so I won't press the point, I think it's good to be merged!

cmoussa1: Thanks! I can look into this further in the near future. My proposal here was just to keep things as similar to what they were beforehand as possible so that it required very little changes within the plugin code itself 😆 . Will set MWP here

PR #417, opened 2024-02-08
plugin: move helper functions for `plugin.query` callback

cmoussa1:
#### Background

The plugin defines a number of helper functions inside the plugin itself to help convert flux-accounting data to JSON format. It would significantly clean up the code if this was moved out of the plugin, especially considering it concerns flux-accounting data specifically.

---

This PR looks to do exactly that. This PR moves the helper functions responsible for creating JSON objects of all of the flux-accounting data in the plugin to `accounting.cpp`.

In the `Association` class, I've added a `.to_json ()` method which converts the object to JSON format. I've also added a new function, `converted_map_to_json ()`, which iterates through all users in the plugin's internal map and creates a JSON object containing all of its information.

As a result, the callback for `plugin.query` is also changed to use this new external function, and the previously defined helper functions inside `mf_priority.cpp` are removed since they are no longer needed. You'll notice that the expected output for the tests for this callback have changed slightly, as I've chosen to include pretty much every attribute of each `Association` object.

##### TODO/might need further feedback or suggestions

- [ ] I noticed that I had to add `-ljansson` to `src/Makefile.am` in order for the unit test file in `src/plugins/test` to run and find all of the jansson methods. Is this the right approach? Perhaps there is something else I am missing or doing something completely wrong.

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -35,3 +35,93 @@ Association* get_association (int userid,
 
     return &bank_it->second;
 }
+
+
+json_t* Association::to_json () const
+{
+    json_t *held_job_ids = json_array ();
+    if (!held_job_ids) {
+        return nullptr;
+    }
+    for (const auto &job_id : held_jobs) {
+        json_t *temp;
+        if (!(temp = json_integer (job_id))
+            || json_array_append_new (held_job_ids, temp) < 0) {
+            json_decref (held_job_ids);
+            return nullptr;
+        }
+    }
+
+    json_t *user_queues = json_array ();
+    if (!user_queues) {
+        json_decref (held_job_ids);
+        return nullptr;
+    }
+    for (const auto &queue : queues) {
+        json_t *temp;
+        if (!(temp = json_string (queue.c_str ()))
+            || json_array_append_new (user_queues, temp) < 0) {
+            json_decref (held_job_ids);
+            json_decref (user_queues);
+            return nullptr;
+        }
+    }
+
+    // 'o' steals the reference for both held_job_ids and user_queues
+    json_t *u = json_pack ("{s:s, s:f, s:i, s:i, s:i, s:i, s:o, s:o, s:i, s:i}",
+                           "bank_name", bank_name.c_str (),
+                           "fairshare", fairshare,
+                           "max_run_jobs", max_run_jobs,
+                           "cur_run_jobs", cur_run_jobs,
+                           "max_active_jobs", max_active_jobs,
+                           "cur_active_jobs", cur_active_jobs,
+                           "held_jobs", held_job_ids,
+                           "queues", user_queues,
+                           "queue_factor", queue_factor,
+                           "active", active);
+
+    if (!u)
+        return nullptr;
+
+    return u;
+}
+
+
+json_t* convert_map_to_json (std::map<int, std::map<std::string, Association>>
+                                &users)
+{
+    json_t *accounting_data = json_array ();
+    if (!accounting_data)
+        return nullptr;
+
+    // each Association in the users map is a pair; the first item is the
+    // userid and the second is a list of banks they belong to
+    for (const auto& association : users) {
+        json_t *banks = json_array ();
+        if (!banks) {
+            json_decref (accounting_data);
+            return nullptr;
+        }
+        for (const auto &bank : association.second) {
+            // bank.second refers to an Association object
+            json_t *b = bank.second.to_json ();
+            if (!b || json_array_append_new (banks, b) < 0) {
+                json_decref (accounting_data);
+                json_decref (banks);
+                return nullptr;
+            }
+        }
+
+        json_t *u = json_pack ("{siso}",
+                               "userid",
+                               association.first,
+                               "banks", banks);
+        if (!u || json_array_append_new (accounting_data, u) < 0) {
+            json_decref (accounting_data);
+            json_decref (banks);
+            return nullptr;
+        }
+    }
+
+    return accounting_data;
+}
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -203,139 +203,6 @@ int check_queue_factor (flux_plugin_t *p,
 }
 
 
-/*
- * Add held job IDs to a JSON array to be added to a bank_info JSON object.
- */
-static json_t *add_held_jobs (
-                            const std::pair<std::string, Association> &b)
-{
-    json_t *held_jobs = NULL;
-
-    // add any held jobs to a JSON array
-    if (!(held_jobs = json_array ()))
-        goto error;
-
-    for (auto const &j : b.second.held_jobs) {
-        json_t *jobid = json_integer (j);
-
-        if (!jobid)
-            goto error;
-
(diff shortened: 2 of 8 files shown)


cmoussa1:
@grondo if you don't mind could you advise on if I have the right approach with including `-ljansson` as a library in `src/Makefile.am`? I noticed I had to include it here in order for the unit tests in `src/plugins/test/` to run. Just want to make sure this is okay, or if maybe there is a better way to handle this.

src/plugins/mf_priority.cpp
```diff
@@ -297,21 +297,21 @@ static int query_cb (flux_plugin_t *p,
                      void *data)
 {
     flux_t *h = flux_jobtap_get_flux (p);
-    json_t *all_users = convert_map_to_json (users);
+    json_t *accounting_data = convert_map_to_json (users);
```
jameshcorbett, line 300: Drop the "I think" from the commit message body maybe?

src/plugins/accounting.cpp
```diff
@@ -35,3 +35,67 @@ Association* get_association (int userid,
...
+
+    json_object_set_new (u, "queue_factor", json_integer (queue_factor));
+    json_object_set_new (u, "active", json_integer (active));
+
+    char *json_str = json_dumps (u, JSON_INDENT (4));
+    std::string result (json_str);
+    free (json_str);
+    json_decref (u);
+
+    return result;
```
jameshcorbett, line 71: There's no error checking in this function. `json_object` for instance can return `NULL` and so can e.g. `json_integer`. I'm not sure what exactly would happen though...

Could you maybe create `held_job_ids` and `user_queues`, checking if either creation fails, add all the elements to them (checking for errors as you go) and then pack everything into `u` with a `json_pack`? You could have the pack steal the references to `held_job_ids` and `user_queues` too. That would combine all your error checking into three locations. Plus one at the end to check that the `json_dumps` succeeds?

src/plugins/accounting.cpp
```diff
@@ -35,3 +35,67 @@ Association* get_association (int userid,
...
+        json_t *u = json_object ();
+        json_object_set_new (u, "userid", json_integer (association.first));
+
+        // the user might belong to multiple banks, so we need to iterate
+        // through each one
+        json_t *banks = json_array ();
+        for (const auto& bank : association.second) {
+            // bank.second refers to an Association object
+            Association a = bank.second;
+            json_t *b = json_loads (a.to_json ().c_str (), 0, nullptr);
```
jameshcorbett, line 92: To avoid writing it out to a string and reading it back in again, I wonder if you could have an `Association` method like `to_json_t` that returns a `json_t` and then have the `to_json` method invoke that method before turning it into a string?

src/plugins/accounting.cpp
```diff
@@ -35,3 +35,67 @@ Association* get_association (int userid,
...
+            json_array_append_new (banks, b);
+        }
+
+        json_object_set_new (u, "banks", banks);
```
jameshcorbett, line 96: There's no error checking in this function either and I think another `pack` might help with error checking rather than setting both `userid` and `banks` independently.

jameshcorbett (commented):
A couple of comments below.

src/plugins/accounting.cpp
cmoussa1: Oh, sure, that's a great suggestion, thanks. Yeah, I think what I could do is move the creation of both `json_array ()`s first (and check for errors upon creation there), and _then_ pack everything all at once into `u` with `json_pack ()`. I'll fix this 👍

src/plugins/accounting.cpp
cmoussa1: Yes, you're absolutely right here - I think what I'll do is convert the `to_json ()` method to return a `json_str *` instead of `std::string`, and this way, I can avoid calling `json_loads ()` altogether.

I'll also change `convert_map_to_json ()` to just use `json_pack ()` like you suggested. 👍

cmoussa1:
Thanks for beginning to take a look at this @jameshcorbett. I've just pushed up some changes to both the `.to_json ()` method and the `convert_map_to_json ()` function per your suggestions. I think I'll just take this PR out of [WIP] too.

src/plugins/accounting.cpp
```diff
@@ -35,3 +35,85 @@ Association* get_association (int userid,
...
+        json_array_append_new (held_job_ids, json_integer (job_id));
+    }
+
+    json_t *user_queues = json_array ();
+    if (!user_queues) {
+        json_decref (held_job_ids);
+        return nullptr;
+    }
+    for (const auto &queue : queues) {
+        json_array_append_new (user_queues, json_string (queue.c_str ()));
```
jameshcorbett, line 56: I think technically the `json_string` call and the `json_array_append_new` should be checked. Like maybe:

```C++

json_t *temp;
if (!(temp = json_string (queue.c_str ())
    || json_array_append_new ( ... ) < 0 ){
return nullptr;
}

```

src/plugins/accounting.cpp
```diff
@@ -35,3 +35,85 @@ Association* get_association (int userid,
...
+                return nullptr;
+            }
+            json_array_append_new (banks, b);
+        }
+
+        json_t *u = json_pack ("{siso}",
+                               "userid",
+                               association.first,
+                               "banks", banks);
+        if (!u) {
```
jameshcorbett, line 109: `if (!u || json_array_append_new (accounting_data, u) < 0) { `

src/plugins/accounting.cpp
```diff
@@ -35,3 +35,85 @@ Association* get_association (int userid,
...
+        }
+        for (const auto &bank : association.second) {
+            // bank.second refers to an Association object
+            json_t *b = bank.second.to_json ();
+            if (!b) {
+                json_decref (accounting_data);
+                json_decref (banks);
+                return nullptr;
+            }
+            json_array_append_new (banks, b);
```
jameshcorbett, line 102: Similar comment as below.

jameshcorbett (approved):
A couple of little nits about checking for errors. Sorry, I know this stuff is really annoying! Other than that LGTM and I think it's ready to go in!

src/plugins/accounting.cpp
jameshcorbett: There's a similar situation a few lines above.

cmoussa1: Oh neat! I didn't know you could combine checks like this in a single line - that's clever. I'll make these changes

cmoussa1:
Just pushed up some changes per @jameshcorbett suggestions - could you give those a quick scan to ensure I understood your suggestion correctly? 😅

jameshcorbett (approved):
Yeah looks good!

cmoussa1:
Appreciate it thanks!! @grondo could you advise on if I have the right approach with including `-ljansson` as a library in `src/Makefile.am`? I noticed I had to include it here in order for the unit tests in `src/plugins/test/` to run. Just want to make sure this is okay, or if maybe there is a better way to do this.

cmoussa1:
OK, I've just pushed up some changes to add a check for `jansson` during `configure` and corrected how I was adding the `jansson` library in the unit test file - thanks @grondo! I'll set MWP here now

PR #418, opened 2024-02-08
plugin: move `split_string ()` out of plugin code

cmoussa1:
#### Background

The `split_string ()` helper function can only accept a list of queues to parse when unpacking user/bank information from the
flux-accounting DB, but there will also exist a need in the future to be able to parse other string lists as well (e.g projects). Since both lists can be parsed the same way, this helper function should be expanded. Also, this helper function is specific to parsing flux-accounting data. It might make more sense to instead place this in `accounting.cpp`.

---

This PR moves the function definition out of the priority plugin code and into `accounting.cpp`. It also expands the `split_string ()` helper function to accept a general string list of items and renames it to `split_string_and_push_back ()`.

A unit test is also added for this function.

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -125,3 +125,18 @@ json_t* convert_map_to_json (std::map<int, std::map<std::string, Association>>
 
     return accounting_data;
 }
+
+
+void split_string_and_push_back (const char *list,
+                                 std::vector<std::string> &vec)
+{
+    std::stringstream s_stream;
+
+    s_stream << list; // create string stream from string
+
+    while (s_stream.good ()) {
+        std::string substr;
+        getline (s_stream, substr, ','); // get string delimited by comma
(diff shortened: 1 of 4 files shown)


src/plugins/accounting.cpp
```diff
@@ -125,3 +125,18 @@ json_t* convert_map_to_json (std::map<int, std::map<std::string, Association>>
 
     return accounting_data;
 }
+
+
+void split_string_and_push_back (const char *list,
+                                 std::vector<std::string> *vec)
```
jameshcorbett, line 131: I think `vec` could be a reference here without losing anything.

jameshcorbett (approved):
LGTM! One minor suggestion.

cmoussa1:
Thanks @jameshcorbett! Just pushed up a change to the function definition so that `vec` is a reference. Will set MWP here

PR #423, opened 2024-02-22
plugin: improve `job.update`/`job.update...queue` callbacks

### LINKED ISSUE #422: `job.update`/`job.update...queue`: use new external methods for association lookup
The callbacks for `job.update` and `job.update.attributes.system.queue` can be cleaned up to use the new external functions for association lookups instead of implementing their own manual lookups.

cmoussa1:
#### Background

Continuing the cleanup of the priority plugin code by utilizing the external functions defined in `accounting.cpp`.

---

This PR continues to clean up the plugin code by improving the callbacks for `job.update` and `job.update.attributes.system.queue`. Specifically, it removes the use of the clunky `get_bank_info ()` helper function defined in the plugin itself and instead uses the lookup function defined in `accounting.cpp`. The function descriptions and comments are updated in both functions, as well as some variable names. Finally, the `get_bank_info ()` helper function and its associated `typedef` and `enum`s are removed from the plugin code, since they are no longer needed.

Fixes #422

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -898,7 +898,6 @@ static int job_updated (flux_plugin_t *p,
                         flux_plugin_arg_t *args,
                         void *data)
 {
-    std::map<std::string, Association>::iterator bank_it;
     int userid;
     char *bank = NULL;
     char *queue = NULL;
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
Looks reasonable to me!

cmoussa1:
Thanks! I'll set MWP here

PR #425, opened 2024-02-26
[WIP] plugin: improve `job.state.priority` callback

cmoussa1:
#### Background

Continuing the improvement process on the priority plugin. The callback for `job.state.priority` uses a manual lookup for associations in the case where we need to re-look up a user's accounting information, but accounting.cpp has a cleanly defined implementation of this now.

---

This PR reformats the `job.state.priority` callback to make use of the new external lookup function defined in `job.state.priority`. It improves some of the comments throughout the function and adds new ones. Finally, it adjusts the test that checks for an exception message in `job.state.priority` to account for the new message format (it matches the exception message in `job.validate`).

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -457,67 +457,58 @@ static int priority_cb (flux_plugin_t *p,
         return -1;
     }
 
-    std::map<int, std::map<std::string, Association>>::iterator it;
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
LGTM

cmoussa1:
Thanks! Setting MWP

PR #427, opened 2024-02-28
plugin: move flux-accounting-specific helper functions, remove unused ones

### LINKED ISSUE #426: plugin: move accounting-specific helper functions to `accounting.cpp`
The priority plugin has defined a number of helper functions that are specific to flux-accounting, and thus would make more sense if they were instead defined in `accounting.cpp`. After a perusal through the plugin code, the functions that make the most sense to move out would be `check_queue_factor ()` and `check_map_for_dne_only ()`.

cmoussa1:
#### Background

Continuing to move/abstract things out of the plugin code and into `accounting.*`.

---

This PR moves the `get_queue_info` and `check_map_for_dne_only ()` definitions out of the plugin code and into the `accounting.*` files. It also converts the `struct queue_info` into a class `Queue` to match the structure of `class Association`.  It improves some of the comments in the `get_queue_info ()` function and cleans up a little bit of the `if-else` checks as well. Unit tests are also now added for the `get_queue_info ()` function.

A couple of unused helper functions previously defined in the plugin code `split_string ()` and `check_queue_factor ()` are also removed from the plugin code since they are no longer needed.

#### TODO

- [x] add unit tests for `check_map_for_dne_only ()`

Fixes #426

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -143,8 +143,8 @@ void split_string_and_push_back (const char *list,
 
 
 int get_queue_info (char *queue,
-                    std::vector<std::string> permissible_queues,
-                    std::map<std::string, struct queue_info> queues)
+                    const std::vector<std::string> &permissible_queues,
+                    const std::map<std::string, Queue> &queues)
 {
     if (queue != NULL) {
         // check #1) the queue passed in exists in the queues map;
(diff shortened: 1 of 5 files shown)


src/plugins/accounting.cpp
```diff
@@ -140,3 +140,37 @@ void split_string_and_push_back (const char *list,
         vec.push_back (substr);
     }
 }
+
+
+int get_queue_info (char *queue,
+                    std::vector<std::string> permissible_queues,
+                    std::map<std::string, struct queue_info> queues)
```
jameshcorbett: Consider taking the last two arguments as const references? Because I think at the moment they're passed as copies, which is a bit wasteful.

jameshcorbett (approved):
Looks good, one minor comment.

src/plugins/accounting.cpp
cmoussa1: Agreed! Thanks for pointing this out. Just pushed up a fix to change those to const references.

cmoussa1:

PR #429, opened 2024-03-08
plugin: add support for updating the bank of a pending job

cmoussa1:
_this is a re-implementation of the changes proposed in #400 now that the priority plugin has been changed to make use of the new classes and external functions defined in `accounting.cpp`._

#### Problem

flux-accounting currently has no support for a user/bank to update the bank they want their pending job to run under.

---

This is a PR that adds support for updating the bank of a pending job. It adds a new callback to the plugin which looks for the `"job.update.attributes.system.bank"` topic string and validates the new bank that the user is trying to update their job to. If they have access to the bank they want to update their job to _but_ this new bank is currently at its max running jobs or active jobs limits, the plugin will reject the update. Since `job.state.depend` is the final place where a job can have dependencies added to it (and adding/removing dependencies is how the plugin handles running job limits), I've chosen to just reject the update and keep the job under its current bank.

Once a new bank has been validated, the job's `Association` object associated with the job is updated in the `job.update` callback. You'll notice that before the new attributes/limits are assigned to the job, the `active_jobs` count of the old bank is decremented. Then, just before the new `Association` object is assigned to the job, the `active_jobs` count of the new bank is incremented (since this count usually gets incremented when the job is in `job.new`).

A new test file, `t1031-mf-priority-update-bank.t` is also added to the test suite which simulates submitting some jobs under default and non-default banks and updating the bank between each, while also making sure that a job being updated to a bank that is already at their current max running jobs limit gets rejected.

I've also added a test file that simulates updating both the queue and the bank of a job at the same time and making sure behavior is as expected.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -733,8 +733,8 @@ static int run_cb (flux_plugin_t *p,
 
 
 /*
- *  Apply an update on a job with regard to its queue once it has been
- *  validated.
+ *  Apply an update on a job with regard to its queue or associated bank once
+ *  it has been validated.
  */
 static int job_updated (flux_plugin_t *p,
                         const char *topic,
@@ -744,15 +744,19 @@ static int job_updated (flux_plugin_t *p,
     int userid;
     char *bank = NULL;
     char *updated_queue = NULL;
+    char *updated_bank = NULL;
     Association *a;
 
+    flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:i, s{s{s{s?s}}}, s:{s?s}}",
+                                "{s:i, s{s{s{s?s}}}, s:{s?s, s?s}}",
                                 "userid", &userid,
-                                "jobspec", "attributes", "system", "bank", &bank,
+                                "jobspec", "attributes", "system", "bank",
+                                &bank,
                                 "updates",
-                                "attributes.system.queue", &updated_queue) < 0)
+                                  "attributes.system.queue", &updated_queue,
+                                  "attributes.system.bank", &updated_bank) < 0)
         return flux_jobtap_error (p, args, "unable to unpack plugin args");
 
     // grab Association object from job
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
LGTM!

src/plugins/mf_priority.cpp
```diff
@@ -824,6 +863,59 @@ static int update_queue_cb (flux_plugin_t *p,
...
+
+    if (a->max_active_jobs > 0 && a->cur_active_jobs >= a->max_active_jobs)
+        // new bank is already at its max active jobs limit; reject update
+        return flux_jobtap_error (p,
+                                  args,
+                                  "mf_priority: new bank is already at "
+                                  "max-active-jobs limit");
+
+    if (a->max_run_jobs > 0 && a->cur_run_jobs == a->max_run_jobs)
+        // new bank is already at its max running jobs limit; reject update
```
grondo, line 909: It may be helpful to expand this comment. If a user can submit a job to a bank that is already it is maximum running jobs limit, then it might be confusing as to why they can't also update a job to that bank. I'm guessing this has something to do with the fact that jobs over the max-run-jobs limit are held in DEPEND state, and there isn't a way to kick a job back to that state?

A comment here detailing that reason (or the real reason if I got it wrong), and a more specific error returned to the user, e.g. "updating to bank %s at max-run-jobs limit not allowed. Try again later" or similar.

src/plugins/mf_priority.cpp
```diff
@@ -824,6 +863,59 @@ static int update_queue_cb (flux_plugin_t *p,
...
+                                       "cannot find user/bank or "
+                                       "user/default bank entry "
+                                       "for uid: %i", userid);
+
+    if (a->max_active_jobs > 0 && a->cur_active_jobs >= a->max_active_jobs)
+        // new bank is already at its max active jobs limit; reject update
+        return flux_jobtap_error (p,
+                                  args,
+                                  "mf_priority: new bank is already at "
+                                  "max-active-jobs limit");
```
grondo, line 906: Suggestion: Probably unnecessary to add `mf_priority:` to the error message. Users won't know what that is and the error is meant primarly for them. (Same for other error message here)

t/t1032-mf-priority-update-bank.t
```diff
@@ -0,0 +1,127 @@
...
+test_expect_success 'submit job under non-default bank' '
+	job6=$(flux python ${SUBMIT_AS} 5001 --setattr=bank=B --urgency=0 sleep 30)
+'
+
+test_expect_success 'updating job to default bank works' '
+	flux update $job6 bank=A &&
+	flux job wait-event -t 30 $job6 priority &&
+	flux job eventlog $job6 > eventlog.out &&
+	grep "attributes.system.bank=\"A\"" eventlog.out
+'
```
grondo, line 121: Is there a way to also check that the mf_priority plugin also sees the updated bank?

grondo (approved):
Nice work! Just a couple suggestions inline.

src/plugins/mf_priority.cpp
cmoussa1: That's a great suggestion, thanks! I went ahead and edited this error message to be a little more descriptive as well as added a slightly more detailed comment.

t/t1032-mf-priority-update-bank.t
cmoussa1: Yes there is, thanks for this suggestion! I added another check in this test file that queries the plugin's map and checks that the intended update is recognized:

```
test_expect_success 'check that plugin also sees the job update' '
	flux jobtap query mf_priority.so > query.json &&
	jq -e ".mf_priority_map[0].banks[0].bank_name == \"A\"" <query.json &&
	jq -e ".mf_priority_map[0].banks[0].cur_active_jobs == 1" <query.json &&
	jq -e ".mf_priority_map[0].banks[1].bank_name == \"B\"" <query.json &&
	jq -e ".mf_priority_map[0].banks[1].cur_active_jobs == 0" <query.json
'
```

cmoussa1:
Thanks for taking a look at this @jameshcorbett and @grondo! I've pushed up some changes via the suggestions above. Will set MWP here shortly

PR #430, opened 2024-03-14
plugin: improve `add_missing_bank_info ()`

cmoussa1:
#### Problem

The `add_missing_bank_info ()` function is defined in the "Callbacks" section of the plugin code when it is really a helper function. The function is also poorly named and not very reflective of what the function is actually doing.

---

This PR improves this function by moving it to the "Helper Functions" section of the plugin code as well as renames it to `add_special_association ()`. It edits the calls to this function to account for the name change and adds a function description as well.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -132,6 +132,35 @@ static int update_jobspec_bank (flux_plugin_t *p, int userid)
 }
 
 
+/*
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
Looks reasonable!

cmoussa1:
Thanks! Setting MWP here

PR #431, opened 2024-03-15
plugin: move `flux_respond ()` to end of functions

cmoussa1:
#### Problem

There are a couple callbacks in the plugin that call `flux_respond ()` before they are done processing.

---

This PR moves these calls to just before the functions call `return`.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -263,9 +263,6 @@ static void rec_update_cb (flux_t *h,
         goto error;
     }
 
-    if (flux_respond (h, msg, NULL) < 0)
-        flux_log_error (h, "flux_respond");
-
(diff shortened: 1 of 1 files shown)


jameshcorbett (commented):
Looks fine but does have any effect other than changing the order of log messages in the event of an error?

cmoussa1:
I don't think so. I think this call is supposed to be placed at the end of the callback in the event where there is no error. This commit was actually part of some work on an old (and now closed PR) that was pointed out by @grondo - maybe he could confirm that this is still ok?

cmoussa1:
@grondo if you have some time this week or next, could you please give this PR a look?

grondo (approved):
LGTM! Nice cleanup!

cmoussa1:
Thanks!! setting MWP here

PR #432, opened 2024-03-15
t: add `active` column, move sample payloads

cmoussa1:
#### Problem(s)

Some sharness tests in flux-accounting's testsuite don't contain the `active` column in the payloads that get sent to the priority plugin.

These sample payloads are starting to grow very large because they have a lot of fields per-user, and they clog up the test files.

---

This PR adds the `active` column to the key-value pairs of the test payloads that get sent to the priority plugin in the sharness tests.

It also moves these key-value pairs to separate JSON files that get loaded into their respective test files instead of creating them within the test directly.


jameshcorbett (approved):
👍

cmoussa1:
Thanks for approving @jameshcorbett! Setting MWP here

PR #434, opened 2024-03-20
plugin: add project information to Association information in plugin

cmoussa1:
#### Background

flux-accounting has the ability to define and list projects for associations in the database, but the plugin does not know about or interact with them. Eventually, the plugin will need to support associating jobs with projects. This is the first step to getting there.

---

This PR looks to start the process of adding project support to the priority plugin by just adding project information to the flux-accounting information that gets sent from the database to the plugin. Each `Association` object now has two new member fields: `projects` and  `def_project`, fields to store their list of projects and their default project, respectively. This info, along with the list of all registered projects, is sent and unpacked and stored in the plugin. At the moment, it is not utilized anywhere in the plugin, it is simply just unpacked and added to each `Association` object. Future PRs will look to integrate actual support for associating jobs with projects.

The other piece of this PR is to add a helper function to `accounting.cpp` to help validate projects for a given association. The plan with this function is to use it when validating a job that charges a certain project. Unit tests are also added for this helper function.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1060,9 +1060,9 @@ static const struct flux_plugin_handler tab[] = {
 extern "C" int flux_plugin_init (flux_plugin_t *p)
 {
     if (flux_plugin_register (p, "mf_priority", tab) < 0
-        || flux_jobtap_service_register (p, "rec_update", rec_update_cb, p)
-        || flux_jobtap_service_register (p, "reprioritize", reprior_cb, p)
-        || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p)
+        || flux_jobtap_service_register (p, "rec_update", rec_update_cb, p) < 0
+        || flux_jobtap_service_register (p, "reprioritize", reprior_cb, p) < 0
+        || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p) < 0
         || flux_jobtap_service_register (p, "rec_proj_update", rec_proj_cb, p)
         < 0)
         return -1;
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -40,6 +40,7 @@ extern "C" {
 std::map<int, std::map<std::string, Association>> users;
 std::map<std::string, Queue> queues;
 std::map<int, std::string> users_def_bank;
+std::vector<std::string> projects;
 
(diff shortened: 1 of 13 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -1007,7 +1063,9 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
     if (flux_plugin_register (p, "mf_priority", tab) < 0
         || flux_jobtap_service_register (p, "rec_update", rec_update_cb, p)
         || flux_jobtap_service_register (p, "reprioritize", reprior_cb, p)
-        || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p) < 0)
+        || flux_jobtap_service_register (p, "rec_q_update", rec_q_cb, p)
+        || flux_jobtap_service_register (p, "rec_proj_update", rec_proj_cb, p)
+        < 0)
```
jameshcorbett, line 1068: Why the `< 0` check only for the last one in this `||` chain?

src/plugins/mf_priority.cpp
```diff
@@ -323,6 +331,54 @@ static void rec_q_cb (flux_t *h,
...
+    if (!data || !json_is_array (data)) {
+        flux_log (h, LOG_ERR, "mf_priority: invalid project info payload");
+        goto error;
+    }
+    num_data = json_array_size (data);
+
+    // clear the projects vector
+    projects.clear ();
+
+    for (int i = 0; i < num_data; i++) {
```
jameshcorbett, line 362: This looks like a good place for `json_array_foreach`?

jameshcorbett (approved):
I don't really know how projects are supposed to work but these changes look reasonable.

src/plugins/mf_priority.cpp
cmoussa1: I think you're right! I converted this for-loop to use `json_array_foreach ()`. Thanks for pointing that out

src/plugins/mf_priority.cpp
cmoussa1: Oops, I guess I forgot to include a `< 0` check for the other calls here. Thanks for catching this! I added another commit to this PR that adds checks for all of the `flux_jobtap_service_register ()` calls here

cmoussa1:
Thanks for approving @jameshcorbett - will set MWP here shortly

PR #435, opened 2024-03-28
`plugin.query`: add projects, def_project to the information returned

cmoussa1:
#### Background

Projects and default project information is now sent to the plugin, but the project information for each association is not returned in the `plugin.query` callback.

---

This PR adds the `projects` and `def_project` fields to the list of information returned for each association in the `plugin.query` callback.

It also adds the `projects` and `def_project` key-value pairs to the expected files for `t1019-mf-priority-info-fetch.t`. It adds some projects to one of the associations in the test and ensures those projects show up when looking at the ouptut of the `plugin.query` callback.

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -67,8 +67,26 @@ json_t* Association::to_json () const
         }
     }
 
+    json_t *user_projects = json_array ();
+    if (!user_projects) {
+        json_decref (held_job_ids);
+        json_decref (user_queues);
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
LGTM, I think the commit message body could use a "problem:" though?

cmoussa1:
You're absolutely right @jameshcorbett - thanks! I pushed up a change to add a **problem** statement. Will set MWP here

PR #437, opened 2024-03-29
plugin: add `max_nodes` as an attribute per-association in plugin

### LINKED ISSUE #436: plugin: send max nodes information per-association
The priority plugin currently does not have any information about max nodes counts per-association, which it will need in order for it to enforce a max nodes limit.

The limit already exists per-association in the flux-accounting DB, so it just needs to be added to the other info that is sent during an update.

cmoussa1:
~~_note: will keep this as [WIP] until after the next flux-accounting release targeted for next week_~~

#### Background

As mentioned in #436, the priority plugin currently does not have any information about max nodes counts per-association, which it will need in order for it to enforce a max nodes limit. The limit already exists per-association in the flux-accounting DB, so it just needs to be added to the other info that is sent during an update.

---

This PR does just that - it adds `max_nodes` to the set of information that is sent per-association from the flux-accounting DB to the plugin. Adjustments to the `Association` class, the callback responsible for unpacking flux-accounting data, the callback responsible for returning loaded flux-accounting information (`plugin.query`), and appropriate tests are made to account for the addition of `max_nodes`.

This PR will set some of the groundwork needed to begin enforcing a `max_nodes` limit via the priority plugin.

Fixes #436

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -28,6 +28,7 @@ extern "C" {
 #include <cinttypes>
 #include <vector>
 #include <sstream>
+#include <cstdint>
 
 // custom bank_info class file
 #include "accounting.hpp"
@@ -152,6 +153,7 @@ static void add_special_association (flux_plugin_t *p, flux_t *h, int userid)
     a->cur_active_jobs = 0;
     a->active = 1;
     a->held_jobs = std::vector<long int>();
+    a->max_nodes = INT16_MAX;
 
     if (flux_jobtap_job_aux_set (p,
                                  FLUX_JOBTAP_CURRENT_JOB,
@@ -207,7 +209,7 @@ static void rec_update_cb (flux_t *h,
                            void *arg)
 {
     char *bank, *def_bank, *assoc_queues, *assoc_projects, *def_project = NULL;
-    int uid, max_running_jobs, max_active_jobs = 0;
+    int uid, max_running_jobs, max_active_jobs, max_nodes = 0;
     double fshare = 0.0;
     json_t *data, *jtemp = NULL;
     json_error_t error;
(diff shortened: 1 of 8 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -152,6 +152,7 @@ static void add_special_association (flux_plugin_t *p, flux_t *h, int userid)
     a->cur_active_jobs = 0;
     a->active = 1;
     a->held_jobs = std::vector<long int>();
+    a->max_nodes = 2147483647;
```
jameshcorbett, line 155: You're going for the maximum integer value here? I think it would be a bit cleaner to `#include <stdint.h>` and use `INT32_MAX` or `<limits.h>` and `INT_MAX`. Or maybe pick a 'nicer' number like 10000000? I was kind of surprised to see this number here seemingly out of nowhere. An inline comment might be helpful.

jameshcorbett (approved):
Looks good, one comment.

src/plugins/mf_priority.cpp
cmoussa1: Oh that's good to know, and a good suggestion - thanks! I think because this is C++, I should use `#include <cstdint>` here. And perhaps I should change `max_nodes` here to be of type `std::int32_t`. Any opinions on this choice?

jameshcorbett: That sounds fine to me, since it looks like `int` is guaranteed to be at least 16 bits but can be more, so `INT_MAX` might vary across systems and you need a stable number for testing. You could also keep it as an `int` and use `INT16_MAX`, up to you.

cmoussa1: I think I'll just use `INT16_MAX`; thanks!

cmoussa1:
Thanks for reviewing @jameshcorbett ! I just pushed up some changes per your suggestions above; will set MWP here

PR #438, opened 2024-03-29
plugin: change `projects`->`assoc_projects` in `rec_update_cb ()`

cmoussa1:
#### Problem

The `projects` variable name in the `rec_update_cb ()` function is poorly named because it can cause confusion with the global variable `projects`. It should be renamed to something clearer, as it represents the projects that the association has access to.

---

This PR renames `projects`->`assoc_projects` in the `rec_update_cb ()` function.


wihobbs (approved):
LGTM!

cmoussa1:
i appreciate it thanks! Setting MWP here

PR #441, opened 2024-04-04
repo: add `pkg.m4`, checks for flux-core libs

cmoussa1:
#### Problem

The flux-accounting repo does not check for certain flux-core libraries that it will eventually need to use, such as the `hostlist` library.

---

This PR adds `pkg.m4` to the `config/` directory which will enable flux-accounting to check for flux-core libraries while it is building. It uncomments the `PKG_CHECK_MODULES` calls in `ax_flux_core.m4` so that flux-accounting actually checks for these libraries, and finally adds output to `configure.ac` to show the location of flux-core's `hostlist` library.


config/ax_flux_core.m4
```diff
@@ -83,9 +85,10 @@ AC_DEFUN([AX_FLUX_CORE], [
   ]
 
   #  Check for other flux-core libraries
-  #PKG_CHECK_MODULES([FLUX_IDSET], [flux-idset], [], [])
-  #PKG_CHECK_MODULES([FLUX_SCHEDUTIL], [flux-schedutil], [], [])
-  #PKG_CHECK_MODULES([FLUX_OPTPARSE], [flux-optparse], [], [])
+  PKG_CHECK_MODULES([FLUX_IDSET], [flux-idset], [], [])
+  PKG_CHECK_MODULES([FLUX_SCHEDUTIL], [flux-schedutil], [], [])
+  PKG_CHECK_MODULES([FLUX_OPTPARSE], [flux-optparse], [], [])
```
grondo, line 90: Probably don't need to check for libflux-schedutil.so and libflux-optparse.so.

grondo (approved):
LGTM, one take-it-or-leave-it comment inline.

config/ax_flux_core.m4
cmoussa1: Thanks! I'll go ahead and remove those.

cmoussa1:
Thanks for reviewing @grondo! Will set MWP here

PR #443, opened 2024-04-12
plugin: add project validation/annotation

cmoussa1:
#### Background

The priority plugin does not do any validation or annotation for a specified _or_ default project associated with an association's jobs.

---

This PR looks to add basic support for validating a project if one is specified when submitting a job as well as adding a default project to jobspec when one is not specified. If a project is specified on the command line, it is validated by looking at the association's list of valid projects. If the specified project cannot be found, the job is rejected.

In the case where no project is specified, the association's default project is added to jobspec via `jobspec-update`, similar to how the default bank for an association is added to jobspec.

Some basic tests are also added that simulate setting project names in multiple cases - where the plugin is still waiting for flux-accounting data to be loaded, where associations set specific projects, or use and/or update their default project.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -588,6 +588,7 @@ static int validate_cb (flux_plugin_t *p,
     int userid;
     char *bank = NULL;
     char *queue = NULL;
+    const char *project = NULL;
     flux_job_state_t state;
     int max_run_jobs, cur_active_jobs, max_active_jobs, queue_factor = 0;
     double fairshare = 0.0;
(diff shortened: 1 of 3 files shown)


cmoussa1:
@ryanday36 if you have some time this week, do you think you could give this PR a high-level look? I want to make sure my understanding of how project support is to work in flux-accounting is as expected, or if you were looking for something else.

ryanday36:
This looks pretty good. My high level question is, are we going to need to set a default project for every association, or can we just set a global default for all associations that don't have a project?

cmoussa1:
All associations that do not have a specified project when they are added to the flux-accounting database will have a default project of `*` assigned to them and their jobs. Does that sound okay?

ryanday36:
That's perfect. Thanks!

cmoussa1:
Thanks! I'll mark this as ready for review then

grondo (approved):
This looks good to me!

cmoussa1:

PR #444, opened 2024-04-18
[WIP] plugin: add max nodes limit support

cmoussa1:
#### Background

The priority plugin does not implement a max nodes limit to be enforced on an association's set of currently running jobs.

---

This PR is built on top of #442 and looks to add basic support for enforcing a max nodes limit on an association's set of running jobs. The method for implementing this limit follows the same template as the max running jobs limit; when a job is in `job.state.depend`, the nodes the submitted job is looking to use is extracted from jobspec and added to the association's current  node count. If a job does not specify any nodes, the plugin just assumes a node count of 1 for the job. If the job would put the association over their limit, a dependency is added to the job and it is held until a currently running job finishes and cleans up.

To do this, I've proposed combining the two limits (max running jobs & max nodes) into one named dependency. So, if an association hits _either_ their max running jobs or max node limit(s), the same dependency is added. 

In the callback for `job.state.inactive`, the logic for releasing a held job due to a flux-accounting limit is slightly reworked. If an association is under their max running jobs limit, the first held job is grabbed and its node count is inspected, similar to how it is checked in `job.state.depend`. If releasing this held job would keep the association under or equal to their max nodes limit, the dependency is removed and the job can move on to being run. If the limit cannot be satisfied, the dependency is not removed and _no_ held jobs are released until another one of the association's currently running jobs finishes and cleans up.

A couple of basic tests are added to `1034-mf-priority-max-nodes.t` to simulate submitting jobs that take up all of the association's node limit and having a job held due to their max nodes limit. Once the currently running job finishes, a test checks that the held job transitions to run.

##### TODO

- [ ] I should probably add some tests that simulate submitting some jobs that don't specify any nodes to make sure it works as expected


src/plugins/mf_priority.cpp
```diff
@@ -1086,6 +1088,20 @@ static int inactive_cb (flux_plugin_t *p,
...
+    // also decrement the cur_node count for the association
+    if (R != NULL) {
+        std::string nodelist = extract_nodelist (R);
+        if (!nodelist.empty ()) {
+            int nnodes = process_nodelist (nodelist);
+            if (nnodes < 0)
+                flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB,
+                                             "mf_priority", 0,
+                                             "job.state.inactive: error "
+                                             "counting nnodes for job");
```
grondo, line 1146: Since the job is inactive, raising an exception probably won't do anything here.
(Also, the decrement of `b->cur_nodes` should be skipped if `nnodes < 0` to avoid instead _incrementing_ it. The same goes for where `b->cur_nodes` is incremented in the previous commit)

Actually I wonder if there's a good way for the plugin to save the `nnodes` for a job in the job aux item list so it doesn't even have to be recalculated here. Let me think on that.

t/t1034-mf-priority-max-nodes.t
```diff
@@ -0,0 +1,103 @@
...
+
+MULTI_FACTOR_PRIORITY=${FLUX_BUILD_DIR}/src/plugins/.libs/mf_priority.so
+SUBMIT_AS=${SHARNESS_TEST_SRCDIR}/scripts/submit_as.py
+DB_PATH=$(pwd)/FluxAccountingTest.db
+
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 5 job -o,--config-path=$(pwd)/conf.d
+
+flux setattr log-stderr-level 1
+
```
grondo, line 17: Note: Due to changes in flux-core you'll now have to enable guest access to the testexec implentation with
```
flux config load <<-EOF
[exec.testexec]
allow-guests = true
EOF
```

src/plugins/mf_priority.cpp
```diff
@@ -202,6 +202,30 @@ int process_nodelist (const std::string& nodelist_str) {
...
+    if (!type || !json_is_string (type)) return -1;
+
+    // if the association did not explicitly request nodes for their job,
+    // just return 1
+    if (strcmp (json_string_value (type), "node") != 0) return 1;
+
+    if (!count || !json_is_integer (count)) return -1;
+    int nnodes = json_integer_value (count);
+    return nnodes;
+}
```
grondo, line 226: Let's consider how to do this more reliably. Probably returning 1 node if a user doesn't explicitly request nodes is not going to work in production because users could bypass all node limits by always requesting only cores:  e.g. `flux alloc -n64000`

Instead we may have to do something more like the flux-core [limit-job-size](https://github.com/flux-framework/flux-core/blob/master/src/modules/job-manager/plugins/limit-job-size.c) plugin, where there is a limit on the total number of cores+nodes+gpus and if a job exceeds any one of those it is deferred.

grondo (commented):
I've made a first pass here.

I think the main thing we have to discuss and consider is expanding the "max nodes" concept to "max size" including cores+nodes+gpus.

Another thing that came to mind while I was looking through this:
 - I wonder if we have to add more helpers to the jobtap interface so that 3rd party plugins don't have to parse jobspec and R (e.g. something like `flux_jobtap_get_jobspec_size()` and `flux_jobtap_resource_size()`) so that every plugin doesn't have to calculate these values individually (and potentially get it wrong)

PR #448, opened 2024-05-13
NEWS: add release notes for v0.32.0

cmoussa1:
#### Problem

flux-accounting has no release notes for `v0.32.0`.

---

This PR adds some release notes.

PR #449, opened 2024-05-13
testsuite: enable guest access to testexec

grondo:
Problem: Guest user access to the job-exec 'testexec' implementation will soon be denied by default, but many tests in the flux-accounting testsuite require such access to test multiuser jobs without access to a system instance.

For all tests that require the testexec implementation, set

 [exec.testexec]
 allow-guests = true

so that these tests continue working after changes in flux-core deny guest access.

This is necessary for flux-accounting tests to pass in flux-framework/flux-core#5961

### DIFF
--- a/t/t1001-mf-priority-basic.t
+++ b/t/t1001-mf-priority-basic.t
@@ -13,6 +13,12 @@ test_under_flux 1 job
 
 flux setattr log-stderr-level 1
 
+test_expect_success 'allow guest access to testexec' '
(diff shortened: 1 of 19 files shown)


cmoussa1 (approved):
Thanks @grondo! This LGTM. If you want, let's get this merged before #448 - I can add it to the release notes for `v0.32.0`.

grondo:
Sure, sounds good. I'll add MWP here.

Issue #451, opened 2024-05-14
flux account commands hang while fairshare is being updated


ryanday36:
I've noticed that, on corona, `flux account` commands hang when the fairshare tasks in the `accounting` cron tab are running. E.g. `flux account view-user ...` usually runs in less than a second, but takes ~2.5 minutes when run at the same time as the accounting cron tab. That's not terrible, so this probably isn't the highest priority, but it would be good if those updates could be done in a way that doesn't block queries.

I also don't know how much scaling testing has been done with the `update-usage` and related scripts, but I do worry that this could be a larger issue as we move Flux to larger systems with, potentially, many more jobs.

cmoussa1:
Thanks for pointing this out. Definitely something to look into further. I know that SQLite is pretty lightweight, and when there are a heavy number of concurrent reads and writes, especially for a fair-share update (which is a lot of writes), it could lead to low concurrency.

I wonder if there is potential to optimize the command that updates the job usage values for all of the associations in a flux-accounting DB. Right now, the command iterates through every row in the `association_table` and if it needs to make an update, it acquires a lock on the database to make that update. Repeat this process for a large number of users, and that's a lot of locks... I'll look into seeing if I can perhaps wrap all of the updates that would happen here into a single transaction that gets written to the database _at once_ instead of for each row.

PR #452, opened 2024-05-14
job-archive interface: wrap job usage updates into a single SQL transaction

cmoussa1:
#### Problem

When flux-accounting needs to do a job usage update for all of the associations in the `association_table`, it needs to make a number of updates to a number of tables in the DB. If there are a lot of associations, this can slow down the performance of a number of other, much simpler commands, such as `flux account view-user`. This is because the job usage updates makes a separate transaction for each UPDATE statement performed, forcing other reads from the DB to wait until the updates are completed.

---

This PR drops the separate `commit()` calls from the various helper functions that execute UPDATE statements in `job_archive_interface.py` and instead wraps the entire set of database updates into a single SQLite transaction, which will only be committed to the DB after all of the updates are done. This should hopefully begin to relieve some of the headache mentioned in #451.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -272,7 +272,6 @@ def update_t_inactive(acct_conn, last_t_inactive, user, bank):
             bank,
         ),
     )
-    acct_conn.commit()
 
 
 def get_last_job_ts(acct_conn, user, bank):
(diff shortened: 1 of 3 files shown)


ryanday36 (approved):
This looks like a good approach to me. I assume there are correctness tests to make sure that the usage is properly calculated. I'm not quite familiar enough with how the update_job_usage code is working to know if, for example, the usage factor changes need to be committed before the bank usage is updated.

cmoussa1:
Thanks for reviewing @ryanday36 - the sharness test added in `t1011-job-archive-interface.t` is in fact checking job usage values for correctness. I was also able to confirm that wrapping up the updates into a single transaction is OK for calculating both user and bank usage values - subsequent updates made into a single transaction should be readable by each other. 👍 I'll set MWP here

PR #453, opened 2024-05-15
database: update schema version

cmoussa1:
#### Problem

The schema version of the flux-accounting database was never updated with the addition of the jobs table in #357. As a result, when newer versions of flux-accounting are installed on a system, the automatic update of the database is never triggered by the background processes of flux-accounting.

---

This PR updates the schema version number of the flux-accounting database.


jameshcorbett (approved):
👍

cmoussa1:
Thanks @jameshcorbett! Setting MWP

PR #454, opened 2024-05-29
testsuite: replace `flux job cancel` --> `flux cancel`

cmoussa1:
#### Problem

There are a number of tests in the flux-accounting testsuite that still use `flux job cancel`, but this is outdated in favor for `flux cancel`.

---

This PR replaces the use of `flux job cancel` with `flux cancel` in the flux-accounting testsuite.

### DIFF
--- a/t/t1005-max-jobs-limits.t
+++ b/t/t1005-max-jobs-limits.t
@@ -94,14 +94,14 @@ test_expect_success 'a submitted job while at max-running-jobs limit will have a
 	jobid3=$(flux python ${SUBMIT_AS} 5011 sleep 60) &&
 	flux job wait-event -vt 60 \
 		--match-context=description="max-running-jobs-user-limit" \
-		$jobid3 dependency-add
+		${jobid3} dependency-add
 '
 
 test_expect_success 'a job transitioning to job.state.inactive should release a held job (if any)' '
-	flux cancel $jobid1 &&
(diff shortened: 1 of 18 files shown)


t/t1005-max-jobs-limits.t
```diff
@@ -161,8 +161,8 @@ test_expect_success 'make sure jobs are still running' '
 '
 
 test_expect_success 'cancel all remaining jobs' '
-	flux job cancel ${jobid1} &&
-	flux job cancel ${jobid2}
+	flux cancel ${jobid1} &&
+	flux cancel ${jobid2}
```
wihobbs, line 165: This is a _**super**_ nit but I noticed in flux we generally do the `${some_variable}` convention in shell scripts/sharness instead of `$some_variable`. Highlighting this one since it's different from the others (but do not feel the need to change it, that's way too pedantic.)

wihobbs (approved):

t/t1005-max-jobs-limits.t
cmoussa1: Good catch, thanks! I just force-pushed up a fix to go through the testsuite and wrap the variables that didn't have brackets with brackets. Hope it's okay that I tag it along in this PR

cmoussa1:
Thanks for reviewing @wihobbs! Going to set MWP here

PR #457, opened 2024-06-05
`inactive_cb`: remove unused iterator variables

cmoussa1:
#### Problem

There are a couple of unused iterator variables left over in the callback for `job.state.inactive` that were never removed when I was cleaning up and abstracting stuff out of the plugin and into an external class.

---

This PR removes these unused iterator variables.


cmoussa1:
thanks for reviewing @wihobbs! Setting MWP here

PR #458, opened 2024-06-06
plugin: initialize factor weights on plugin load

cmoussa1:
#### Problem

Mentioned as possible cleanup in #455, the weight of each factor is checked every time the priority of a job is calculated. It would be cleaner to put this check somewhere else so it does not need to be repeated for every job.

---

This PR creates defaults for the weights of each factor used in the priority calculation of a job and initializes each weight when the plugin is first loaded. In the callback for `conf.update`, it then checks to see if these weights have been configured; if not, it will just continue to use the defaults defined at the top of the file.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -38,6 +38,11 @@ extern "C" {
 // flux-accounting
 #define BANK_INFO_MISSING 999
 
+// default weights for each factor in the priority calculation for a job
+#define DEFAULT_FSHARE_WEIGHT 100000
+#define DEFAULT_QUEUE_WEIGHT 10000
+#define DEFAULT_AGE_WEIGHT 1000
+
 std::map<int, std::map<std::string, Association>> users;
 std::map<std::string, Queue> queues;
 std::map<int, std::string> users_def_bank;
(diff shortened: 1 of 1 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -201,8 +201,8 @@ static int conf_update_cb (flux_plugin_t *p,
     }
 
     // assign unpacked weights into priority_weights map
-    priority_weights["fairshare"] = fshare_weight;
-    priority_weights["queue"] = queue_weight;
+    if (fshare_weight != -1) priority_weights["fairshare"] = fshare_weight;
+    if (queue_weight != -1) priority_weights["queue"] = queue_weight;
```
grondo, line 205: I find single line conditionals a little hard to read, I'd suggest breaking these into two lines for readability.
I see we didn't comment on this earlier so this is just a suggestion not required.

grondo (approved):
This is great! Just one very minor comment.

src/plugins/mf_priority.cpp
cmoussa1: No problem - I think that's perfectly reasonable. I'll push up a fix to break these up into two lines. 👍

cmoussa1:
Thanks for reviewing @grondo! I'll set MWP here

PR #460, opened 2024-06-27
job archive interface: clean up a couple helper functions

cmoussa1:
#### Problem

There are a few functions in `job_archive_interface.py` that could use some improvements to their error handling and parameter and string construction. There's also some opportunity for some duplicate code cleanup.

---

This PR just makes some minor cleanup in a couple of functions in `job_archive_interface.py`. Notably, it reduces a couple of unnecessary checks in `get_job_records()` and makes use of f-strings for the query construction.

Some duplicate code is cleaned up in the helper functions for filtering jobs by default/secondary banks by reducing them into just one `filter_jobs_by_bank()` function.

Some more explicit error handling is added to `add_job_records()` to handle the case where the construction of a `ResourceSet` object fails while parsing job records.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -156,30 +156,17 @@ def check_jobspec(jobspec, bank):
     )
 
 
-# we are looking for jobs that were submitted under a secondary bank, so we'll
-# only add jobs that have the same bank name attribute in the jobspec
-def sec_bank_jobs(job_records, bank):
+# Filter job records based on the specified bank. For a default bank,
+# it includes jobs that either specify the default bank or do not
+# specify any bank at all.
+def filter_jobs_by_bank(job_records, bank, is_default_bank=False):
     jobs = []
     for job in job_records:
         jobspec = json.loads(job[7])
 
         if check_jobspec(jobspec, bank):
             jobs.append(job)
-
-    return jobs
-
-
-# we are looking for jobs that were submitted under a default bank, which has
-# two cases: 1) the user submitted a job while specifying their default bank,
-# or 2) the user submitted a job without specifying any bank at all
-def def_bank_jobs(job_records, default_bank):
-    jobs = []
-    for job in job_records:
-        jobspec = json.loads(job[7])
-
-        if check_jobspec(jobspec, default_bank):
-            jobs.append(job)
-        elif "bank" not in jobspec["attributes"]["system"]:
+        elif is_default_bank and "bank" not in jobspec["attributes"]["system"]:
             jobs.append(job)
 
     return jobs
@@ -225,15 +212,14 @@ def get_job_records(conn, bank, default_bank, **kwargs):
     if bank is None and default_bank is None:
         # special case for unit tests in test_job_archive_interface.py
         return add_job_records(result)
-    
-    if bank != default_bank:
-        jobs = sec_bank_jobs(result, bank)
-    else:
-        jobs = def_bank_jobs(result, default_bank)
 
-    job_records = add_job_records(jobs)
+    # find out if we are fetching jobs from a user's default bank or under
+    # one of their secondary banks; this will determine how we filter the
+    # job records we've found
+    is_default_bank = bank == default_bank
+    jobs = filter_jobs_by_bank(result, bank, is_default_bank)
 
-    return add_job_records(result)
+    return add_job_records(jobs)
 
 
 def output_job_records(conn, output_file, **kwargs):
--- a/src/bindings/python/fluxacct/accounting/job_archive_interface.py
+++ b/src/bindings/python/fluxacct/accounting/job_archive_interface.py
@@ -123,7 +123,13 @@ def add_job_records(rows):
     job_records = []
 
     for row in rows:
-        rset = ResourceSet(row[6])  # fetch R
+        try:
+            # attempt to create a ResourceSet from R
+            rset = ResourceSet(row[6])
+            nnodes = rset.nnodes
+        except (ValueError, TypeError):
+            # can't convert R to a ResourceSet object; skip it
+            continue
 
         job_record = JobRecord(
             row[0],  # userid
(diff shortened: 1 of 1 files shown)


src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -180,69 +180,49 @@ def def_bank_jobs(job_records, default_bank):
...
-    if bank != default_bank:
-        jobs = sec_bank_jobs(result, bank)
-    else:
-        jobs = def_bank_jobs(result, default_bank)
+        return add_job_records(result)
 
     job_records = add_job_records(jobs)
 
-    return job_records
+    return add_job_records(jobs)
```
jameshcorbett, line 221: It looks like you call `add_job_records(jobs)` twice in a row, lines 223 and 225?

src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -123,7 +123,13 @@ def add_job_records(rows):
...
 
     for row in rows:
-        rset = ResourceSet(row[6])  # fetch R
+        try:
+            # attempt to create a ResourceSet from R
+            rset = ResourceSet(row[6])
+            nnodes = rset.nnodes
+        except (ValueError, TypeError):
+            # can't convert R to a ResourceSet object; skip it
+            continue
```
jameshcorbett, line 132: Is skipping it a better option than using an empty `ResourceSet` or something? Not like I would know, just curious.

src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -226,7 +213,10 @@ def get_job_records(conn, bank, default_bank, **kwargs):
         # special case for unit tests in test_job_archive_interface.py
         return add_job_records(result)
 
-    job_records = add_job_records(jobs)
+    is_default_bank = bank == default_bank
+    jobs = filter_jobs_by_bank(
+        result, bank if not is_default_bank else default_bank, is_default_bank
```
jameshcorbett, line 218: maybe `result, bank if bank != default_bank else default_bank, is_default_bank`? I found this a bit hard to read

jameshcorbett: Actually how is this different than `filter_jobs_by_bank(result, bank)`? Unless there's some important difference between `bank == default_bank` and `bank is default_bank`?

src/bindings/python/fluxacct/accounting/job_archive_interface.py
jameshcorbett: Wait, this comment only applies to the first commit. The line is removed in the third. Maybe could you move that change down to the first commit? It's a bit puzzling otherwise

jameshcorbett (changes requested):
I am puzzled by some things 😖 but it otherwise looks good

src/bindings/python/fluxacct/accounting/job_archive_interface.py
cmoussa1: I think in flux-accounting's case, it probably would be best to skip, the reason being that it needs information from `ResourceSet` (specifically `nnodes`) to calculate job usage, so if it can't retrieve that information from the job, it should not be factored in.

This sort of check is kind of reinforcing the check made in the script that actually fetches job records from Flux using the `job-list` module, which already filters job that have actually run. So while I don't expect this kind of `Exception` to happen very often, I think it would be best to just skip adding the job if it does occur.

src/bindings/python/fluxacct/accounting/job_archive_interface.py
cmoussa1: I think I should just break these up into separate lines; it'll make it clearer - sorry about that. To the best of my knowledge there isn't an important difference between `bank == default_bank` and `bank is default_bank`.

src/bindings/python/fluxacct/accounting/job_archive_interface.py
cmoussa1: Sorry about the confusion. I just force-pushed up some changes to the first and third commit that I'm hoping makes this a bit clearer? Can you let me know if I am understanding you correctly?

src/bindings/python/fluxacct/accounting/job_archive_interface.py
jameshcorbett: I just realized I missed the third argument to `filter_jobs_by_bank()`, where you pass in `is_default_bank`. My bad. It makes more sense now.

src/bindings/python/fluxacct/accounting/job_archive_interface.py
jameshcorbett: Sounds good!

src/bindings/python/fluxacct/accounting/job_archive_interface.py
jameshcorbett: Yeah I think it's OK, it was never a big deal, I was just confused by some of logic at the end of the function you were cleaning up in the first commit. But regardless that function was overhauled in the third commit.

jameshcorbett (approved):
One nit, otherwise things are good! I was just confused 😖

src/bindings/python/fluxacct/accounting/job_archive_interface.py
```diff
@@ -225,15 +212,15 @@ def get_job_records(conn, bank, default_bank, **kwargs):
...
-        jobs = sec_bank_jobs(result, bank)
-    else:
-        jobs = def_bank_jobs(result, default_bank)
 
-    job_records = add_job_records(jobs)
+    # find out if we are fetching jobs from a user's default bank or under
+    # one of their secondary banks; this will determine how we filter the
+    # job records we've found
+    is_default_bank = bank == default_bank
+    bank = bank if bank != default_bank else default_bank
```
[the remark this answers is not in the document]
cmoussa1: 🤦 you're absolutely right. Sorry about that. I'll remove this line. Thanks for pointing that out!

cmoussa1:
Thanks for reviewing this @jameshcorbett and for putting up with this old code 😆 gonna set MWP here

cmoussa1:
@mergifyio refresh

cmoussa1:
@Mergifyio requeue

cmoussa1:
@Mergifyio dequeue

cmoussa1:
@Mergifyio requeue

cmoussa1:
Looks like re-pushing helped get this merged. Great suggestion @grondo!

PR #462, opened 2024-07-02
t: move python unit tests to `t/python/` directory

cmoussa1:
#### Problem

The unit tests for the Python bindings for flux-accounting live in a separate directory from the rest of the test files for the repository, which can sometimes be tedious to keep track of. They were also located there before flux-accounting had a `python/` directory in the test directory.

---

This PR moves the tests to `t/python/`. I've also given them slightly shorter names than what they used to be.


jameshcorbett (approved):
LGTM!

cmoussa1:
Thanks for reviewing so quickly @jameshcorbett! Setting MWP here :)

PR #463, opened 2024-07-09
python: clean job-archive interface code

cmoussa1:
#### Problem

The code for both viewing job records in the flux-accounting database and calculating/updating job usage values is in the same file, `job_archive_interface.py`. This causes the file to be very large and hard to read; it would be best to separate these two different features (viewing job records & calculating usage) into two, smaller files, as well as clean up some of the existing code for viewing job records.

---

This PR moves the code for viewing job records into a new file called `jobs_table_subcommands.py`. It also makes some slight improvements to the function names and adds some descriptions to a couple of those moved functions to better represent what they are doing.

The code in `job_archive_interface.py` (renamed to `job_usage_calculation.py`), the flux-accounting service, and unit tests are adjusted to account for this moved code.


src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
```diff
@@ -88,7 +88,9 @@ def write_records_to_file(job_records, output_file):
             )
 
 
-def fetch_job_records(job_records):
+# Convert the results of a query to the jobs table to a readable string
+# that can either be output to stdout or written to a file.
```
jameshcorbett, line 92: This is the style police 👮‍♂️ 🚔 you have violated [the law](https://peps.python.org/pep-0257/)

jameshcorbett (approved):
Looks good to me. It looks like a lot of the python code in this repo has function descriptions like
```
# description
def function():
    pass
```

or

```
def function():
    # description
    pass
```

but I think the usual Python style is

```
def function():
    """description"""
```

I noticed it in some of the code in this PR and then saw that it was pretty consistent across the repo.

src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
cmoussa1: 👐 i swear i'm innocent!

I will change the style of the function descriptions that I've created in this PR. If it's okay, I'll open a separate one to fix the descriptions in the rest of the repo.

cmoussa1:
Thanks for reviewing @jameshcorbett. I've pushed up a fix to the function descriptions I wrote in this PR. I'll open up an issue to step through the Python code in this repo and update the rest of the function descriptions. Setting MWP here

PR #465, opened 2024-07-10
`conf.update`: add missing bracket in format string

cmoussa1:
#### Problem

The format string for unpacking the various weights for priority factors in a config file is missing a closing bracket.

---

This PR adds it.

As an aside, I'm not sure why the testsuite never caught this before. I have a dedicated testing file for `conf_update_cb ()` (`t1034-mf-priority-config.t`) that **should** be testing different configuration files, but apparently it passed all of those. Is there maybe a different kind of test I could add that maybe comprehensively checks the format string? 🤷

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -191,7 +191,7 @@ static int conf_update_cb (flux_plugin_t *p,
     // unpack the various factors to be used in job priority calculation
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s?{s?{s?{s?i, s?i}}}",
+                                "{s?{s?{s?{s?i, s?i}}}}",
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM. Did we figure out why the testsuite did not catch this?

cmoussa1:
@grondo the test file that is supposed to test loading configuration files `t1034-mf-priority-config.t` has tests that involve creating a config TOML file and loading it, so I'm not sure why the testsuite did not catch it.

Is there a general format string test (or something similar) that I could add?

grondo:
Let me take a look at the test

cmoussa1:
I've added a commit to this PR that `return -1` after calling `flux_log_error ()` in the callback for `conf.update`. This probably explains why the testsuite never caught the error.

Since this is approved, I'll set MWP here (and create a new release)

PR #467, opened 2024-07-10
testsuite: fix on systems with flux-accounting already installed

grondo:
Problem: When flux-accounting is already installed on a system, the sharness tests pull in the installed version of flux-accounting Python modules instead of the test version. This occurs because sharness prepends the source tree Python path to PYTHONPATH, which is then subverted by flux-core's `py-runner.py`, which prepends the system path in front of that before executing Python subcommands like the accounting service etc.

Use FLUX_PYTHONPATH_PREPEND instead of PYTHONPATH, which is preserved by the py-runner.py wrapper.


cmoussa1 (approved):
LGTM! This would probably be good to tag into `v0.35.0` as well. Thanks!

grondo:
I'll set MWP. thanks.

PR #469, opened 2024-07-16
plugin: add estimation of cores-per-node count on system during initialization

cmoussa1:
#### Problem

The priority plugin does not know about basic system information it will need in order to enforce a max-cores limit per association, such as the number of cores on a node.

---

This PR adds an estimation of a cores-per-node count estimate during the initialization of the priority plugin by fetching `resource.R` from the KVS. It stores this estimate in a global variable in the plugin. The plan is to use this count when calculating the number of cores used by a job when only `nnodes` are specified. I think that this count might not be exactly right for systems where the core count per-node might be different throughout the system, but I figure this could at least be a good estimate and a start for tracking and enforcing resource limit per-association across all of their running jobs (see conversation in flux-framework/flux-core#6091).

This count is also added to the list of information returned in the callback for `plugin.query`.


src/plugins/mf_priority.cpp
```diff
@@ -1116,6 +1144,42 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
...
+
+    if (flux_kvs_lookup_get_unpack (f,
+                                    "{s{s[{s{s:s}}]}}",
+                                    "execution",
+                                      "R_lite",
+                                        "children",
+                                          "core", &core) < 0) {
+        flux_log_error (h, "flux_kvs_lookup_unpack");
+        return -1;
+    }
```
grondo, line 1169: Normally I would request that this synchronous get be replaced with an asynchronous `flux_future_then(3)` and the parsing of R handled in a callback. However, it seems like the result is required for validation of jobs, and you probably don't want to let some jobs through erroneously while waiting for `resource.R`, so perhaps this is actually the right solution. The job manager will block here while waiting for R when `mf_priority.so` is loaded, but it _should_ be a very short time, and while this is occurring job management will also pause. Most of the time this will occur during job manager module load, which has other synchronous work anyway.

I'd at least suggest a comment here describing why a synchronous get is used in this case.

@garlick: any other thoughts?

src/plugins/mf_priority.cpp
```diff
@@ -1116,6 +1144,42 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
...
+        flux_log_error (h, "flux_kvs_lookup");
+        return -1;
+    }
+
+    if (flux_kvs_lookup_get_unpack (f,
+                                    "{s{s[{s{s:s}}]}}",
+                                    "execution",
+                                      "R_lite",
+                                        "children",
+                                          "core", &core) < 0) {
```
grondo, line 1166: R may have multiple entries in the R_lite array. To handle possible heterogeneity,  you could iterate each entry and use the maximum number of cores found. 

Probably ok if this is just a first cut, though. If so, I'd put a comment stating that "equal number of cores on all nodes in R is assumed, so we only look at the first entry" or simimlar.

src/plugins/mf_priority.cpp
```diff
@@ -1116,6 +1144,42 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
...
+
+    if (core == NULL) {
+        flux_log_error (h,
+                        "mf_priority: could not get system "
+                        "cores-per-node information");
+        return -1;
+    }
+
+    // calculate number of cores-per-node on system
+    ncores_per_node = calculate_range (core);
```
grondo, line 1179: Instead of parsing the idset with custom code here, you can use libflux-idset.so exported by flux-core.
See [idset_decode(3)](https://flux-framework.readthedocs.io/projects/flux-core/en/stable/man3/idset_decode.html), then just use [idset_count(3)](https://flux-framework.readthedocs.io/projects/flux-core/en/stable/man3/idset_create.html) to get the number of cores.

t/t1036-mf-priority-resource-counting.t
```diff
@@ -0,0 +1,33 @@
...
+
+test_expect_success 'check that mf_priority plugin is loaded' '
+	flux jobtap list | grep mf_priority
+'
+
+test_expect_success 'check that cores-per-node count is correct' '
+	flux jobtap query mf_priority.so > query.json &&
+	test_debug "jq -S . <query.json" &&
+	cat query.json &&
+	jq -e ".ncores_per_node == 2" <query.json
```
grondo, line 30: Will this check work on all systems? If you need to check the actual cores/node you could use `nproc` or possibly `flux resource R --include=0 | flux R decode --count=core` (sorry if I missed someplace where you've guaranteed 2 cores per node)

grondo (commented):
Some early comments on a first pass-through

src/plugins/mf_priority.cpp
garlick: The FLUX_KVS_WATCH flag should be dropped.  That's only used when you want to receive a response for every change to the key, and here the future is being destroyed after the first response.

IRL, we only load `mf_priority.so` in the system instance so `resource.R` should exist already and the KVS lookup should be fast.   FLUX_KVS_WAITCREATE is probably is needed in test where the plugin is loaded in a test instance and  `resource.R` is dynamically discovered though.

Yes a comment would be good since synchronous activities always raise eyebrows.

cmoussa1:
Thank you both for the first pass! I've force-pushed up some changes to drop `FLUX_KVS_WATCH`, use libflux-idset, and change how the test is structured to instead use `flux resource R --include=0 | flux R decode --count=core` (this check makes a lot more sense than what I had).

I've also added some comments in code where suggested.

PR #471, opened 2024-07-29
src: remove `flux_account_shares.cpp` in favor of just using `-t` option with `view-bank`

cmoussa1:
#### Problem

flux-accounting has two identical ways to print out the hierarchy of its accounting database: the `flux account-shares command` (a
`.cpp` file) and an optional argument passed into the `view-bank` command. It would be less confusing to have two different options that do the same thing, especially since the `.cpp` file requires a dash `-` in between "account" and "shares", which doesn't match the rest of the flux-accounting commands.

---

This PR removes `flux_account_shares.cpp` from flux-accounting and converts the various calls in its testsuite to just use the `-t` option of the `view-bank` command. It also converts the expected files used for these tests to match the higher precision output from the `view-bank` command, particularly the job-usage values. I've added a `-P`/`--parsable` option to `view-bank` to print out hierarchies in a pipe-delimited (`|`) format.

I've also broken up `t1000-print-hierarchy.t` into multiple, smaller tests.


t/Makefile.am
```diff
@@ -1,7 +1,6 @@
 # This list is included in both TESTS and dist_check_SCRIPTS.
 TESTSCRIPTS = \
 	t0000-sharness.t \
-	t1000-print-hierarchy.t \
```
jameshcorbett, line 4: It looks like this file isn't deleted, even though it's removed from the testsuite?

src/bindings/python/fluxacct/accounting/bank_subcommands.py
```diff
@@ -176,6 +176,56 @@ def print_hierarchy(cur, bank, hierarchy_str, indent=""):
...
+                    + bank
+                    + "|"
+                    + str(user[0])
+                    + "|"
+                    + str(user[1])
+                    + "|"
+                    + str(user[2])
+                    + "|"
+                    + str(user[3])
+                    + "\n"
```
jameshcorbett, line 207: did you consider `f"{indent} {bank}|{user[0]}|{user[1]}|{user[2]}|{user[3]}\n"`?

jameshcorbett (approved):
Looks good, two small comments. 

Also, depending on what your use-case is, have you considered an option to output the hierarchy info as JSON or to allow users to specify what information they want printed out and in what format? Like `flux jobs` does:

```
flux jobs ƒ2WFJyd8B -o"{id}|{username}"
JOBID|USER
ƒ2WFJyd8B|corbett8
```

t/Makefile.am
cmoussa1: Oh yeah, I forgot to delete this file. Thanks for catching that

src/bindings/python/fluxacct/accounting/bank_subcommands.py
cmoussa1: I didn't even consider f-strings 🤦 thank you for pointing that out! I'm gonna switch to using these instead; much cleaner

cmoussa1:
Thanks for reviewing @jameshcorbett! I've force-pushed up some fixes/changes based on your suggestions. 

> Also, depending on what your use-case is, have you considered an option to output the hierarchy info as JSON or to allow users to specify what information they want printed out and in what format?

This is a good suggestion. `view-user` currently has support to output its info in `JSON` format but it is not available for bank information yet. I'll look into adding some support for this in the near future.

Setting MWP here shortly

PR #473, opened 2024-08-01
`bank_table`: add a new `list-banks` command

### LINKED ISSUE #472: `flux view-account --list-all-banks-in-json-please`
Steph was asking (for myLC/Lorenz) if there's a command in accounting that can be used to list all of the banks (excluding the expired ones) on a host. I assume that if there isn't, this is probably pretty low-hanging fruit...?

By my understanding she has a workaround already that yields a tree structure with all of the banks under root, but that requires a lot of string parsing.

cmoussa1:
#### Problem

There is currently no way to list the current banks defined in the `bank_table` in the flux-accounting database, especially in a
convenient format like JSON.

---

This PR adds a new command to the flux-accounting command suite called `list-banks`. When called, this command will create a JSON object containing all of the active banks in the `bank_table`. Below is example output of `flux account list-banks`:

```json
[
  {
    "bank_id": 1,
    "bank": "root",
    "parent_bank": "",
    "shares": 1,
    "job_usage": 0.0
  },
  {
    "bank_id": 2,
    "bank": "A",
    "parent_bank": "root",
    "shares": 1,
    "job_usage": 0.0
  },
  {
    "bank_id": 3,
    "bank": "B",
    "parent_bank": "root",
    "shares": 1,
    "job_usage": 0.0
  }
]
```

Passing `--inactive` will include *all* banks, even those which have been disabled. The fields output in the JSON object can be customized with a list of fields to be included with `--fields`. Below is example output of `flux account list-banks --fields=bank_id,bank`:

```json
[
  {
    "bank_id": 1,
    "bank": "root"
  },
  {
    "bank_id": 2,
    "bank": "A"
  },
  {
    "bank_id": 3,
    "bank": "B"
  }
]
```

A couple of basic tests calling these different optional arguments are also added to the test suite. I've included them with the other sharness tests for various bank commands.

Fixes #472 

---

@wihobbs, does Steph have a GitHub account? Perhaps she should take a high-level look at this PR to see if it matches her needs. :-)

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: LGPL-3.0
 ###############################################################
 import sqlite3
+import json
 
 from fluxacct.accounting import user_subcommands as u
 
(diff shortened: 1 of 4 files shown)


wihobbs (approved):
Thanks for doing this! In offline discussion with Steph it seems like this would suffice her use case.

cmoussa1:
Thanks! Setting MWP here

PR #475, opened 2024-08-05
`fetch-job-records`: add integrity check for records

cmoussa1:
#### Problem

There is a case where a job can have an `R` but not have entered `RUN` state, and thus, does not receive a `t_run` time. The `flux account-fetch-job-records` script does not verify that the timestamp key-value pairs exist before attempting to insert them into the `jobs` table in the flux-accounting DB, which can result in a `KeyError`.

---

This PR adds two integrity checks for job records fetched by this script:

- When initially fetching jobs that checks that each key-value pair exists before adding it to a list of fetched job records, if a key-value pair is not valid, skip adding the job.

- When attempting to insert a job record into the `jobs` table, wrap the `INSERT` in a try-except block in case a `KeyError` is raised. If so, just skip adding the job.

### DIFF
--- a/src/cmd/flux-account-fetch-job-records.py
+++ b/src/cmd/flux-account-fetch-job-records.py
@@ -90,6 +90,24 @@ def fetch_new_jobs(last_timestamp=0.0):
         if data["jobspec"] is not None:
             single_record["jobspec"] = data["jobspec"]
 
+        required_keys = [
+            "userid",
+            "t_submit",
+            "t_run",
+            "t_inactive",
+            "ranks",
+            "id",
+            "R",
+            "jobspec",
+        ]
+        if not all(
+            key in single_record and single_record.get(key) is not None
+            for key in required_keys
+        ):
+            # job does not have all required fields to be added to jobs table
+            # in DB; skip this entry
+            continue
+
         # append job to job_records list
         job_records.append(single_record)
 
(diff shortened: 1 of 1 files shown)


src/cmd/flux-account-fetch-job-records.py
```diff
@@ -90,6 +90,24 @@ def fetch_new_jobs(last_timestamp=0.0):
...
+            "t_submit",
+            "t_run",
+            "t_inactive",
+            "ranks",
+            "id",
+            "R",
+            "jobspec",
+        ]
+        if not all(
+            key in single_record and single_record[key] is not None
```
grondo, line 104: Just a suggestion, but I think you check if `key in single_record and single_record[key] is not None` in one call with `single_record.get(key)` (returns falsy if key doesn't exist or is `None`)

grondo (approved):
LGTM! Just one suggestion inline.

src/cmd/flux-account-fetch-job-records.py
jameshcorbett: I hope I'm not misunderstanding you but I think the check would have to be `single_record.get(key) is not None`: https://docs.python.org/3/library/stdtypes.html#dict.get

jameshcorbett (approved):
👍

src/cmd/flux-account-fetch-job-records.py
jameshcorbett: Oh wait, maybe you're right, assuming that all the values are truthy if they aren't `None`...

grondo: > I hope I'm not misunderstanding you but I think the check would have to be single_record.get(key) is not None: https://docs.python.org/3/library/stdtypes.html#dict.get

Very good point @jameshcorbett - I should keep my mouth shut about Python stuff.

jameshcorbett: booo everyone is wrong sometimes

cmoussa1:
Thanks @jameshcorbett and @grondo for the catch and for approving! I'll set MWP here

PR #477, opened 2024-08-06
plugin: add instance owner info to plugin

### LINKED ISSUE #476: the flux user cannot run jobs when flux-accounting is installed and flux is not in the database
Problem: It seems like the `flux`(instance owner) user should always be allowed to run, to support "system jobs".  (@jameshcorbett has a use case for this now).  But flux-accounting prevents this unless the flux user is in the database.

I was thinking maybe flux-core could do something to ensure this always works but I'm not sure how we would do that generally - it might need to go in flux-accounting.  But anyway,  maybe others have some ideas.

cmoussa1:
#### Problem

The instance owner cannot run jobs when the priority plugin is loaded because they are not a part of the flux-accounting database. flux-accounting should be able to handle this automatically so that administrators do not have to manually add instance owner information.

---

This PR sends instance owner information to the priority plugin along with the rest of the flux-accounting DB information during an accounting update. It leverages Flux's `handle.attr_get("security.owner")` to get the uid and corresponding username of the instance owner. Once retrieved, it is sent and stored by the plugin, which will allow the instance owner to submit jobs with the plugin loaded. I set the limits values very high for the instance owner, but just gave it a default fair-share value of 0.5. I am definitely open to making this value higher, however.

I had to adjust a few of the existing sharness tests that created users with conflicting uid's as the instance owner in the testsuite, so the second commit in this PR is just some cleanup and adjusting of some of those tests to use a different uid. I also refined a couple of the tests that use `jq` to search for key-value pairs by first calling `select()` with a uid instead of manually indexing the JSON object.

Finally, I added a short set of tests that loads the plugin and and submits a job as the instance owner and makes sure it receives a priority.

Fixes #476


t/t1019-mf-priority-info-fetch.t
```diff
@@ -62,8 +61,8 @@ test_expect_success 'add some projects to the DB' '
 '
 
 test_expect_success 'add a user with two different banks to the DB' '
-	flux account add-user --username=user1001 --userid=1001 --bank=account1 --max-running-jobs=2 &&
-	flux account add-user --username=user1001 --userid=1001 --bank=account2
+	flux account add-user --username=user5001 --userid=5001 --bank=account1 --max-running-jobs=2 &&
+	flux account add-user --username=user5001 --userid=5001 --bank=account2
```
grondo, line 65: Is there something in the testsuite that ensures the user running it is not 5001 (or for that matter is 1001). Or is the testsuite only expected to be run in certain environments?

One idea would be to add a check for the current uid in `sharness.d/flux-accounting.sh` and set up some environment variables with userids instead of explicitly using 1001 and 5001 (in case those conflict with system uids)

This could be done in a future cleanup PR if this is working for now.

grondo (approved):
LGTM!  I did wonder if there was some way the  instance owner data could be sent only once, but I'm not sure how that would work.

I guess an alternative would be to have the mf_priority plugin manually set up the instance owner associations internally at startup, but the benefit of this approach is that it is simple...

t/t1019-mf-priority-info-fetch.t
cmoussa1: I guess there isn't anything explicit that is checking it (I just checked the logs from the tests that failed via GitHub actions), but that's a good idea to set up some environment variables so that I don't have to define them every single time I create a new test. I'll open an issue on this so I can get this cleaned up in the future.

cmoussa1:
Thanks for reviewing @grondo - I'll set MWP here so I can get this landed in time for some testing

PR #479, opened 2024-08-06
`view-user`: add a new `--list-banks` optional argument

### LINKED ISSUE #478: `view-user`: add an option to concisely list banks a user belongs to
There currently isn't an easy way to see a list of all the banks a user might belong to without having to parse through the rest of the accounting information. I should add an optional argument to `view-user` that simply lists the banks a user belongs to.

cmoussa1:
#### Problem

Mentioned in #478, there is no concise way to just view a list of banks that a user belongs to without having to parse through the rest of the accounting information associated with it.

---

This PR adds a new optional argument to `view-user` called `--list-banks`, which just outputs a newline-delimited list of banks that a user belongs to.

```console
$ flux account view-user testuser --list-banks
bankA
bankB
bankC
```

A basic test file is also added which calls this optional argument. For now, this is just built on top of #477, but I can always make it separate in case this gets reviewed first. :-)

Fixes #478

### DIFF
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -306,3 +306,18 @@ class AssociationFormatter(AccountingFormatter):
         super().__init__(
             cursor, error_msg=f"user {self.username} not found in association_table"
         )
+
+    def list_banks(self):
+        """
+        Return all of the banks that the user belongs to with each bank
+        on its own line.
+        """
+        self.cursor.execute(
+            "SELECT bank FROM association_table WHERE username=?", (self.username,)
+        )
+        result = self.cursor.fetchall()
+        banks = ""
+        for bank in result:
+            banks += f"{str(bank[0])}\n"
+
+        return banks
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -233,7 +233,7 @@ def clear_projects(conn, username, bank=None):
 #                   Subcommand Functions                      #
 #                                                             #
 ###############################################################
-def view_user(conn, user, parsable=False, cols=None):
+def view_user(conn, user, parsable=False, cols=None, list_banks=False):
     # use all column names if none are passed in
     cols = cols or fluxacct.accounting.ASSOCIATION_TABLE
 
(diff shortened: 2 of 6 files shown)


t/t1040-issue478.t
```diff
@@ -0,0 +1,50 @@
...
+
+test_expect_success 'add a user' '
+    flux account add-user --username=testuser --bank=bankA &&
+    flux account add-user --username=testuser --bank=bankB &&
+    flux account add-user --username=testuser --bank=bankC
+'
+
+test_expect_success 'call view-banks --list-banks' '
+	flux account view-user testuser --list-banks > list_banks.out &&
+    grep -E "bankA|bankB|bankC" list_banks.out
```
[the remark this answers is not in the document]
cmoussa1: Sorry that I didn't catch this earlier! I usually double check this but it slipped through.

I just force-pushed up a change to also just separate the `grep` check into three different checks for all three banks instead of combining all of them on just one line (that was my misunderstanding of the `-E` option, sorry!). Can you let me know if you are still seeing that weird output behavior when calling `view-user --list-banks`? I wasn't able to reproduce that locally (yet) and I tried calling `view-user` with every optional argument...

cmoussa1:
I force-pushed to catch up after #526, but I'm thinking now that flux-accounting has its own formatter class, perhaps I should hold off on getting this PR landed until I restructure `view-user` to use the new class. Then adding this feature probably becomes cleaner

src/bindings/python/fluxacct/accounting/formatter.py
```diff
@@ -306,3 +306,21 @@ def __init__(self, cursor, username):
...
             cursor, error_msg=f"user {self.username} not found in association_table"
         )
+
+    def list_banks(self):
+        """
+        Return all of the banks that the user belongs to with each bank
+        on its own line.
+
+        Args:
+            username: the name of the user.
```
jameshcorbett, line 316: Docstring misrepresents arguments to this function?

jameshcorbett (approved):
LGTM one small docstring comment

src/bindings/python/fluxacct/accounting/formatter.py
cmoussa1: Thank you for catching this!! I've updated the docstring.

cmoussa1:

PR #482, opened 2024-08-07
`Makefile`: remove left over compile instructions for `flux_account_shares`

### LINKED ISSUE #481: src/Makefile.am: remove `flux_account_shares` targets
The targets for `flux_account_shares` are still lingering in the top-level `Makefile.am`. These should be removed

cmoussa1:
#### Problem

`flux_account_shares.cpp` was removed in #471, but there is still a compile definition for it in the top-level `Makefile.am` that is left over.

---

This PR just removes the left over compilation instructions.

Fixes #481


jameshcorbett (approved):
💯

cmoussa1:
🙏 setting MWP here

PR #484, opened 2024-08-14
configure: add `jansson` as a dependency check

### LINKED ISSUE #364: build: dependencies may need adjustment
Problem: it was noted in #363 that flux-accounting's `configure.ac` might be missing some dependency checks.

Actually, the only one that I'm pretty sure is missing is jansson.
```diff
diff --git a/configure.ac b/configure.ac
index 81ee424..8ddbe18 100644
--- a/configure.ac
+++ b/configure.ac
@@ -33,6 +33,7 @@ AC_PROG_MAKE_SET
 AM_PROG_CC_C_O
 AX_CODE_COVERAGE
 PKG_CHECK_MODULES([SQLITE], [sqlite3], [], [])
+PKG_CHECK_MODULES([JANSSON], [jansson >= 2.10], [], [])
 
 if test "$GCC" = yes; then
   WARNING_CFLAGS="-Wall -Werror -Werror=missing-field-initializers -Wno-error=deprecated-declarations"
```

I didn't look too hard though...  If any non-builtin python packages are required, we should add those.  See flux-core's [cofnigure.ac](https://github.com/flux-framework/flux-core/blob/master/configure.ac#L246) for examples.

cmoussa1:
#### Problem

As noted in #364, flux-accounting does not check for `jansson` when running `./configure` even though it uses it in the priority plugin.

---

This PR adds a check for `jansson` in `configure.ac`.

Fixes #364

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,6 +34,7 @@ AM_PROG_CC_C_O
 AX_VALGRIND_H
 AX_CODE_COVERAGE
 PKG_CHECK_MODULES([SQLITE], [sqlite3], [], [])
+PKG_CHECK_MODULES([JANSSON], [jansson >= 2.10], [], [])
 
 if test "$GCC" = yes; then
   WARNING_CFLAGS="-Wall -Werror -Werror=missing-field-initializers -Wno-error=deprecated-declarations"


grondo (approved):
LGTM!

cmoussa1:
Thanks! Setting MWP here

PR #485, opened 2024-08-23
doc: add example error message when creating DB after starting systemd service

cmoussa1:
#### Problem

The flux-accounting service depends on the creation of the flux-accounting DB before starting since it needs to establish a connection to the DB in order to run `flux account` commands. If this is done out of order, confusing behavior can result from running the commands, such as an `attempt to write to a readonly database` error.

---

This PR just adds a note to the flux-accounting guide about making sure the flux-accounting DB has been created before starting the flux-accounting service and gives an example of the error message that can potentially occur.

### DIFF
--- a/doc/guide/accounting-guide.rst
+++ b/doc/guide/accounting-guide.rst
@@ -77,7 +77,9 @@ the following prerequisites must be met:
 
 1. A flux-accounting database has been created with ``flux account create-db``.
 The service establishes a connection with the database in order to read from
-and write to it.
+and write to it. If the service has been started before the creation of the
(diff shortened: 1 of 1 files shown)


wihobbs (approved):
Note that you can use `` instead of "" in commit messages if you backslash out of them or open the editor. However that's IMO personal preference. 

Thanks for getting this plumbed correctly (and on short notice) on fluke yesterday! :)

cmoussa1:
TIL! Thanks for the commit message tip. Glad we got it figured out. Setting MWP here

PR #486, opened 2024-08-27
`cmd`: add `export-db` as a flux account command

cmoussa1:
#### Problem

The `export-db` command is located in a separate file in the `src/cmd/` directory and thus requires a different method of calling it than the other flux-accounting commands, i.e you need to write `flux account-export-db` instead of `flux account export-db`.

---

This PR moves this file to the directory containing the other Python bindings for flux-accounting and adds it as a regular command like the other bindings so that you can just call it like `flux account export-db`.

It removes the methods and `--path` command line argument that deal with establishing a connection to the flux-accounting DB since the systemd service will handle establishing that connection.

Finally, it adjusts the calls in `t1016-export-db.t` and top-level README for `export-db` to account for the change.


jameshcorbett (approved):
Looks reasonable!

cmoussa1:
Thanks! Setting MWP here

PR #487, opened 2024-08-29
`cmd`: add `pop-db` as a flux account command

cmoussa1:
#### Problem

The `pop-db` command is located in a separate file in the `src/cmd/` directory and thus requires a different method of calling it than the other flux-accounting commands, i.e you need to write `flux account-pop-db` instead of `flux account pop-db`.

---

This PR moves this file to the directory containing the other Python bindings for flux-accounting and adds it as a regular command like the other bindings so that you can just call it like `flux account pop-db`.

It removes the methods and `--path` command line argument that deal with establishing a connection to the flux-accounting DB since the systemd service will handle establishing that connection.

I've also added both `pop-db` and `export-db` as commands that are tested in `t1026-flux-account-perms.t` to ensure that they require admin privileges to be run.

Finally, it adjusts the calls in `t1009-pop-db.t` and `t1016-export-db.t` and top-level README for `pop-db` to account for the change.


jameshcorbett (approved):
LGTM!

cmoussa1:
thanks! setting MWP here

PR #489, opened 2024-09-16
JobRecord: remove `username` from `__init__()` of JobRecord object

cmoussa1:
#### Problem

The `username` field is automatically populated when `userid` is passed into the construction of a `JobRecord` object, but it is listed as an attribute during initialization.

---

This PR removes `username` from the list of attributes in `__init__()`. It also explicitly labels which fields are being assigned to which member of the `JobRecord` object instead of labeling them via comments.

I've also snuck in a spelling correction commit at the end of this PR.


wihobbs (approved):
Much clearer, thanks!

cmoussa1:
Thank you! Setting MWP here

Issue #402, opened 2023-12-01
per-queue user limits


ryanday36:
we've had a request for more limits that we can set on specific queues. We can currently set limits on how many resources a specific job can use (max nodes, min nodes, wall time). We'd also like to set limits on how many resources a specific user's running jobs can use in a given queue. Specifically, we'd like to be able set the following for a queue:

max running jobs per user
max nodes per user (across all running jobs, similar to #349)

I've also been giving some thought to whether some sort of 'max committed node-hours per user' (total nodes*requested walltime of all of a users running jobs) would be useful. I'm not so sure about that though.

```[tasklist]
### Tasks
- [ ] support enforcing a max running jobs limit per association in a specific queue
- [ ] support enforcing a max node count limit across an association's running jobs in a specific queue
```

cmoussa1:
> max nodes per user (across all running jobs, similar to https://github.com/flux-framework/flux-accounting/issues/349)

I could also be wrong here, but I _believe_ enforcing a max nodes per-user limit, both per-queue and in general, would require some coordination between flux-accounting and other Flux components, as nicely summarized by @grondo in a comment in #349: 

> Thanks @cmoussa1. Re-reading above it seems like the current summary is:
> - as  a first cut, implement holistic limits in flux-accounting instead of a max-nodes limit. I.e. impose a max-nodes+max-cores limit for users across all jobs
> - as a prerequisite, the accounting plugin will need access to the actual resource counts assigned jobs, so that nnodes for core-only requests and ncores for nodes-only requests can be accounted. Therefore flux-framework/flux-core#3851 should be solved first.
>
> Edit: Forgot to mention that more design work is needed on how to loop the scheduler into these limits so that a real max-nodes limit could be imposed.
>
> Also, there is probably a race condition we should consider if the flux-accounting jobtap plugin is enforcing these liimits. E.g. a max-nodes limit could be exceeded or hit during one job's `job.stat.run` callback, while at the same time the scheduler is allocating more nodes to that user before the plugin has a chance to hold all pending jobs for the user.

It has been a little while since we've discussed this, though, so perhaps we are better suited to tackle this now than before.

ryanday36:
[...]

PR #491, opened 2024-09-27
plugin: add enforcement of max running jobs limit for a queue per-association

cmoussa1:
#### Problem

The priority plugin does not support enforcement of a max running jobs limit in a queue for an association, i.e the only max running jobs limit currently enforced for an association is one across _all_ of their running jobs, regardless of queue.

---

This PR looks to add enforcement of a new limit to the priority plugin: max running jobs in a queue per-association. To achieve this, new members are added to the `Queue` class and the `Association` class:

##### Queue

* `max_running_jobs`: the max number of running jobs an association can have in a queue

##### Association

* `queue_usage`: a hash-map storing the name of the queue and the number of running jobs the association has in that queue
* `queue_held_jobs`: a hash-map storing the name of the queue and a list of any held jobs the association has in that queue

If a queue is specified on submission, the priority plugin will keep track of these running jobs per-association by incrementing the association's running jobs counter for that queue when a job enters `job.state.run` and decrementing when it enters `job.state.inactive`.

If an association has the max number of running jobs in a queue and submits another job to that queue, a *queue* running jobs dependency is added in `job.state.depend`. The job ID will be stored in the `Association` object and held until a currently running job in that queue transitions to `job.state.inactive`.

In `job.state.inactive`, if a queue is specified for a job, a check is performed to see if the association has any other jobs waiting to be run in that queue. If one is specified (and the association is now under the max running jobs limit), the dependency is removed from the first held job in that queue and it can proceed to run.

I've added some basic tests that simulate an association submitting the max number of jobs to a queue and having a dependency added to another submitted job. But, while at this limit, the association can successfully submit jobs to other queues and have them run. Once a currently running job in the queue where a job is held completes, the held job can transition to run.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -922,12 +922,15 @@ static int run_cb (flux_plugin_t *p,
     int userid;
     Association *b;
     json_t *jobspec = NULL;
+    char *queue = NULL;
 
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:o}",
-                                "jobspec", &jobspec) < 0) {
+                                "{s:o, s{s{s{s?s}}}}",
+                                "jobspec", &jobspec,
+                                "jobspec", "attributes", "system",
+                                "queue", &queue) < 0) {
         flux_log (h,
                   LOG_ERR,
                   "flux_plugin_arg_unpack: %s",
@@ -948,7 +951,12 @@ static int run_cb (flux_plugin_t *p,
         return -1;
     }
 
-    // increment the user's current running jobs and resources counts
+    if (queue != NULL)
+        // a queue was passed-in; increment counter of the number of
+        // queue-specific running jobs for this association
+        b->queue_usage[std::string (queue)]++;
+
+    // increment the user's current running jobs count
     b->cur_run_jobs++;
     if (jobspec == NULL) {
         flux_jobtap_raise_exception (p,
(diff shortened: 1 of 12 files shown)


cmoussa1:
OK, I've added a couple more tests here and I think this might be ready for some initial review. I don't expect this to make this month's release, so no rush here. Let me know if the second and third commit could perhaps be squashed? I tried to keep them separate so they'd be easier to review, but can squash if we feel like it'd make more sense.

cmoussa1:
@ryanday36 would you like to give this PR just a high-level look and confirm that this solves the case you brought up in #402? Let me know if I understood the requirements correctly.

cmoussa1:
@ryanday36 this has now been rebased if you want to give this PR a look :-)

ryanday36:
Thanks @cmoussa1! Sorry I never got back to you on this in October. It sounds like it's doing what we want.

src/plugins/mf_priority.cpp
```diff
@@ -895,6 +898,38 @@ static int depend_cb (flux_plugin_t *p,
...
+            // might not be defined in the flux-accounting DB, so just set
+            // it to a large number
+            queue_max_run_jobs = std::numeric_limits<int>::max ();
+        }
+
+        // look up the association's current number of running jobs;
+        // if queue cannot be found, an entry in the Association object will be
+        // initialized with a current running jobs count of 0
+        int assoc_cur_run_jobs = b->queue_usage[queue];
+        if (assoc_cur_run_jobs == queue_max_run_jobs) {
```
jameshcorbett, line 917: Could be `>=` here right? Shouldn't hurt? or help I guess

jameshcorbett (approved):
NICE

src/plugins/mf_priority.cpp
cmoussa1: Oh yeah, good suggestion, you're right. Shouldn't hurt!

PR #492, opened 2024-09-27
`job.state.inactive`: add `return -1` to exception

cmoussa1:
#### Problem

The callback for `job.state.inactive` does not `return -1` in the event of an exception being raised due to failing to remove a dependency from a held job for an association.

---

Add a `return -1` call after `flux_jobtap_raise_exception ()` in the callback for `job.state.inactive`.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1138,8 +1138,11 @@ static int inactive_cb (flux_plugin_t *p,
         if (flux_jobtap_dependency_remove (p,
                                            jobid,
                                            "max-running-jobs-user-limit") < 0)
+        {
             flux_jobtap_raise_exception (p, jobid, "mf_priority",
                                          0, "failed to remove job dependency");
+            return -1;
+        }
 
         b->held_jobs.erase (b->held_jobs.begin ());
     }


src/plugins/mf_priority.cpp
```diff
@@ -1135,11 +1135,14 @@ static int inactive_cb (flux_plugin_t *p,
     if ((b->held_jobs.size () > 0) && (b->cur_run_jobs < b->max_run_jobs)) {
         long int jobid = b->held_jobs.front ();
 
-        if (flux_jobtap_dependency_remove (p,
-                                           jobid,
-                                           "max-running-jobs-user-limit") < 0)
+        if (flux_jobtap_dependency_remove (
+                                          p,
```
jameshcorbett, line 1139: Just a formatting nit, I think 

```
        if (flux_jobtap_dependency_remove (p,
```

would be the format more in line with the rest of the project

jameshcorbett (approved):
Looks good one small nit!

src/plugins/mf_priority.cpp
cmoussa1: Thanks for catching this! Force pushed up a fix for formatting. Will set MWP

PR #494, opened 2024-10-04
`view-user`: make "parsable" spelling consistent

cmoussa1:
#### Problem

There are multiple `--parsable` optional arguments throughout the flux-accounting command suite, but the spelling for it in the view-user command is not consistent and is spelled `--parseable`. Technically, both spellings are correct, but they should be made consistent throughout the command suite.

---

This PR just changes the spelling of the `--parseable` optional argument for `view-user` to `--parsable`.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -129,10 +129,10 @@ def create_json_object(conn, user):
     return user_info_json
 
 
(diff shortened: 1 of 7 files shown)


wihobbs (approved):
Indeed you are correct: https://www.oed.com/dictionary/parsable_adj?tab=factsheet#32035960

cmoussa1:
thanks @wihobbs - setting MWP here

PR #495, opened 2024-10-07
projects: fix unit tests for project subcommands, `--projects` reset capability

cmoussa1:
#### Problem 1

For some reason, the unit tests for the project subcommands was just a duplicate of the functions for the subcommands themselves instead of the actual unit tests, and I never noticed. I must have botched something when I moved them from the `accounting` directory to the `t/` directory.

#### Problem 2

Passing `-1` for the `--projects` optional argument in the `edit-user` command does not work because the `edit_user()` function thinks that `-1` is a project name, so it doesn't actually reset the projects list to just `*`.

---

This PR does a couple of minor things. It adds the missing unit tests to `t1005_project_cmds.py`. And it adds a check for `-1` in the projects portion of `edit_user()` to reset the `projects` field to just `*`. I've added both a unit test and sharness test to make sure this works as expected.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -280,6 +280,26 @@ def clear_queues(conn, username, bank=None):
     return 0
 
 
+def clear_projects(conn, username, bank=None):
+    update_stmt = "UPDATE association_table SET projects='*' WHERE username=?"
+    if bank is None:
+        conn.execute(update_stmt, (username,))
+    else:
+        update_stmt += " AND bank=?"
+        conn.execute(
+            update_stmt,
+            (
+                username,
+                bank,
+            ),
+        )
+
+    update_mod_time(conn, username, bank)
+    conn.commit()
+
+    return 0
+
+
 ###############################################################
 #                                                             #
 #                   Subcommand Functions                      #
@@ -480,6 +500,10 @@ def edit_user(
                 except ValueError as bad_queue:
                     raise ValueError(f"queue {bad_queue} does not exist in queue_table")
             if field == "projects":
+                if str(params[field]) == "-1":
+                    # reset the association's projects list to just "*"
+                    clear_projects(conn, username, bank)
+                    return 0
                 try:
                     params[field] = validate_project(conn, params[field])
                 except ValueError as bad_project:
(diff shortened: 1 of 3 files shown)


src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -480,6 +500,10 @@ def edit_user(
                 except ValueError as bad_queue:
                     raise ValueError(f"queue {bad_queue} does not exist in queue_table")
             if field == "projects":
+                if params[field] == -1 or params[field] == "-1":
```
wihobbs, line 503: You could check `int(params[field])` and avoid the double-check?

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -280,6 +280,26 @@ def clear_queues(conn, username, bank=None):
     return 0
 
 
+def clear_projects(conn, username, bank=None):
+    update_stmt = "UPDATE association_table SET projects='*' WHERE username=?"
+    if bank is None:
+        conn.execute(update_stmt, (username,))
```
wihobbs, line 286: SQL noob here but stray comma?

wihobbs (approved):
Minor stuff, one might not even be wrong. Take it or leave it

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: Actually, `Connection` objects take a tuple for the parameter substitution when it is executing a `SQL` statement: 

> Instead, use the DB-API’s parameter substitution. To insert a variable into a query string, use a placeholder in the string, and substitute the actual values into the query by providing them as a [tuple](https://docs.python.org/3/library/stdtypes.html#tuple) of values to the second argument of the cursor’s [execute()](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.execute) method.

reference: https://docs.python.org/3/library/sqlite3.html

src/bindings/python/fluxacct/accounting/user_subcommands.py
wihobbs: Excellent point. An invalid cast in accounting would not be ideal...

cmoussa1:
Thanks for reviewing @wihobbs! Force-pushed to catch up after #494. Setting MWP here

PR #496, opened 2024-10-07
command suite: add new `list-projects` command

cmoussa1:
#### Problem

There currently exists no way to list all of the currently registered projects in the flux-accounting database.

---

This PR adds a new command to the flux-accounting command suite, `list-projects`, that creates a table of all of the currently registered projects in the flux-accounting database.

I've also added a couple sharness tests to `t1025-flux-account-projects.t` that check the output of calling `list-projects` with just the default project listed and after registering a couple of other projects.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/project_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/project_subcommands.py
@@ -115,3 +115,32 @@ def delete_project(conn, project):
         return warning_stmt
 
     return 0
+
+
+def list_projects(conn):
+    """
+    List all of the available projects registered in the project_table.
+    """
+    cur = conn.cursor()
+
+    cur.execute("SELECT * FROM project_table")
+    rows = cur.fetchall()
+
+    # fetch column names and determine width of each column
+    col_names = [description[0] for description in cur.description]
+    col_widths = [
+        max(len(str(value)) for value in [col] + [row[i] for row in rows])
+        for i, col in enumerate(col_names)
+    ]
+
+    def format_row(row):
+        return " | ".join(
+            [f"{str(value).ljust(col_widths[i])}" for i, value in enumerate(row)]
+        )
+
+    header = format_row(col_names)
+    separator = "-+-".join(["-" * width for width in col_widths])
+    data_rows = "\n".join([format_row(row) for row in rows])
+    table = f"{header}\n{separator}\n{data_rows}"
+
+    return table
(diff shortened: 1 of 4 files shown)


t/t1025-flux-account-projects.t
```diff
@@ -43,6 +43,16 @@ test_expect_success 'add some queues to the DB' '
 	flux account add-queue special --priority=99999
 '
 
+test_expect_success 'list contents of project_table before adding projects' '
+	flux account list-projects > project_table.test &&
+	cat <<-EOF >project_table.expected
+	project_id | project | usage
+	-----------+---------+------
+	1          | *       | 0.0
```
wihobbs, line 51: Usage is reported in what unit?

src/bindings/python/fluxacct/accounting/project_subcommands.py
```diff
@@ -115,3 +115,32 @@ def delete_project(conn, project):
         return warning_stmt
 
     return 0
+
+
+def list_projects(conn):
```
wihobbs, line 120: Okie sorry for another Noob question but... In my brainmap projects were subsidiaries of banks. So does this list only the projects for one bank, or can projects cut across banks altogether?

wihobbs (approved):
My two comments are my misunderstanding and nothing you did wrong. Go ahead with MWP!

t/t1025-flux-account-projects.t
cmoussa1: Usage in accounting is calculated as a product of `nnodes` and `t_elapsed` (which is defined as `t_inactive - t_run`). I have some more detailed explanations and examples [here](https://github.com/flux-framework/flux-accounting/tree/master/src/bindings/python/fluxacct/accounting).

src/bindings/python/fluxacct/accounting/project_subcommands.py
cmoussa1: Projects are created and managed separately from banks. My understanding of projects is that they can be thought of as annotations on a job that multiple users (who might belong to different banks) can make. So, this command will list all registered projects that the flux-accounting database knows about, i.e ones that are added via `flux account add-project my_project`.

cmoussa1:
Thanks for reviewing @wihobbs! Setting MWP here too

src/bindings/python/fluxacct/accounting/project_subcommands.py
wihobbs: Thank you so much @cmoussa1 for all the clarifications. I created more work for you by trying to use this review to learn about the interfaces we might use to solve some issues raised by our users at Sandia. I really appreciate the explanations!

PR #499, opened 2024-10-09
plugin: add callback prefixes to exception messages

cmoussa1:
#### Problem

The priority plugin uses the same exception message for jobspec updates in more than one callback because there is more than one instance where that message would be appropriate. However, these exception messages are not prefixed with the callback the exception is being raised from, which can make it hard to debug when looking at a job's eventlog.

---

This PR just adds a prefix of the name of the callback the exception message is coming from.


jameshcorbett (approved):
👍

cmoussa1:

PR #500, opened 2024-10-09
doc: add note about manually loading plugin

### LINKED ISSUE #497: documentation: add note about `flux jobtap load`, `flux config reload` for priority plugin setup
The documentation detailing the setup of the priority plugin could use some additional clarification. More specifically, notes about calling the `flux jobtap load mf_priority.so` command should be added in the case where the plugin is not listed to load in a config file for the job-manager.

cmoussa1:
#### Problem

The flux-accounting guide does not detail how to manually load the priority plugin if the path to the plugin is not configured in the job-manager config file.

---

This PR just adds a note that the plugin can be loaded manually with `flux jobtap load`. It includes the `flux account-priority-update` command so admins know to send flux-accounting information to the plugin once it is loaded.

Fixes #497

### DIFF
--- a/doc/guide/accounting-guide.rst
+++ b/doc/guide/accounting-guide.rst
@@ -130,6 +130,14 @@ configure the ``job-manager`` to load this plugin.
 
 See also: :core:man5:`flux-config-job-manager`.
 
+The plugin can also be manually loaded with ``flux jobtap load``. Be sure to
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
👍

cmoussa1:

PR #504, opened 2024-10-11
ci: add spellchecker to flux-accounting

cmoussa1:
#### Problem

flux-accounting does not have a spellchecker that runs on pull requests.

---

This PR adds a spellchecker as another builder during CI. As a result of the spellchecker running (and my horrendous spelling), I've also fixed all of the spelling mistakes flagged by the spellchecker throughout flux-accounting. Like flux-core, I've removed `IPv4` and `IPv6` from the dictionary. Finally, I've added `"dne"`, `"DNE"`, and `"fairshare"` as acceptable words.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,13 @@ jobs:
     - run: git fetch origin master
     - uses: flux-framework/pr-validator@master
 
+  spelling:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check Spelling
+      uses: crate-ci/typos@bcafd462cb07ef7ba57e34abf458fe20767e808b # v1.19.0
+
   python-format:
     name: python format
     runs-on: ubuntu-latest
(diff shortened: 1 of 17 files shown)


config/tap-driver.sh
```diff
@@ -402,7 +402,7 @@ function setup_result_obj(line)
   sub("^(not )?ok[ \t]*", "", line)
 
   # If the result has an explicit number, get it and strip it; otherwise,
-  # automatically assing the next progresive number to it.
```
garlick, line 405: This file comes from an outside source and is already excluded in `.typos.toml` so should this change be dropped?

jameshcorbett (approved):
I like @garlick's suggestions, otherwise LGTM!

cmoussa1:
Thanks @garlick and @jameshcorbett for the review! I've gone ahead and split out the changes into separate commits to 1) fix the spelling mistakes, 2) fix the spellchecker, and 3) add the spellchecker as a builder. I can set MWP here shortly

PR #506, opened 2024-10-14
doc: add "Database Administration" section, update README to point to docs site

cmoussa1:
#### Problem

flux-accounting has its own docs site now, but still has some detailed sections in its top-level README that would probably be more suited for its own sections on the docs site.

---

This PR continues to make additions to flux-accounting's ReadTheDocs site by moving some of the sections from the top-level README to the documentation site. Specifically, the section talking about interacting with the flux-accounting DB now has its own page on ReadTheDocs where it goes into slightly more detail and with more examples.

I've also added a link to the ReadTheDocs site on the top-level README as well as a new badge for the build status of the documentation.


doc/components/database-administration.rst
```diff
@@ -0,0 +1,174 @@
...
+
+flux-accounting provides an interface to the SQLite database containing all of
+the aforementioned data for associations, banks, queues, and projects. Commands
+can be run to add/remove all of these, edit certain attributes, and view this
+information in different formats. See ``flux account --help`` to list all of
+the available flux-accounting commands.
+
+The bank/association database structure assumes one that is hierarchical in
+structure, meaning that there is one root bank from which all banks and
+associations belong under:
```
jameshcorbett, line 40: How about 

> assumes a tree structure, meaning that there is one root bank which all other banks and associations descend from.

jameshcorbett (approved):
Looks good, one sentence I thought was a little confusing.

doc/components/database-administration.rst
cmoussa1: That's a great suggestion, thanks!

cmoussa1:
Thanks for approving @jameshcorbett! Setting MWP here

PR #509, opened 2024-10-15
`add-bank`: add a check when adding a root bank

### LINKED ISSUE #508: `add-bank`: multiple root banks can be added
The structure of the flux-accounting database hierarchy of banks and users assumes there is only one root bank, but multiple root banks can be added:

```console
$ flux account add-bank root 1
$ flux account add-bank second_root 1
```

There should be some safeguards to check that if there is already a root bank, another one cannot be added.

cmoussa1:
#### Problem

Mentioned in #508, the structure of the flux-accounting database hierarchy of banks and users assumes there is only one root bank (e.g when calculating fair-share), but multiple root banks can be added with the `add-bank` command.

---

This PR adds a check when adding a root bank that one does not already exist. If so, an exception is raised.

A couple of existing tests that did not make this assumption have been edited to account for this new check, and I've also added a simple sharness test that ensures an exception is raised when trying to add more than one root bank.

Fixes #508

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -219,6 +219,12 @@ def print_parsable_hierarchy(cur, bank, hierarchy_str, indent=""):
 def add_bank(conn, bank, shares, parent_bank=""):
     cur = conn.cursor()
 
+    if parent_bank == "":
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
LGTM!

cmoussa1:
Thanks! Setting MWP here

PR #510, opened 2024-10-15
doc: reorganize top-level site, add License and Support page

cmoussa1:
#### Problem

The ReadTheDocs site just lists all pages on one level and doesn't organize the site into guides, general information, and man pages. It would be better if it was oriented more like flux-core, which is organized. The site is also missing a **License & Support** page.

---

This PR adds a **License & Support** page to flux-accounting's ReadTheDocs site, as well as does a general re-orientation of the top-level index to match the organization of flux-core's. Specifically, it just adjusts the man page index to match flux-core's orientation and man5 title.


wihobbs (approved):

cmoussa1:
Thanks @wihobbs, setting MWP here

PR #512, opened 2024-10-16
`flux-account.py`: get rid of dictionary initialization

cmoussa1:
#### Problem

`flux-account.py` manually creates the dictionary of arguments that is sent via RPC to the flux-accounting service, but since the service already unpacks these arguments, `flux-account.py` can just create a dictionary of the arguments using `vars()` and send the dictionary without having to define each key-value pair before sending the RPC.

---

This PR changes `flux-account.py` to just use `vars()` on the parsed args to create a dictionary before sending the RPC to the appropriate flux-accounting command.

### DIFF
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -331,7 +331,7 @@ class AccountingService:
             val = b.list_banks(
                 self.conn,
                 msg.payload["inactive"],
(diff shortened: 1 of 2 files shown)


wihobbs (approved):
Nice improvement! LGTM

cmoussa1:
Thanks! Setting MWP here

PR #513, opened 2024-10-16
doc: add priority equation to accounting guide

garlick:
Problem: the accounting guide uses words to describe how priority is calculated but math would be clearer.

Add "queue" to the lists of factors and format the list as a dictionary rather than a bullet list.

Use a math expression for the priority value.

I briefly looked at the code and priority appeared to be calculated this way - did I get it right?


garlick:
(oops fixed a typo and forced a push)

cmoussa1 (approved):
Looks good! Thanks for adding this @garlick

garlick:
Just to confirm, so the urgency factor has a weight of 1 and we subtract 16 from it before adding?

cmoussa1:
You are correct, it has a weight of 1.

garlick:
OK, I'll set MWP then. Thanks!

Issue #517, opened 2024-10-17
t1011-job-archive-interface.t relies on deprecated module


garlick:
Problem: `t1011-job-archive-interface.t` relies on the job-archive module in flux-core which is being considered for removal in flux-framework/flux-core#6378 since flux-accounting now has that capability on its own.

Possibly it's too early to remove this module?

If we can remove it, but we need to keep tests for migration around for a while we could maybe do something like this
```diff
index a2f18ea..726b5f1 100755
--- a/t/t1011-job-archive-interface.t
+++ b/t/t1011-job-archive-interface.t
@@ -4,6 +4,20 @@ test_description='test fetching jobs and updating the fair share values for a gr
 
 . $(dirname $0)/sharness.sh
 
+have_module() (
+    eval $(flux env)
+    IFS=:
+    for dir in "$FLUX_MODULE_PATH"; do
+        test -f $dir/$1 && return 0
+    done
+    return 1
+)
+
+if ! have_module job-archive.so; then
+    skip_all="Skipping job-archive tests as module is unavailable"
+    test_done
+fi
+
 DB_PATH=$(pwd)/FluxAccountingTest.db
 ARCHIVEDIR=`pwd`
 ARCHIVEDB="${ARCHIVEDIR}/jobarchive.db"
```

I wasn't sure if the test was covering other important things as well.

Thoughts?

cmoussa1:
I think the main reason for keeping the job-archive load in the test suite (specifically in this file) was just for testing that the flux-accounting script that was responsible for fetching inactive jobs had the capability of querying the job-archive DB if needed. :-) I think it would probably be okay to remove the module if you guys wanted to. I could edit `t1011-job-archive-interface.t` to no longer test querying the job-archive DB for job records. What do you think?

garlick:
Accounting is the only user, so it's past needing it then it probably makes sense to remove the module in core.

If the test script is doing other things, the diff above could be modified to do a
```sh
if have_module job-archive.so; then
    test_set_prereq JOB_ARCHIVE
```

and then any sub-tests that require the module could declare that they need that prereq e.g.
```sh
test_expect_success JOB_ARCHIVE 'test description...'
```

`t/python/t1006_job_archive.py` would need attention too.

garlick:
No urgency here btw.

cmoussa1:
[...]

PR #518, opened 2024-10-17
t: skip t1011 if job-archive module not detected, add new tests for `fetch-job-records`

cmoussa1:
#### Problem

Mentioned in #517, `t1011-job-archive-interface.t` relies on the job-archive module
 in flux-core which is being considered for removal in
 https://github.com/flux-framework/flux-core/pull/6378 since flux-accounting now has that
 capability on its own.

---

This PR adds a check in `t1011-job-archive-interface.t` that will skip the test file in the event that the job-archive module does not exist.

I've also added a new file to the test suite that very similarly runs through the same tests in `t1011` but does not rely at all on the detection of the job-archive module.

### DIFF
--- a/t/t1011-job-archive-interface.t
+++ b/t/t1011-job-archive-interface.t
@@ -5,8 +5,6 @@ test_description='test fetching jobs and updating the fair share values for a gr
 . $(dirname $0)/sharness.sh
 
 DB_PATH=$(pwd)/FluxAccountingTest.db
-ARCHIVEDIR=`pwd`
-ARCHIVEDB="${ARCHIVEDIR}/jobarchive.db"
 QUERYCMD="flux python ${SHARNESS_TEST_SRCDIR}/scripts/query.py"
 NO_JOBS=${SHARNESS_TEST_SRCDIR}/expected/job_usage/no_jobs.expected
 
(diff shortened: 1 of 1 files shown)


chu11:
Why not just remove the tests that rely on job-archive, since flux-accounting is no longer using job-archive?  Then perhaps this could just be one commit, saying something like "we aren't using job-archive anymore, rework to rely on ..."?

cmoussa1:
I think we thought about it briefly in the thread #517, but we are not 100% sure if there are old-enough versions of flux-accounting that are still running on our systems that still rely on the job-archive module? i.e `v0.33.0` and older. Sorry if I am misunderstanding you.

garlick:
My concern was that we don't necessarily want to lose coverage for reading and parsing the `job-archive` database if we still need to migrate some production systems to the new job table.  If we're past that then agreed we can probably just drop the tests.

chu11:
> If we're past that then agreed we can probably just drop the tests.

I was going with this assumption that we were long past systems using `job-archive`.  But if we aren't, then we should keep the tests around a little longer and just create a "TODO" issue for later on.

cmoussa1:
I think I'll just go ahead and drop those tests - I don't think versions of flux-accounting `v0.33.0` or older are still in use anywhere. I'll edit this PR to just have one commit!

garlick (approved):
LGTM!

cmoussa1:
Thanks! Setting MWP here

PR #522, opened 2024-10-29
`view_jobs()`: adjust helper function to actually return a string

cmoussa1:
#### Problem

The `convert_to_str()` function used to convert a list of `JobRecord` objects to a string does not actually convert it to a string, but rather just appends separate strings to a list and then returns a list. Thus, `flux-account.py` has to have some special logic for just the `view-job-records` command to parse the returned data from the function.

---

This PR adds a `.join()` call to the `convert_to_str()` function to actually construct a string of all of the job records returned by `view_jobs()`. It also removes the special logic in `flux-account.py` that iterates through the list of job records and prints it out since it is no longer needed. I've also adjusted the unit tests in `t1006_job_archive.py` to use a better function in `jobs_table_subcommands.py`, as well as fixed a test description for one of the unit tests.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
@@ -123,7 +123,7 @@ def convert_to_str(job_records):
             )
         )
 
-    return job_record_str
+    return "\n".join(job_record_str)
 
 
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
Looks reasonable!

cmoussa1:
Thanks! Setting MWP here

PR #524, opened 2024-11-01
`list-banks`: use `AccountingFormatter` class

cmoussa1:
#### Problem

The `list-banks` command defines its own formatting for the output of the command, but the Python bindings now has its own common formatter and SQL utilities.

---

This PR edits the `list-banks` command to make use of the new formatter and external SQLite utilities. It also changes the default for the `--fields` optional argument to be `None` if it is not specified.


wihobbs (approved):
Nice cleanup!

PR #525, opened 2024-11-04
python: create new `BankFormatter` subclass, restructure `view-bank` to use new class

cmoussa1:
#### Problem

The `AccountingFormatter` class has good overarching methods for printing the results of a query to the flux-accounting database, but there is some functionality specific to viewing bank information from the database that could use its own subclass.

---

This PR adds a new subclass called `BankFormatter`, which contains unique methods for viewing bank/user information in hierarchical and parsable formats with the `view-bank` command, particularly with the `--tree`, `--users`, and `-P` options. A new `--fields` option is also added to match the customization available in the `list-banks` command. It removes the helper functions previously defined in `bank_subcommands.py` in favor of using this new subclass.

I've added some new unit tests for the output of the reworked `view-bank` command to the testsuite as well as adjusted a number of expected output files throughout the testsuite to account for the use of the new `BankFormatter` subclass.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -13,7 +13,7 @@ import json
 
 
 class AccountingFormatter:
-    def __init__(self, cursor):
+    def __init__(self, cursor, error_msg="no results found in query"):
         """
         Initialize an AccountingFormatter object with a SQLite cursor.
 
@@ -26,7 +26,7 @@ class AccountingFormatter:
 
         if not self.rows:
             # the SQL query didn't fetch any results; raise an Exception
-            raise ValueError("no results found in query")
+            raise ValueError(error_msg)
 
     def get_column_names(self):
         """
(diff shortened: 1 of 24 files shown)


src/bindings/python/fluxacct/accounting/formatter.py
```diff
@@ -94,3 +94,183 @@ def as_json(self):
...
+                # reached a bank with no sub banks, so get associations under this bank
+                cur.execute(
+                    "SELECT username,shares,job_usage,fairshare FROM association_table WHERE bank=?",
+                    (bank,),
+                )
+                users = cur.fetchall()
+                if users:
+                    for user in users:
+                        hierarchy += (
+                            f"{indent} {bank}|{user[0]}|{user[1]}|{user[2]}|{user[3]}\n"
```
wihobbs, line 234: Beware the magic numbers here (and in the lines below...)

t/t1036-hierarchy-small-no-tie-db.t
```diff
@@ -54,7 +54,7 @@ test_expect_success 'view database hierarchy' '
 '
 
 test_expect_success 'view database hierarchy in a parsable format' '
-	flux account view-bank -P root > small_no_tie_parsable.test &&
+	flux account view-bank --tree -P root > small_no_tie_parsable.test &&
```
wihobbs, line 57: commit nit: hierarchies

wihobbs (approved):
Nice work!

cmoussa1:
Thanks @wihobbs! I've rebased this branch to catch up to `master`; setting MWP here

PR #527, opened 2024-11-06
`view-user`: create new `AssociationFormatter` subclass for viewing associations

cmoussa1:
#### Problem

The `view_user()` function manually defines its own print functionality for viewing information about an association in the
flux-accounting database, but the Python bindings have its own formatting class that does this. It's also tedious to add new formatting features to the `view_user()` function.

---

This PR creates a new subclass of `AccountingFormatter` called `AssociationFormatter`. It has a unique error message when the user cannot be found in the `association_table`, but otherwise inherits the default JSON and table outputs that the `AccountingFormatter` class already provides. Thus, the various helper functions used to create output for the `view_user()` functions could be removed.

I've edited a number of tests in the testsuite that check for specific output from the `view-user` command, as well as added unit tests for the `view_user()` function.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -286,3 +286,23 @@ class BankFormatter(AccountingFormatter):
             return info
         except ValueError:
             return info + f"\n\nno users under {bank}"
+
+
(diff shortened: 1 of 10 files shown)


wihobbs (approved):

cmoussa1:
rebased to catch up after #525; setting MWP here

PR #528, opened 2024-11-06
t: change which user is deleted from association_table

cmoussa1:
#### Problem

There is a test in `t1023-flux-account-banks.t` that makes sure the default bank of a user is updated when one of their user/bank rows is deleted from the database by removing it in the test prior, but the wrong user is deleted.

---

This PR just changes which user is deleted in `t1023-flux-account-banks.t`.

### DIFF
--- a/t/t1023-flux-account-banks.t
+++ b/t/t1023-flux-account-banks.t
@@ -110,13 +110,14 @@ test_expect_success 'add a user to two different banks' '
 '
 
 test_expect_success 'delete user default bank row' '
-	flux account delete-user user5013 E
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
Looks reasonable but shouldn't this test have failed then?

cmoussa1:
The test continued to pass because the `grep` check was lazy (and my fault); `grep -w "username: user5015\|bank: F\|default_bank: F" new_default_bank.out` was checking for a whole-word match for one of the patterns instead of all three. I changed this `grep` check to be more strict and check that all three were present.

Setting MWP here

PR #532, opened 2024-11-11
`add-user`: make `--username` and `--bank` required arguments

### LINKED ISSUE #531: `add-user`: error message is confusing when a bank is not specified
Was working on some documentation this morning when I noticed the error message for `add-user` being somewhat inconspicuous:

```console
$ flux account add-user --username=moose
OSError: [Errno 22] error in add_user: Bank None does not exist in bank_table
```

This is because the contents of the `bank` argument in `add_user()` is not checked before looking up the argument's value in `bank_table` with `validate_bank()`.

~~A check should be added at the top of the `add_user()` function to ensure that `bank` is not `None`, and if so, raise a `ValueError`.~~

A simple fix could be to just set `required=True` for both of these arguments in `add_add_user_arg(subparsers)`.

cmoussa1:
#### Problem

The `--username` and `--bank` arguments for `add-user` should be required arguments, but they are currently treated as optional arguments. This means that if someone calls add-user without passing `--username` or `--bank`, the corresponding `add_user()` function will try to run the function with `None` values. Requirements should be added for both of these arguments at the top level so that the call fails if either one of these is not present.

---

This PR adds `required=True` to both the `--username` and `--bank` arguments in `add-user`. I've also added two tests to `t1007-flux-account-users.t` to make sure errors are raised when either one of these is missing from the `add-user` call.

Fixes #531

### DIFF
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -171,16 +171,16 @@ class AccountingService:
     def add_user(self, handle, watcher, msg, arg):
         try:
             val = u.add_user(
-                self.conn,
-                msg.payload["username"],
-                msg.payload["bank"],
-                msg.payload["userid"],
-                msg.payload["shares"],
-                msg.payload["max_running_jobs"],
-                msg.payload["max_active_jobs"],
-                msg.payload["max_nodes"],
-                msg.payload["queues"],
-                msg.payload["projects"],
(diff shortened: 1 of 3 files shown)


wihobbs (approved):

cmoussa1:
Rebased to catch up here too. Setting MWP

PR #534, opened 2024-11-12
`export-db/pop-db`: extend functionality to return more data, allow import of any table in database

cmoussa1:
#### Problem

The `export-db` and `pop-db` commands only handle a couple of columns from just the `association_table` and `bank_table`. These commands would be more robust if they included export and import support from all of the tables.

---

This PR restructures the `export_db_info()` function to return data from *all* of the tables in the flux-accounting DB into separate `.csv` files, labeled with their table names in the database. It also restructures the `populate_db()` function to handle populating any of the tables in the flux-accounting DB with a corresponding `.csv` file. It adds an optional argument to the command which allows the user to specify which columns to include from the file when populating the table.

As a result of the improvements made to both functions, I've also restructured the tests for both of these commands slightly, mostly just:

- updated the filenames to match the table names in the flux-accounting DB
- include the column names in the first line of the `.csv` files
- drop the use of the `--banks` and `--users` optional arguments

### DIFF
--- a/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
@@ -13,93 +13,97 @@ import csv
 import sqlite3
 import json
 
-from fluxacct.accounting import bank_subcommands as b
-from fluxacct.accounting import user_subcommands as u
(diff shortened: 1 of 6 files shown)


jameshcorbett (approved):
Looks reasonable to me!

PR #535, opened 2024-11-13
`edit-user`: unify reset behavior, `**kwargs` for editable fields

cmoussa1:
#### Problem

The `edit_user()` function is clunky and could stand to undergo some improvements. Namely:

- the reset behavior for editable fields is inconsistent. For example, an association's `projects` list can be reset with `-1`, but their `queues` list is reset with an empty string.
- the function parameters define every editable field in the signature when it really could just use `**kwargs`.
- The function allows `bank` to be edited for an association, which should not be allowed since associations, once created, are not deleted because they can continue to contribute to a bank's fair-share (up to the job usage reset period).

---

This PR changes the function's signature to use `**kwargs` for the editable fields and removes "bank" as an editable field for an association. It also unifies the reset behavior for any of the editable fields by making `-1` the keyword to reset the field to its default value.

Finally, I've added a function docstring description.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -464,24 +464,39 @@ def delete_user(conn, username, bank):
     return 0
 
 
-def edit_user(
-    conn,
-    username,
-    bank=None,
-    userid=None,
-    default_bank=None,
-    shares=None,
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
A lot of different small changes in one commit but they are all related I guess so LGTM!

PR #538, opened 2024-11-21
doc: add man(1) page support, entries for DB administration subcommands

cmoussa1:
#### Problem

flux-accounting has no support for adding man(1) pages to the repository, which is needed for man pages for all of the various `flux account` subcommands.

---

This PR begins to add `man(1)` entries to the repository that can be accessed with `man flux-account`. To avoid posting a gigantic PR with many `.rst` files to read and review, I've started with adding just `man(1)` entries for all of the subcommands related to DB administration: `create-db`, `export-db`, and `pop-db`. These man pages can be accessed either with `man flux-account create-db`, `man flux-account-create-db`, or `man flux account-create-db`.

### DIFF
--- a/doc/man1/flux-account.rst
+++ b/doc/man1/flux-account.rst
@@ -1,0 +1,121 @@
+.. flux-help-section: flux account
+
+===============
+flux-account(1)
+===============
+
+
+SYNOPSIS
+========
+
+**flux** **account** [*COMMAND*] [OPTIONS]
+
+DESCRIPTION
+===========
+
+.. program:: flux account
+
+:program:`flux account` provides an interface to the SQLite database containing
+information regarding banks, associations, queues, projects, and archived jobs.
+It also provides administrative commands like exporting and populating the DB's
+information to and from ``.csv`` files, updating the database when new versions
+of flux-accounting are released, and more.
+
+DATABASE ADMINISTRATION
+=======================
+
+create-db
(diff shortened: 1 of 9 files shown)


jameshcorbett (approved):
Nice!

doc/man1/flux-account.rst
```diff
@@ -0,0 +1,121 @@
...
+edit-bank
+^^^^^^^^^
+
+list-banks
+^^^^^^^^^^
+
+QUEUE ADMINISTRATION
+====================
+
+view-queue
```
wihobbs, line 84: Sorry I didn't add a review @cmoussa1. I didn't know if this was WIP or not since a lot of these commands were blank, though its obvious what some of them do.

My only confusion is that (I assume) delete-queue doesn't correspond to a flux queue but rather an accounting construct of queues. Might want to clarify that. And again, so sorry I missed this!

cmoussa1: No problem at all @wihobbs - these are left blank on purpose until they valid entries in them. I was going to go piece by piece when adding these man pages to avoid a huge PR adding slightly detailed man pages for every command.

You make a great point though about `delete-queue`, so I'll be sure to include this when adding man pages for the `queue-*` commands.

cmoussa1:
Thanks for the review!! I'll set MWP

PR #541, opened 2024-11-25
priority plugin: add support for unpacking priorities for queues in TOML config

### LINKED ISSUE #540: plugin: read priority configuration information from TOML files
The priority plugin already reads basic configuration properties for general priority factor weights, but not priority attributes for queues specifically, e.g their associated priorities. The plugin should be expanded to support reading attribute properties about queues (such as priorities) from a config file.

cmoussa1:
#### Problem

The priority plugin cannot unpack any defined integer priorities for queues in a configuration file, but it was brought up in an offline conversation that it would be nice to be able to set these priorities in a TOML config file.

---

This PR adds an unpack of an `accounting.queue-priorities` table in the callback for `conf.update` in the priority plugin. The table can have any queues and their associated integer priorities. If the queue already exists in the plugin's internal `queues` map, it will update the associated priority of the queue. If the queue does not exist, it will add the queue to the plugin's internal `queues` map.

I've added some basic tests that go through defining and changing a priority for a specific queue in a TOML file as well as updated the `flux-config-accounting` man page to explain how to set up this table.

Fixes #540

### DIFF
--- a/doc/man5/flux-config-accounting.rst
+++ b/doc/man5/flux-config-accounting.rst
@@ -6,6 +6,9 @@ flux-config-accounting(5)
 DESCRIPTION
 ===========
 
+accounting.factor-weights
+-------------------------
+
 The flux-accounting priority plugin can be configured to assign different
 weights to the different factors used when calculating a job's priority.
 Assigning a higher weight to a factor will result in it having more
@@ -24,7 +27,7 @@ The ``accounting.factor-weights`` sub-table may contain the following keys:
 
 
 KEYS
-====
+^^^^
 
 fairshare
     Integer value that represents the weight to be associated with an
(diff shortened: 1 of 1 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -234,6 +237,37 @@ static int conf_update_cb (flux_plugin_t *p,
...
+            } else {
+                // queue does not exist; create a new queue and insert it
+                // and set the limits to their default values defined in
+                // the flux-accounting DB
+                Queue new_queue;
+                new_queue.min_nodes_per_job = 1;
+                new_queue.max_nodes_per_job = 1;
+                new_queue.max_time_per_job = 60;
+                new_queue.priority = priority;
+                queues[queue_name] = new_queue;
```
grondo, line 266: Can you remind me if it is an error if a queue exists in the Flux broker config, but does not exist in the flux-accounting database?

I'm wondering if these defaults make sense if, say, an admin wanted to configure a temporary queue with a high priority. The defaults of max nodes (1)  and max time (60m) would not necessarily make sense in this case...

cmoussa1: Good question. I don't think it is an error - I think the plugin will just return `UNKNOWN_QUEUE`, meaning it can't find any information about the queue, so it will just apply a priority of `0` for that queue. Perhaps I should just leave out the initialization of max nodes and max time when adding this queue to the map?

grondo: That's what I was thinking (just leave out initialization of maximums here). However, we should ask @kkier and @ryanday36 if there is a use case for adding queue priorities for queues  not in the accounting db.

cmoussa1: Cool. FWIW, those maximums are not currently enforced anywhere in the plugin (yet). But probably good to get @ryanday36 and @kkier's thoughts on it too.

ryanday36: I can confirm that it is not an error for queues to exist in the broker config but not the accounting database. I've generally only added queues to the accounting database when I have a specific need to do so (i.e. I need to limit access to the queue to specific users). The use case that I can see for queue priorities without the accounting db would be pre-production systems where we're not otherwise worried about banks / fairshare / per-user limits / etc. It might be useful to be able to configure what the accounting plugin is enforcing a bit more (e.g. enforce queue limits, but don't require banks or calculate fairshare). It's pretty easy to set up a dummy-ish bank (`guests`) and add everyone to it though, which is what we're usually actually doing.

PR #542, opened 2024-11-25
doc: add example on configuring priorities for queues

cmoussa1:
#### Problem

The accounting guide makes mention of setting priorities for a queue so that it can affect the priority calculation for a submitted job, but there are no examples this configuration anywhere in the docs.

---

This PR adds a subsection in the **Multi-Factor Priority Plugin** section on how configuring priorities for certain queues affects a job's calculated priority. I've also written a short example on how the priority for a job changes depending on this configuration.


grondo (approved):
Nice!!

PR #547, opened 2024-12-16
github: fix ubuntu version for `"python-format"` action

cmoussa1:
#### Problem

The `"python-format"` GitHub action fails with:

> Error: The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04.

because Python 3.7 has reached its EOL: https://github.com/actions/setup-python/issues/962.

---

This PR fixes the version of Ubuntu for the `"python-format"` action to `20.04` instead of using `ubuntu-latest`.

PR #548, opened 2024-12-16
testsuite: add longer test descriptions for some of the more complex test scenarios

cmoussa1:
#### Problem

Some of the tests in the testsuite can be somewhat complex and simulate a scenario that might take multiple steps; it would be helpful if descriptions were added that went into more detail than what is described in the one-liner for each test.

---

This PR makes a number of minor test description improvements in the flux-accounting testsuite, most notably adding some more thorough test descriptions in a couple of the test files explaining what is being tested, particularly for the scenarios that are broken up over multiple tests.

I've also fixed a couple of mistakes in the test description and in already-existing test comments.

Lastly, I've just added a space between the `"allow guest access..."` tests to match the format of the rest of the testsuite.

Built on top of #547

### DIFF
--- a/t/t1001-mf-priority-basic.t
+++ b/t/t1001-mf-priority-basic.t
@@ -19,6 +19,7 @@ test_expect_success 'allow guest access to testexec' '
 	allow-guests = true
 	EOF
 '
+
 test_expect_success 'load multi-factor priority plugin' '
 	flux jobtap load -r .priority-default ${MULTI_FACTOR_PRIORITY}
 '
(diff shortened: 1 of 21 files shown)


t/t1001-mf-priority-basic.t
```diff
@@ -156,6 +161,9 @@ test_expect_success 'submit a job using default bank' '
 	flux cancel ${jobid}
 '
 
+# The following two tests ensure job subimssions are rejected when a user
```
wihobbs, line 164: sp nit: submissions

wihobbs (approved):

cmoussa1:
Thanks for the review @wihobbs! Force-pushed a fix for the spelling mistake and rebased to catch up after #527; setting MWP here

PR #551, opened 2024-12-18
doc: add "Projects" page

cmoussa1:
#### Problem

There is no documentation in flux-accounting talking about how to configure and use projects for associations and how to submit jobs or view completed job records using projects.

---

This PR adds a new page to flux-accounting's ReadTheDocs site which outlines what projects are used for, how to configure them, and how to submit and view jobs by project.

### DIFF
--- a/doc/components/projects.rst
+++ b/doc/components/projects.rst
@@ -1,0 +1,122 @@
+.. _projects:
+
+########
+Projects
+########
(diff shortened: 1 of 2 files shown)


cmoussa1:
appreciate you @jameshcorbett! setting MWP here

PR #556, opened 2024-12-19
bank_table: add `max-preempt-after` column

### LINKED ISSUE #553: `bank_table`: add `max-preemptible-after` column
The `bank_table` does not have a way to track a max `preemptible-after` value for a certain bank. This kind of information is needed to enforce a limit on association-set `preemptible-after` values on their jobs as well as set a default on their job in the case where the association does not set one on their own.

Because this might be specific to a bank, its default should be to require no preemption. This would also follow the `preemptible-after` attribute defined in [RFC 14](https://github.com/flux-framework/rfc/blob/master/spec_14.rst).

The column in the `bank_table` should be a `real` number that has no default value (which would indicate that there is no preemption required for this bank and should retroactively support banks already defined), but when defined, represents:

> A floating point number greater than or equal to zero representing job run time after which the job MAY be canceled in order to make way for other jobs. A value of zero indicates that the job is immediately preemptible. The attribute SHALL NOT be set if the job is never preemptible.

This value should be able to be specified on the command line when first adding a bank as well as be edited after the bank is added.

---

```[tasklist]
### Tasks
- [ ] add `max-preemptible-after` column to `bank_table`
- [ ] update flux-accounting database schema
- [ ] add `max-preemptible-after` column to output of `view-bank`
- [ ] add `--max-preemptible-after` optional arguments to `add-bank`, `edit-bank`
```

cmoussa1:
#### Problem

The `bank_table` does not have a way to track a max `preemptible-after` value for a certain bank. This kind of information is needed to enforce a limit on association-set `preemptible-after` values on their jobs as well as set a default on their job in the case where the association does not set one on their own.

---

This PR begins to add support for enforcing `preemptible-after` limits on a per-bank basis by adding a new column to the `bank_table` called `max_preempt_after`, which is a floating point number representing the max number of seconds in which a job submitted under this bank can run before being preempted. It does not require a value and does not have a default value.

`--max-preempt-after` optional arguments are also added to the `add-bank` and `edit-bank` commands to allow an admin to both define and change the `max-preempt-after` attribute for a bank. The expected format for this attribute in both the `add-bank` and `edit-bank` commands is [Flux Standard Duration (FSD)](https://github.com/flux-framework/rfc/blob/master/spec_23.rst). A bank's `max-preempt-after` can also be cleared by passing `-1` as a value if the admin no longer wants to have a max `preemptible-after` attribute defined for a particular bank.

I've added some basic tests to a new test file that adds multiple banks with different FSD values as well as resets its value after setting a value initially.

Fixes #553

### DIFF
--- a/src/bindings/python/fluxacct/accounting/__init__.py.in
+++ b/src/bindings/python/fluxacct/accounting/__init__.py.in
@@ -1,6 +1,6 @@
 DB_DIR = "@X_LOCALSTATEDIR@/lib/flux/"
 DB_PATH = "@X_LOCALSTATEDIR@/lib/flux/FluxAccounting.db"
-DB_SCHEMA_VERSION = 29
+DB_SCHEMA_VERSION = 30
 
 PRIORITY_FACTORS = ["fairshare", "queue", "bank"]
 FSHARE_WEIGHT_DEFAULT = 100000
@@ -28,7 +28,7 @@ ASSOCIATION_TABLE = [
     "projects",
     "default_project",
 ]
-BANK_TABLE = ["bank_id", "bank", "active", "parent_bank", "shares", "job_usage", "priority"]
+BANK_TABLE = ["bank_id", "bank", "active", "parent_bank", "shares", "job_usage", "priority", "max_preempt_after"]
 QUEUE_TABLE = [
     "queue",
     "min_nodes_per_job",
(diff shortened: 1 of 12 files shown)


cmoussa1:
@ryanday36 when you have some time, could you give this a look and let me know if this looks like something you'd be good with in terms of managing banks in the flux-accounting database? Note that this doesn't actually enforce the limit in the priority plugin (yet); it's just the first part in being able to define and set this limit in the DB.

cmoussa1:
Going to close this for now.

PR #560, opened 2025-01-07
`association_table`: add `max_cores` attribute, send information to plugin

cmoussa1:
#### Problem

The `association_table` has no way to track the max number of cores an association can have across all of their running jobs. This is needed in order to track the max number of resources an association has across all of their running jobs.

---

This PR adds a new column to the `association_table` called `max_cores`. This field is configurable both when adding or editing an association in the flux-accounting database. This attribute is added to the rest of the information sent per-association to the priority plugin, where it is stored as another member in the `Association` class.

Existing tests are updated to account for the addition of the new column, and the flux-accounting database from the previous schema version is also added to the testsuite now that the schema has been updated.


jameshcorbett (approved):
Looks good! I don't really understand 9f04c000647ccf4927e760c47b8e05c5fc2bb91d but that's probably OK 😊

cmoussa1:
Thanks @jameshcorbett - that commit is just giving an admin (or whoever is using flux-accounting) the ability to configure the `max_cores` for an association in the database (i.e, set their value to some value _other_ than the default). Do you think that commit would make more sense if it was squashed onto 119acdf?

cmoussa1:
OH, oh I gotcha - that commit is adding the DB schema _before_ adding the new `max_cores` column to the `association_table` to the testsuite so that it can be tested via the `update-db` command. This is needed because there are versions of the flux-accounting database that are already in production that will not have the most up-to-date DB schema once this PR lands, so I just need to make sure that the current DBs out there will be able to be updated when a new version of flux-accounting gets released. I hope that makes sense??

jameshcorbett:
Got it that makes sense!

cmoussa1:
Thanks!! Okay, setting MWP here

PR #561, opened 2025-01-09
plugin: track the resources used across all of an association's running jobs

cmoussa1:
#### Problem

The priority plugin does not keep track of the resources used by an association across all of their running jobs, which will be needed when the plugin begins to enforce limits on how many resources they can have across all of their running jobs.

---

This PR looks to add the _tracking_ component of an association's resource consumption across all of their running jobs. It does this by copying over the [jj](https://github.com/flux-framework/flux-core/blob/master/src/common/libjob/jj.h) (along with a required [streq](https://github.com/flux-framework/flux-core/blob/02ad07cf4041d81253ca5765e5933c722b66d898/src/common/libccan/ccan/str/str.h#L10) function) code over from flux-core and utilizing the `jj_get_counts_json ()` function to get a total `nnodes` and `ncores` count. When a job enters `job.state.run`, the association's current `nnodes` and `ncores` count will get incremented accordingly, and subsequently decremented when the job enters `job.state.inactive`.

I've also added some basic tests that check tracking resources across an association's set of jobs as they go through their lifecycle.

~~I probably need to hold this PR as [WIP] until I properly add the license for the code I copied over from `ccan` over in flux-core; for now, I just manually copied the `streq ()` function to the `jj.h` file just to get a working solution, but should clean this up.~~

### DIFF
--- a/src/plugins/jj.hpp
+++ b/src/plugins/jj.hpp
@@ -1,0 +1,62 @@
+/************************************************************\
+ * Copyright 2014 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+extern "C" {
+#ifndef HAVE_JJ_H
+#define HAVE_JJ_H 1
+
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <jansson.h>
+#include <stdbool.h>
+
+#define JJ_ERROR_TEXT_LENGTH 256
+
+/**
+ * streq - Are two strings equal?
+ * @a: first string
+ * @b: first string
+ *
+ * This macro is arguably more readable than "!strcmp(a, b)".
+ *
+ * Example:
+ *	if (streq(somestring, ""))
+ *		printf("String is empty!\n");
+ */
+#define streq(a,b) (strcmp((a),(b)) == 0)
+
+struct jj_counts {
+    int nnodes;    /* total number of nodes requested */
+    int nslots;    /* total number of slots requested */
+    int slot_size; /* number of cores per slot        */
+    int slot_gpus; /* number of gpus per slot        */
+
+    bool exclusive;  /* enable node exclusive allocation if available */
+
+    double duration; /* attributes.system.duration if set */
+
+    char error[JJ_ERROR_TEXT_LENGTH]; /* On error, contains error description */
+};
+
+/*  Parse jobspec from json string `spec`, return resource request summary
+ *   in `counts` on success.
+ *  Returns 0 on success and -1 on failure with errno set and jj->error[]
+ *   with an error message string.
+ */
+
+int jj_get_counts (const char *spec, struct jj_counts *counts);
+
+/*  Identical to jj_get_counts, but take json_t  */
+int jj_get_counts_json (json_t *jobspec, struct jj_counts *counts);
+
+#endif /* !HAVE_JJ_H */
+}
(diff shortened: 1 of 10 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -1131,8 +1195,19 @@ static int inactive_cb (flux_plugin_t *p,
         return 0;
 
     // this job was running, so decrement the current running jobs count
-    // and look to see if any held jobs can be released
+    // and the resources count and look to see if any held jobs can be released
     b->cur_run_jobs--;
+    if (jobspec) {
```
jameshcorbett, line 1200: under what conditions would `jobspec` be NULL? Just wondering because then a user might get stuck with the system permanently thinking they have active resources when in fact they don't.

src/plugins/jj.hpp
```diff
@@ -0,0 +1,66 @@
...
+ * @a: first string
+ * @b: first string
+ *
+ * This macro is arguably more readable than "!strcmp(a, b)".
+ *
+ * Example:
+ *	if (streq(somestring, ""))
+ *		printf("String is empty!\n");
+ */
+#define streq(a,b) (strcmp((a),(b)) == 0)
```
jameshcorbett, line 39: The license is needed just for this macro?

jameshcorbett (approved):
Looks good 👍 two little questions.

src/plugins/mf_priority.cpp
cmoussa1: The only case I can think of where `jobspec` might be NULL is if for some reason the jobtap plugin failed to unpack the `jobspec` in `flux_plugin_arg_unpack ()`. Perhaps the plugin should raise an exception on the job if it fails to unpack it?

src/plugins/jj.hpp
cmoussa1: Correct 👍

garlick: Eh, you can safely not include a license for that one IMHO., and leave off the comment for that matter.

src/plugins/mf_priority.cpp
garlick: Pretty sure that can't happen - the "pack" operation would fail and not call this.

garlick: Oh you said unpack (sorry).  You probably want to immediately fail if the unpack fails (like due to out of memory).

garlick:
For running jobs, would it make more sense to grab _R_ instead of jobspec?  In jobspec, the node count can be ambiguous if the user only requested cores, for example.  But R always contains the actual count.

cmoussa1:
@garlick that's a good suggestion! I picked jobspec because I was able to use the `jj_get_counts()` function to get the resource counts, but I didn't know I could also do this with _R_. Is that the case?

Also, when the plugin ultimately needs to enforce a limit due to exceeding a certain max, I believe this check would need to happen in `job.state.depend`. I think the [jobtap](https://github.com/flux-framework/flux-core/blob/master/doc/man7/flux-jobtap-plugins.rst) man page states that the _R_ can only be accessed in RUN state or later. So, if I were to use _R_ in one callback and jobspec in another, is there a chance I could end up with a discrepancy in the actual resource count? Apologies if I'm misunderstanding.

garlick:
Yes, R is only available at RUN state.

On node scheduled, homogeneous systems, jobspec should be fine.  That's everything we have in production today AFAIK.

However for core scheduled and/or homogeneous systems (different node:core counts depending on the node), if the user requests N cores, you may not know how many nodes that is until the scheduler assigns them.

Maybe that's OK for now where we are now!

cmoussa1:
Thank you for the clarification! Yes, the latter case is absolutely something to consider - I think at this point, just to make some progress on this front, maybe using jobspec (and/or making an estimation as best we can) should be okay.

PR #562, opened 2025-01-15
plugin: enforce max resource limits per-association

cmoussa1:
#### Problem

The priority plugin tracks resource usage across an association's set of running jobs, but it does not enforce any limits
on this resource usage in the event where an association goes over their limit.

---

This PR looks to add limit enforcement on an association's set of running jobs according to their resource usage, specifically with `ncores` and `nnodes`. A new method is added to the `Association` class called `under_max_resources`, which takes one argument (a `Job`) and looks to see if the job's requested resources would put an association over either their `max_nodes` or `max_cores` limits and returns a `bool`, similar to the other dependency checks for an association's `max_run_jobs` and a queue's `max_run_jobs` limit. This method is called in `job.state.depend` to determine whether a job should have a `max-resources` per-association dependency added to it, as well as in the series of checks in the `check_and_release_held_jobs ()` helper function to determine whether a held job could be released.

A couple of unit tests are added to `dependencies_test03.cpp` to check that a held job could be released due to potentially exceeding _both_ `max_nodes` and `max_cores` limits _or_ potentially exceeding just one of those limits.

I've expanded the test file that originally just tracked resources across an association's set of jobs to also test checking adding and removing dependencies. Different scenarios are walked through to make sure an association's set of jobs have the correct dependencies added to it and are removed correctly, such as having a job only have a `max-running-jobs` or `max-resources` dependency added to it, have _both_ dependencies added to it at the same time, only have one of the two dependencies removed from a held job, etc.


cmoussa1:
@ryanday36 when you get the chance, could you give this PR a high-level look and let me know if this looks good from your side? I think I'd mostly be curious if the way that the plugin is enforcing the resource limits looks good to you.

ryanday36:
This looks like a good approach to me @cmoussa1. I'm not sure I understand where `b->curr_nodes` and `b->curr_cores` get updated. Is that already happening when jobs start / end? or is that still to come?

src/plugins/mf_priority.cpp
```diff
@@ -1225,24 +1329,102 @@ static int inactive_cb (flux_plugin_t *p,
...
+                                             FLUX_JOBTAP_CURRENT_JOB,
+                                             "mf_priority",
+                                             0,
+                                             "job.state.inactive: unable to " \
+                                             "count resources from jobspec " \
+                                             "of held job");
+                goto error;
+            }
+            int held_job_nnodes = counts.nnodes;
+            int held_job_ncores = counts.nslots * counts.slot_size;
```
grondo, line 1370: Hm, it might be good to consider saving the estimated nnodes/ncores for held jobs in the map somehow (sorry I haven't digested the code enough to actually suggest how to do that) instead of reparsing the jobspec.

Also, side note: Does the code currently only check to see if the held job at the front of the list can be released based on current limits? If there are more held jobs, should they also be checked to see if they can be released. E.g. if a user is over a max nodes limit of 10, and a 2 node job becomes inactive, if the held job at the front of the list is 5 nodes, and the next job is only 2 nodes, should the second one be released?

In anticipation of this, perhaps abstract the check if a held job is eligible for release into a separate function?

(hope that makes sense.

t/t1044-mf-priority-resource-limits.t
```diff
@@ -91,4 +94,188 @@ test_expect_success 'cancel job; check resource counts' '
...
+	flux cancel ${job2} &&
+	flux cancel ${job4}
+'
+
+test_expect_success 'make sure association has no held jobs in their Association object' '
+	flux jobtap query mf_priority.so > query.json &&
+	test_debug "jq -S . <query.json" &&
+	jq -e ".mf_priority_map[] | select(.userid == 5001) | .banks[0].held_jobs | length == 0" <query.json
+'
+
```
grondo, line 280: Nice exhaustive testing BTW!

grondo (commented):
I've just made a quick first pass and have some comments and suggestions.

I'll have to study the code further and perhaps experiment before I can do a full review.

src/plugins/mf_priority.cpp
cmoussa1: That's a good thought. I might have to think about how best to do that - perhaps I could extend the existing map structure to store the resource counts of the held job as well. Maybe by creating a class for held jobs that has a `jobid`, `nnodes`, and `ncores` members. I'll think about that - thanks!

> Does the code currently only check to see if the held job at the front of the list can be released based on current limits?

Yes, this was a deliberate choice to only check the held job at the front of the list (i.e the first job that was held due to a flux-accounting dependency) is eligible to be released. I chose this to avoid turning the priority plugin into its own "mini scheduler" by considering multiple jobs. I could be open to perhaps changing this, though, to consider multiple jobs for release. What do you think? Maybe @ryanday36 has thoughts on this too.

grondo: > Maybe by creating a class for held jobs that has a jobid, nnodes, and ncores members. I'll think about that - thanks!

One benefit of extracting some of the code into a class is that you could then move this to its own source file, then create a unit test based on that

t/t1044-mf-priority-resource-limits.t
```diff
@@ -91,4 +94,188 @@ test_expect_success 'cancel job; check resource counts' '
...
+# Scenario 2:
+# The association submits enough jobs to take up the max-nodes limit, but they
+# still have one core available to use, so they can submit a job that takes up
+# the last core and it will run right away.
```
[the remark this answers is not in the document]
cmoussa1: Yes, a job should be held if it causes the association to go over either limit. In the plugin, I am just relying on what the `jj_get_counts ()` function returns, and while it is not a perfect approximation, it is trying to serve at least as an estimate.

I think before I had started working on this, I had talked to @ryanday36 about this exact scenario before that you mentioned (a job requesting a bunch of cores without specifying nodes and being able to pass through), but from what I _think_ I remember from that conversation is that this case is pretty rare in practice. He should definitely comment though and clarify in case I am misremembering.

The second scenario in the sharness test is mostly just showing that the plugin will keep track of exactly what `jj_get_counts ()` returns and only enforces limits based on what resources it counts from that function; essentially, a job could be submitted that specifies an exact resource and can perhaps "get around" a set limit by only specifying nodes or only specifying cores. I'm sorry if I misunderstood your question or if I'm not making sense

src/plugins/mf_priority.cpp
cmoussa1: Cool!! Yeah, I'll play around with this and see if I can clean this up (particularly the resource counting part) by moving this into its own class.

t/t1044-mf-priority-resource-limits.t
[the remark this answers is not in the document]
grondo: > Yes, a job should be held if it causes the association to go over either limit.

> a job could be submitted that specifies an exact resource and can perhaps "get around" a set limit by only specifying nodes or only specifying cores. I'm sorry if I misunderstood your question or if I'm not making sense

Ok, I think I understand now. If a job submits a jobspec that doesn't ask for nodes, then `jj_get_counts ()` reports `nnodes=0` (which actually is impossible, since a job cannot run on 0 nodes), so the nnodes limit is not exceeded _and_ neither is the ncores limit since the association is still below that limit.

I wonder if would be more practical to set `nnodes = 1` when `jj_get_counts()` doesn't detect an explicit node request?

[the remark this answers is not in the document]
cmoussa1: That might make sense. Please correct me if I am misunderstanding, but would setting `nnodes = 1` still be accurate on a system that is not node-exclusive scheduled? Maybe it applies in both cases (node-exclusive scheduling and non-node-exclusive scheduling)?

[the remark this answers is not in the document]
grondo: Good question! I'd ask @ryanday36 for clarification. However, even on a core-scheduled system, a new job is guaranteed to use at least 1 node. However, it may not use a _new_ node when it is actually scheduled, so this is kind of a tough one! 

This is where to really properly support these limits, it might have to be the scheduler that is applying them. Only the scheduler knows if it allocates resources for a non-exclusive job whether it will use a new node and add to the associations node count. Or, maybe the node limit just doesn't make sense on core scheduled clusters?

src/plugins/mf_priority.cpp
cmoussa1: I think I was able to get somewhere here. I can instead create a `Job` class that has properties like its `id` and its `nnodes` and `ncores` counts that it retrieves in `job.state.depend`, as well as any flux-accounting-specific dependencies added to it. Then the plugin does not even need to unpack the held job's jobspec a second time in `job.state.inactive`; it can just retrieve the job's resources from its entry in a map and use it for comparison. Much cleaner!

cmoussa1: OK, I've just force-pushed up a rework to the way dependencies are added and removed in `job.state.depend` and `job.state.inactive` to make use of a newly-added `Job` class - notably, not having to count the resources for a held job a second time. The internal map that the plugin uses to keep track of held jobs uses this `Job` class as the value of the map to avoid having to do a second lookup. The callback for `job.state.inactive` is hopefully a little cleaner now as a result.

In the near future, I think I could definitely rework this to convert the `held_jobs` vector in the `Association` class to just be this same map (and this way, the plugin only needs to keep track of held jobs in one place), but that will probably be enough work to warrant its own PR.

cmoussa1:
Going to re-work this PR now that #610 has landed and we now have a `Job` class and a more structured approach to how dependencies are handled in the plugin.

cmoussa1:
OK, I've restructured this PR based on the changes implemented in #610. The addition of a `max-resources` limit is now added via new method in the `Association` class to see if a job would put an association over either their `max_cores` _or_ `max_nodes` limit, and is referenced both in `job.state.depend` as well as the helper function used to loop through jobs in `job.state.inactive`. I've also added some new unit tests to the unit test file added in #610 to simulate going over max resources limits.

t/t1044-mf-priority-resource-limits.t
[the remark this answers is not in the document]
cmoussa1: @ryanday36 if you have some time, do you think we could get your expertise here ^? It would probably be helpful to know if this approach would be okay before we move on with an official review for this PR.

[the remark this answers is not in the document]
ryanday36: > Or, maybe the node limit just doesn't make sense on core scheduled clusters?

I think this is right. Node limits don't really make sense on core-scheduled clusters, so hopefully admins won't be setting them. I guess it's still tricky if flux-accounting doesn't know the actual number of cores that the scheduler will assign when someone does something like `-N 1 --exclusive` on a core scheduled cluster. Eventually we'll need to solve that, but just setting `nnodes=1` when you don't have an explicit node count seems like a good place to start.

cmoussa1:
Thanks for the feedback above @ryanday36 - I've slightly restructured this PR to instead account for the fact that the plugin will set `nnodes = 1` when a job does not explicitly request `nnodes`. I ended up adding a commit in the beginning of this PR to update the helper functions that count the resources for those kinds of jobs to set `nnodes = 1`, as well as when considering a job in `job.state.depend`. With that, I think I'll re-ask for a review on this PR whenever convenient.

ryanday36:
@grondo, can you take a look at this review request? Folks are looking for these kinds of limits on elcap.

cmoussa1:
Thanks for all of the review and help here @grondo! Setting MWP

PR #563, opened 2025-01-24
`formatter`: add new `JobsFormatter` class, restructure `view-job-records` to use new class

cmoussa1:
#### Problem

The `formatter` module in flux-accounting does not have a class for formatting the output of job records which would be useful because the flux-accounting Python bindings manually define the formatting of these job records within its bindings.

---

This PR creates a new class called `JobsFormatter`, a subclass of `flux.util.OutputFormat`, which can take a Python format string to allow a user to customize their output, similar to flux-jobs:

```console
flux account view-job-records -o "{userid:<8} | {nnodes:<6} | {bank:<8}" 
userid   | nnodes | bank    
5001     | 1      | bankB   
5002     | 1      | bankB   
5002     | 1      | bankB   
5001     | 1      | bankA   
5001     | 1      | bankA 
```

`view-job-records` is restructured to use the new `JobsFormatter` class instead of defining its own print function as well as modified to include a `-o, --format` optional argument to the command to allow the user to specify a format string to customize their output.

Finally, a couple of the `view-job-records` tests are cleaned up to account for the use of the new formatter class.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -10,6 +10,9 @@
 # SPDX-License-Identifier: LGPL-3.0
 ###############################################################
 import json
+import string
+
+import flux.util
 
 
 class AccountingFormatter:
@@ -321,3 +324,74 @@ class AssociationFormatter(AccountingFormatter):
             banks += f"{str(bank[0])}\n"
 
         return banks
+
+
+class JobsFormatter(flux.util.OutputFormat):
+    """
+    Store a parsed version of an output format string for JobRecord objects,
+    allowing the fields to iterated without modifiers, building
+    a new format suitable for headers display, etc...
+    """
+
+    # mapping of legal format fields and their header names
+    headings = {
+        "userid": "userid",
+        "username": "username",
+        "jobid": "jobid",
+        "t_submit": "t_submit",
+        "t_run": "t_run",
+        "t_inactive": "t_inactive",
+        "nnodes": "nnodes",
+        "resources": "resources",
+        "project": "project",
+        "bank": "bank",
+    }
+
+    def __init__(self, fmt, headings=None):
+        """
+        Parse the input format fmt with string.Formatter.
+        Save off the fields and list of format tokens for later use,
+        (converting None to "" in the process)
+
+        Throws an exception if any format fields do not match the allowed
+        list of headings above.
+
+        Special case for annotations, which may be arbitrary
+        creations of scheduler or user.
+        """
+        format_list = string.Formatter().parse(fmt)
+        for _, field, _, _ in format_list:
+            if field and field in self.headings:
+                self.headings[field] = field
+        super().__init__(fmt)
+
+    def build_table(self, items):
+        """
+        Handle constructing a table of job records with the current format.
+
+        Sort items via any sort keys as set by set_sort_keys() or
+        via a ``sort:`` prefix in the supplied format.
+
+        Args:
+            items (iterable): list of items to format
+        """
+        # preprocess original format by processing with filter():
+        newfmt = self.filter(items)
+        # create new instance of the current class from filtered format:
+        formatter = type(self)(newfmt, headings=self.headings)
+
+        items = self.sort_items(items)
+
+        output_str = f"{formatter.header()}\n"
+        for item in items:
+            line = formatter.format(item)
+            if not line or line.isspace():
+                continue
+            try:
+                output_str += f"{line}\n"
+            except UnicodeEncodeError:
+                output_str += (
+                    f"{line.encode('utf-8', errors='surrogateescape').decode()}"
+                )
+
+        return output_str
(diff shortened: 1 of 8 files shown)


src/cmd/flux-account.py
```diff
@@ -293,6 +293,13 @@ def add_view_job_records_arg(subparsers):
         help="bank",
         metavar="BANK",
     )
+    subparser_view_job_records.add_argument(
+        "--fields",
+        type=str,
+        help="list of fields to include in output",
+        default="userid,username,jobid,t_submit,t_run,t_inactive,nnodes,project,bank",
+        metavar="USERID,USERNAME,JOBID,T_SUBMIT,T_RUN,T_INACTIVE,NNODES,PROJECT,BANK",
```
jameshcorbett, line 301: Does having this long `metavar` not make the help/usage message kind of weird looking?

src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
```diff
@@ -123,41 +124,32 @@ def write_records_to_file(job_records, output_file):
...
+        "jobid",
+        "t_submit",
+        "t_run",
+        "t_inactive",
+        "nnodes",
+        "project",
+        "bank",
+    ]
+    fields = fields or default_fields
+    invalid_fields = [field for field in fields if field not in default_fields]
```
jameshcorbett, line 144: If I'm reading this right, the user can suppress some fields if they choose but there aren't any hidden fields they can see by changing the format?

jameshcorbett (commented):
This looks good but I wonder if for consistency with `flux jobs` the subcommand might instead take an option like (taken from `flux jobs`)
```
 -o, --format=FORMAT        Specify output format using Python's string format syntax or a defined format by name (use 'help' to get a list of names)
```
example usage:
```
[corbett8@rzadams1:~]$ flux jobs -A -o "{id} |  {username}"
JOBID |  USER
f2DVdVNa66iB | user1
f2DbuK95EHuq |  user2
[corbett8@rzadams1:~]$ flux jobs -A -o "{id} |  {username} {foo}"
flux-jobs: ERROR: Error in user format: Unknown format field: foo
```

src/cmd/flux-account.py
cmoussa1: I don't think so? But maybe it's just my terminal?

```console
usage: flux-account.py view-job-records [-h] [-u USERNAME] [-j JOBID]
                                        [-a START TIME] [-b END TIME]
                                        [--project PROJECT] [--bank BANK]
                                        [--fields USERID,USERNAME,JOBID,T_SUBMIT,T_RUN,T_INACTIVE,NNODES,PROJECT,BANK]
```

I could probably just break this into a multi-line string though in `flux-account.py`. FWIW, it's less than 90 characters which I think is compatible with `black` formatting.

src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
cmoussa1: Yes. The user can choose to only include a couple of the fields listed, but can _only_ choose from those fields.

cmoussa1:
Thanks for reviewing this @jameshcorbett! I think that'd be a good improvement to this. Let me think about how I might be able to take a format string. It might take some extra thought because I believe `flux jobs` just `print`s stuff but these subcommands need to return strings in the function... I'll do some digging!

cmoussa1:
@jameshcorbett I went ahead and restructured this PR a little bit to have the `JobsFormatter` class actually be a subclass of `flux.util.OutputFormat`, which allows it to take a format string (like `flux jobs`). You'll notice that I had to create a mapping for the headings (I referenced [this](https://github.com/flux-framework/flux-core/blob/11ad017622ac2f0f4b942b23bed60dc1ae4be4e3/src/bindings/python/flux/job/info.py) class in flux-core), so I just kept the headings the same across the board.

src/bindings/python/fluxacct/accounting/formatter.py
```diff
@@ -321,3 +324,76 @@ def list_banks(self):
...
+
+        Args:
+            items (iterable): list of items to format
+        """
+        # preprocess original format by processing with filter():
+        newfmt = self.filter(items)
+        # get the current class for creating a new formatter instance:
+        cls = self.__class__
+        # create new instance of the current class from filtered format:
+        formatter = cls(newfmt, headings=self.headings, prepend=self.prepend)
```
jameshcorbett, line 383: I noticed this comes from flux-core but I think I would prefer `type(self)` instead of `self.__class__`, maybe `type(self)(newfmt, headings=...`

src/bindings/python/fluxacct/accounting/formatter.py
```diff
@@ -321,3 +324,76 @@ def list_banks(self):
...
+        list of headings above.
+
+        Special case for annotations, which may be arbitrary
+        creations of scheduler or user.
+        """
+        format_list = string.Formatter().parse(fmt)
+        for _, field, _, _ in format_list:
+            if field and field in self.headings:
+                self.headings[field] = field
+        super().__init__(fmt)
```
jameshcorbett, line 366: Should `prepend` and `headings` be passed on to the superclass' `__init__`?

cmoussa1:
Thanks for reviewing @jameshcorbett! After an offline discussion it seems like the `prepend` attribute is probably not needed by this subclass, so I just removed it altogether. Will set MWP here

PR #567, opened 2025-01-27
`view-job-records`: accept multiple timestamp formats for `after-start-time`, `before-end-time` optional args

### LINKED ISSUE #565: time format for --before-end-time and --after-start-time for view-job-records
It's not clear what the time format should be for `flux account view-job-records -a / -b`. I'm guessing seconds since the epoch, since that's what it outputs, but acceptance of common datetime formats as well as some input validation would be helpful.

cmoussa1:
#### Problem

The `--after-start-time` and `--before-end-time` optional args for the `view-job-records` command only takes one format: a seconds-since-epoch timestamp, which isn't super user friendly.

---

This PR makes use of the `flux.util.parse_datetime()` function to accept multiple, more human-readable format strings as values to the `--after-start-time` and `--before-end-time` optional args.

It adds a helper function to `jobs_table_subcommands` to perform the parsing of the timestamp. If parsing the timestamp string fails, it just attempts to pass it as a seconds-since-epoch timestamp to the SQLite query (which will return no job records if the timestamp is invalid).

I've also updated the help description to both of the optional arguments to provide examples of human-readable formats as well as added some basic sharness tests that pass different timestamp formats to the flux-accounting job-archive.

Fixes #565

### DIFF
--- a/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
@@ -14,6 +14,7 @@ import csv
 import json
 
 from flux.resource import ResourceSet
+from flux.util import parse_datetime
 
 
 def get_username(userid):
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
Looks like a nice improvement!

cmoussa1:
Thanks! I'll set MWP here

PR #568, opened 2025-01-28
`job.state.priority`: remove raising exception when no aux item found

### LINKED ISSUE #575: plugin: jobs in SCHED state lose their `Association` aux item on flux-restart
I noticed a couple of pending jobs on `elcap` had an exception raised on them in `job.state.priority` because the plugin could not find their `Association` aux_item set on the job, and I think this is due to the jobs being sent from SCHED back to PRIORITY after a flux-restart. When the instance is restarted, the aux items set on the job are cleared, and any pending jobs are sent back to PRIORITY:

Image

Currently, one of the first checks the plugin makes in `job.state.priority` is that it can unpack the `Association` aux item set on the job and will raise an exception on the job if it can't unpack it. But, since the aux items are cleared on a restart, these jobs will fail _even if_ the plugin already has the flux-accounting information for the association that submitted the job:

```c++
b = static_cast<Association *> (flux_jobtap_job_aux_get (
                                                p,
                                                FLUX_JOBTAP_CURRENT_JOB,
                                                "mf_priority:bank_info"));

if (b == NULL) {
    flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
                                 0, "internal error: bank info is missing");

    return -1;
}
```

Since the plugin already performs a lookup in `job.state.priority` for associations who submitted jobs _before_ the plugin was loaded, this check for `NULL` (or probably more correctly, `nullptr`) should be moved a few lines down to the following `if`-condition to be a part of an `or` condition:

```c++
if (b->max_run_jobs == BANK_INFO_MISSING) {
```

cmoussa1:
#### Scenario

A job held in PRIORITY and waiting to be allocated resources gets reprioritized and sent back to `job.state.priority` or `job.priority.get` _after_ a restart (clearing its aux items) when a plugin is already loaded and seen the job.

#### Problem

The multi-factor priority plugin raises an exception on a job in `job.state.priority` when it cannot find the aux item containing the association information for the job. However, when a system instance is reloaded, the aux items set on a job are cleared. When jobs are reprioritized and are sent back to `job.state.priority` or `job.priority.get`, the plugin will raise an exception on the job because it can't unpack the aux item containing the flux-accounting information associated with that job. See the below eventlog for an example:

```
...
{"timestamp":1737671575.5923743,"name":"priority","context":{"priority":19668}}
{"timestamp":1737736372.7067723,"name":"priority","context":{"priority":19620}}
{"timestamp":1737757970.5174642,"name":"priority","context":{"priority":19743}}
{"timestamp":1737801156.5343764,"name":"priority","context":{"priority":19865}}
{"timestamp":1737822769.2824628,"name":"priority","context":{"priority":19160}}
{"timestamp":1737998525.3946743,"name":"exception","context":{"type":"mf_priority","severity":0,"note":"internal error: bank info is missing","userid":767}}
{"timestamp":1737998525.3947248,"name":"clean"}
```

---

This PR removes raising a job exception in `job.state.priority` in the case where the job does not have an aux item for the accounting information associated with the job. Instead, it attempts to perform another lookup for the flux-accounting information for the association that submitted the job, which is already what it does when a job is submitted before the plugin is loaded with any flux-accounting information.

Fixes #575

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -544,18 +544,15 @@ static int priority_cb (flux_plugin_t *p,
                                                     FLUX_JOBTAP_CURRENT_JOB,
                                                     "mf_priority:bank_info"));
 
-    if (b == NULL) {
-        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
-                                     0, "internal error: bank info is missing");
-
-        return -1;
-    }
-
-    if (b->max_run_jobs == BANK_INFO_MISSING) {
-        // the association that this job is submitted under could not be found
-        // in the plugin's internal map when the job was first submitted and is
-        // held in PRIORITY by assigning special temporary values to the job;
-        // try to look up the association for this job again
+    if (b == nullptr || b->max_run_jobs == BANK_INFO_MISSING) {
+        /*
+         * The flux-accounting information associated with this job could not
+         * be found by the time this job got to job.state.priority. This could
+         * be due to Flux being restarted and pending jobs having their aux
+         * items reset or the plugin being reloaded and jobs being "re-seen"
+         * (see Flux RFC 21). Attempt to look up the association again and
+         * attach its information to the job with aux_set.
+         */
         Association *assoc = get_association (userid,
                                               bank,
                                               users,
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -47,6 +47,7 @@ TESTSCRIPTS = \
 	t1045-issue478.t \
 	t1046-issue565.t \
 	t1047-issue564.t \
+	t1048-issue575.t \
 	t5000-valgrind.t \
 	python/t1000-example.py \
 	python/t1001_db.py \
--- a/t/t1048-issue575.t
+++ b/t/t1048-issue575.t
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+test_description='test handling pending jobs sent back to PRIORITY after aux items cleared'
+
+. `dirname $0`/sharness.sh
+MULTI_FACTOR_PRIORITY=${FLUX_BUILD_DIR}/src/plugins/.libs/mf_priority.so
+SUBMIT_AS=${SHARNESS_TEST_SRCDIR}/scripts/submit_as.py
+DB_PATH=$(pwd)/FluxAccountingTest.db
+
+test_under_flux 1 job
+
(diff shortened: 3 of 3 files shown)


cmoussa1:
@grondo I believe I've found the source of those job failures I was telling you about a few days ago with the jobs that failed to load the `Association` aux item after a system instance restart (I wrote up a description with my thoughts in #575). I'm in the process of trying to add a sharness test for this case but I can't seem to figure out how to simulate this scenario. I've added a [WIP] commit to this PR to get me started, but I could use some advice on how to actually execute the restart and send the pending job back to PRIORITY. Do you have any advice or are there any examples that I can look at in flux-core for this?

grondo:
You can take a look at `t3200-instance-restart.t` which uses a series of `flux start` commands to simulate restarts.
O/w, you could try just reloading the job-manager module and restart the flux-accounting python service to see if that reproduces the original issue.

cmoussa1:
OK, I believe I have a working test file now to reproduce the behavior I was describing in #575, so I've dropped [WIP] from that commit.

src/plugins/mf_priority.cpp
```diff
@@ -544,18 +544,11 @@ static int priority_cb (flux_plugin_t *p,
...
-    if (b->max_run_jobs == BANK_INFO_MISSING) {
-        // the association that this job is submitted under could not be found
-        // in the plugin's internal map when the job was first submitted and is
-        // held in PRIORITY by assigning special temporary values to the job;
-        // try to look up the association for this job again
+    if (b == nullptr || b->max_run_jobs == BANK_INFO_MISSING) {
+        // The flux-accounting information associated with this job could not
+        // be found by the time this job got to job.state.priority. Attempt to
+        // look up the association again and attach its information to the job
+        // with aux_set
```
grondo, line 551: Suggestion: Amend the comment to indicate why the aux item may not be set or the bank info is missing. This may help a future developer. E.g. "This could be due to either ..."

grondo (approved):
This LGTM and seems to be a nice solution! Just one comment suggestion inline.

src/plugins/mf_priority.cpp
cmoussa1: Good suggestion, thanks! I've just force-pushed up an amendment to this comment as well as referenced RFC 21 for the job state diagram; the image in that RFC is super helpful.

cmoussa1:
Thanks for reviewing this @grondo! I'll set MWP here

PR #569, opened 2025-01-28
`edit-user`: make `fairshare` an editable field for an association

cmoussa1:
#### Problem

Since admins are the only ones who have privileges to run the `edit-user` command, they should be able to edit the `fairshare` column for an association manually if they so choose, but there is currently no way to do so via `edit-user`; they have to connect to a SQLite shell and create and execute a SQL statement to update the column manually.

---

This PR makes the `fairshare` column an optional argument in the `edit-user` command. I've updated the `fairshare` column in the `association_table` to match the other editable fields, as well as added a unit and sharness test.


jameshcorbett (approved):
👍

cmoussa1:
Thanks @jameshcorbett! Rebased to catch up after #570; setting MWP here

PR #572, opened 2025-01-30
`delete-user`: add `--force` option to actually remove a row from the `association_table`

cmoussa1:
#### Problem

The `delete-user` command does not actually delete a row from the `association_table` but instead sets the 'active' column for a row to 0. There might be a use case where an admin would actually want to remove a row from the table.

---

This PR adds a new `--force` optional argument to the `delete-user` command which will actually remove the row from the `association_table`. I've also added a function description to `delete_user()` to describe more fully what the function does (i.e. explaining disabling vs. actually deleting a user).

I've also added a couple of unit tests and sharness tests for calling `delete-user --force`.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -378,13 +378,29 @@ def add_user(
         raise sqlite3.IntegrityError()
 
 
-def delete_user(conn, username, bank):
-    cur = conn.cursor()
+def delete_user(conn, username, bank, force=False):
+    """
+    Deactivate a user row in the association_table by setting its 'active' status to 0.
+    If force=True, actually remove the user row from the association_table. If the
+    association belongs to multiple banks and the row being deactivated or removed is
+    the user's default bank, update other rows to set the new default bank to one that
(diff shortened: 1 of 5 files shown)


src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -151,7 +151,9 @@ def get_default_bank(cur, username):
     cur.execute(select_stmt, (username,))
     result = cur.fetchall()
 
-    return result[0][0]
+    if len(result) > 0:
```
jameshcorbett, line 154: I think `if result:` is the Pythonic way of making this conditional (because a collection is `False` if it doesn't have any elements) but I'm kind of ambivalent about it so...

jameshcorbett (approved):
Looks reasonable! One little comment

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: Thanks @jameshcorbett, that's a good suggestion! I just force-pushed to use `if result:`. Will set MWP here

PR #573, opened 2025-01-30
`delete-bank`: add `--force` option to actually remove a row from the `bank_table`

cmoussa1:
#### Problem

Similar to #572, the `delete-bank` command does not actually delete a row from the `bank_table` but instead sets the 'active' column for a row to 0. There might be a use case where an admin would actually want to remove a row from the table.

---

This PR adds a new `--force` optional argument to the `delete-bank` command which will actually remove the row from the `bank_table`. I've also added a function description to `delete_bank()` to describe more fully what the function does (i.e. explaining disabling vs. actually deleting a bank).

I've also added a couple of unit tests and sharness tests for calling `delete-bank --force`, both with and without sub-banks and users in those banks.

### DIFF
--- a/t/python/t1003_bank_cmds.py
+++ b/t/python/t1003_bank_cmds.py
@@ -153,6 +153,23 @@ class TestAccountingCLI(unittest.TestCase):
         rows = cur.fetchall()
         self.assertEqual(rows[0][0], 1)
 
+    # actually remove a bank row from the bank_table
+    def test_13_delete_bank_force(self):
+        b.delete_bank(acct_conn, bank="G", force=True)
+        cur.execute("SELECT * FROM bank_table WHERE bank='G'")
+        rows = cur.fetchall()
+        self.assertEqual(len(rows), 0)
+
+    # actually delete multiple banks from bank_table by deleting a parent bank
+    def test_14_delete_parent_bank_force(self):
+        b.delete_bank(acct_conn, bank="C", force=True)
+        cur.execute("SELECT * FROM bank_table WHERE bank='C'")
+        rows = cur.fetchall()
+        self.assertEqual(len(rows), 0)
+        cur.execute("SELECT * FROM bank_table WHERE parent_bank='C'")
+        rows = cur.fetchall()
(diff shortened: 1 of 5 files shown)


t/python/t1003_bank_cmds.py
```diff
@@ -153,6 +153,23 @@ def test_12_reactivate_bank(self):
...
+        rows = cur.fetchall()
+        self.assertEqual(len(rows), 0)
+
+    # actually delete multiple banks from bank_table by deleting a parent bank
+    def test_14_delete_parent_bank_force(self):
+        b.delete_bank(acct_conn, bank="C", force=True)
+        cur.execute("SELECT * FROM bank_table WHERE bank='C'")
+        rows = cur.fetchall()
+        self.assertEqual(len(rows), 0)
+        cur.execute("SELECT * FROM bank_table WHERE parent_bank='C'")
```
jameshcorbett, line 169: Why does deleting the parent bank delete the sub-banks? Is that something SQLite does automatically because you've defined a relationship between the two entries? Or is there something else going on that I've missed?

jameshcorbett (approved):
Looks good, one question.

t/python/t1003_bank_cmds.py
cmoussa1: Good question. The reason that sub-banks also get deleted when the parent bank gets deleted is because of the definition of the structure of the "flux-accounting" hierarchy. Basically, if a parent bank gets deleted, then the sub-bank now has no parent, and the hierarchy is invalid. As an example:

```
root
├── A
│   ├── user1
│   └── user2
├── B
│   ├── user3
│   └── user4
└── C
    |─ D
    |── user2
    └── user5
```

If parent bank C gets deleted, then D no longer has a parent bank unless we implement behavior to just move all of the following sub-banks up a level automatically... but when I was designing the DB I wasn't sure if I wanted to make that assumption and just leave the responsibility to the administrators

cmoussa1:
Thanks @jameshcorbett for the review! I'll set MWP here

PR #574, opened 2025-01-30
doc: add note about permanently deleting rows with `--force`, update `view-user` examples

cmoussa1:
##### problem 1

The `delete-user` and `delete-bank` commands now offer a `--force` option to permanently remove a row from the `association_table` or `bank_table`, but there is no mention of this in the flux-accounting documentation.

##### problem 2

The example output of the `view-user` command is out-of-date with the updates to the default and customizable formatting that `view-user` now offers.

---

This PR adds a note about permanently deleting rows in the flux-accounting DB as well as updates the example output to reflect the new changes to `view-user`.

This PR should not be landed until #573 is merged.


jameshcorbett (approved):
🥇

PR #579, opened 2025-02-06
python: use column names when accessing results of a query instead of row indices

### LINKED ISSUE #578: sqlite bindings: change query result handling by accessing columns by name instead of by index
Currently, a lot of the functions that retrieve data from the SQLite database use a raw SQL query. The results of this query are accessed using positional indexing (e.g., `row[0]`, `row[1]`, etc.), which requires prior knowledge of the column order. This approach is problematic because it makes the code harder to read and maintain, and if the database schema changes (e.g., column order is modified), the function could break or return incorrect values.

The SQLite `Connection` object has a `row_factory` property that can be set, which allows row values to be accessed by _column name_ instead of by index, so I can do something like `row["userid"]` instead of `row[1]`.

I should update the references throughout the bindings to use this property.

cmoussa1:
#### Problem

Currently, a lot of the functions that retrieve data from the SQLite database use a raw SQL query. The results of this query are accessed using positional indexing (e.g., `row[0]`, `row[1]`, etc.), which requires prior knowledge of the column order. This approach is problematic because it makes the code harder to read and maintain, and if the database schema changes (e.g., column order is modified), the function could break or return incorrect values.

---

This PR sets the `row_factory` property for the SQLite `Connection` object created in the flux-accounting service, the bulk-update script, the `formatter` module, and the Python unit tests. As a result, a number of functions are edited to access the results of a SQL query by column name instead of by row index.

I've broken this PR into multiple commits for clarity, but since they all do basically the same thing, I wouldn't mind squashing these into just one commit. I'm open either way.

Fixes #578

### DIFF
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -246,21 +246,26 @@ class BankFormatter(AccountingFormatter):
                 if users:
                     for user in users:
                         hierarchy += (
-                            f"{indent} {bank}|{user[0]}|{user[1]}|{user[2]}|{user[3]}\n"
+                            f"{indent} {bank}|{user['username']}|{user['shares']}|"
+                            f"{user['job_usage']}|{user['fairshare']}\n"
                         )
+
             else:
                 # continue traversing the hierarchy
                 for sub_bank in sub_banks:
-                    hierarchy += f"{indent} {str(sub_bank[0])}||{str(sub_bank[1])}|{str(sub_bank[2])}\n"
+                    hierarchy += (
+                        f"{indent} {str(sub_bank['bank'])}||{str(sub_bank['shares'])}|"
+                        f"{str(sub_bank['job_usage'])}\n"
+                    )
                     hierarchy = construct_parsable_hierarchy(
-                        cur, sub_bank[0], hierarchy, indent + " "
+                        cur, sub_bank["bank"], hierarchy, indent + " "
                     )
 
             return hierarchy
 
         # construct a hierarchy string starting with the bank passed in
         hierarchy = "Bank|Username|RawShares|RawUsage|Fairshare\n"
-        hierarchy += f"{self.rows[0][1]}||{str(self.rows[0][4])}|{str(round(self.rows[0][5], 2))}\n"
+        hierarchy += f"{self.rows[0]['bank']}||{str(self.rows[0]['shares'])}|{str(round(self.rows[0]['job_usage'], 2))}\n"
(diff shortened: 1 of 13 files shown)


src/bindings/python/fluxacct/accounting/formatter.py
```diff
@@ -246,21 +246,26 @@ def construct_parsable_hierarchy(cur, bank, hierarchy, indent=""):
...
-                        cur, sub_bank[0], hierarchy, indent + " "
+                        cur, sub_bank["bank"], hierarchy, indent + " "
                     )
 
             return hierarchy
 
         # construct a hierarchy string starting with the bank passed in
         hierarchy = "Bank|Username|RawShares|RawUsage|Fairshare\n"
-        hierarchy += f"{self.rows[0][1]}||{str(self.rows[0][4])}|{str(round(self.rows[0][5], 2))}\n"
+        hierarchy += f"{self.rows[0]['bank']}||{str(self.rows[0]['shares'])}|{str(round(self.rows[0]['job_usage'], 2))}\n"
```
jameshcorbett, line 268: Use double quotes for consistency?

jameshcorbett (approved):
amazing incredible 

but it doesn't look like the changes in commit 45bd1c88cc9d1db71959d6af6a2d597a5c1e799c are necessary? Like they aren't having any effect? But maybe I'm wrong.

cmoussa1:
Thanks @jameshcorbett! The changes aren't totally necessary (except for in `t1003_bank_cmd.py`), but for the sake of consistency of the `sqlite3.Connection` object across the flux-accounting project, I figured it would be nice to set that property everywhere. But I would be happy to drop the property from the other unit tests since they aren't really making a difference.

src/bindings/python/fluxacct/accounting/formatter.py
cmoussa1: I don't think I can use double quotes here, right? Because I already wrapped the f-string in double quotes. But I could switch those (i.e use single quotes for the f-string and double quotes for the keys in the curly braces) if you'd like? I'll defer to your preference.

jameshcorbett: Oh yeah you're right, I was totally missing that it was inside double quotes already.

jameshcorbett:
> But I would be happy to drop the property from the other unit tests since they aren't really making a difference.

Up to you, I think it's fine.

cmoussa1:
I'll go ahead and leave them there, then. For the sake of solidarity 😅 thanks! Setting MWP here

PR #581, opened 2025-02-11
`update-fshare`: change job usage variable from 32-bit `int` to `double`

cmoussa1:
#### Problem

The variable that stores job usage values for banks in the update-fshare script is a 32-bit `int`, and it cannot keep up with larger job usage values that are generated on big machines.

---

This PR changes the variable type of the job usage value from a 32-bit `int` to a 64-bit `double`. This variable probably should have been a `double` this whole time because the job usage value for an association or bank is not always guaranteed to be a whole number. 

I've also added a test that updates the fair-share value for an association and bank where the job usage value is gigantic.


grondo (approved):
LGTM!

cmoussa1:

PR #585, opened 2025-02-13
fair-share calculation: change `shares`, `usage` types to more appropriate types

cmoussa1:
#### Problem

The `shares` attribute for the `Account` class is of type `uint64_t`, but `data_reader_db` is extracting the shares value for an
association as a string and then casting it back to a `long long`. It would make more sense to just set this variable type as a `uint64_t` from the start.

The `usage` attribute used throughout the fair-share code is of type `uint64_t`, but it is of type `real` (i.e a floating-point value) in
the SQLite database. This means that the job usage values are getting truncated when they are being passed through to the actual weight --> fair-share calculation. It would be more precise to keep this `usage` value as type `double` throughout the fair-share calculation code.

---

This PR changes the variable type of the `shares` value to `uint64_t` to match its declared type in the **Account** class. It also changes the type of `usage` throughout the fair-share calculation code to be of type `double`.

You'll also notice that I've changed the comparison of `usage` to be a comparison to an epsilon value instead of a direct comparison to `0` since its type has changed to `double`.

### DIFF
--- a/src/fairness/reader/data_reader_db.cpp
+++ b/src/fairness/reader/data_reader_db.cpp
@@ -66,7 +66,7 @@ association's data and adds it to the weighted tree.
 */
 int data_reader_db_t::add_assoc (const std::string &username,
                                  uint64_t shrs,
-                                 const std::string &usg,
+                                 double usg,
                                  double fshare,
                                  std::shared_ptr<weighted_tree_node_t> &node)
 {
@@ -75,7 +75,7 @@ int data_reader_db_t::add_assoc (const std::string &username,
                                                              username,
                                                              true,
                                                              shrs,
-                                                             std::stoll (usg));
+                                                             usg);
     user_node->set_fshare (fshare);
     return node->add_child (user_node);
 }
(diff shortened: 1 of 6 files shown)


src/fairness/reader/data_reader_db.cpp
```diff
@@ -245,8 +245,7 @@ std::shared_ptr<weighted_tree_node_t> data_reader_db_t::get_sub_banks (
         while (rc == SQLITE_ROW) {
             std::string username = reinterpret_cast<char const *> (
                 sqlite3_column_text (c_assoc, 0));
-            std::string shrs = reinterpret_cast<char const *> (
-                sqlite3_column_text (c_assoc, 1));
+            uint64_t shrs = sqlite3_column_int64 (c_assoc, 1);
```
jameshcorbett, line 248: Is there no danger that the shares might grow big enough that the signed/unsigned conversion here becomes a problem?

jameshcorbett (approved):
Looks good, just one question about whether all 64 bits of `shares` might ever be used 🫨

src/fairness/reader/data_reader_db.cpp
jameshcorbett: Yeah that seems reasonable!

cmoussa1: Cool! I'll set MWP here. Thanks for the review

PR #586, opened 2025-02-19
python: create new `QueueFormatter` class, refactor `view_queue()` to use new class

cmoussa1:
#### Problem

The `view-queue` command defines its own formatting for the output, but flux-accounting now has a common `AccountingFormatter` class that is uniform across all of the `view-*` commands.

---

This PR creates a new `QueueFormatter` subclass to use for printing out queue information from the flux-accounting database. I've also added some basic sharness tests for testing the output of calling `view-queue`.


cmoussa1:
thanks @jameshcorbett!! setting MWP here

PR #588, opened 2025-02-19
cmd: add `list-queues` command

cmoussa1:
#### Problem

There is no way to concisely list a set of defined queues in the flux-accounting database, similar to the `list-banks` and `list-projects` commands.

---

This PR adds a new command to the command suite called `list-queues`, which will list all of the queues defined in the `queue_table`. It offers the same options as `list-banks` to customize the output.

I've also added a couple of basic tests that call `list-queues` both in JSON and in a table format with customization.


jameshcorbett (approved):
Nice!

cmoussa1:
thanks! setting MWP here

PR #590, opened 2025-02-26
plugin: return more specific error messages when updating bank attribute

cmoussa1:
#### Problem

The error message that `update_bank_cb ()` raises when it cannot find flux-accounting information for a particular association is vague and hard to debug, for there could be multiple reasons: the plugin could still be waiting on flux-accounting data to be loaded in OR there exists no flux-accounting data for that particular association.

---

This PR reworks the handling in `update_bank_cb ()` in the case where the plugin cannot find information for a particular association; specifically, it checks to see if the plugin has ANY flux-accounting data before returning an error, and customizes the error messages returned to be more specific depending on the scenario, and, in the second case, also includes the userid and the bank passed in.

I've edited a couple of existing tests that check for the old error message to account for the update as well as added a new sharness test that checks that both kinds of error messages are raised when updating the bank attribute after a plugin reload.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1130,12 +1130,26 @@ static int update_bank_cb (flux_plugin_t *p,
 
     // look up association
     a = get_association (userid, bank, users, users_def_bank);
-    if (a == nullptr)
-        return flux_jobtap_reject_job (p,
-                                       args,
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
Looks really nice!

cmoussa1:
Thanks @jameshcorbett! setting MWP here

PR #592, opened 2025-03-06
doc: add health/sanity checklist for flux-accounting

### LINKED ISSUE #591: documentation: add health/sanity checklist for flux-accounting
@wihobbs made a great suggestion offline yesterday that it would be helpful if some of the Flux projects had a checklist or some documentation surrounding how the sanity of the flux-* project is checked when it's deployed and running on a system. This would be a useful thing for flux-accounting particularly as more and more users get familiar with using flux-accounting. It would also be useful for me because sometimes I forget some steps on how _I_ check on how flux-accounting is running on a system.

Some things to include might be:

- [ ] checking the presence/configuration of the flux-accounting DB (association hierarchy)
- [ ] see if job records can be fetched
- [ ] check if priority plugin is loaded and bank/project attributes are being successfully recorded in the jobspec

cmoussa1:
#### Problem

flux-accounting contains a guide for setting up flux-accounting, but it does not contain any documentation for checking that the various components of flux-accounting are up, configured, and running correctly.

---

This PR adds a short "Health/Sanity Checklist" to flux-accounting that walks through checking the various components of the accounting system and making sure it is running as expected.

Fixes #591

### DIFF
--- a/doc/guide/sanity-checklist.rst
+++ b/doc/guide/sanity-checklist.rst
@@ -1,0 +1,101 @@
+.. _sanity-checklist:
+
+#######################################
+Flux Accounting Health/Sanity Checklist
+#######################################
+
+********
+Overview
+********
+
+As flux-accounting is deployed and running on a system, there are a number of
+components an administrator can check to ensure that the module is loaded and
+running as expected. The following attempts to list the major tasks an
+administrator can look to complete after building, installing, and configuring
+flux-accounting on a system.
+
+*****************************
+Database Integrity and Access
+*****************************
+
+Ensure users, banks, queues, and projects can be added, viewed, and edited with
+various ``flux account`` commands:
+
+.. code-block:: console
+
+    $ flux account add-bank --parent-bank=root giants 1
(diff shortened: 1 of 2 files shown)


doc/guide/sanity-checklist.rst
```diff
@@ -0,0 +1,98 @@
...
+
+Ensure users, banks, queues, and projects can be added, viewed, and edited with
+various ``flux account`` commands:
+
+.. code-block:: console
+
+    $ flux account add-bank --parent-bank=root giants 1
+    $ flux account add-user --username=crawford --bank=giants
+    $ flux account add-queue orange
+    $ flux account add-projects shortstops
```
wihobbs, line 29: Nice work! Maybe add a note that after the sanity check is complete, all of these commands work backwards with `flux account delete-*`?

wihobbs (approved):

cmoussa1:
Thanks @wihobbs! I added a small note about `delete-*` commands. Setting MWP here

PR #594, opened 2025-03-11
`apply_decay_factor ()`: fix iteration through usage periods, actually `commit` SQL statements

### LINKED ISSUE #593: `update-usage`: past usage factors not actually getting applied, not going all the way back to last usage period
While doing some random experimentation on `fluke`, I noticed that the past usage period job usage factors for associations was not actually getting updated. Only the most current period was being calculated. I think this is because the `apply_decay_factor()` function was not actually committing any of the `UPDATE` statements it was executing. I need to add a `.commit()` to the end of the `UPDATE`s.

I also noticed that the last usage period column was never being updated because of the way the past usage factors were being indexed in the `apply_decay_factor()` function. I need to take a look at how this is being indexed in the new half-life period scenario.

cmoussa1:
#### Problem

As noted in #593, the past usage factors for an association are not actually getting calculated and applied to associations and in the flux-accounting database because the `UPDATE` statements are not being `commit`ted in the `apply_decay_factor ()` function. The iteration through usage periods is also incorrect because the `for`-loop is starting at the wrong index; it *should* be starting at the first index and going to the second-to-last index instead of starting at the *second* index.

---

This PR changes the starting index for the `for`-loop in `apply_decay_factor()` to start at the *first* index instead of the second. It also adds a `commit()` statement near the end of the function to actually apply the `UPDATE` statements that are executed in the flux-accounting database. Finally, a unit test is adjusted to account for the change to how past usage factors are calculated and applied to the database.

Fixes #593

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
+++ b/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
@@ -117,10 +117,11 @@ def apply_decay_factor(decay, acct_conn, user=None, bank=None):
     for power, usage_factor in enumerate(usg_past, start=1):
         usg_past_decay.append(usage_factor * math.pow(decay, power))
 
-    # update job_usage_factor_table with new values, starting with period-2;
-    # the last usage factor in the table will get discarded after the update
+    # update job_usage_factor_table with new values, starting with the second usage
+    # period and working back to the oldest usage period since the most recent usage
+    # period is updated separately
     period = 1
-    for usage_factor in usg_past_decay[1 : len(usg_past_decay) - 1]:
+    for usage_factor in usg_past_decay[0:-1]:
         update_stmt = (
             "UPDATE job_usage_factor_table SET usage_factor_period_"
             + str(period)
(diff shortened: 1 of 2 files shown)


src/bindings/python/fluxacct/accounting/job_usage_calculation.py
```diff
@@ -120,7 +120,7 @@ def apply_decay_factor(decay, acct_conn, user=None, bank=None):
     # update job_usage_factor_table with new values, starting with period-2;
     # the last usage factor in the table will get discarded after the update
     period = 1
-    for usage_factor in usg_past_decay[1 : len(usg_past_decay) - 1]:
+    for usage_factor in usg_past_decay[0 : len(usg_past_decay) - 1]:
```
jameshcorbett, line 123: Is the comment above here about "starting with period-2" still correct after this change?

Also I think the line could be simplified to `usg_past_decay[0:-1]` since negative indices are counted from the end:

```
>>> x
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> x[0:len(x)-1]
[0, 1, 2, 3, 4, 5, 6, 7, 8]
>>> x[0:-1]
```

jameshcorbett (approved):
Looks good, one small comment and a question.

PR #595, opened 2025-03-17
`jj_read_vertex ()`: ignore unknown resources

cmoussa1:
#### Problem

As mentioned in flux-framework/flux-core#6710, the `jj_read_vertex ()` function returns an error if it encounters a resource type it does not recognize certain types like 'ssd'. Since flux-coral2 can create non-v1 jobspecs, flux-accounting needs to be able to handle these types of resources instead of raising an exception on the job.

---

This PR removes the `else`-condition from `jj_read_vertex ()` to raise an error when it encounters an unknown resource type, and instead just ignore unknown resources.

### DIFF
--- a/src/plugins/jj.cpp
+++ b/src/plugins/jj.cpp
@@ -56,11 +56,7 @@ static int jj_read_vertex (json_t *o, int level, struct jj_counts *jj)
         jj->slot_size = count;
     else if (streq (type, "gpu"))
         jj->slot_gpus = count;
-    else {
-        sprintf (jj->error, "Unsupported resource type '%s'", type);
-        errno = EINVAL;
-        return -1;
-    }
+    // ignore unknown resources
     if (with)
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM. I wonder if we can somehow add a test using the resources section flux-coral2 would provide...

jameshcorbett:
Yeah I want to add some tests to at least `t/t0022-jj-reader.t` in flux-core like Mark mentioned in https://github.com/flux-framework/flux-core/issues/6710. Because flux-coral2 both adds `ssd` entries to the jobspec and adds a second `slot` entry, which some core tools complain about.

cmoussa1:
That sounds like a great plan @jameshcorbett - can you tag me in those whenever you come up with those tests? They could definitely be added here.

cmoussa1:
Closing now that #603 has been landed.

PR #597, opened 2025-03-19
cmd: add new `list-users` command

### LINKED ISSUE #596: feature request: `flux account list-users`
I've recently encountered a couple of use cases where the ability to list all users in the Flux accounting database would be useful. The specific use cases are finding all users who have access to a specific queue in order to clean up after DATs and finding all users with a specific default bank after adding new banks or before retiring old ones. I can also imagine wanting to get al users with a given node limit, etc.

So, the first level of this request would be something like a `flux account list-users` command that lists all users. The second level would be to add flags that would display a subset of users. E.g. `flux account list-users --default-bank=guests` would list all users whose default bank is `guests`.

I could also see this as being a modification to the `view-user` command rather than a new command. I.e. make the `USERNAME` argument to `view-user` optional so that `flux account view-user` with no other arguments would list all users and `flux account view-user --default-bank=guests` would list all users whose default bank is `guests`.

cmoussa1:
#### Problem

As noted in #596, there is no way to list any of the rows in the `association_table`, similar to the `list-banks` command. It would also be nice if the table could be filtered by certain columns in the table, such as by queue, project, or limit.

---

This PR adds a new command to the flux-accounting command suite called `list-users`, which by default will list all of the rows in the `association_table` in a human-readable table format. An optional `--json` argument is also added to output the results in a JSON format:

```console
$ flux account list-users
username | bank | default_bank | shares
---------+------+--------------+-------
user1    | A    | A            | 1     
user2    | A    | A            | 1     
user2    | B    | A            | 1     
user3    | B    | B            | 100   
user4    | C    | C            | 100  
```

It accepts multiple optional arguments to filter the table by certain columns in the `association_table`.

```console
$ flux account list-users --shares=100
username | bank | default_bank | shares
---------+------+--------------+-------
user3    | B    | B            | 100   
user4    | C    | C            | 100
```

Its output can also be customized by column, e.g the person calling this command only wants to see the username of the association.

```console
$ flux account list-users --shares=100 --fields=username,bank
username | bank
---------+-----
user3    | B   
user4    | C   
```

Fixes #596

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -269,6 +269,61 @@ def view_user(conn, user, parsable=False, cols=None, list_banks=False):
         raise ValueError(f"view-user: {exc}")
 
 
+def list_users(conn, cols=None, json_fmt=False, **kwargs):
+    """
+    List all associations in the association_table in the flux-accounting DB. If
+    filters are passed in, limit the associations returned to the ones which fit
+    all filters.
+
+    Args:
+        conn: a SQLite connection object
+        cols: a list of columns from the table to include in the output. By default, all
+            columns are included.
+        filters: a list of optional constraints passed-in to filter the
+            association_table by.
+    """
+    # use all column names if none are passed in
+    cols = cols or fluxacct.accounting.ASSOCIATION_TABLE
+
+    # if any filters are passed in, make sure they are valid columns
+    table_filters = {key: val for key, val in kwargs.items() if val is not None}
+
+    try:
+        cur = conn.cursor()
+
+        sql.validate_columns(cols, fluxacct.accounting.ASSOCIATION_TABLE)
+        # construct SELECT statement
+        select_stmt = f"SELECT {', '.join(cols)} FROM association_table"
+        # filter by any constraints passed in
+        where_clauses = []
+        filters_list = []
+        for table_filter in table_filters:
+            if table_filter in ("queues", "projects"):
+                # we are filtering the table with a string; append wildcards ('%') to
+                # the string so we can match multiple cases (e.g the association belongs
+                # to more than one queue or project)
+                where_clauses.append(f"{table_filter} LIKE ?")
+                filters_list.append(f"%{table_filters[table_filter]}%")
+            else:
+                where_clauses.append(f"{table_filter} = ?")
+                filters_list.append(table_filters[f"{table_filter}"])
+
+        if where_clauses:
+            select_stmt += " WHERE " + " AND ".join(where_clauses)
+
+        cur.execute(select_stmt, tuple(filters_list))
+
+        # initialize AccountingFormatter object
+        formatter = fmt.AccountingFormatter(cur)
+        if json_fmt:
+            return formatter.as_json()
+        return formatter.as_table()
+    except sqlite3.Error as err:
+        raise sqlite3.Error(f"list-users: an sqlite3.Error occurred: {err}")
+    except ValueError as exc:
+        raise ValueError(f"list-users: {exc}")
+
+
 def add_user(
     conn,
     username,
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -95,6 +95,7 @@ class AccountingService:
             "view_project",
             "list_projects",
             "list_queues",
+            "list_users",
         ]
 
         privileged_endpoints = [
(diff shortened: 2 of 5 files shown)


grondo:
One quick comment is that the table output should probably be called "human readable" instead of "parseable", since it is the JSON output that is machine parsable. Unless I'm misunderstanding something. 

Also, for me I'd want a list users command that just dumps a list of users so it could be used in a shell loop, e.g.
```sh
for user in $(flux account list-users); do <something>; done
```
However, that's a minor point because you could presumably get that from the JSON output?

cmoussa1:
Oh that's a good thought! Maybe I'll rename the table output to `--table` or something?

I can also change the default behavior to output _just_ usernames and keep the ability to customize any additional columns as an optional argument? What do you think?

grondo:
We'll probably want @ryanday36's feedback, but what I'd expect from a `list-users` command is a list of all unique users by default, with the filtering options as Ryan suggested. I do like the ability to dump the info from the db by user as a table (for human consumption) or JSON (for use with tools like `jq`), so maybe this could be another subcommand (if there isn't one already) or the `--table` and `--json` options like you suggest.

One other suggestion is that it might be good to standardize the behavior of the `flux account` subcommands in this regard. For example, maybe all `list-*` commands just output a straight list of items, with `--table` and `--json` options to get extended output in table and json forms. 

Just a suggestion at this point though! Definitely up to you and @ryanday36!

ryanday36:
First off, this looks really good. I think that having a human-readable table output as the default and the option to output JSON for parsing in scripts would fit how I expect to use this best. If I'm just doing something quick and dirty I'd like to be able to visually inspect the output without adding flags or piping into `jq`, whereas if I'm consuming the output in a script I don't mind typing out extra flags to get exactly what I want. I'd also lean toward displaying all of the columns by default since I can get just the users (or other individual columns) pretty easily with that `--fields` flag.

For the future PR standardizing output behavior, it might be good to also think about making that `--fields` flag act like the output strings from `flux-core` commands (and maybe call it `-o/--format` too). I often find myself doing things like `<some Slurm command> | awk '{print $1}'` because the Slurm DB commands use different keywords and formats to specify their output than the commands that query the slurmctld and just piping into awk is an easier mental shortcut for me than remembering how a particular command needs its output flags formatted.

cmoussa1:
@ryanday36 this is super helpful, thanks. OK, why don't I go ahead and adjust this PR to the following:

* Make the default output a human-readable table with JSON output an optional argument. I'll keep the behavior to output all columns by default and keep the `--fields` flag as is.
* I'll open up an issue to track the standardization of the `list-*` (and perhaps the `view-*`?) commands. There might be some helpful discussion/ideas when thinking about using `-o/--format` for the flux-accounting commands (I already have this implemented for the `view-job-records` command).

cmoussa1:
Just force-pushed those fixes mentioned above.

src/cmd/flux-account-service.py
```diff
@@ -171,6 +172,36 @@ def view_user(self, handle, watcher, msg, arg):
...
+            )
+
+            payload = {"list_users": val}
+
+            handle.respond(msg, payload)
+        except KeyError as exc:
+            handle.respond_error(msg, 0, f"missing key in payload: {exc}")
+        except Exception as exc:
+            handle.respond_error(
+                msg, 0, f"a non-OSError exception was caught: {str(exc)}"
```
jameshcorbett, line 202: why non-`OSError`? Couldn't the exception caught have been an `OSError`?

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -269,6 +269,61 @@ def view_user(conn, user, parsable=False, cols=None, list_banks=False):
         raise ValueError(f"view-user: {exc}")
```
jameshcorbett, line 269: commit message is out of date, default is no longer for JSON and there is no `--parsable` option?

t/t1051-list-users.t
```diff
@@ -0,0 +1,150 @@
...
+	cat multiple_filters_users.json &&
+	test_debug "jq -S . <multiple_filters_users.json" &&
+	jq -e "length == 1" multiple_filters_users.json
+'
+
+# In the following set of tests, we customize the output to only include
+# certain columns in the output.
+test_expect_success 'customize JSON output of list-users' '
+	flux account list-users --json \
+		--bank=A --fields=username,bank,default_bank,shares
```
jameshcorbett, line 134: Should there maybe be a check that this is outputting the expected results?

jameshcorbett (approved):
Looks good, some small comments.

PR #600, opened 2025-03-25
`view-*`/`list-*` commands: add `-o/--format` optional argument

### LINKED ISSUE #598: `list-*`/`view-*` commands: consider standardizing, adding `-o/--format` optional args like flux-core
A while ago I began creating classes for the output format for all of the `view-*` and `list-*`, but they are still a little different in their own ways and it can be tedious to know how to use each one. I should really consider standardizing the output for all of these commands so that they are intuitive and easy to use across the board.

Better yet, adopting the `-o/--format` optional argument that flux-core offers (and the `view-job-records` command already offered by flux-accounting) might make these commands more in line with the rest of the Flux command suite.

cmoussa1:
#### Problem

There is no way to customize the output of a query to the flux-accounting DB using a Python format string. It would be nice if
flux-accounting offered a way to customize the output with a format string, similar to a lot of the commands offered by flux-core.

---

This PR adds a new method to the `AccountingFormatter` class called `as_format_string()`, which will return the results of a query to the flux-accounting DB as a formatted string. As a result, the `view-*` and `list-*` commands now offer an `-o/--format` optional argument, which allows the user to pass in a formatted string to customize their output:

```console
$ flux account view-user -o "{userid:<8} | {bank:<12}" user5011
userid   | bank        
5011     | A
```

I've also added some basic tests for the commands that now have this format string option.

Fixes #598

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -282,6 +282,7 @@ def list_banks(
     inactive=False,
     cols=None,
     table=False,
+    format_string="",
 ):
     """
     List all banks in bank_table.
@@ -293,6 +294,8 @@ def list_banks(
             columns are included.
         table: output data in bank_table in table format. By default, the format of any
             returned data is in JSON.
+        format_string: a format string defining how each row should be formatted. Column
+            names should be used as placeholders.
     """
     # use all column names if none are passed in
     cols = cols or fluxacct.accounting.BANK_TABLE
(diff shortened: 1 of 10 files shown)


cmoussa1:
@ryanday36 could you let me know if this looks like it would be a useful addition for the `list-*` and `view-*` commands? I know we had talked briefly about adding a `-o/--format` flag to these kinds of commands.

ryanday36:
yes! This looks very useful.

cmoussa1:
Thanks @ryanday36 and @jameshcorbett! Setting MWP here

PR #602, opened 2025-03-27
bindings: add `ProjectFormatter` subclass, `-o/--format` options to `view-project`/`list-projects`

cmoussa1:
#### Problem

The `view-project` and `list-projects` commands manually define their own formatting for their respective commands, which is different from the rest of the command suite, which are subclasses of a common `AccountingFormatter` class to standardize all output.

The `view-project` and `list-projects` commands also do not take an optional `-o/--format` argument to format the output using a Python format string, which is inconsistent with the rest of the flux-accounting command suite.

---

This PR adds a new `ProjectFormatter` subclass to the Python bindings so that the output of the `view-project` and `list-projects` commands are standard with the rest of the flux-accounting command suite. I've also added `--table` and `--fields` options to the `list-projects` command and a `--parsable` option to the `view-projects` command. Lastly, I've added `-o/--format` optional arguments to both commands.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/project_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/project_subcommands.py
@@ -44,7 +44,7 @@ def project_is_active(cur, project):
 ###############################################################
 
 
-def view_project(conn, project, parsable=False):
+def view_project(conn, project, parsable=False, format_string=None):
     try:
         cur = conn.cursor()
         # get the information pertaining to a project in the DB
@@ -52,6 +52,8 @@ def view_project(conn, project, parsable=False):
 
         formatter = fmt.ProjectFormatter(cur, project)
 
+        if format_string is not None:
+            return formatter.as_format_string(format_string)
         if parsable:
             return formatter.as_table()
         return formatter.as_json()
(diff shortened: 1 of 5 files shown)


src/bindings/python/fluxacct/accounting/project_subcommands.py
```diff
@@ -134,6 +136,8 @@ def list_projects(conn, cols=None, table=False):
 
         # initialize AccountingFormatter object
         formatter = fmt.AccountingFormatter(cur)
+        if format_string != "":
```
jameshcorbett, line 139: why not make `format_string` default to `None`? And then have this check be `if format_string is not None`

jameshcorbett (approved):
Looks good, some suggestions about argument defaults

src/cmd/flux-account.py
```diff
@@ -738,6 +745,19 @@ def add_list_projects_arg(subparsers):
     )
 
     subparser_list_projects.set_defaults(func="list_projects")
+    subparser_list_projects.add_argument(
+        "--fields",
+        type=str,
```
jameshcorbett, line 756: you could remove this `type=str` but it's insignificant. Your call. Otherwise 👍

jameshcorbett (approved):

PR #603, opened 2025-03-28
mf_priority: update jj library from flux-core

jameshcorbett:
Problem: the `jj` library from flux-core was recently updated (https://github.com/flux-framework/flux-core/pull/6727) to handle the non-v1 jobspecs produced by flux-coral2.

Copy over the new version from flux-core.

### DIFF
--- a/src/plugins/jj.cpp
+++ b/src/plugins/jj.cpp
@@ -19,9 +19,9 @@ extern "C" {
 
 #include "jj.hpp"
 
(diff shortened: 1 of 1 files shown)


cmoussa1 (approved):
LGTM! Thanks! I should probably close #595 now that this is proposed.

jameshcorbett:
I completely forgot that was a thing

PR #605, opened 2025-04-01
plugin: clean up error handling, comments in `job.state.depend` 

cmoussa1:
#### Problem

The `dependency_add ()` calls in the callback for `job.state.depend` are a little tough to read in the case that an exception needs to be raised because they are broken up over multiple lines to follow the 80 character-per-line limit. As more dependencies are added to this callback, this function will get even longer and perhaps more tedious to read through.

---

This PR adds an `error` label to the `job.state.depend` callback that is called when adding a dependency fails. It moves the `flux_jobtap_raise_exception ()` call to this label and sets the name of the dependencies before going there to include it in the exception message.

I've also moved a comment to below one of the `if`-statements as well as removed an errant `return 0` from one of the dependency checks in `job.state.depend`.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -927,9 +927,9 @@ static int depend_cb (flux_plugin_t *p,
         }
     }
 
-    // if user has already hit their max running jobs count, add a job
-    // dependency to hold job until an already running job has finished
     if ((b->max_run_jobs > 0) && (b->cur_run_jobs == b->max_run_jobs)) {
+        // association is already at their max running jobs count; add a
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
Another amazing PR from the greatest coder and baseball coach I know. However I wonder if there should be a test for a job being in both dependencies?

grondo (approved):
LGTM!

cmoussa1:
Thanks to you both for the review! I agree that a test should be added to cover a job with both dependencies added to it. I've just force-pushed a commit to add a set of tests that walks through the scenario to cover a job hitting both dependencies at the same time. Will set MWP here

PR #606, opened 2025-04-01
plugin: create label for `dependency_remove < 0`

cmoussa1:
#### Problem

Continuing the cleanup started in #605, the `dependency_remove ()` calls in the callback for `job.state.inactive` are a little tough to read in the case that an exception needs to be raised because they are broken up over multiple lines. As more dependencies are added to this callback, this function will get even longer and perhaps more tedious to read through.

---

This PR just adds an `error` label to the `job.state.inactive` callback that is called when removing a dependency fails. It moves the `flux_jobtap_raise_exception ()` call to this label and sets the name of the dependencies before going to this label to include it in the exception message.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1226,6 +1226,7 @@ static int inactive_cb (flux_plugin_t *p,
     Association *b;
     json_t *jobspec = NULL;
     char *queue = NULL;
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
Nice!

cmoussa1:
thanks! setting MWP here

PR #608, opened 2025-04-08
plugin: add a new `Job` class 

cmoussa1:
#### Problem

As more and more flux-accounting dependencies are added to the priority plugin, it will be harder to keep track of certain attributes about jobs (and subsequently track them against an association's flux-accounting limits) without having to unpack those attributes in multiple callbacks. This gets cumbersome and makes tracking an association's jobs and limits very tedious.

---

This PR attempts to begin to alleviate this by adding a new `Job` class to be compiled with the rest of the plugin. A `Job` object stores things like the job's ID, counted resources, and any flux-accounting dependencies added to the job. Helper functions to check a job's resources and list of dependencies are also added to this class.

The goal with this class to eventually convert the `job.state.depend` and `job.state.inactive` callbacks to use this new class and its functions to clean up how dependencies and limits are checked, added, and removed from an association's set of running and active jobs (which I have a working implementation of, but will hold off for a subsequent PR barring this one's approval).

I've added basic unit tests for the `Job` class, its members, and checking and removing certain dependencies from a `Job` object.

### DIFF
--- a/src/plugins/job.hpp
+++ b/src/plugins/job.hpp
@@ -1,0 +1,60 @@
+/************************************************************\
+ * Copyright 2025 Lawrence Livermore National Security, LLC
+ * (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+ *
+ * This file is part of the Flux resource manager framework.
+ * For details, see https://github.com/flux-framework.
+ *
+ * SPDX-License-Identifier: LGPL-3.0
+\************************************************************/
+
+// header file for the Job class
+extern "C" {
+#if HAVE_CONFIG_H
+#include "config.h"
+#endif
+#include <flux/core.h>
+#include <flux/jobtap.h>
+#include <jansson.h>
+}
+
+#ifndef JOB_H
+#define JOB_H
+
+#include <vector>
+#include <string>
+#include <map>
+#include <iterator>
+#include <sstream>
+#include <algorithm>
+
+// custom job resource counting file
+#include "jj.hpp"
+
+class Job {
+public:
+    // attributes
+    flux_jobid_t id = 0;               // the ID of the job
+    std::vector<std::string> deps; // any dependencies on job
+    int nnodes = 0;                // the number of nodes requested
+    int ncores = 0;                // the number of cores requested
+    std::string queue;             // the queue the job was submitted under
+
+    // constructor
+    Job () = default;
+
+    // methods
+    // count the resources requested for a job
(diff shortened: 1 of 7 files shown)


src/plugins/job.hpp
```diff
@@ -0,0 +1,56 @@
...
+};
+
+// count the resources requested for a job
+int count_resources (Job &job, json_t *jobspec);
+
+// determine if a job contains a certain dependency
+bool contains_dep (const Job &job, const std::string &dep);
+
+// remove a job dependency from a job's list of dependencies
+void remove_dep (Job &job, const std::string &dep);
```
grondo, line 54: I'm not the greatest C++ developer, but could these functions should be public members of the Job class, instead of functions that take a `Job` object?

Also, should there be an `add_dep()` method as well as `remove_dep()`? Then perhaps the `deps` vector could be private.

jameshcorbett (approved):
Unacceptable garbage I love it

src/plugins/job.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <sstream>
+#include <algorithm>
+
+// custom job resource counting file
+#include "jj.hpp"
+
+class Job {
+public:
+    // attributes
+    long int id = 0;               // the ID of the job
```
grondo, line 37: Should this be of type `flux_jobid_t`?

grondo (commented):
Just peeked at this and had a couple high level comments. More questions, really.
Feel free to ignore if not applicable.

jameshcorbett:
Ohhhhh a real developer has commented with real suggestions. I second his suggestions. Obviously the only reason I didn't make them myself was that I was too lazy

src/plugins/job.hpp
cmoussa1: Those are great suggestions, thanks! I'll go ahead and change these just to be members of the `Job` class. I'll also go ahead and add an `add_dep ()` method too.

I think there might be a use case or two where the plugin might benefit from being able to access both the size of the `deps` vector as well as seeing the order in which dependencies are placed onto this vector. Would it be okay if I left this public?

grondo: Oh yeah, that makes sense!

cmoussa1:
Thanks @jameshcorbett and @grondo for the reviews! I've force-pushed up some changes based on the feedback above to make those functions methods of the `Job` class and changed the type of `id` to `flux_jobid_t`. Will set MWP here

PR #609, opened 2025-04-09
plugin: improve `Queue` class

cmoussa1:
#### Problem

The `Queue` class definition is below the `Association` class definition, but there will eventually be a need to pass a `Queue` object to method(s) in the `Association` class. There is also no default initialization for the attributes of a `Queue` object. Additionally, the `Queue` class has no attribute for the name of the queue, which would be nice to have.

---

This PR moves some macro definitions as well as the `Queue` class definition to the top of the `accounting.hpp` file. It also adds a `name` attribute to the `Queue` class as well as add some default values to the existing attributes. As a result, I've adjusted the callback in the plugin that unpacks queue information from the flux-accounting DB as well as some unit tests to also include the name of the queue.

I've also added a couple of unit tests to `accounting_test01` to simulate comparing an Association's current running jobs count in a queue to a queue's max running jobs limit.

Like #608, this is laying the groundwork for improving the way flux-accounting dependencies are checked and handled in the priority plugin.

### DIFF
--- a/src/plugins/accounting.hpp
+++ b/src/plugins/accounting.hpp
@@ -28,8 +28,36 @@ extern "C" {
 #include <sstream>
 #include <algorithm>
 #include <unordered_map>
+#include <cmath>
 
 #include "job.hpp"
+
+// - UNKNOWN_QUEUE: a queue is specified for a submitted job that flux-accounting
+// does not know about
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
Nice

cmoussa1:
Thank you good sir. Setting MWP here

PR #610, opened 2025-04-10
plugin: improve how dependency logic works with new `Job` class

cmoussa1:
#### Problem

The priority plugin has to manually extract and check info about both the job and the association that is submitting the job for
any potential flux-accounting limits in both `job.state.depend` and `job.state.inactive`. As more and more dependencies are added to keep track of in flux-accounting, it will get more cumbersome to have to do these checks in the callbacks without the use of a `Job` object that can store this information across multiple callbacks.

---

This PR changes the how `job.state.depend` and `job.state.inactive` callbacks handle adding & removing dependencies from jobs as they encounter various (and future) flux-accounting limits. It condenses the need to store held jobs _both_ in a vector of just job IDs and a map that stores which job IDs are stored with which queue into the `held_jobs` vector, which instead of being a vector of just job IDs, is now a vector of class `Job`, which contains attributes such as its job IDs, the queue it was submitted in, counted resources, and any flux-accounting dependencies associated with the job.

The `job.state.depend` callback now pushes back any and all dependencies potentially added to a Job's `deps` vector. In `job.state.inactive` all held jobs for an association are iterated through to see if they satisfy any and all requirements to have their dependencies removed - for now, there are just two: the per-queue max-running-jobs limit and the per-association max-running-jobs limit. But, as more flux-accounting-specific dependencies are removed, these can be added to the callback (or potentially abstracted into one or more helper functions).

I've also restructured the `held_jobs` member of each association's set of information returned with `flux jobtap query` to include the entire `Job` object instead of just the job ID.

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -252,14 +252,3 @@ int get_project_info (const char *project,
 
     return 0;
 }
-
-
-int max_run_jobs_per_queue (const std::map<std::string, Queue> &queues,
-                            const std::string &queue)
-{
-    auto it = queues.find (queue);
-    if (it == queues.end ())
-        return -1;
-
-    return it->second.max_running_jobs;
-}
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -39,146 +39,119 @@ Association* get_association (int userid,
 
 json_t* Association::to_json () const
 {
-    json_t *user_queues = json_array ();
-    if (!user_queues) {
-        return nullptr;
-    }
+    json_t *user_queues = nullptr;
+    json_t *user_projects = nullptr;
+    json_t *temp = nullptr;
+    json_t *queue_usage_json = nullptr;
+    json_t *hj_json = nullptr;
+    json_t *job_json = nullptr;
+    json_t *deps_array = nullptr;
+    json_t *dep_str = nullptr;
+    json_t *u = nullptr;
+
+    user_queues = json_array ();
+    if (!user_queues)
+        goto error;
     for (const auto &queue : queues) {
-        json_t *temp;
         if (!(temp = json_string (queue.c_str ()))
-            || json_array_append_new (user_queues, temp) < 0) {
-            json_decref (user_queues);
-            return nullptr;
-        }
+            || json_array_append_new (user_queues, temp) < 0)
+            goto error;
     }
+    // set temp to nullptr here to avoid a double free in case of an error
+    temp = nullptr;
 
-    json_t *user_projects = json_array ();
-    if (!user_projects) {
-        json_decref (user_queues);
-        return nullptr;
-    }
+    user_projects = json_array ();
+    if (!user_projects)
+        goto error;
     for (const auto &project : projects) {
-        json_t *temp;
         if (!(temp = json_string (project.c_str ()))
-            || json_array_append_new (user_projects, temp) < 0) {
-            json_decref (user_queues);
-            json_decref (user_projects);
-            return nullptr;
-        }
+            || json_array_append_new (user_projects, temp) < 0)
+            goto error;
     }
+    temp = nullptr;
 
-    json_t *queue_usage_json = json_object ();
-    if (!queue_usage_json) {
-        json_decref (user_queues);
-        json_decref (user_projects);
-        return nullptr;
-    }
+    queue_usage_json = json_object ();
+    if (!queue_usage_json)
+        goto error;
     for (const auto &entry : queue_usage) {
         if (json_object_set_new (queue_usage_json,
                                  entry.first.c_str (),
-                                 json_integer (entry.second)) < 0) {
-            json_decref (user_queues);
-            json_decref (user_projects);
-            json_decref (queue_usage_json);
-            return nullptr;
-        }
+                                 json_integer (entry.second)) < 0)
+            goto error;
     }
 
-    json_t *hj_json = json_object ();
-    if (!hj_json) {
-        json_decref (user_queues);
-        json_decref (user_projects);
-        json_decref (queue_usage_json);
-        return nullptr;
-    }
+    hj_json = json_object ();
+    if (!hj_json)
+        goto error;
 
     for (const auto &entry : held_jobs) {
         const Job &job = entry;
-        json_t *job_json = json_pack ("{s:i, s:i, s:s, s:o}",
-                                      "nnodes", job.nnodes,
-                                      "ncores", job.ncores,
-                                      "queue", job.queue.c_str (),
-                                      "deps", json_array ());
+        job_json = json_pack ("{s:i, s:i, s:s, s:o}",
+                              "nnodes", job.nnodes,
+                              "ncores", job.ncores,
+                              "queue", job.queue.c_str (),
+                              "deps", json_array ());
 
-        if (!job_json) {
-            json_decref (user_queues);
-            json_decref (user_projects);
-            json_decref (queue_usage_json);
-            json_decref (hj_json);
-            return nullptr;
-        }
+        if (!job_json)
+            goto error;
 
-        json_t *deps_array = json_array ();
-        if (!deps_array) {
-            json_decref (user_queues);
-            json_decref (user_projects);
-            json_decref (queue_usage_json);
-            json_decref (hj_json);
-            json_decref (job_json);
-            return nullptr;
-        }
+        deps_array = json_array ();
+        if (!deps_array)
+            goto error;
 
         for (const auto &dep : job.deps) {
-            json_t *dep_str = json_string (dep.c_str ());
-            if (!dep_str || json_array_append_new (deps_array, dep_str) < 0) {
-                json_decref (user_queues);
-                json_decref (user_projects);
-                json_decref (queue_usage_json);
-                json_decref (hj_json);
-                json_decref (deps_array);
-                json_decref (job_json);
-                return nullptr;
-            }
+            dep_str = json_string (dep.c_str ());
+            if (!dep_str || json_array_append_new (deps_array, dep_str) < 0)
+                goto error;
         }
 
-        if (json_object_set_new (job_json, "deps", deps_array) < 0) {
-            json_decref (user_queues);
-            json_decref (user_projects);
-            json_decref (queue_usage_json);
-            json_decref (hj_json);
-            json_decref (deps_array);
-            json_decref (job_json);
-            return nullptr;
-        }
+        if (json_object_set_new (job_json, "deps", deps_array) < 0)
+            goto error;
 
         // use job id (flux_jobid_t) as string key
         char keybuf[32];
         snprintf (keybuf, sizeof(keybuf), "%ld", entry.id);
-        if (json_object_set_new (hj_json, keybuf, job_json) < 0) {
-            json_decref (user_queues);
-            json_decref (user_projects);
-            json_decref (queue_usage_json);
-            json_decref (hj_json);
-            return nullptr;
-        }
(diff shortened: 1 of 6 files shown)


cmoussa1:
One of the things I think some feedback would be especially helpful for would be how `job.state.inactive` iterates through jobs as they complete. You'll notice that I use a `while` loop here to iterate through an association's inactive jobs (if there is more than one  held job) in the case that maybe the first job does not meet all requirements to be released by the plugin.

I think that with this logic, _multiple_ jobs could potentially meet all requirements to be released by the plugin, which could make sense, e.g. in the future when resource limits are enforced and one really big job from an association completes and two small ones can now run. However, I don't know if the current implementation could introduce a race condition if a completed job finishes and potentially releases more than one job by the time jobs are released and running job counters are incremented.

src/plugins/accounting.cpp
```diff
@@ -103,10 +103,71 @@ json_t* Association::to_json () const
...
+        json_object_set_new (job_json, "deps", deps_array);
+
+        // use job id (flux_jobid_t) as string key
+        char keybuf[32];
+        snprintf (keybuf, sizeof(keybuf), "%ld", entry.first);
+        if (json_object_set_new (hj_json, keybuf, job_json) < 0) {
+            json_decref (user_queues);
+            json_decref (user_projects);
+            json_decref (queue_usage_json);
+            json_decref (hj_json);
```
jameshcorbett, line 143: Should decref `job_json` here too? Because if this function returns -1, I think you can assume the reference wasn't stolen.

src/plugins/accounting.cpp
```diff
@@ -103,10 +103,71 @@ json_t* Association::to_json () const
...
+                                      "queue", job.queue.c_str (),
+                                      "deps", json_array ());
+
+        if (!job_json) {
+            json_decref (user_queues);
+            json_decref (user_projects);
+            json_decref (queue_usage_json);
+            json_decref (hj_json);
+            return nullptr;
+        }
```
jameshcorbett, line 109: I think you could reduce a bunch of these duplicated `json_decref` blocks if you put them all at the end of the function in a `goto error` type thing. I think you would need to initialize all of the variables to `nullptr` at the beginning of the function though.

```C
if (!job_json) {
    goto error;
}

...

error:
    json_decref (user_queues);
    json_decref (user_projects);
    ...
    return nullptr;
```

jameshcorbett (approved):
Looks reasonable. I had a hard time making sense of the second commit but that might just be me being slow today. Maybe @grondo should take a look, especially because I don't know how to respond to your comment above about the `while` loop and `job.state.inactive`. I had some minor comments on the other commits.

src/plugins/accounting.cpp
cmoussa1: These are great suggestions, thanks. I kind of hate the duplicated code to `json_decref` every object as the function progresses, so just creating an `error` label makes more sense. Should I lump this fix in with the same commit that adds the new object to this method? Or make it a separate commit?

cmoussa1: I'm leaning towards making this just a separate commit if that is okay.

jameshcorbett: yeah that's fine!

cmoussa1:
> How does `std::map` keep track of the order in which jobs are held? The [docs](https://cplusplus.com/reference/map/map/begin/) state that "begin points to the element that goes first following the container's [sorting criterion](https://cplusplus.com/map::key_comp)."

I _think_ by default, this is sorted using a `<` operator, so smaller job IDs will typically be held first. Now that I've been giving this more thought, though, I'm not sure why I can't just change the `held_jobs` vector to a vector of `Job` objects instead of adding a new `map` attribute. By using a vector, I'm pretty sure we can still preserve the order in which jobs are held without needing to sort (since a `Job` object will store its job ID as well). Thanks for pointing this out! I can restructure the PR slightly to just change `held_jobs` to a vector of `Job`s instead of introducing a new `hj` map? What do you think?

> `held_jobs` is not that much longer than `hj`, but is more searchable and obvious what `b->held_jobs` is. Just a suggestion to perhaps use the longer name in this instance. (You may have a good reason for the short name, though, so feel free to ignore the suggestion :-))

I agree with you here, and if I stick with the slight restructure I mentioned above, then I can just keep the `held_jobs` name.

> In the future, it would be really nice if there were unit tests for the management of these dependencies on Job and Association objects. One way to do that might be to abstract all the code that manages the dependency tracking into a set of interfaces that don't require the libflux/jobtap API. This could be kept separate from the plugin and have its own unit tests that simulate jobs being submitted from different users/banks with different limits, and ensure the code is working correctly, no memory errors etc. It is also much easier to review future changes to interfaces like that because they have been designed to be testable and we can see from the tests how they are meant to operate. Just a thought!

This is a great suggestion! I started writing some [unit tests](https://github.com/flux-framework/flux-accounting/blob/master/src/plugins/test/job_test02.cpp) for the `Job` class that were added in #608, but they are not anything super detailed, and adding some unit tests that involve multiple associations with different limits is definitely a good idea. I could look into adding some more unit tests here. In terms of abstracting the code that manages dependency tracking, are you referring to the various `if` checks in both `job.state.depend` and `job.state.inactive` that check each limit?

grondo:
> are you referring to the various if checks in both job.state.depend and job.state.inactive that check each limit?

I was mainly noticing that there's a quite large and complex chunk of code added directly to the `inactive_cb` in the plugin. My first inclination would be to move that code to a function. Unfortunately, I'd need to study this and other code to make a more concrete suggestion, but it seems like if you could abstract the functionality of this and related code into a class or function that could then have unit tests exercising various scenarios, it would give us a lot of confidence that the code is correct, especially when making future changes.

cmoussa1:
OK, that seems like a reasonable suggestion. I'll go ahead and push up the revisions based on the other suggestions you've made in the meantime while I think about how I might abstract the dependency checks in `job.state.inactive`.

cmoussa1:
One thought that comes to mind after playing around with it for a couple hours is potentially abstracting the various dependency checks into methods in the `Association` class that do not depend on calling anything in the flux jobtap API:

```c++
bool Association::under_max_run_jobs ()
{
    bool under_assoc_max_run_jobs = cur_run_jobs < max_run_jobs;

    return under_assoc_max_run_jobs;
}

bool Association::under_queue_max_run_jobs (
                                const std::string &queue,
                                const std::map<std::string, Queue> queues) {
    bool under_queue_max_run_jobs = queue_usage.at (queue) <
                                        queues.at (queue).max_running_jobs;

    return under_queue_max_run_jobs;
}
```

I could add some unit tests that could add `Job` objects to an association's list of held jobs without needing to actually submit them and exercise determining whether a job meets none, one, or all requirements.

cmoussa1:
Just pushed up some [WIP] commits so that I can pull these on my other machine tomorrow and continue working on them there

cmoussa1:
OK, I've pushed up some new commits and fixes to existing commits to this PR that do the following:

- I've added methods to the `Association` class that do the actual comparison of flux-accounting limits, i.e. the comparison between an association's current running jobs count and their `max_run_jobs` count as well as the comparison between an association's current running jobs count in a specific queue and the queue's `max_run_jobs` count.
- These methods are used in a set of newly-created unit tests (`dependencies_test03.cpp`) that simulates holding an association's jobs due to one or both limits without actually submitting the job (i.e. manually setting the association's current and max-running jobs counts) and calling those new methods and the `Job` methods to add/check/remove dependencies.
- As a result, the `job.state.depend` and `job.state.inactive` callbacks in the plugin are slightly reworked to use these new methods in the `Association` class. I'm not sure if the check in `job.state.inactive` is made any clearer, and I'm wondering if there's room for even more abstraction here. I'm wondering how I might be able to do this in the plugin while also ensuring that `flux_jobtap_dependency_remove ()` is called appropriately.

The rest of the commits (the ones that @jameshcorbett helped review) are basically the same.

cmoussa1:
Just to move more code directly out of the `job.state.inactive` callback even further, I've pushed up a commit (`"plugin: abstract loop for association's held jobs"`) to its own helper function. Now, the callback contains a check like this:

```c++
if (!b->held_jobs.empty ()) {
    // the Association has at least one held Job; begin looping through
    // held Jobs and see if they satisfy the requirements to be released
    if (check_and_release_held_jobs (p, b) < 0)
        goto error;
}
```

which hopefully looks more readable and provides better context to what is actually happening in this section of the callback than before.

src/plugins/test/accounting_test01.cpp
```diff
@@ -73,10 +72,10 @@ void initialize_map (
 {
     Association user1 = {"bank_A", 0.5, 5, 0, 7, 0, {},
                          {}, 0, 1, {"*"}, "*", 2147483647, 2147483647, 0, 0,
-                         {}, {}};
+                         {}};
     Association user2 = {"bank_A", 0.5, 5, 0, 7, 0, {},
                          {}, 0, 1, {"*"}, "*", 2147483647, 2147483647, 0, 0,
-                         {}, {}};
+                         {}};
```
grondo, line 79: Are these whitespace changes in the right commit?

src/plugins/accounting.cpp
```diff
@@ -39,146 +39,115 @@ Association* get_association (int userid,
...
+
+error:
+    json_decref (user_queues);
+    json_decref (user_projects);
+    json_decref (queue_usage_json);
+    json_decref (hj_json);
+    json_decref (deps_array);
+    json_decref (job_json);
+    json_decref (u);
+    return nullptr;
```
grondo, line 150: I was having trouble following the diff here, so I could be wrong, but does a `json_decref (temp);` need to be added in this error block? (Basically any local variable that is set to `nullptr` above should have a corresponding `decref` here?

If you add that, be sure to set `temp = NULL` after the reference for `temp` is taken by `json_array_append_new()`

grondo (approved):
LGTM and nice cleanup! Just a couple minor comments.

src/plugins/test/accounting_test01.cpp
cmoussa1: Those changes are actually removing an empty map from each `Association` object because its corresponding member in the Association object (`queue_held_jobs`) was also removed in this commit.

src/plugins/accounting.cpp
cmoussa1: Oh, yes, good catch. Thank you!

> If you add that, be sure to set `temp = NULL` after the reference for `temp` is taken by `json_array_append_new()`

Do I need to set `temp = NULL` in each iteration of both `for`-loops? Or after each for-loop? i.e.:

```diff
    user_queues = json_array ();
    if (!user_queues)
        goto error;
    for (const auto &queue : queues) {
        if (!(temp = json_string (queue.c_str ()))
            || json_array_append_new (user_queues, temp) < 0)
            goto error;
    }
+   temp = nullptr;

    user_projects = json_array ();
    if (!user_projects)
        goto error;
    for (const auto &project : projects) {
        if (!(temp = json_string (project.c_str ()))
            || json_array_append_new (user_projects, temp) < 0)
            goto error;
    }
+   temp = nullptr;
```

src/plugins/test/accounting_test01.cpp
grondo: Oh, duh, I see that now. I misread the diff :facepalm:

src/plugins/accounting.cpp
grondo: I think so. Otherwise any call to `goto error` after the ownership of `temp` is passed to the array, you'll have a double free.

cmoussa1:
Thanks @grondo for the review and advice here! I've just force-pushed up some fixes based on the above suggestions. Will set MWP here

grondo:
Nice work @cmoussa1 !

PR #611, opened 2025-04-14
view-* man pages: add -o/--format optional arg

cmoussa1:
#### Problem

The `view-bank` and `view-user` man pages do not list the `-o/--format` optional argument in the command descriptions which was added in #600.

---

This PR adds the `-o/--format` optional argument to both the `view-bank` and `view-user` man pages.


wihobbs (approved):
🍝

PR #612, opened 2025-04-14
`delete-queue`: add warning statement when deleting a queue still referenced by associations

cmoussa1:
#### Problem

A queue can be deleted from the `queue_table` but still be referenced by associations in the `association_table`, i.e. an
association can still list that queue in its `queues` column. An admin who calls the `delete-queue` command should be warned that they should also edit any association rows in the `association_table` to account for deleting a queue.

---

This PR adds a check in `delete_queue()` to see if any associations have the queue being deleted still referenced in their `queues` attribute. If so, it issues a warning statement letting the admin know they should also edit any and all associations' rows to account for deleting a queue.

I've also added a basic test to make sure that the warning statement gets issued when removing a queue still referenced by an association.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/queue_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/queue_subcommands.py
@@ -70,11 +70,29 @@ def add_queue(
 
 
 def delete_queue(conn, queue):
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
💯

cmoussa1:
Thanks! Setting MWP here

PR #613, opened 2025-04-14
doc: add `man(1)` pages for `*-queue` commands

cmoussa1:
#### Problem

flux-accounting does not have any `man(1)` pages for the commands involving queue administration.

---

This PR adds `man(1)` pages for the `*-queue` and `list-queues` commands.


doc/man1/flux-account-add-queue.rst
```diff
@@ -0,0 +1,40 @@
...
+
+**flux** **account** **add-queue** QUEUE [OPTIONS]
+
+DESCRIPTION
+===========
+
+.. program:: flux account add-queue
+
+:program:`flux account add-queue` will add a queue to the ``queue_table`` in
+the flux-accounting database. Different properties and limits can be set for
```
wihobbs, line 19: Somewhere (but I'm having trouble thinking of where it makes sense) I would like to see something along the lines of 

> Note that flux-accounting queues are for job accounting purposes, and a separate, disparate construct from [queues in flux-core](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man5/flux-config-queues.html) and [queue policies in flux-sched](https://flux-framework.readthedocs.io/projects/flux-sched/en/latest/man5/flux-config-sched-fluxion-qmanager.html).

wihobbs (approved):

doc/man1/flux-account-add-queue.rst
cmoussa1: That's a good thought! Maybe I should add it as a note at the top of these man `*-queue` pages.

wihobbs: In my note above, s/queue policies in flux-sched/_queuing_ policies in flux-sched/

cmoussa1:
Thanks for the review @wihobbs - I pushed up some a fix to add a note to the top of the man page based on your suggestion above! Setting MWP here

PR #614, opened 2025-04-23
`construct_hierarchy ()`: use column names

cmoussa1:
#### Problem

The `construct_hierarchy ()` method uses column indices instead of column names when placing the results of a SQL query into the string constructed for a flux-accounting bank/user hierarchy. This makes the code hard to parse and understand exactly what fields are being placed where in the string.

---

This PR changes the column indices to instead use column names when accessing the results of a SQL query and formatting the resulting string.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -187,10 +187,10 @@ class BankFormatter(AccountingFormatter):
                             indent
                             + " "
                             + bank.ljust(20)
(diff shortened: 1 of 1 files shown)


wihobbs (approved):
"Generate a cartoon of a baseball player refactoring SQL code"

ChatGPT Image Apr 23, 2025, 12_56_41 PM

cmoussa1:
LOL. Maybe I should come to work in my baseball uniform next time. Setting MWP here

PR #615, opened 2025-04-29
`add-user`: add `--fairshare` as an optional argument

cmoussa1:
#### Problem

The `fairshare` attribute for associations is not configurable in the `add-user` command, meaning that every association that gets added to the flux-accounting database **has** to have a `fairshare` value of 0.5. There might be a case where an admin might want to add an association with a different fairshare value, but in order to do this, they have to
open a SQLite shell, connect to the DB, and execute a SQL statement.

---

This PR adds `fairshare` as an optional argument to the `add-user` command, which allows an admin to customize the `fairshare` value for an association if they want to. The default is kept at 0.5.

I've also added the optional argument to the `add-user(1)` man page.

### DIFF
--- a/doc/man1/flux-account-add-user.rst
+++ b/doc/man1/flux-account-add-user.rst
@@ -41,6 +41,11 @@ be defined upon user creation.
     The amount of available resources their organization considers they should
     be entitled to use relative to other competing users.
 
+.. option:: --fairshare
+
+    The ratio between the amount of resources an association is allocated
+    versus the amount actually consumed.
+
 .. option:: --max-running-jobs
 
     The max number of running jobs the association can have at any given time.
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
Looks good!

cmoussa1:
Thanks! Setting MWP here

cmoussa1:
something went wonky here. going to try restarting the merge queue

cmoussa1:
@mergifyio requeue

cmoussa1:
@mergifyio dequeue

cmoussa1:
@mergifyio requeue

cmoussa1:
@mergifyio dequeue

cmoussa1:
That was odd. I think when I force-pushed some changes to this PR earlier, the normal suite of CI checks (`validate commits`, `el8`, `distcheck`, etc.) never triggered? So mergify never saw the condition succeed and thus never merged the PR. Re-pushing the commits seemed to trigger the normal suite of checks and then allowed mergify to merge the PR.

PR #616, opened 2025-04-30
`list-users`: add `--default-project` optional argument, man page

cmoussa1:
#### Problem

The `list-users` command does not take `default_project` as an optional argument, but it would be nice to have in case a user wants to filter the table by `default_project`.

There is also no `man(1)` entry for the `list-users` command.

---

This PR adds `default_project` as an optional argument and fixes the docstring comment for the `list_users()` function. I've also added a man page for the command.

I also snuck in a minor spelling fix commit for the `edit-user(1)` page.


jameshcorbett (approved):
👍

cmoussa1:
thanks! setting MWP here

PR #618, opened 2025-05-01
`add-user(1)`: clarify how default project is set

cmoussa1:
#### Problem

It is not totally clear from reading `add-user(1)` that the first project listed will become the association's default project when passing `--projects` to the `add-user` command.

---

This PR just adds a note to the man page to clarify how the default project is set.


jameshcorbett (approved):
Hate it, needs a complete rework

cmoussa1:
setting MWP here (regrettably)

PR #619, opened 2025-05-01
`add-user`: add `--default-project` as an optional argument

cmoussa1:
#### Problem

The `add-user` command does not have an option to specify a default project when adding an association for the first time. The
behavior currently is to pass a list of projects where the first one is treated as the default. It would be convenient to have an option to also just specify a default project by itself.

---

This PR (built on top of #618) adds a new optional argument to the add-user command: `--default-project`, which takes a string representing the name of the default project an association will submit jobs under. It checks for the existence of this argument when determining the default project for an association before taking the first project listed in `projects` to keep backwards compatibility.

It a default project is specified, it makes sure that it is also present in `projects`, and if not, it adds it.

I've added a simple test that exercises use of this option as well as updated `add-user(1)` with the new option.

### DIFF
--- a/doc/man1/flux-account-add-user.rst
+++ b/doc/man1/flux-account-add-user.rst
@@ -72,6 +72,10 @@ be defined upon user creation.
     the association's default project. The association's default can be changed
     with :man1:`flux-account-edit-user`.
 
+.. option:: --default-project
+
+    The default project an association will run jobs under.
+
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
Nice!

cmoussa1:
rebased to catch up after #618. setting MWP here

PR #622, opened 2025-05-05
`flux-account`: standardize prefixes for error messages

cmoussa1:
#### Problem

A lot of the exception handlers in `flux-account-service.py` either have the wrong prefix or no prefix at all. Now that
flux-accounting has better formatting for printing out error messages (i.e. it disables the huge stack trace by default), it should take better care of making sure that error messages raised by flux-accounting have an appropriate format.

---

This PR standardizes the prefixes in the exception handlers for the different commands in the flux-accounting service. For functions in the Python bindings that have the command prefix in their `except` block, I've removed them, since they are now all located in `flux-account-service.py`.

I've also removed the prefix for the general `Exception` block since it was pointed out in #597 that a function could in fact return an `OSError`.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -170,9 +170,9 @@ def view_bank(
             return formatter.with_users(bank)
         return formatter.as_json()
     except sqlite3.Error as err:
-        raise sqlite3.Error(f"view-bank: an sqlite3.Error occurred: {err}")
+        raise sqlite3.Error(err)
     except ValueError as exc:
-        raise ValueError(f"view-bank: {exc}")
+        raise ValueError(exc)
 
 
 def delete_bank(conn, bank, force=False):
@@ -318,6 +318,6 @@ def list_banks(
             return formatter.as_table()
         return formatter.as_json()
     except sqlite3.Error as err:
-        raise sqlite3.Error(f"list-banks: an sqlite3.Error occurred: {err}")
+        raise sqlite3.Error(err)
     except ValueError as exc:
-        raise ValueError(f"list-banks: {exc}")
+        raise ValueError(exc)
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -163,17 +163,8 @@ class AccountingService:
             handle.respond(msg, payload)
         except KeyError as exc:
             handle.respond_error(msg, 0, f"view-user: missing key in payload: {exc}")
-        except ValueError as val_err:
-            handle.respond_error(msg, 0, f"view-user: {val_err}")
-        except sqlite3.OperationalError as sql_err:
-            handle.respond_error(
-                msg, 0, f"view-user: sqlite3.OperationalError: {sql_err}"
-            )
         except Exception as exc:
-            # fall through to a non-OSError exception
-            handle.respond_error(
-                msg, 0, f"a non-OSError exception was caught: {str(exc)}"
-            )
+            handle.respond_error(msg, 0, f"view-user: {type(exc).__name__}: {exc}")
 
     # pylint: disable=no-self-use
     def list_users(self, handle, watcher, msg, arg):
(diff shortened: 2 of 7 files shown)


src/cmd/flux-account-service.py
```diff
@@ -202,6 +202,8 @@ def list_users(self, handle, watcher, msg, arg):
             handle.respond(msg, payload)
         except KeyError as exc:
             handle.respond_error(msg, 0, f"list-users: missing key in payload: {exc}")
+        except (ValueError, sqlite3.Error) as exc:
+            handle.respond_error(msg, 0, f"list-users: {exc}")
```
grondo, line 204: Just checking on this commit. The commit message states:

> Condense the two except blocks to handle them on the same line instead of using two separate except blocks.

I'm probably missing something in the diff, but I just see this _additional_ except block added, but no `except` removed, so just making sure this does what the commit message says?

grondo: Also, should the fall-through (`except Exception`) case also have the `list-user: ` prefix added?

grondo (approved):
LGTM! Not necessarily a comment but left a question inline.

src/cmd/flux-account-service.py
```diff
@@ -659,11 +679,11 @@ def list_queues(self, handle, watcher, msg, arg):
...
+            handle.respond_error(msg, 0, f"list-queues: missing key in payload: {exc}")
         except ValueError as exc:
-            handle.respond_error(msg, 0, f"{exc}")
+            handle.respond_error(msg, 0, f"list-queues: {exc}")
         except sqlite3.Error as exc:
-            handle.respond_error(msg, 0, f"a SQLite error occurred: {exc}")
+            handle.respond_error(msg, 0, f"list-queues: a SQLite error occurred: {exc}")
         except Exception as exc:
             handle.respond_error(
                 msg, 0, f"a non-OSError exception was caught: {str(exc)}"
```
jameshcorbett: Instead of so many different types of exceptions, since the handling all seems to be pretty similar, I wonder if you might do something like:

```python
except Exception as exc:
    handle.respond_error(msg, 0, f"list-queues: {type(exc).__name__}: {exc}
```

jameshcorbett (commented):
Two comments below that I think could reduce some code duplication.

src/cmd/flux-account-service.py
cmoussa1: Oh duh, thank you for catching this 🤦 I got mixed up while editing one of these commits and never fixed the commit message. This should instead say something along the lines of "the `list-users` command was never handling either exception, but either could occur in the function...."

I'll edit this commit message.

src/bindings/python/fluxacct/accounting/bank_subcommands.py
```diff
@@ -318,6 +318,6 @@ def list_banks(
             return formatter.as_table()
         return formatter.as_json()
     except sqlite3.Error as err:
-        raise sqlite3.Error(f"list-banks: an sqlite3.Error occurred: {err}")
+        raise sqlite3.Error(err)
     except ValueError as exc:
-        raise ValueError(f"list-banks: {exc}")
+        raise ValueError(exc)
```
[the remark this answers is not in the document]
cmoussa1: This is a good suggestion. I think I should definitely go back through the bindings and perhaps re-work a lot of these `try/except` blocks, since now the flux-accounting service handles these. These `try/except` blocks were leftover back when the function itself was responsible for handling these errors and before a flux-accounting service was introduced.

Would you be okay with me opening an issue on this for a future PR?

src/cmd/flux-account-service.py
cmoussa1: I like this suggestion a lot. Perhaps I should refactor these `try/except` blocks to follow this format instead?

src/bindings/python/fluxacct/accounting/bank_subcommands.py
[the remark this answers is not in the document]
jameshcorbett: Sure I don't see why not!

cmoussa1:
Just force-pushed up a slight restructure to this PR to include @jameshcorbett's suggestion of condensing the multiple `except` blocks into just one `except Exception` that includes the specific exception type in the message. This looks much cleaner now! Thanks. I'll go ahead and set MWP here

PR #625, opened 2025-05-06
bindings: remove redundant `try/except` blocks

### LINKED ISSUE #624: bindings: remove `try/except` blocks if possible and just raise `Exception`
A lot of the bindings still have `try/except` blocks from when flux-accounting did not have a systemd service and something that could actually catch any exceptions raised from the functions themselves. Thus, there is a lot of code in the bindings that look like:

```python
except Exception:
    raise Exception
```

Now that the flux-accounting service has a more robust way of capturing and propagating errors that occur, these bindings should be refactored to just `raise` the exception.

cmoussa1:
#### Problem

As noted in #624, a lot of the bindings still have `try/except` blocks from when flux-accounting did not have a systemd service and something that could actually catch any exceptions raised from the functions themselves. Thus, there is a lot of code in the bindings that looks like:

```python
except Exception:
    raise Exception
```

Now that the flux-accounting service has a more robust way of capturing and propagating errors that occur, these bindings should be refactored to just `raise` the exception.

---

This PR removes a lot of the `try/except` blocks in the python bindings that do not provide a custom error message and just re-`raise`s the error to the service.

I've also moved one of the multi-line comments that describes some of the reasons for `sqlite3.Error`s to the service since the comment was removed from the bindings.

Fixes #624

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -100,8 +100,6 @@ def add_bank(conn, bank, shares, parent_bank=""):
             validate_parent_bank(cur, parent_bank)
     except ValueError as bad_parent_bank:
         raise ValueError(f"parent bank {bad_parent_bank} not found in bank table")
-    except sqlite3.OperationalError as exc:
-        raise sqlite3.OperationalError(exc)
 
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
Looks good!

cmoussa1:
thanks! setting MWP here

PR #626, opened 2025-05-08
`list-*` commands: make table output default

cmoussa1:
#### Problem

The `list-users` command outputs in a table format by default, which was discussed by multiple Flux developers to be the preferred default (see https://github.com/flux-framework/flux-accounting/pull/597). However, the rest of the `list-*` command outputs in JSON by default instead of in a table format.

---

This PR changes the default behavior for the `list-*` commands to output in a table format by default instead of JSON format to match the `list-users` command.

I've also adjusted tests and man pages to reflect the change in the behavior for these commands.


jameshcorbett (approved):
Looks reasonable!

cmoussa1:
thanks! setting MWP here

PR #628, opened 2025-05-12
`update-usage`: add INFO-level logging

cmoussa1:
#### Problem

The `update-usage` command (the script responsible for updating job usage values) can take upwards of a minute or two to run because it needs to update job usage values for every association and bank in the database, i.e. executing an `UPDATE` statement in the database. This requests a lock on the DB, and other flux-accounting commands executed during this time might hang while this is running. But since there is no logging of when this is running, figuring out why flux-accounting commands are hanging can be confusing.

---

This PR adds `INFO`-level logging to the function called by the `update-usage` command when the function starts/stops to give an idea of how long the function takes to run.


jameshcorbett (approved):
Looks good although I would suggest not using the root logger, so instead of `logging.info` do `logging.getLogger(__name__).info(...)`. It's easier to configure logging later that way, see https://docs.python.org/3.12/library/logging.html and the following advice from the docs:

> For very short scripts or quick demonstrations of logging facilities, debug and the other module-level functions may be convenient. However, most programs will want to carefully and explicitly control the logging configuration, and should therefore prefer creating a module-level logger and calling [Logger.debug()](https://docs.python.org/3.12/library/logging.html#logging.Logger.debug) (or other level-specific methods) on it, as described at the beginnning of this documentation.

jameshcorbett:
> I see now that the message is prefixed with the actual module name, which is nice

The format of the log message is completely configurable, there's an example here: https://docs.python.org/3.12/library/logging.html#logging.Logger.debug

PR #630, opened 2025-05-13
`calc_usage_factor()`

cmoussa1:
#### Problem

In the case where no new jobs are found in a new half-life period, the `update-usage` command does not update the "current" job usage period for an association, which means that there is a discrepancy between what the `job_usage_factor_table` shows for an association's historical job usage and their total historical job usage reported with `view-user`.

---

This PR adds an update to the current job usage period for an association in the case where no new jobs are found in a new half-life period (i.e. their current job usage should be 0). I've also added a test to make sure the current job usage value is properly set when no new jobs are found in a new half-life period.

Finally, I've updated some comments in `calc_usage_factor()` to further expand what is happening in each `if`-condition when determining how to calculate the job usage value for a user in different scenarios.


jameshcorbett (approved):
Looks good!

cmoussa1:
Thanks! Setting MWP here

Issue #631, opened 2025-05-14
`update-usage`: parent banks accumulating incorrect usage during update


ryanday36:
This might still be related to #585, but on elcap and Tuolumne, there are a bunch of parent banks with large numbers for their usage even though all of their children have no usage. E.g.

```
[day36@tuolumne2152:~]$ flux account view-bank -t cmetal
Bank                            Username           RawShares            RawUsage           Fairshare
cmetal                                                    73       3171414442.24
 cbronze                                                  44                 0.0
 csilver                                                  29                 0.0

[day36@tuolumne2152:~]$
```

I see the same thing for a number of banks on elcap, but the rawusage value is different there (5598345332.6). There is one bank on Tuolumne where the rawusage  value of 3171414442.24 appears to be correct based on the sum of its children (the 'overhead' parent bank). 

Actually, looking at the rawusage values for other banks, it looks like all of them other than overhead are getting more rawusage than the sum of their own children. E.g. 'wmetal' appears to be 'overhead' + its children ('wbronze'), and 'gc' appears to be 'wmetal' + the children of 'gc' ('bcwc' and 'latticgc').

Here's a tree view of all of the banks without any users:

```
[day36@tuolumne2152:~]$ bankinfo -T root -v 
Name                     Shares  Norm Shares         Usage   Norm Usage   Level FS   Priority  Type
ROOT                          1     0.000000  3863035860.7     0.000000         --         --  Bank
 asc                     331000     0.122365  3550484205.4     0.919092         --         --  Bank
  hpcic2b                  2373     0.000877           0.0     0.000000         --         --  Bank
  wci                    328627     0.121488  3550484205.4     0.919092         --         --  Bank
   cmetal                    73     0.000027  3171414442.2     0.820964         --         --  Bank
    cbronze                  44     0.000016           0.0     0.000000         --         --  Bank
    csilver                  29     0.000011           0.0     0.000000         --         --  Bank
   gmetal                 18788     0.006946  3171414442.2     0.820964         --         --  Bank
    gbronze               10324     0.003817           0.0     0.000000         --         --  Bank
    ggold                   793     0.000293           0.0     0.000000         --         --  Bank
    gsilver                7671     0.002836           0.0     0.000000         --         --  Bank
   pmetal                 19270     0.007124  3171414442.2     0.820964         --         --  Bank
    pbronze               19270     0.007124           0.0     0.000000         --         --  Bank
[...]

PR #632, opened 2025-05-14
`update-usage`: fix usage aggregation in multi-level bank hierarchies

cmoussa1:
#### Problem:

As mentioned in #631, banks in the flux-accounting database were reporting inflated usage due to shared state of a total usage value during recursion, which caused some parent banks to sum unrelated sub bank usage.

---

This PR refactors `calc_parent_bank_usage()` to isolate the `total_usage` variable within each recursive call so that each bank now accumulates usage only from its direct sub banks and updates its own `job_usage` field.

I've added a sharness test to ensure that when the `update-usage` script is run, the `view-bank --tree` command reports correct usage for the root bank _as well as any_ sub banks. I've also added a unit test to simulate a half-life period taking place and making sure that all banks and users have their job usage values appropriately decayed by calling `update_job_usage()` again.


jameshcorbett (approved):
Nice fix! Lots of tests! 🧑‍💻 🧪 🥇

cmoussa1:
Just when you think you have a test for everything... 😆 thanks for the review! Setting MWP here

PR #633, opened 2025-05-15
python: improve logger format in `update-usage`, create module-level logger in `create-db`

cmoussa1:
#### Problem

The `create-db` command logs status of creating tables in the flux-accounting database, but it uses the root logger instead of one custom to the module.

The log messages in the `update-usage` command don't include a timestamp of when those messages were actually logged, which might make analyzing run times of the `update-usage` command more difficult in the future.

---

This PR changes the `create-db` command to use a logger as a constant and uses it to log the status of creating various tables in the flux-accounting DB. I've also configured the logger in `update-usage` to prefix its log messages with a datetime.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/create_db.py
+++ b/src/bindings/python/fluxacct/accounting/create_db.py
@@ -18,6 +18,8 @@ import time
 
 import fluxacct.accounting
 
+LOGGER = logging.getLogger(__name__)
+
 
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
Nice!

cmoussa1:
thanks! setting MWP here

PR #636, opened 2025-05-20
commands: add shorter options for some longer optional arguments

### LINKED ISSUE #635: commands: look into adding more short options for optional arguments
A lot of the flux-accounting commands (e.g. `add-user`) do not offer short options for the optional arguments, which would be pretty convenient to have when the long form options are in fact _very_ long.

cmoussa1:
#### Problem

A number of commands that deal with users and queues in the flux-accounting database do not offer short-form options, which results in some inconvenient typing when specifying a lot of the optional args. It would be nicer if these optional arguments had a short-form option.

---

This PR adds some short options for a lot of the optional arguments for the commands that deal with users and queues. I also added a small sharness test file that calls these commands with the short options.

Fixes #635

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -78,6 +78,7 @@ TESTSCRIPTS = \
 	t1073-mf-priority-init-basic.t \
 	t1074-mf-priority-init-depend.t \
 	t1075-custom-usage-reporting.t \
+	t1076-cmds-short-options.t \
 	t5000-valgrind.t \
 	python/t1000-example.py \
 	python/t1001_db.py \
--- a/t/t1076-cmds-short-options.t
+++ b/t/t1076-cmds-short-options.t
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+test_description='test using short options for various flux-accounting commands'
+
+. `dirname $0`/sharness.sh
+
+mkdir -p conf.d
+
+DB_PATH=$(pwd)/FluxAccountingTest.db
+
+export TEST_UNDER_FLUX_SCHED_SIMPLE_MODE="limited=1"
+test_under_flux 1 job -o,--config-path=$(pwd)/conf.d
+
+flux setattr log-stderr-level 1
+
+test_expect_success 'create flux-accounting DB' '
+	flux account -p ${DB_PATH} create-db
+'
+
+test_expect_success 'start flux-accounting service' '
+	flux account-service -p ${DB_PATH} -t
+'
+
+test_expect_success 'add a root bank' '
+	flux account add-bank root 1
+'
+
+test_expect_success 'add a bank with a short option' '
+	flux account add-bank --parent-bank=root A 1
+'
+
+test_expect_success 'add a queue with short options' '
+	flux account add-queue gold \
+		-N 64 \
+		-t 3600 \
+		-P 1000 &&
+	flux account view-queue gold
+'
+
+test_expect_success 'edit a queue with short options' '
+	flux account edit-queue gold \
+		-N 32 \
+		-t 1800 \
+		-P 500 &&
+	flux account view-queue gold
+'
+
+test_expect_success 'add a project' '
+	flux account add-project project1
+'
+
+test_expect_success 'add an association with short options' '
+	flux account add-user \
+		-u user1 \
+		-i 50001 \
+		-B A \
+		-N 1000 \
+		-c 2000 \
+		-q gold \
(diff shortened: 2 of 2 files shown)


garlick:
Just a thought but would command completions like [flux-core's](https://github.com/flux-framework/flux-core/tree/master/etc/completions) be another way to solve this problem?

The short option space gets used up pretty quickly, and then it's hard to remember what goes with what.

wihobbs:
> Just a thought but would command completions like [flux-core's](https://github.com/flux-framework/flux-core/tree/master/etc/completions) be another way to solve this problem?

Neat, TIL! But AFAICT flux-core provides some straight up short options too like `-B` for bank on submission commands and `-j` for json output on `flux-job(1)` and I think it'd be nice to preserve that across projects.

This is kind of hazy feedback: I think some of these make sense like `-j` for json but `-c` for active isn't something I'd remember offhand, so maybe lean on command completion for that. @garlick's point about the short option space is well reasoned, and it'd be unfortunate to have to reassign short options later (or impossible after we get to a 1.0 release).

src/cmd/flux-account.py
```diff
@@ -104,42 +106,52 @@ def add_list_users_arg(subparsers):
...
         "--max-active-jobs",
         metavar="MAX_ACTIVE_JOBS",
     )
     subparser_list_users.add_argument(
+        "-N",
         "--max-nodes",
         metavar="MAX_NODES",
     )
     subparser_list_users.add_argument(
+        "-n",
```
wihobbs, line 139: I'd make this `-c` and drop `-c`, `-s`, `-a`, and `-k`

src/cmd/flux-account.py
```diff
@@ -153,73 +165,85 @@ def add_add_user_arg(subparsers):
...
     )
     subparser_add_user.add_argument(
+        "-i",
         "--userid",
         help="userid",
         default=65534,
         metavar="USERID",
     )
     subparser_add_user.add_argument(
+        "-b",
```
wihobbs, line 182: I'd change this to `-B` to be consistent with core (above too)

src/cmd/flux-account.py
```diff
@@ -662,34 +703,39 @@ def add_edit_queue_arg(subparsers):
...
     subparser_edit_queue.add_argument(
+        "-a",
         "--min-nodes-per-job",
         type=int,
         help="min nodes per job",
         default=None,
         metavar="MIN NODES PER JOB",
     )
     subparser_edit_queue.add_argument(
+        "-m",
```
wihobbs, line 714: I'd change this to `-N` and drop `-a` and `-r`

src/cmd/flux-account.py
```diff
@@ -595,30 +631,35 @@ def add_add_queue_arg(subparsers):
     subparser_add_queue.set_defaults(func="add_queue")
     subparser_add_queue.add_argument("queue", help="queue name", metavar="QUEUE")
     subparser_add_queue.add_argument(
+        "-a",
```
wihobbs, line 634: Same as edit_queue.

src/cmd/flux-account.py
wihobbs: And here I'd drop `-s`, `-f`, `-r`, `-a`, `-k`.

src/cmd/flux-account.py
```diff
@@ -153,73 +165,85 @@ def add_add_user_arg(subparsers):
...
     )
     subparser_add_user.add_argument(
+        "-N",
         "--max-nodes",
         help="max number of nodes a user can have across all of their running jobs",
         default=2147483647,
         metavar="MAX_NODES",
     )
     subparser_add_user.add_argument(
+        "-n",
```
wihobbs, line 225: I'd make this `-c`

src/cmd/flux-account.py
```diff
@@ -273,65 +298,76 @@ def add_edit_user_arg(subparsers):
...
     )
     subparser_edit_user.add_argument(
+        "-P",
         "--projects",
         help="projects the user is allowed to submit jobs under",
         default=None,
         metavar="PROJECTS",
     )
     subparser_edit_user.add_argument(
+        "-k",
```
wihobbs, line 370: Basically, same feedback as above.

wihobbs (changes requested):
Some of these I think could be really useful, but some are confusing. I commented which ones I'd keep, change, and drop.

PR #643, opened 2025-05-22
`bank_table`: add `priority` column

### LINKED ISSUE #639: `bank_table`: add `priority column`
The `bank_table` in the flux-accounting database does not have a `priority` column, i.e. a value associated with each bank that could affect a job's overall priority. This column should be configurable when adding a bank for the first time and when editing a bank's properties in the database.

cmoussa1:
#### Problem

As mentioned in #639, the `bank_table` in the flux-accounting database does not have a priority column, i.e. a value associated with each bank that could affect a job's overall priority.

---

This PR adds a new column to the `bank_table`: **priority**, a floating-point number associated with each bank that could affect a job's overall priority if it is submitted under that bank. `--priority` is also added as an optional argument to the `add-bank` and `edit-bank` commands so that new banks can specify a priority when they are first added and/or changed later on.

Note that this PR looks to just add the ability to define a priority for a bank in the flux-accounting database. Subsequent PRs will be submitted to actually add support to the plugin to actually integrate a bank's priority to a job's priority calculation.

Fixes #639


cmoussa1:
@ryanday36 one question before I ask for this to be officially reviewed: would you like this `priority` associated with banks to just be an integer? Would you rather it be a floating-point number for more fine-tuned control? I guess one benefit to instead making it a floating-point number is if we decide to change how priority is calculated, we could perhaps get more control and more of an affect out of how we set these priority values per-bank. I'll defer to your preference here.

ryanday36:
I assume this bank priority is going to be another factor in the multi-factor priority that can have its own weight, etc. I can't really think of a scenario where I would use the more fine-grained control that a floating-point value would give, but if it's no harder to use a float than an int, it might be worth doing just in case someone else has a different use case for it.

jameshcorbett (approved):
👍

cmoussa1:
thanks @jameshcorbett! setting MWP here

PR #645, opened 2025-05-27
plugin: add `Bank` class, send bank priority information to plugin

### LINKED ISSUE #640: plugin: send bank `priority` to plugin
The integer priority for each bank needs to be sent, unpacked, and stored in the priority plugin in the case that the bank a job is submitted in affects the overall priority for a job.

### LINKED ISSUE #641: plugin: create `Bank` class to store various potential attributes that affect a job
The priority plugin has previously only had to keep track of a bank's name in order to associate jobs with banks, but now with the request to potentially configure an integer priority with a bank (and configure max `preemptible-after` limits like in #552), creating a `Bank` class that can be used in the priority plugin might be a nice improvement to tracking certain attributes for each bank in the plugin.

cmoussa1:
#### Problem

The priority plugin does not unpack or have a way to store certain attributes about a bank, such as its associated priority, which was added in #643.

---

This PR adds a new class called `Bank`, which will store basic attributes about banks defined in the flux-accounting DB, particularly their priority, if one is set. I've also added a helper function to perform a lookup in a map of `Bank` objects and retrieve its associated priority. In the plugin, an unpack of bank information defined in the flux-accounting DB is also added so that bank names and their associated priorities are stored in the plugin. The ability to also define and unpack a `weight` for the bank factor is also added to the plugin.

I've also added some basic unit tests for the `Bank` class and `get_bank_priority ()` helper function.

Fixes #640
Fixes #641

### DIFF
--- a/src/cmd/flux-account-priority-update.py
+++ b/src/cmd/flux-account-priority-update.py
@@ -74,6 +74,7 @@ def bulk_update(path):
     bulk_user_data = []
     bulk_q_data = []
     bulk_proj_data = []
+    bulk_bank_data = []
 
     # fetch all rows from association_table (will print out tuples)
     for row in cur.execute(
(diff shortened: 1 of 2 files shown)


cmoussa1:
Thanks! Setting MWP here

PR #646, opened 2025-05-27
doc: remove `jinja2 < 3.1.0` requirement

cmoussa1:
#### Problem

GitHub dependabot is reporting moderate security vulnerabilities detected with `jinja2 < 3.1.4`, but flux-accounting has `jinja2 < 3.1.0` listed in its `requirements.txt` file.

---

Remove the `jinja2` version requirement from `requirements.txt`.


grondo (approved):
LGTM. Should we also try bumping the sphinx version to the latest?
That could be done separately since this fixes a known issue.

PR #647, opened 2025-05-28
plugin: utilize `bank` priority when calculating priority for a job

### LINKED ISSUE #642: plugin: add `bank` factor to priority calculation
The priority calculation for a job only takes three factors: fair-share, the queue the job was submitted in, and a self-set urgency. Once the bank gets a configurable priority attribute and it is stored in the plugin, this priority should be added to the set of factors used to generate a job's priority. Once this is added, the appropriate [documentation](https://flux-framework.readthedocs.io/projects/flux-accounting/en/latest/guide/accounting-guide.html#multi-factor-priority-plugin) should be updated to reflect the addition.

cmoussa1:
#### Problem

Banks can now have an associated floating-point priority assigned to them, but they are not utilized anywhere in the plugin.

---

This PR unpacks the priority associated with a bank and utilizes it when calculating the priority for a job. If the bank is updated while a job is still pending, the plugin will unpack the new associated priority and use it when re-calculating the overall priority for a job.

I've added a new set of tests that simulate submitting jobs under multiple banks with different priorities to showcase how a job's priority can be affected depending on how bank priorities are configured. There is also an example in there of a job's priority changing as the result of updating their bank while the job is still pending.

Fixes #642

### DIFF
--- a/src/plugins/accounting.hpp
+++ b/src/plugins/accounting.hpp
@@ -77,6 +77,7 @@ public:
     std::vector<Job> held_jobs;        // vector to keep track of held Jobs
     std::vector<std::string> queues;   // list of accessible queues
     int queue_factor;                  // priority factor associated with queue
+    double bank_factor;                // priority factor associated with bank
     int active;                        // active status
     std::vector<std::string> projects; // list of accessible projects
     std::string def_project;           // default project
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
Looks good!

cmoussa1:
Thank you! Setting MWP here

PR #652, opened 2025-06-02
doc: add note about active per-queue limits

cmoussa1:
#### Problem

There are no notes in flux-accounting documentation about configuring max running jobs limits on a per-queue basis to be enforced by the priority plugin.

---

This PR adds a section in the flux-accounting guide making a note about how to set a max running jobs limit for a queue and how it is enforced on an association's set of running jobs.


jameshcorbett (commented):
👍

cmoussa1:
Oh shoot. @jameshcorbett I think you might need to hit "approve" here officially. Sorry about that!

jameshcorbett:
Oh my bad!

jameshcorbett (approved):

PR #653, opened 2025-06-05
flux-accounting service: use `.get()` instead of direct key access

cmoussa1:
#### Problem

The flux-accounting service uses direct dict key access to fetch parameters for the various flux-accounting functions. This can be problematic in the even that a key-value pair is not present in the payload (like in the case of an optional argument), and a `KeyError` will be raised.

---

This PR converts the direct key access in the payload dictionary to use `.get()`, which will return `None` in the case that a key-value pair does not exist in the dictionary instead of raising a `KeyError` exception. This makes using `.get()` standard across all of the flux-accounting commands defined in the service since right now some commands use `.get()` and some do not.

### DIFF
--- a/src/cmd/flux-account-service.py
+++ b/src/cmd/flux-account-service.py
@@ -151,10 +151,12 @@ class AccountingService:
             val = u.view_user(
                 self.conn,
                 msg.payload["username"],
-                msg.payload["parsable"],
-                msg.payload["fields"].split(",") if msg.payload.get("fields") else None,
-                msg.payload["list_banks"],
-                msg.payload["format"],
+                msg.payload.get("parsable"),
+                msg.payload.get("fields").split(",")
+                if msg.payload.get("fields")
+                else None,
+                msg.payload.get("list_banks"),
+                msg.payload.get("format"),
             )
 
             payload = {"view_user": val}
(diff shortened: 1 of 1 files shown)


src/cmd/flux-account-service.py
```diff
@@ -150,11 +150,13 @@ def view_user(self, handle, watcher, msg, arg):
...
                 self.conn,
-                msg.payload["username"],
-                msg.payload["parsable"],
-                msg.payload["fields"].split(",") if msg.payload.get("fields") else None,
-                msg.payload["list_banks"],
-                msg.payload["format"],
+                msg.payload.get("username"),
+                msg.payload.get("parsable"),
+                msg.payload.get("fields").split(",")
+                if msg.payload.get("fields")
```
jameshcorbett, line 156: `if "fields" in msg.payload`?

jameshcorbett (approved):
This looks fine to me, but if a payload field is required, it seems to me like the RPC should fail with some kind of error? What is the advantage of this change?

src/cmd/flux-account-service.py
jameshcorbett: I wasn't clear but my suggestion was to replace `if msg.payload.get("fields")` with `if "fields" in msg.payload`. They're not exactly equivalent though, because if `fields` is set to the empty string, the first conditional would fail but the second one would succeed. So I guess it depends on how you want empty strings to be handled. My general thought was just that `if key in map` is clearer.

> the script that actually creates these commands (flux-account.py) is already making sure that if arguments are required, it fails before it actually issues the RPC with a more helpful error message

Yeah I think my thought is, this change seems to me like it removes all the payload verification in the RPC recipient. I like that the RPC sender (`flux-account.py`) verifies the payload before it sends it, but suppose now that you start modifying the RPC sender and somehow you accidentally make it drop some field from the payload, like `username`. Instead of the RPC recipient immediately throwing a `KeyError` when it receives the RPC, an error would presumably occur somewhere down the line. I don't feel like the benefit of standardizing the way things are unpacked outweighs the potential for errors to creep in 🤷 but idk that's me and maybe I'm overlooking some other factor

Issue #655, opened 2025-06-09
`update-usage`: move job usage update to its own script instead of part of flux-accounting service


cmoussa1:
The performance of job-usage updates on clusters with lots of users and banks can slow down and halt the efficiency of other flux-accounting commands, even simple ones like `flux account-view-user`. After some digging and experimenting, I believe I have narrowed down the root cause.

The `update_job_usage ()` is part of the flux-accounting systemd service. The service establishes its own connection to the DB through which most, if not all, of the flux-accounting commands that interact with the DB go through. And I believe that since `update-usage` is a part of this service, while it is running, it is holding the other commands that interact with this service until the `update-usage` command has finished running (even with concurrency modes enabled on the database itself). I was able to reproduce this behavior in a Docker container while calling `update-usage` on a very large flux-accounting DB. Even the `view-user` command would halt until the script finished running.

Moving the `update_job_usage ()` call to its own script (which establishes its _own_ SQLite connection to the DB) enables the other flux-accounting commands to be executed concurrently while the job usage values are being executed.

I'm thinking I should append #654 to also move `update_job_usage ()` out of the flux-accounting service and into its own script.

cmoussa1:
This is now completed by #657.

PR #657, opened 2025-06-10
job-usage update: move update out of flux-accounting service and into own script

cmoussa1:
#### Problem

As mentioned in #655, the `update-usage` command is a part of the flux-accounting service, which means that when it is running (especially on a large flux-accounting DB that can take a lot of time), it will halt the execution of other flux-accounting commands (even ones that just read from the DB) until the `update-usage` command finishes running. This command should be moved to its own script that establishes a separate connection to the DB so other flux-accounting commands can run while job usage values are being updated.

---

This PR creates a new Python command script in the `src/cmd/` directory called `flux-account-update-usage.py`, a script that will just call `update_job_usage()`, pretty much exactly similar to what the `update-usage` command in the flux-accounting service was doing. 

As a result, the `update-usage` command is removed from the flux-accounting service and both the testsuite and the documentation have been updated to reflect the new change (i.e. calling `flux account-update-usage` instead of `flux account update-usage`).

### DIFF
--- a/doc/guide/accounting-guide.rst
+++ b/doc/guide/accounting-guide.rst
@@ -179,7 +179,7 @@ The scripts should be run by :core:man1:`flux-cron`:
 
  # /etc/flux/system/cron.d/accounting
 
- 30 * * * * bash -c "flux account-fetch-job-records; flux account update-usage; flux account-update-fshare; flux account-priority-update"
+ 30 * * * * bash -c "flux account-fetch-job-records; flux account-update-usage; flux account-update-fshare; flux account-priority-update"
 
 Periodically fetching and storing job records in the flux-accounting database
 can cause the DB to grow large in size. Since there comes a point where job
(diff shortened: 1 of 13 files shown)


grondo:
One idea would be to make a special case for `"update_usage"` in the `flux-account.py` tool. Instead of sending an RPC for `accounting.update_usage`, issue a warning like
```
flux-account: WARNING: update-usage is deprecated. Use "flux account-update-usage" instead.
flux-account: running "flux account-update-usage" locally...
``` 
Then use `subprocess.run()` to run the script locally. Perhaps also check that the current broker rank is rank 0 as expected, otherwise issue an error?

This would just be in place temporarily for one release or so.

Does that sound reasonable?

cmoussa1:
OK, I've pushed up another commit on this PR to call `flux account-update-usage` as a subprocess within `flux-account.py`, and after some interactive testing it looks like this still allowed for both parallel reads/writes from other commands in the flux-accounting service while this usage update was running 👍

Issue #659, opened 2025-06-12
configuration: define priority factor weights in database instead of directly in plugin


cmoussa1:
While working on #658, I noticed that it was going to be super annoying to have to keep consistency with the way factor weights are configured in the priority plugin. Right now, the default weights for each factor used in the priority plugin are defined in the plugin code itself:

```c++
// default weights for each factor in the priority calculation for a job
#define DEFAULT_FSHARE_WEIGHT 100000
#define DEFAULT_QUEUE_WEIGHT 10000
#define DEFAULT_BANK_WEIGHT 0
```

An admin can change these weights by configuring the `[accounting.factor-weights]` table in a TOML file and reloading the config, for which a plugin has a callback ready to update any of the factors with a new weight. But, the more I think about this, the more it might be nice to just define these weights in the flux-accounting database itself, perhaps in its own table. If it is stored in the DB, any flux-accounting commands (especially one that outputs these factors like the command being discussed in #658) can easily just read from the DB to get an updated value, and any admins can easily view and modify the weight with a simple command: 

```console
flux account set-factor-weight --fairshare=100
```

This data could then just be packaged up with the rest of the DB information already sent to the plugin and unpacked and stored there.

If I were to go this route, I wonder if it might be worth deprecating the ability to define these weights in TOML. I'm honestly not sure if it's gotten any use, and having both options might lead to inconsistencies if an admin defines a weight in a TOML config, defines a different weight in the DB, and forgets to change one of them.

cmoussa1:
I think the weights would get applied immediately after a `flux config reload`; at least, that's how I think they worked in the [sharness tests](https://github.com/flux-framework/flux-accounting/blob/master/t/t1034-mf-priority-config.t) here.

You're absolutely right that if we change this to storing these value in the DB, they will not get updated until the next bulk update (which right now is happening every 30 minutes), so we should definitely update any documentation to make this clear. I'd need to update the flux-accounting guide and probably remove the `flux config-accounting(5)` man page.

I agree that getting @ryanday36's opinion here is a good idea.

cmoussa1:
OK, I believe every sub-issue has now been closed and the existing priority factors (`fairshare`, `queue`, and `bank`) can now be modified and listed in the flux-accounting DB, which will be sent to the plugin and utilized when calculating the priority for a job. I can re-open more specific issues if they come up, but I'll go ahead and close this one as completed.

PR #665, opened 2025-06-12
database: add new table to store priority factors and their weights

### LINKED ISSUE #660: add table to flux-accounting DB to store priority factor weights
There should be a table in the flux-accounting DB that can store the different priority factors and their associated weights in the DB. At a minimum, it should store the name of the factor and its associated weight.

### LINKED ISSUE #661: commands: add commands to view/edit/list all of the priority factors
All of the different priority factors should be able to be viewed, listed, and modified with front-end commands. The output of these commands should follow a similar format to the other flux-accounting commands that interact with the DB.

cmoussa1:
#### Problem

As mentioned in #659, the priority factors and their associated weights are manually defined in the priority plugin and nowhere else in flux-accounting. As we look to expand the configuration options with these weights (e.g. changing the way job priority is calculated to not consider fair-share) and accessibility to how these weights affect a job's priority, they need to be stored in a place where all of the flux-accounting components can easily reach them.

---

This PR begins this process by creating a new table in the flux-accounting DB called `priority_factor_weight_table`, a table to store the integer weights of the various factors used to calculated a job's priority in the priority plugin. These factors and their weights are added when the DB is created.

```
factor    | weight
----------+-------
fairshare | 100000  
queue     | 10000 
bank      | 0  
```

New commands to view and modify an individual factor's weight as well as list all of these factors at once have been added: `view-factor`, `edit-factor`, and `list-factors`.

```console
$ flux account view-factor fairshare
factor    | weight
----------+-------
fairshare | 100000
$ flux account edit-factor --factor=fairshare --weight=999
$ flux account view-factor fairshare
factor    | weight
----------+-------
fairshare | 999 
$ flux account list-factors
factor    | weight
----------+-------
fairshare | 100000  
queue     | 10000 
bank      | 0  
```

Logic has also been added to the `update-db` command to automatically insert these factors into the `priority_factor_weight_table` in the case where the table is being added for the first time. If the table and the factors already exist, the `INSERT` is ignored so that we don't potentially override any modified weights for existing factors.

Tests have been added both for the functions that the commands call as well as for the commands themselves.

Fixes #660 
Fixes #661

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -57,6 +57,7 @@ TESTSCRIPTS = \
 	t1052-mf-priority-queue-limits.t \
 	t1053-issue631.t \
 	t1054-mf-priority-bank-priorities.t \
+	t1055-flux-account-priorities.t \
 	t5000-valgrind.t \
 	python/t1000-example.py \
 	python/t1001_db.py \
@@ -68,7 +69,8 @@ TESTSCRIPTS = \
 	python/t1007_formatter.py \
 	python/t1008_banks_output.py \
 	python/t1009_users_output.py \
-	python/t1010_issue631.py
+	python/t1010_issue631.py \
+	python/t1011_priorities.py
 
 dist_check_SCRIPTS = \
 	$(TESTSCRIPTS) \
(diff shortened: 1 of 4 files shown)


ryanday36:
This looks good for working with the weights in the db.

jameshcorbett (approved):
Very nice!

cmoussa1:

PR #667, opened 2025-06-16
doc: add new "Limits" page

### LINKED ISSUE #666: documentation: update docs to include information on resource limits
Now that #562 has landed, the docs need to be updated with information on how per-association resource limits are enforced in the priority plugin and how they can be configured both in the database. Perhaps a separate page just on limits could be worthwhile.

cmoussa1:
####  Problem

There is a little description in the flux-accounting guide about on limits in the priority plugin, but it is incomplete (it is
missing the max resources per-association limit) and could use some more explanation and perhaps an example.

---

This PR adds a new page to the flux-accounting docs about how limits work in flux-accounting. It provides an example on the difference between soft limits and hard limits and creates a reference to the new "Limits" page in the flux-accounting guide.

Fixes #666

### DIFF
--- a/doc/components/limits.rst
+++ b/doc/components/limits.rst
@@ -1,0 +1,104 @@
+.. _limits:
+
+######
+Limits
+######
+
+********
+Overview
+********
+
+The multi-factor priority plugin in flux-accounting not only performs
+validation for associations and priority calculation for jobs; it can also
+administer and enforce limits on a per-association and per-queue basis. Limits
+are another way to ensure fair behavior of systems that see a lot of
+simultaneous activity from a large number of users.
+
+Hard Limits vs. Soft Limits
+===========================
+
+There are two different kinds of limits in flux-accounting: **hard limits** and
+**soft limits**. Hard limits will prevent an association from submitting a job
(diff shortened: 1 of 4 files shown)


doc/components/limits.rst
```diff
@@ -0,0 +1,103 @@
...
+
+``buster`` can submit a job and the priority plugin will generate a priority
+for this job and pass it on to the scheduler to begin running. If ``buster``
+submits a second job while the first one is running, this second job will be
+held until the first job completes. Specifically, a dependency_ will be added
+to the job to hold it in the ``DEPEND`` state until the first job has
+transitioned to the ``INACTIVE`` state. The name of the dependency can be found
+in the job's eventlog:
+
+.. code-block:: console
```
jameshcorbett, line 67: Maybe say something like "example output from `flux job eventlog JOBID`:..."

jameshcorbett (approved):
Nice!

doc/components/limits.rst
cmoussa1: This is a great suggestion, thank you! I've force-pushed up a fix to this commit to add this prompt to the code block.

PR #670, opened 2025-06-18
plugin: unpack priority factor weights from database

### LINKED ISSUE #662: plugin: send priority factor weight information to priority plugin
The information in the table that stores priority factors and its weights should be included with the other flux-accounting DB information that is sent to the priority plugin with a bulk update. There, it can be be unpacked and stored in the existing `priority_weights` map, very similar to how the plugin is currently looking to unpack priority weights if they are defined in a TOML file (i.e. the `conf.update` callback).

### LINKED ISSUE #663: plugin: deprecate `conf.update` callback, `man(5)` entry for `flux-config-accounting`
I spoke with @ryanday36 today a little bit about the lack of use that the actual TOML configuration of priority weights and it seems like this was not getting much use at all, which is actually probably a pretty good thing if we're going to add support for listing and modifying priority factor weights in the flux-accounting DB. So I should deprecate the callback in the plugin so that priority factor weights cannot be modified via TOML as well as remove the `man(5)` entry for `flux-config-accounting`.

cmoussa1:
#### Problem

The flux-accounting database is now the central location for storing the priority factor weights, but this information is not sent to the plugin to actually be used when calculating the priority for a job. This will be used **instead of** configuring the weights for each factor in a TOML file and reloading config.

---

This PR removes the `conf.update` callback and replaces it with a callback that unpacks the priority factor weight information from the database. As a result, the tests that modify the weights for these factors are also changed to use the new method of executing flux-accounting commands to change the weights and send these updated values to the plugin.

Finally, the documentation regarding `flux-config-accounting` has been removed to indicate it as a feature that is no longer supported. Any examples showing configuration of priority factor weights have been edited to use the new `flux account edit-factor` commands.

Fixes #662
Fixes #663


jameshcorbett (approved):
Looks good, is the new behavior documented?

cmoussa1:
Thanks! Yes, it is documented and ready to be proposed in a subsequent PR. I held off from adding `man(1)` pages here to prevent it from being a gauntlet to review in just one PR. I will propose it shortly! Setting MWP here in the meantime

PR #672, opened 2025-06-18
doc: add `man(1)` pages for priority factor commands

### LINKED ISSUE #664: documentation: add `man(1)` entries for various priority factor commands, examples on configuring weights
There should be `man(1)` entries for the new commands that deal with viewing and modifying the various priority factors, as well as probably an in-depth example of how to configure the weights for the priority factors and how they affect a job's priority.

cmoussa1:
#### Problem

There are no man pages for the commands that deal with viewing and modifying the priority factors in the flux-accounting database.

---

This PR adds `man(1)` pages for all of the commands that deal with viewing and modifying priority factor weights.

Fixes #664


jameshcorbett (approved):
🎉

cmoussa1:

PR #673, opened 2025-06-19
plugin: add weight for `urgency` factor, update docs with new equation

### LINKED ISSUE #671: allow urgency weight to be reconfigurable
Problem: when the flux accounting priority plugin is in use, the `urgency` (described in [RFC 30](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_30.html) has a very small impact on priority, which means that users cannot meaningfully lower the priority of their jobs (other than hold) relative to others,  and sys admins cannot  raise the priority of jobs.

Suggestion: give the `urgency` a weight that gives it *more* impact than any other factors by default.

Extra credit: allow the weight to be site configurable.

cmoussa1:
#### Problem

As noted in #671, when the flux-accounting priority plugin is in use, the urgency (described in [RFC 30](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_30.html) has little to no impact on a job's priority, which means that users cannot significantly lower the priority of their own jobs and sys admins cannot easily raise the priority of jobs other than setting them to `expedite`. This is because the urgency factor is not multiplied by an integer weight, similar to the other factors.

---

This PR adds a weight option to the urgency factor to make it more significant in the priority calculation for a job. By default, urgency has a weight of 1000 (but am completely open to lowering or raising this default), making it so a user can significantly decrease the priorities of their own jobs _without_ unintentionally holding their job, and admins can significantly increase the priorities of jobs without having to just go straight to `expedite`.

A new sharness test file is added to showcase how urgency can now significantly affect a job's priority if not set at a default urgency.

I've also updated the flux-accounting guide with the addition of a new integer weight to be associated with the urgency factor as well as adding a new table with all of the priority factors and their default weights.

For extra credit, I've also added urgency as a configurable priority factor in the flux-accounting database, meaning that an admin can change the weight to be associated with urgency when calculating priorities for jobs.

Fixes #671


src/plugins/mf_priority.cpp
```diff
@@ -114,7 +116,7 @@ int64_t priority_calculation (flux_plugin_t *p,
     priority = round ((fshare_weight * fshare_factor) +
                       (queue_weight * queue_factor) +
                       (bank_weight * bank_factor) +
-                      (urgency - 16));
+                      (urgency_weight * (urgency - 16)));
```
garlick, line 119: Not strictly related to this PR so feel free to ignore but FLUX_JOB_URGENCY_DEFAULT could be used in place of the value 16 here.

doc/guide/accounting-guide.rst
```diff
@@ -368,7 +368,12 @@ urgency
 
 Thus the priority :math:`P` is calculated as follows:
 
-:math:`P = (F_{fairshare} \times W_{fairshare}) + (F_{queue} \times W_{queue}) + (F_{bank} \times W_{bank}) + (F_{urgency} - 16)`
+.. math::
+
+  P = (F_{fairshare} \times W_{fairshare})
+      + (F_{queue} \times W_{queue})
+      + (F_{bank} \times W_{bank})
+      + (W_{urgency} \times (F_{urgency} - 16))
```
garlick, line 376: This might be slightly nicer visually if the factor comes before the weight like the other expressions.  Totally OK as is though!

garlick (approved):
Thanks, this looks good to me!
I noted  a couple minor things that you can feel free to ignore.

src/plugins/mf_priority.cpp
cmoussa1: This is a great suggestion - definitely makes this more clear. I'll go ahead and push up an extra commit to use this macro instead of just 16.

cmoussa1:
Thanks for reviewing this @garlick! I've force-pushed up some changes based on the above suggestions, and will set MWP

PR #674, opened 2025-06-19
commands: add new `jobs` command to view priority breakdowns of jobs

cmoussa1:
#### Problem

There is no way to see a convenient breakdown of how various job priorities for an association are calculated. Since there are multiple factors with different weights, it might be helpful to list this breakdown on the command line instead of making someone consult documentation and calculate the priorities themselves.

---

This PR (built on top of #673) adds a new command called `jobs`, which takes one argument (`username`) and lists a breakdown of all of their jobs' priorities:

```console
JOBID          USER     BANK    BANKPRIO  BANKFACT  QUEUE   QPRIO  QFACT  FAIRSHARE FSFACTOR  URGENCY URGFACT PRIORITY
foEvbno        moussa1  A       100.0     1000      bronze  1      10000  0.99      100000    16      1000    209000  
fqyZGwh        moussa1  B       10.0      1000      bronze  1      10000  0.5       100000    16      1000    70000   
ftiBx6b        moussa1  C       1.0       1000      bronze  1      10000  0.08      100000    16      1000    19000
``` 

This command accepts optional arguments (`--bank` and `--queue`) to filter jobs that are running under a certain bank, queue, or both. It also accepts an `-o/--format` argument in case you want to customize the output.

Tests are added walking through editing certain factors and submitting different jobs to different banks and queues to give different examples of how job priorities are calculated.

A new page is added to the flux-accounting documentation going into more detail about how job priorities are calculated. Since I've added this page, I've moved some of the explanation out of the flux-accounting guide and instead referenced this new page.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -94,6 +94,7 @@ EXTRA_DIST = \
 	components/job-usage-calculation.rst \
 	components/projects.rst \
 	components/limits.rst \
+	components/job-priorities.rst \
 	$(RST_FILES) \
 	man1/index.rst
 
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -33,7 +33,8 @@ MAN1_FILES_PRIMARY = \
 	man1/flux-account-view-factor.1 \
 	man1/flux-account-edit-factor.1 \
 	man1/flux-account-list-factors.1 \
-	man1/flux-account-reset-factors.1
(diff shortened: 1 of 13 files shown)


jameshcorbett (approved):
This looks great!

cmoussa1:
Thanks @jameshcorbett! I know this was kind of a chunky PR so I appreciate you taking the time to review this. Setting MWP here

PR #675, opened 2025-06-23
update-usage: remove extra `.commit()` from helper function

cmoussa1:
#### Problem

The `apply_decay_factor()` helper function has a `.commit()` call at the end of the function, which will write all changes to disk. However, this helper function is called in the middle of a SQLite TRANSACTION, which will already call `.commit()` when the transaction is finished running. Thus, this hinders performance of the `update-usage` command by performing an extra write to disk for every association in the `association_table`.

---

This PR removes the extra `.commit()` from the end of `apply_decay_factor()`.


jameshcorbett (approved):
LGTM although maybe could use a docstring stating that the function should be called inside of a transaction?

cmoussa1:
Thanks @jameshcorbett! That's a good suggestion. I force-pushed up a fix to add a docstring to that function to clarify. I'll go ahead and set MWP here

PR #677, opened 2025-06-25
database: add graph commands for displaying job usage values for associations, banks

cmoussa1:
#### Problem

Some of the numerical data in flux-accounting, such as job usage, would be cool to be visualized and compared directly on the command line for admins or users who are interested in looking at what users or banks are consuming the most usage. This would be a lot more convenient than calling `list-users` or `list-banks` and manually parsing tables to look at this kind of data.

---

This PR adds a new class to the Python bindings (`RenderBarChart`) that can create a horizontal bar chart using data from the flux-accounting database. To start, there are just two subclasses: classes for displaying the top _n_ associations or top _n_ banks in terms of job usage.

A new command is added to the flux-accounting command suite called `show-usage`, which will render a graph for either associations or banks:

```console
moussa1@docker-desktop:/usr/src$ flux account list-usage associations --limit=5
user1 / guests  | █████████████████████████████████████████████████████████████████████████████  8814977.53
user2 / guests  | █████████████████████████████████████████████████████████                      6614967.62
user3 / wbronze | █████████████████████████████████████                                          4419446.80
user4 / guests  | ████████████████████████████                                                   3395406.50
user5 / guests  | ██████████████████████████                                                     3184829.87
```

The graph is automatically sized to the width of the terminal the person running the command on and will fallback to a `#` character in the case the █ character cannot be rendered. An optional `--limit` argument is also available to limit the number of rows displayed.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/visuals.py
+++ b/src/bindings/python/fluxacct/accounting/visuals.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2025 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import shutil
+import sys
+
+
+class RenderBarGraph:
+    """A class for displaying a horizontal bar graph."""
+
+    def __init__(self, conn, query, query_params=(), bar_char="█", fallback_char="#"):
+        """
+        Initialize the graph renderer.
+
+        Parameters:
+            conn: A SQLite Connection object.
+            query: The query to the flux-accounting DB to actually fetch the data used
+                in the graph.
+            query_params: Any parameters specified for the SQLite query.
+            bar_char: The primary character to draw the bars (default: block char).
+            fallback_char: A fallback ASCII character if primary char can't be encoded in
+                the terminal.
+        """
+        self.conn = conn
+        self.bar_char = bar_char
+        self.fallback_char = fallback_char
+        self.query = query
+        self.query_params = query_params
+
+    def fetch_rows(self):
+        cur = self.conn.cursor()
+        if self.query_params != ():
+            cur.execute(self.query, self.query_params)
+        else:
+            cur.execute(self.query)
+        return cur.fetchall()
+
+    def make_label(self, row):
+        """Return string label for a row; override in subclasses."""
+        raise NotImplementedError
+
+    def _bar_char(self):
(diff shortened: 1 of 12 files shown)


src/bindings/python/fluxacct/accounting/visuals.py
```diff
@@ -0,0 +1,162 @@
...
+    def draw(self):
+        rows = self.fetch_rows()
+        if not rows:
+            return "no data to display"
+
+        labels = [self.make_label(row) for row in rows]
+        # last column will be the numerical value to display
+        values = [row[-1] for row in rows]
+
+        term_width = shutil.get_terminal_size((80, 20)).columns
```
jameshcorbett, line 68: It looks like the default for `get_terminal_size` is `(80,24)`, and you don't check the number of rows so I think you could remove the argument to this function?

t/python/t1012_visuals.py
```diff
@@ -0,0 +1,123 @@
...
+    @classmethod
+    def tearDownClass(self):
+        conn.close()
+        os.remove("FluxAccountingVisuals.db")
+
+
+def suite():
+    suite = unittest.TestSuite()
+
+    return suite
```
jameshcorbett, line 117: unnecessary?

jameshcorbett (approved):
This is very cool! It doesn't entirely feel like a "list" to me, like I'm not sure "list-usage" is the most intuitive name for the command for me. I might prefer "show-usage" or "plot-usage" or something. But not a big deal of course.

src/cmd/flux-account.py
```diff
@@ -1046,6 +1046,31 @@ def add_jobs_arg(subparsers):
...
+        help="list job usage values from flux-accounting database",
+        formatter_class=flux.util.help_formatter(),
+    )
+    subparsers_visuals.set_defaults(func="list_usage")
+    subparsers_visuals.add_argument(
+        "table",
+        help=(
+            "the type of job usage data to visualize "
+            "(available options: 'associations','banks')"
+        ),
```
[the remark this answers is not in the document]
cmoussa1: TIL! This is super helpful. And the error message is much clearer now:

```console
moussa1@docker-desktop:/usr/src$ flux account show-usage foo
usage: flux-account.py show-usage [-h] [-n LIMIT] {associations,banks}
flux-account.py show-usage: error: argument table: invalid choice: 'foo' (choose from 'associations', 'banks')
```

t/python/t1012_visuals.py
cmoussa1: unnecessary and copied over from the other python unit tests 🤦 will fix this

cmoussa1:
Thanks for the review and the helpful feedback @jameshcorbett! I ended up changing this command's name to just `show-usage` - I agree with you that it might be a better name for it than `list-usage`. I also force-pushed up some changes based on your suggestions above. Will set MWP here shortly

PR #680, opened 2025-07-01
update-usage: move one-time SQL query out of `for`-loop, combine per-association queries

cmoussa1:
#### Problem

The `calc_usage_factor()` function issues a query to fetch the timestamp of the end of the current half-life period for the _cluster_, but it does this for every association, when it really only needs to perform this query once.

The function also issues a `SELECT` query to fetch a timestamp for every association in order to determine how to filter the `jobs` table to fetch new jobs for each association. However, the parameters used to get this timestamp (username and bank) are already used before calling the `calc_usage_factor()` function for every association. So, there is opportunity to combine these two queries.

---

This PR moves the query to fetch the timestamp of the end of the current half-life period for the cluster to _before_ entering the `for`-loop to calculate the new job usage values for every association.

It also combines the queries:

1) fetching every association from the `association_table`
2) fetching the last seen timestamp for every association in the `association_table`

into just one query by using a `LEFT JOIN` **before** iterating through each association and calculating their new job usage value.

The unit tests in `t1006_job_archive.py` have been adjusted to account for the new parameters expected in the functions responsible for calculating job usage values for associations.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
+++ b/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
@@ -179,20 +179,10 @@ def get_curr_usg_bin(acct_conn, user, bank):
     return float(row[0])
 
 
-def calc_usage_factor(conn, pdhl, user, bank, default_bank):
+def calc_usage_factor(conn, pdhl, user, bank, default_bank, end_hl):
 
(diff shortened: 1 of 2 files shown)


cmoussa1:
Thanks! Setting MWP here

PR #644, opened 2025-05-22
codecov: include Python bindings in codecov report

cmoussa1:
#### Problem

The Python bindings are not included in coverage reporting by Codecov.

---

This PR copies over flux-core's `checks_run.sh` to include the Python bindings in the coverage report from Codecov.

PR #682, opened 2025-07-01
[WIP] checks-run: remove duplicate `coverage.rc` file creation

cmoussa1:
#### Problem

`checks_run.sh` has two instances of creating a `coverage.rc` file.

---

This PR removes the duplicate creation of `coverage.rc`. Also looking to see if this might fix the missing Python bindings coverage for this repository.


cmoussa1:
No, not a problem. I opened this PR mostly as an experiment to see if removing the duplicate `coverage.rc` that I had in flux-accounting's `checks_run.sh` (I think due to my my negligence in #644) would actually trigger the Python bindings from actually getting picked up by Codecov, but I'm still not seeing coverage there. 🤷

grondo:
I wonder if this error is the problem:
```
  No source for code: '/usr/src/usr/src/src/bindings/python/fluxacct/accounting/__init__.py'.
```

I'm not sure why the extra `/usr/src`, but maybe you can see if you add `ignore_errors = True` to the `coverage.rc` if you can get past that?

PR #683, opened 2025-07-02
flux-accounting service: use `DB_SCHEMA_VERSION` constant instead of hard-coded value

cmoussa1:
#### Problem

The flux-accounting service hard-codes the schema version of the flux-accounting DB when checking to see if the DB is out-of-date and needs to be updated. However, there is a constant that exists that will remain up to date event when the schema version number is updated.

---

This PR updates the flux-accounting service to use `fluxacct.accounting.DB_SCHEMA_VERSION` in place of a hard-coded
schema version number when checking whether the database needs to be updated.

I've also snuck in a couple of minor formatting commits.


cmoussa1:
thanks @wihobbs! setting MWP here

PR #686, opened 2025-07-08
`jobs`: use fair-share at priority calculation instead of association's current fair-share

### LINKED ISSUE #685: `jobs` command shows inaccurate priority when fair-share value changes
One thing I failed to consider when creating the `jobs` command was that the `FAIRSHARE` value reported for an association's jobs won't always be accurately reflected in the calculated priority for a job, i.e. the command will just report the association's _current_ fairshare value instead of their fairshare value _at the time the job was submitted_. This is especially confusing when looking at an association's inactive jobs:

```
JOBID          USER     BANK    BANKPRIO  BANKFACT  QUEUE   QPRIO  QFACT  FAIRSHARE FSFACTOR  URGENCY URGFACT PRIORITY
fdrA5AGemE7    keilbart guests  0.0       0         batch   0      10000  0.004739  100000    16      1000    474       
fdqsKXBEzAB    keilbart guests  0.0       0         batch   0      10000  0.004739  100000    16      1000    948     
fdfPAiAKaAK    keilbart guests  0.0       0         batch   0      10000  0.004739  100000    16      1000    90995
```

If it's a lightweight enough solution, maybe I could store the association's fairshare value _at the time of submission_ via `jobspec-update`, and just unpack it later in the Python code that creates the above table.

cmoussa1:
#### Problem

Described in #685, the `jobs` command pulls the association's *current* fair-share value instead of what their fair-share value was at the time of priority calculation for a job, which leaves inactive jobs with inaccurate priority calculations in the output, particularly as an association's fair-share value changes over time.

---

This PR adds a `flux_jobtap_event_post_pack ()` call post a `"memo"` event to add the association's fair-share value at the time of priority calculation for a particular job. Then, when generating output in the `flux jobs` command, this value is unpacked by looking at the job's `annotations`.

I've added a basic test to make sure that an association's fair-share value stays consistent even after their fair-share value is updated as time goes on.

Fixes #685


src/plugins/mf_priority.cpp
```diff
@@ -109,6 +109,13 @@ int64_t priority_calculation (flux_plugin_t *p, int urgency)
     queue_factor = b->queue_factor;
     bank_factor = b->bank_factor;
 
+    // pack fairshare using memo event
+    flux_jobtap_event_post_pack (p,
+                                 FLUX_JOBTAP_CURRENT_JOB,
+                                 "memo",
+                                 "{s:f}",
+                                 "fairshare", b->fairshare);
```
grondo, line 117: _Probably_ should check the return code here and raise an error if fails (or add a comment that the error is purposefully ignored)

grondo (approved):
This LGTM. One comment about checking the return code of `flux_jobtap_event_post_pack()`

src/plugins/mf_priority.cpp
grondo: Yeah, that sounds fine to me. Put a note that the error shouldn't be fatal in a comment too.

cmoussa1:
Thanks for reviewing this @grondo! I force-pushed up a change to check the return code of `flux_jobtap_event_post_pack ()` as well as added a comment explaining that an error there is not fatal. I'll go ahead and set MWP here

PR #695, opened 2025-07-11
database: add `max_nodes` column to `queue_table`

### LINKED ISSUE #689: database: store max nodes as a per-queue attribute
The flux-accounting database does not keep track of max nodes as an attribute on a per-queue basis; only a minimum and maximum nodes _per job_. I should add a new column to `queue_table` to keep track of a total `max_nodes` property. Its default should probably be the same as the `max_nodes` attribute in `association_table`. This column should also be able to be modified with the `add-queue` and `edit-queue` commands.

cmoussa1:
#### Problem

Mentioned in #689, `queue_table` does not keep track of a queue's `max_nodes` attribute, or an attribute that represents how many nodes an association can have across all of their running jobs in a particular queue.

---

This PR adds support for a `max_nodes_per_assoc` attribute for a queue. `max_nodes_per_assoc` is an integer that represents how many nodes an association can have across all of their running jobs in a particular queue. Its default is the same value as the `max_nodes` attribute for `association_table`.

`--max-nodes-per-assoc` optional arguments are added to the `add-queue` and `edit-queue` commands to allow this attribute to be modified for each queue in the table.

Fixes #689


cmoussa1:
@ryanday36 if you have some time, could you give this PR a high-level look? I want to make sure the interface for specifying a `max_nodes` limit with queues in flux-accounting is what you expect or if you have anything else you might want to see here before I start integrating this into the priority plugin to actually enforce this limit.

cmoussa1:

PR #696, opened 2025-07-16
plugin: shorten line that is greater than 80 characters

cmoussa1:
#### Problem

There is a line in the callback for job.state.run that is longer than 80 characters, which does not follow the coding guidelines outlined in [RFC 7](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_7.html).

---

This PR just shortens the line to be at less than 80 characters.


wihobbs (approved):
Time for a .clang-format file?

PR #698, opened 2025-07-17
doc: add note about case sensitivity for names

### LINKED ISSUE #697: doc: add a note that usernames/bank names are case sensitive
Had a meeting this morning where a discussion at SNL was brought up about some confusion surrounding specifying banks when submitting jobs, and it turns out that users were not obeying case sensitivity rules for their bank, i.e. passing `--bank=my_bank` when it should have been `--bank=My_Bank`.

A note should probably be added to the flux-accounting documentation that usernames and bank names are case sensitive, and admins/users should be aware of this when creating banks.

cmoussa1:
#### Problem

As mentioned in #697, there is some confusion about having to reference usernames and bank names by their exact case, i.e. using capital letters for bank names that contain a capital letter, and there is no documentation making this clear for users and/or administrators.

---

This PR adds a note to the Database Administration document about case sensitivity and being careful about including capital letters in usernames or bank names.

Fixes #697

### DIFF
--- a/doc/components/database-administration.rst
+++ b/doc/components/database-administration.rst
@@ -75,6 +75,12 @@ From here, users can be added to these banks to create *associations*, a
 
  $ flux account add-user --username=user_1 --bank=bank_A
 
+.. note::
+    Both usernames and bank names are case sensitive. Make sure that if a name
+    contains a capital letter, it is referenced by its exact case throughout
+    the various flux-accounting components that take these names as a
+    parameter.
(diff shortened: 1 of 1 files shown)


wihobbs (approved):
Can you open a PR to update this on the core manpage too? Thanks!

cmoussa1:
Thanks @wihobbs! Is there a specific manpage you were thinking of?

PR #700, opened 2025-07-17
command: add `edit-all-users` command

cmoussa1:
#### Problem

There is no way to efficiently edit a column (or attribute) for every association in `association_table`. This might come in handy in the case that an admin wants to update something for every association, such as increasing the `max_active_jobs` limit for every association.

---

This PR adds a new command to the flux-accounting command suite called `edit-all-users`, which takes optional arguments for a number of columns in `association_table` and executes an `UPDATE` statement for every association in the table at once.

Sharness tests and documentation for the command are also added.


cmoussa1:
Thanks for reviewing @jameshcorbett! I pushed up some fixes based on the suggestions above. Will set MWP here

PR #702, opened 2025-07-22
priority-update: send `max_nodes_per_assoc` to plugin, add `max_nodes_per_assoc` attribute to `Queue` class

### LINKED ISSUE #690: database: send max nodes queue property to plugin to be unpacked and stored with queue information
The max nodes information for a particular queue should be sent along with the other queue information already sent, unpacked, and stored in the plugin. The `Queue` class will need to be modified to store this attribute.

cmoussa1:
#### Problem

(continuing the work in #695) `max_nodes_per_assoc` is not sent along with the other queue information to the priority plugin during an update of flux-accounting DB information. Further, a `Queue` object in the priority plugin has no way to store and keep track of how many nodes an association can have across all of their running jobs in that queue.

---

This PR adds `max_nodes_per_assoc` to the set of per-queue information sent and unpacked in the priority plugin. To store this value, a new `max_nodes_per_assoc` attribute is added to the `Queue` class, which has the same default as queues do in the flux-accounting DB.

Note that this does not actually add support for tracking an association's node usage in a particular queue nor does it enforce a limit to avoid posting a gigantic PR. This kind of functionality will be incrementally added in future PRs.

Fixes #690

### DIFF
--- a/src/plugins/accounting.hpp
+++ b/src/plugins/accounting.hpp
@@ -63,6 +63,7 @@ public:
     int max_time_per_job = std::numeric_limits<int>::max ();
     int priority = 0;
     int max_running_jobs = std::numeric_limits<int>::max ();
+    int max_nodes_per_assoc = 2147483647;
 };
 
 // all attributes are per-user/bank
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
👍

PR #703, opened 2025-07-23
plugin: add `QueueUsage` class to track an association's node usage per-queue

### LINKED ISSUE #691: plugin: update `Association` class' `queue_usage` attribute to keep track of multiple pieces of data
If the plugin is to enforce a per-queue max nodes limit, it will need to keep track of both an association's number of running jobs in that queue _as well as_ their total number of nodes across their running jobs in that queue. The `queue_usage` attribute in the `Association` class just keeps track of the number of running jobs currently.

There is opportunity here to create a class that can keep track of multiple pieces of data pertaining to an association's usage within a particular queue as well as provide the ability to expand this in the future. A `QueueUsage` class can be defined that stores an association's number of running jobs, their node usage across those running jobs, and any other attributes that deal with an association's usage within a particular queue.

```c++
// a class to track an association's usage in a particular queue
class QueueUsage {
public:
    int cur_run_jobs = 0; // the number of running jobs in queue
    int cur_nodes = 0;    // the number of nodes used in queue
    // future attributes if needed...
};
```

### LINKED ISSUE #692: plugin: track an association's number of nodes in a particular queue
The plugin currently keeps track of an association's resource usage (i.e. their current core and node count across all of their running jobs) _in general_, but it will need to do this for a particular queue as well. Since the plugin already has helper functions to increment their total node and core counts across their running jobs, this should be able to be expanded to _also_ track their total node and core counts across their running jobs under a particular queue.

cmoussa1:
#### Problem

The `queue_usage` attribute of the `Association` class currently _only_ keeps track of the queue an association is running their jobs under and how many running jobs they have under that particular queue. However, there is also a need to know how many resources (i.e. nodes) an association has across all of their running jobs under that queue, which the plugin does not currently keep track of.

---

This PR changes the `queue_usage` attribute of the `Association` class to instead be a map where the key remains the same (a string containing the name of the queue) but with a value of a new `QueueUsage` class, a class that tracks both an association's number of currently running jobs as well as their total node usage across all of their running jobs in that queue. Node usage per-queue is now tracked among an association's set of running jobs. Note that _no_ limit enforcement based on an association's per-queue node usage is added in this PR, but will be added in a future PR.

The plugin is updated to change the way the current number of running jobs per-queue is accessed to use the new class. The `to_json ()` method for `Association` objects is also updated to pack all of the `QueueUsage` class attributes into a JSON object when constructing JSON output for each association.

Tests are added to check different scenarios of an association submitting different-sized jobs to one and/or multiple queues and ensuring that resource counts are correct.

Fixes #691
Fixes #692

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -43,6 +43,7 @@ json_t* Association::to_json () const
     json_t *user_projects = nullptr;
     json_t *temp = nullptr;
     json_t *queue_usage_json = nullptr;
+    json_t *usage_object = nullptr;
     json_t *hj_json = nullptr;
     json_t *job_json = nullptr;
     json_t *deps_array = nullptr;
@@ -74,9 +75,16 @@ json_t* Association::to_json () const
     if (!queue_usage_json)
         goto error;
     for (const auto &entry : queue_usage) {
+        const QueueUsage &usage = entry.second;
+        usage_object = json_pack ("{s:i, s:i}",
+                                  "cur_run_jobs", usage.cur_run_jobs,
+                                  "cur_nodes", usage.cur_nodes);
+        if (!usage_object)
+            goto error;
+
         if (json_object_set_new (queue_usage_json,
                                  entry.first.c_str (),
-                                 json_integer (entry.second)) < 0)
+                                 usage_object) < 0)
             goto error;
     }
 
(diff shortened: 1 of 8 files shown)


cmoussa1:
thanks @jameshcorbett! Setting MWP here

PR #706, opened 2025-07-29
`view-user`: add `-J/--job-usage` optional arg

cmoussa1:
#### Problem

An association's job usage value is made up of multiple job usage values (i.e. it is the sum of both current and older job usage factors that are stored in `job_usage_factor_table`), but this data is not viewable on the command line.

---

This PR adds a new optional argument to the view-user command: `-J/--job-usage`, which will return all the data stored for an association in `job_usage_factor_table`. Docs are updated with a description of the optional argument itself as well as an example of how to use it.

### DIFF
--- a/doc/man1/flux-account-view-user.rst
+++ b/doc/man1/flux-account-view-user.rst
@@ -35,3 +35,11 @@ well as customizing which fields are returned.
 .. option:: -o/--format
 
     Specify output format using Python's string format syntax.
+
+.. option:: -j/--job-usage
+
+    List all of the past job usage factors that make up an association's
+    historical job usage value.
+
+    .. note::
+        This optional argument cannot be combined with ``--fields``.
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -236,19 +236,38 @@ def clear_projects(conn, username, bank=None):
 #                                                             #
 ###############################################################
 def view_user(
-    conn, user, parsable=False, cols=None, list_banks=False, format_string=""
+    conn,
+    user,
+    parsable=False,
+    cols=None,
+    list_banks=False,
+    format_string="",
+    job_usage=False,
 ):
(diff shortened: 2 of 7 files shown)


src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -236,19 +236,38 @@ def clear_projects(conn, username, bank=None):
...
+    cols=None,
+    list_banks=False,
+    format_string="",
+    job_usage=False,
 ):
+    if job_usage and cols:
+        # job_usage is setting a flag to fetch user data from another table different
+        # from association_table, so raise an Exception
+        raise ValueError(f"-J/--job-usage cannot be combined with --fields")
+
```
jameshcorbett, line 251: This looks great, you could also make `cols` and `job-usage` mutually exclusive to do some client-side validation: https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

doc/man1/flux-account-view-user.rst
```diff
@@ -35,3 +35,8 @@ well as customizing which fields are returned.
 .. option:: -o/--format
 
     Specify output format using Python's string format syntax.
+
+.. option:: -j/--job-usage
+
+    List all of the past job usage factors that make up an association's
+    historical job usage value.
```
jameshcorbett, line 42: Maybe note incompatibility with `--fields` here?

jameshcorbett (approved):
Looks great!

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: Oh this is a great suggestion! Thanks for pointing this out. Perhaps I should refactor `view_bank()` to also use a mutually exclusive group...

cmoussa1:
Thanks for reviewing this @jameshcorbett! I pushed up some fixes based on your suggestions above. Will set MWP here

PR #710, opened 2025-08-04
plugin: reject jobs that exceed an association's or queue's max resources limits

### LINKED ISSUE #709: plugin: jobs greater than an association's max resources limit are held without context
Jobs on tuolumne were being held with the `max-resources-user-limit` (which are added to a job if an association's currently running jobs are equal to or greater than either their `max_nodes` or `max_cores` limits), but there is no validation that if a single job is greater than either one of these limits, that the job is _rejected_. For example, if an association has a `max_nodes` limit of 100 but submits a job that is > 100 nodes, the job is still accepted and immediately put in DEPEND state, which is incorrect. The job should be _rejected_ with a more helpful message that the job exceeds the association's max resources limit(s).

cmoussa1:
#### Problem

The priority plugin does not reject jobs in `job.validate` if it exceeds a max resources limit  either per-association or per-queue, and instead will hold the job in `DEPEND`, even though the job is not waiting for a currently running job to finish running. It instead should reject the job during validation with a message saying that the job size is greater than the limit defined per-association or per-queue.

---

This PR adds a check in `job.validate` to compare the size of the job to both the association's `max_nodes` or `max_cores` limits and the queue's `max_nodes_per_association` limit, and if the job would exceed any of these limits, the job is rejected.

Some tests that reproduced the original issue are added to ensure that a job is rejected in `job.validate` instead of held if it exceeds the configured limits.

Fixes #709

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -882,18 +882,22 @@ static int validate_cb (flux_plugin_t *p,
     double fairshare = 0.0;
     bool only_dne_data;
     Association *a;
+    json_t *jobspec = NULL;
+    Job job;
+    std::string queue_str;
 
     // unpack the attributes of the user/bank's submitted job when it
     // enters job.validate and place them into their respective variables
     flux_t *h = flux_jobtap_get_flux (p);
     if (flux_plugin_arg_unpack (args,
                                 FLUX_PLUGIN_ARG_IN,
-                                "{s:i, s:i, s{s{s{s?s, s?s, s?s}}}}",
+                                "{s:i, s:i, s{s{s{s?s, s?s, s?s}}}, s:o}",
                                 "userid", &userid,
                                 "state", &state,
                                 "jobspec", "attributes", "system",
                                 "bank", &bank, "queue", &queue,
-                                "project", &project) < 0) {
+                                "project", &project,
+                                "jobspec", &jobspec) < 0) {
         return flux_jobtap_reject_job (p, args, "unable to unpack bank arg");
     }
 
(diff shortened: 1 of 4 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -950,6 +954,49 @@ static int validate_cb (flux_plugin_t *p,
...
+        if (job.count_resources (jobspec) < 0) {
+            return flux_jobtap_reject_job (p,
+                                           args,
+                                           "job.validate: unable to count " \
+                                           "resources for job");
+        }
+        if (job.ncores > 0 && job.nnodes == 0)
+            // the job specified cores but no nodes, so we need to set
+            // nnodes == 1 here
+            job.nnodes = 1;
```
grondo, line 974: Suggestion: add braces around this block even though unnecessary to enclose the code along with the multiline comment. This is also consistent with the conditionals below.

grondo: Also, just a question: should this (adjusting nnodes to 1) b be done inside of `job.count_resources()` by default? Sorry if that's a dumb question...

src/plugins/mf_priority.cpp
```diff
@@ -950,6 +954,49 @@ static int validate_cb (flux_plugin_t *p,
...
+        if (it != queues.end ()) {
+            if (job.nnodes > queues[queue_str].max_nodes_per_assoc) {
+            // the job size is greater than the max nodes per-association limit
+            // configured for this queue; reject the job
+            return flux_jobtap_reject_job (p,
+                                           args,
+                                           "job.validate: job size is greater "
+                                           "than max resources limit "
+                                           "configured for queue");
+            }
```
grondo, line 987: Is it possible (or desirable I guess) to report the limit and the current job size in this error? e.g.

 nodes requested (24) greater than maximum (16) configured for queue

Also, I'd drop the `job.validate` prefix from this error since it will be returned to the user (and the error will already say something like "submission failed")

src/plugins/mf_priority.cpp
```diff
@@ -950,6 +954,49 @@ static int validate_cb (flux_plugin_t *p,
...
+        }
+        if ((job.nnodes > a->max_nodes) || (job.ncores > a->max_cores)) {
+            // the job size is greater than the max resources limits (max nodes
+            // OR max cores) configured for this association; reject the job
+            return flux_jobtap_reject_job (p,
+                                           args,
+                                           "job.validate: job size is greater "
+                                           "than max resources limits "
+                                           "configured for association");
+        }
```
grondo, line 997: similar to above, would it make sense to include the detected job size and the limit (or at least just the limit). Also, I suggest dropping the `job.validate: ` prefix.

grondo (approved):
LGTM! A couple comments and questions that are pretty minor inline.

src/plugins/mf_priority.cpp
grondo: Putting it in this PR makes sense to me!

cmoussa1:
Thanks for reviewing so quickly @grondo! I've pushed up an additional commit to move that assignment of `nnodes = 1` to inside the `count_resources ()` method, as well as added more context to the rejection messages about what limits are currently configured. I'll go ahead and set MWP here

PR #712, opened 2025-08-05
plugin: improve error message context in `job.validate`

cmoussa1:
#### Problem

In order to provide more context for an error message as to why an association can't run a job under a certain queue or project, the list of the association's valid projects or queues should also be returned. Because this list is stored as a vector in the Association class, there needs to be a way to join these strings into just one comma-delimited string.

Secondarily, since these error messages now include more context, they span multiple lines within the `flux_jobtap_reject_job ()` function in order to keep < 80 characters long. It might be cleaner to define these error messages as macros and use those in the function calls.

---

This PR moves the error messages returned in `job.validate` to macro definitions in `accounting.hpp` and uses these macros instead of defining the error messages within the `flux_jobtap_reject_job ()` function call. I've also added more context to these error messages to give more context as to _why_ a job was rejected due to an invalid queue or project (i.e. returning which queues and projects the association _does_ belong to).

I've also cleaned up the map lookup of a queue to drop the declaration of an iterator variable in favor of just comparing the lookup directly to the end of the map.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1001,8 +1001,7 @@ static int validate_cb (flux_plugin_t *p,
         }
         // look up queue in queues map to see if it has a defined
         // max_nodes_per_association limit
-        auto it = queues.find(queue_str);
-        if (it != queues.end ()) {
+        if (queues.find (queue_str) != queues.end ()) {
             int queue_max_nodes = queues[queue_str].max_nodes_per_assoc;
             if (job.nnodes > queue_max_nodes) {
                 // the job size is greater than the max nodes per-association limit


src/plugins/mf_priority.cpp
```diff
@@ -918,7 +918,8 @@ static int validate_cb (flux_plugin_t *p,
                                            args,
                                            "cannot find user/bank or "
                                            "user/default bank entry "
-                                           "for uid: %i", userid);
+                                           "for uid: %i",
+                                           userid);
```
jameshcorbett, line 941: Is there an automatic formatter for C and C++ code for this repo?

src/plugins/mf_priority.cpp
```diff
@@ -947,19 +947,30 @@ static int validate_cb (flux_plugin_t *p,
         return flux_jobtap_reject_job (p, args, "user/bank entry has been "
```
jameshcorbett, line 947: This commit could maybe be merged with d50f8bf1, I think it actually would have made reviewing a bit easier.

jameshcorbett (approved):
Looks great!

src/plugins/mf_priority.cpp
cmoussa1: There is not, but @wihobbs has made that suggestion before too. I should look into adding one! I can open an issue on this to track it.

cmoussa1:
Thanks for reviewing @jameshcorbett! I squashed those two commits you mentioned and force-pushed. Will set MWP here

PR #718, opened 2025-08-11
`jobs`: add requested duration to `jobs` table, return both requested and actual duration in `view-job-records`

### LINKED ISSUE #715: `jobs` table: add requested duration, actual duration of job to jobs table for tracking
Spinning off some tangible items from the discussion in #714. While we don't have to necessarily enforce any penalties or make adjustment to job priority calculation based on an association's history of requesting duration, we probably _can_ at least keep track of it in a more tractable way in flux-accounting. To start, we could probably just add some columns to the `jobs` table that store both the requested duration on the job and the actual duration of the job.

This would make it pretty easy to track an association's average difference of their requested duration vs. actual duration of their jobs in case we did want to do some reporting or policy enforcement on it in the future.

cmoussa1:
#### Problem

The requested duration for jobs submitted by associations is stored in jobspec, but it is not conveniently stored anywhere for use throughout flux-accounting. It would be nice if this value was easily available for tracking on a per-association or other basis.

---

This PR adds a new `requested_duration` column to the `jobs` table, a floating-point value that stores the job duration requested by the association. If duration is not configured for queues or not specified when a job is run, a duration of `0.0` is returned for the `requested_duration` for a job.

In `view-job-records`, both the requested duration and the _actual_ duration for the job are now returned when looking at old jobs.

Finally, some basic tests for submitting jobs both with and without setting a duration and submitting a job to a queue with a configured duration are added.

Fixes #715


jameshcorbett (approved):
LGTM!

PR #719, opened 2025-08-13
`view-job-records`: add filters for requested duration and actual duration in `jobs` table

cmoussa1:
#### Problem

There is no way to filter the `jobs` table by requested or actual duration.

---

This PR adds new optional arguments to the `view-job-records` command that can take one or more filters for both the `requested_duration` and `actual_duration` columns via expressions. For example, passing:

```console
$ flux account view-job-records -d "< 60"
```

will filter for jobs with a requested duration of less than 60 seconds. Multiple expressions can be passed to either optional argument to form filters bound with an AND statement in the SQL query:

```console
$ flux account view-job-records -d "> 60" "< 180"
```

will filter for jobs with a requested duration of greater than 60 seconds AND less than 180 seconds. The available operators to filter these columns by are: `<`, `<=`, `=`, `>=`, and `>`.

A helper function is added to validate the expressions passed in for both optional arguments to ensure a valid operator is passed **and** that it is an expression with just one value.

Basic tests are added that test filtering the `jobs` table by requested duration, actual duration, and both of them at the same time. The documentation for `view-job-records(1)` is also updated to include the new optional arguments.


jameshcorbett (commented):
This seems like a very cool feature but I think it inherently exposes you to a SQL injection attack. I'm not a super clever SQL injection attacker but looking at the code I'm pretty sure I could write something like `view-job-records -d "<60;DROP\nTABLE\n\IF\nEXISTS\nsometable;UPDATE\nsomeothertable\nWHERE\nsomething< 50"`. You could fix various vulnerabilities by making your code check for more things, but I think a better approach would be to not insert any user-specified strings except with SQL's native `?` insertion. 

If you want to keep the interface the same, I think you could do something like:
```
if userstring.startswith("<"):
    operator = "<"
elif userstring.startswith("<="):
    operator = "<="
else:
    ...
return (operator, operand)

query = f"SELECT a from b WHERE duration {operator} ?", operand
```

jameshcorbett (approved):
LGTM!

PR #723, opened 2025-08-19
`update-usage`:  remove extra `SELECT` query in favor of retrieving past job usage information beforehand

cmoussa1:
#### Problem

`calc_usage_factor()` issues a `SELECT` query (by calling `fetch_usg_bins()`) to locate an association's past job usage factors for every association, but the table that stores this information is already queried before the function is called. It would save additional queries to just fetch and store these factors in the initial query instead of issuing another `SELECT` query while inside `calc_usage_factor()`.

---

This PR does two things. First, it change the function definition of `calc_usage_factor()` to accept a dictionary of association information instead of taking all of the various fields related to an association as separate arguments. Second, it removes the `fetch_usg_bins()` function in favor of just fetching all of the past job usage factors in the initial query ahead of calling `calc_usage_factor()` for every association and adding it to the `association_info` dictionary. 

The unit tests in `python/t1006_job_archive.py` have been adjusted to account for the structure of the `association_info` function argument passed to `calc_usage_factor()`.


src/bindings/python/fluxacct/accounting/job_usage_calculation.py
```diff
@@ -110,7 +89,7 @@ def apply_decay_factor(decay, acct_conn, user=None, bank=None):
     usg_past = []
     usg_past_decay = []
 
-    usg_past = fetch_usg_bins(acct_conn, user, bank)
+    usg_past = usage_factors
```
jameshcorbett, line 92: It looks like `usg_past` is defined and then initialized to something else, could you just remove the `usg_past` variable and use `usage_factors` directly instead?

src/bindings/python/fluxacct/accounting/job_usage_calculation.py
```diff
@@ -175,24 +161,24 @@ def calc_usage_factor(
...
     elif (last_t_inactive - float(end_hl)) < hl_period:
         # found new jobs in the current half-life period; we need to 1) add the
         # new jobs to the current usage period, and 2) update the historical usage
         # period
-        usg_current += usage_period_0
+        usg_current += usage_factors[0]
 
         # usage_user_past = sum of the older usage factors
-        usg_past = fetch_usg_bins(conn, user, bank)
+        usg_past = usage_factors
```
jameshcorbett, line 181: Similarly, as far as I can tell `usg_past` is just an alias for `usg_factors`?

jameshcorbett (approved):
Looks great, two minor questions/comments.

PR #726, opened 2025-08-26
`update-usage`: add `.rollback()` in case of error

cmoussa1:
#### Problem

The function that's called when updating the job usage values for associations and banks does not do any sort of error handling or rollback in the case of an error occurring in the middle of the update, which could potentially lead to dangling connections and misleading `"Database is locked"` errors from other flux-accounting commands also trying to write to the flux-accounting DB.

---

This PR uses the Connection object as a context manager, which handles committing the changes made to the database as a result of the job usage update automatically *and* rolling back any changes in the case of an exception
occurring during the update. It removes explicitly declaring the beginning of a TRANSACTION and committing changes after the update since the context manager handles this.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
+++ b/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
@@ -259,75 +259,75 @@ def update_job_usage(acct_conn, pdhl=1):
     )
     acct_conn.row_factory = sqlite3.Row
     cur = acct_conn.cursor()
-    # fetch timestamp of the end of the current half-life period
-    s_end_hl = """
-        SELECT end_half_life_period FROM t_half_life_period_table WHERE cluster='cluster'
+
+    with acct_conn:
+        # fetch timestamp of the end of the current half-life period
+        s_end_hl = """
+            SELECT end_half_life_period FROM t_half_life_period_table WHERE cluster='cluster'
+            """
+        cur.execute(s_end_hl)
+        row = cur.fetchone()
+        end_hl = row[0]
+
+        # begin transaction for all of the updates in the DB
+        acct_conn.execute("BEGIN TRANSACTION")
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
I think you could use the connection as a context manager instead of having such a big try/except block? https://docs.python.org/3/library/sqlite3.html#how-to-use-the-connection-context-manager . And then maybe catch exceptions and log errors from the caller of the function? Up to you though.

Otherwise looks good.

cmoussa1:
I could be wrong, but I think if I use the connection as a context manager, I still have to wrap the entire block inside of the `with connection:` statement, yes? I guess the benefit of doing this, though, is that it will automatically handle the rollback and commit of the transaction.

jameshcorbett:
Yeah, it wouldn't save you any indentation. It might be a bit easier to read though?

jameshcorbett (approved):
Looks great! I'm not sure whether or not the "BEGIN TRANSACTION" might still be needed, it depends on how the `acct_conn` is initialized with `isolation_level`: https://docs.python.org/3.9/library/sqlite3.html#controlling-transactions. But I think if you default-initialize it it should be fine.

cmoussa1:
Oh, I think you're absolutely right:

> The underlying sqlite3 library operates in autocommit mode by default, but the Python [sqlite3](https://docs.python.org/3.9/library/sqlite3.html#module-sqlite3) module by default does not.
>
> autocommit mode means that statements that modify the database take effect immediately. A BEGIN or SAVEPOINT statement disables autocommit mode, and a COMMIT, a ROLLBACK, or a RELEASE that ends the outermost transaction, turns autocommit mode back on.

I think leaving the `BEGIN TRANSACTION` is the right call here. We probably want to keep a transaction here because of the amount of potential writes to `association_table`, so I'll go ahead and add that back in.

PR #731, opened 2025-08-28
python: add general utility file for duplicate function definitions, general helper functions

cmoussa1:
#### Problem

Some of the Python files in flux-accounting have duplicate function definitions across their files, which makes them harder to
maintain in case the definition ever changes.

---

This PR creates a new general-use utility file that stores general helper functions that can be used across the different flux-accounting Python files. It adds the set of functions that are duplicative (the functions pertaining to fetching and setting userid and usernames) to this file. Finally, it removes the duplicate definitions from some of the Python files themselves in favor of just `import`-ing the new utility file and calling the functions from there.

### DIFF
--- a/t/python/t1006_job_archive.py
+++ b/t/python/t1006_job_archive.py
@@ -15,6 +15,7 @@ import sqlite3
 import time
 import sys
 from collections import defaultdict
+from collections import namedtuple
 
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


src/bindings/python/fluxacct/accounting/util.py
```diff
@@ -0,0 +1,60 @@
...
+
+        try:
+            if isinstance(fetched_uid, int):
+                uid = fetched_uid
+            else:
+                raise KeyError
+        except KeyError:
+            uid = 65534
+
+    return uid
```
jameshcorbett, line 60: I don't understand this function and I don't think the docstring is accurate. It does nothing unless the passed uid is `65534`? Also it raises a `KeyError` just to catch it immediately?

jameshcorbett (approved):
Looks good, one comment/question

jameshcorbett (approved):
LGTM!

PR #734, opened 2025-09-03
plugin: don't initialize "queues" map when checking limits

### LINKED ISSUE #733: unexpected 'queue not valid for user' errors with flux accounting 0.49
In earlier versions of flux accounting, if a queue wasn't defined in the flux accounting db, any user could submit to it. After updating tioga to 0.49, users started reporting that they were getting 'queue not valid for user' errors for the default queue. I've worked around this by adding the queue to the flux accounting db and adding all of the users to it.

I was able to reproduce the issue on fluke, but it's weird. `flux account list-queues` and `view-queues` confirm that no queues are defined in the accounting db, but I get 'queue not valid for user' errors when I try to `flux run`. If I run 'flux account-priority-update', I'm able to run one job, but then subsequent jobs get the 'queue not valid for user' errors. e.g.

```
[day36@fluke1:~]$ flux run -n1 hostname
flux-run: ERROR: Queue not valid for user: batch
[day36@fluke1:~]$ sudo -u flux flux account-priority-update
[day36@fluke1:~]$ flux run -n1 hostname
flux-job: fpX9pubPLeb started                                                                                                                                                                    00:00:02
fluke126
[day36@fluke1:~]$ flux run -n1 hostname
flux-run: ERROR: Queue not valid for user: batch
[day36@fluke1:~]$
```

cmoussa1:
#### Problem

As mentioned in #733, when checking to see if an association is under a queue's max resources limit for a particular job, the "queues" map gets initialized with a `Queue` object unintentionally (especially if queues are not configured in the flux-accounting DB), and subsequently submitted jobs get rejected with a `"Queue not valid for user"` error message.

---

This PR reworks the `under_queue_max_resources ()` method to perform a safer lookup of the "queues" map with the passed-in queue instead of passing the queue directly in the map (which will initialize the map if the queue does not already exist). It adjusts the calls to this method to account for the change in which parameters are passed to `under_queue_max_resources ()`. I've added a simple test file to ensure that configured queues in Flux (particularly with a default queue) do not reject submitted jobs from associations when there are no queues configured in flux-accounting.

I've also added some minor cleanups to drop initialization of some of the `Association` class' queue attributes with empty strings in the case where queues are not configured in flux-accounting.

Fixes #733

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -219,6 +219,13 @@ void split_string_and_push_back (const char *list,
 }
 
 
+bool has_text (const char *s) {
+    if (!s) return false;
+    while (*s && std::isspace (static_cast<unsigned char> (*s))) ++s;
+    return *s != '\0';
+};
+
+
 int get_queue_info (char *queue,
                     const std::vector<std::string> &permissible_queues,
                     const std::map<std::string, Queue> &queues)
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -332,11 +332,22 @@ bool Association::under_max_resources (const Job &job)
     return under_max_resources;
 }
 
-bool Association::under_queue_max_resources (const Job &job,
-                                             const Queue &queue)
+bool Association::under_queue_max_resources (
+                                    const Job &job,
+                                    const std::string &queue,
+                                    const std::map<std::string, Queue> &queues)
 {
-    bool under_max_nodes = (queue_usage[queue.name].cur_nodes + job.nnodes)
-                           <= queue.max_nodes_per_assoc;
+    auto qit = queues.find (queue);
+    if (qit == queues.end ())
+        // queue is unknown to flux-accounting; skip check
+        return true;
+    const int queue_max_nodes_per_assoc = qit->second.max_nodes_per_assoc;
 
-    return under_max_nodes;
+    // look up current per-queue node usage for the association
+    int cur_nodes_in_queue = 0;
+    auto uit = queue_usage.find (queue);
+    if (uit != queue_usage.end ())
+        cur_nodes_in_queue = uit->second.cur_nodes;
+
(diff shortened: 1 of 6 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -241,8 +241,9 @@ static int increment_resources (Association *b,
     b->cur_cores = b->cur_cores + (counts.nslots * counts.slot_size);
 
     // increment cur_nodes for queue
-    b->queue_usage[queue].cur_nodes = b->queue_usage[queue].cur_nodes +
-                                      counts.nnodes;
+    if (queue != "")
+        b->queue_usage[queue].cur_nodes = b->queue_usage[queue].cur_nodes +
+                                          counts.nnodes;
```
grondo, line 246: Definitely not a C++ expert, but would it be more idiomatic (and potentially a little faster) to use `!queue.empty()` here?

src/plugins/accounting.cpp
```diff
@@ -332,11 +332,22 @@ bool Association::under_max_resources (const Job &job)
     return under_max_resources;
 }
 
-bool Association::under_queue_max_resources (const Job &job,
-                                             const Queue &queue)
+bool Association::under_queue_max_resources (
+                                    const Job &job,
+                                    const std::string &queue,
+                                    const std::map<std::string, Queue> queues)
```
grondo, line 338: Another C++ newbie question: should `queues` be passed by reference here to avoid a copy?

grondo (approved):
LGTM! A couple questions inline.

jameshcorbett (approved):
Looks reasonable!

src/plugins/mf_priority.cpp
cmoussa1: I think you're absolutely right, thanks for this suggestion. I'll change to just use `.empty ()`.

cmoussa1:
Thank you for the helpful feedback here @grondo and @jameshcorbett - I believe I've addressed your feedback here, so I'll go ahead and set MWP

PR #737, opened 2025-09-05
plugin: add more information returned in output of `flux jobtap query`

### LINKED ISSUE #736: plugin: return stored queues, projects information in `jobtap query` callback
It would be nice if the plugin could return the information it has about any known queues or projects it has stored in the output of `flux jobtap query` for the multi-factor priority plugin. This would make debugging any potential hiccups related to what info the plugin has a little easier in the future.

cmoussa1:
#### Problem

The information stored in the priority plugin's `queues` map and `projects` vector are not returned in the output of `flux jobtap query`, which makes it difficult to diagnose what queue and/or project information the plugin has on hand without inserting print statements directly within the plugin.

---

This PR adds the information stored in the `queues` map and the `projects` vector to the information returned in the output of `flux jobtap query`. It adds helper functions to construct and pack the objects into JSON objects, similar to how `Association` objects are already constructed for the output of `flux jobtap query`.

I've added some basic tests to check for both queue and project information in the output of `flux jobtap query`.

Fixes #736

### DIFF
--- a/src/plugins/accounting.cpp
+++ b/src/plugins/accounting.cpp
@@ -351,3 +351,37 @@ bool Association::under_queue_max_resources (
 
     return (cur_nodes_in_queue + job.nnodes) <= queue_max_nodes_per_assoc;
 }
+
+json_t* convert_queues_to_json (const std::map<std::string, Queue> &queues)
+{
+    json_t *root = json_object ();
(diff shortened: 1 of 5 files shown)


cmoussa1:

PR #742, opened 2025-09-12
`view-job-records`: add filter for duration delta of jobs

### LINKED ISSUE #738: add filter for duration delta in `view-job-records` command
Now that the `jobs` table stores both requested and actual duration for jobs, it would be nice if there was a convenient way to view the difference between the two in its own column in the `view-job-records` command, as well as a way to filter for jobs of a certain duration delta.

cmoussa1:
#### Problem

There is no convenient way to get the difference between the requested duration and the actual duration of jobs in the `jobs` table without manually calculating the difference yourself. There is also no way to filter the `jobs` table by this duration delta.

---

This PR adds a new property to the `JobRecord` objects that populate the `jobs` table: **duration_delta**, which is just the difference between a job's requested duration and its actual duration. Similar to the `--requested-duration` and `--actual-duration` fields, the `jobs` table can be filtered by the duration delta to look for jobs that fit a certain criteria in relation to an expression passed in (e.g. jobs with a duration delta greater than 30 seconds, or jobs with a duration delta less than 30 seconds).

Some basic tests are added to test filtering the `jobs` table by this field, and the `view-job-records(1)` page is updated with the newly-added option.

I also snuck in a minor commit to fix an incorrect option in one of the help messages for another option in the `view-job-records` command as well as add a missing period to one of the descriptions of the options in `view-job-records(1)`.

Fixes #738

### DIFF
--- a/doc/man1/flux-account-view-job-records.rst
+++ b/doc/man1/flux-account-view-job-records.rst
@@ -59,9 +59,9 @@ job records.
 .. option:: -e/--actual-duration
 
     Return jobs that fit a certain actual duration's criteria by passing an
-    expression Multiple filters can be passed to refine the ``actual_duration``
-    column by. Expressions can be started with any of the following operators:
-    ``<``, ``<=``, ``=``, ``>=``, or ``>``.
+    expression. Multiple filters can be passed to refine the
+    ``actual_duration`` column by. Expressions can be started with any of the
+    following operators: ``<``, ``<=``, ``=``, ``>=``, or ``>``.
 
 .. option:: -D/--duration-delta
 
--- a/src/cmd/flux-account.py
+++ b/src/cmd/flux-account.py
@@ -480,7 +480,7 @@ def add_view_job_records_arg(subparsers):
         nargs="+",
         help=(
             "the actual duration for a job; multiple expressions can be passed as "
-            "filters, e.g. -d '< 60' '> 120'"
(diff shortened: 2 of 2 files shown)


doc/man1/flux-account-view-job-records.rst
```diff
@@ -63,6 +63,13 @@ job records.
     column by. Expressions can be started with any of the following operators:
     ``<``, ``<=``, ``=``, ``>=``, or ``>``.
 
+.. option:: -D/--duration-delta
+
+    Return jobs that fit a certain duration delta's criteria by passing an
+    expression. Multiple filters can be passed to refine the ``duration_delta``
+    column by. Expressions can be started with any of the following operators:
+    ``<``, ``<=``, ``=``, ``>=``, or ``>``.
+
```
jameshcorbett: It might be nice to provide an example here?

jameshcorbett (approved):
Looks great! One suggestion.

doc/man1/flux-account-view-job-records.rst
cmoussa1: Great suggestion! I went ahead and force-pushed up a change to add a short example at the bottom of the man page.

cmoussa1:
Thanks @jameshcorbett! I've force-pushed up a fix to the man page to add an example based on your suggestion above. Will set MWP here

PR #743, opened 2025-09-14
testsuite: stop using flux job cancelall

garlick:
Problem: flux job cancelall is deprecated but flux accounting is still using an old rc1-job in the test suite that uses it.

Update rc1-job.


garlick:
Thanks! I'll go head and set MWP on this.

PR #751, opened 2025-09-18
`view-user`/`edit-user`: use `"unlimited"` for to display and set `max_nodes` and `max_cores` to their default values

### LINKED ISSUE #748: `view-user`: display `MAX` instead of actual integer when showing an association's limits
The `view-user` command displays the actual integer (`2147483647`) for an association's `max_nodes` or `max_cores` limits, but it would probably be a lot cleaner and more intuitive if it displayed a string like `MAX` or `UNLIMITED` when an association's limits are set to this value.

cmoussa1:
#### Problem

The `view-user` command displays the actual integer (2147483647) for an association's `max_nodes` or `max_cores` limit(s), but it would be a lot cleaner and more intuitive if it displayed a string when an association's limits are set to this value.

The `max_nodes` and `max_cores` do not have an easy way to specify the max integer for either field for an association other than
specifying `"-1"`, which isn't totally backwards compatible with much older versions of flux-accounting databases where `NOT NULL` constraints are properly handled for either field.

---

This PR adds a helper function to `util.py` to check the values being output for the max integer for `max_nodes` and `max_cores` and replaces it with `unlimited` if so. It uses this function when displaying output both in JSON and in table format.

It also adds a `"unlimited"` keyword for the `max_nodes` and `max_cores` fields in the `edit-user` command, which allows a user to set either field to the large integer without having to specify it on the command line.

I've added some basic tests to `t1007-flux-account-users.t` for both of these cases.

Fixes #748


src/bindings/python/fluxacct/accounting/util.py
```diff
@@ -54,3 +54,15 @@ def parse_timestamp(timestamp):
...
+
+def format_value(val):
+    """
+    Replace a max value in the database with a string.
+
+    Args:
+        val: the value being evaluated
+    """
+    if isinstance(val, int) and val == 2147483647:
+        return "MAX"
```
jameshcorbett, line 67: Could this number be saved as a constant somewhere, like `INTEGER_MAX`? Also is the `isinstance` check accomplishing anything here?

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -556,6 +556,16 @@ def edit_user(conn, username, bank=None, **kwargs):
                     clear_queues(conn, username, bank)
                 elif field == "projects":
                     clear_projects(conn, username, bank)
+                elif field in ("max_nodes", "max_cores"):
+                    # set value to max integer
+                    update_stmt = f"UPDATE association_table SET {field}=2147483647 WHERE username=?"
+                    tup = (username,)
```
jameshcorbett, line 562: Another place where I feel like it would be clearer to have a named constant rather than the integer directly.

jameshcorbett (approved):
LGTM, some minor comments. I wonder if it might be clearer to users if the string they saw were "INF" rather than "MAX." If I ran `view-user` on myself and I saw "MAX" as the value for my `max_nodes`, I might wonder something like, "well, what is the MAX though? Maybe it's 16 or something." Whereas I think if it were "INF" I would know right away I had no limit to my max nodes.

On the other hand there might be some users who don't know that "INF" stands for "infinity," whereas nobody would be confused about what "MAX" means.

src/bindings/python/fluxacct/accounting/util.py
cmoussa1: Ah, this is a great suggestion. A constant would probably be cleaner to read here and elsewhere. Thanks for this.

> is the isinstance check accomplishing anything here?

🤦 I don't believe it is; I could probably just remove it. Thanks!

cmoussa1:
Thanks @jameshcorbett - yeah, I think I see your point here with `"MAX"` being potentially less clear than something like `"INF"`. Both have their benefits. To me, the implication with `"MAX"` feels like it would be set to a value where I wouldn't really even need to think about it, but your point of potentially wanting to dig further to see what "max" really is 100% valid.

I wonder if maybe we should just ditch a three-character abbreviation altogether and go with something like `"unlimited"`? That might be a little clearer. That was another good suggestion made by @grondo - what do you think?

jameshcorbett (approved):
LGTM!

cmoussa1:

PR #752, opened 2025-09-19
plugin: check dependencies of held jobs after `flux account-priority-update`

### LINKED ISSUE #749: plugin: check that jobs in `DEPEND` state meet their requirements to have dependencies lifted after an update
The priority plugin does not consider that jobs held in `DEPEND` state could potentially have their limits removed if an association's limits have changed after an update. When `flux account-priority-update` is called, all jobs in `SCHED` or `PRIORITY` are reprioritized with a call to `flux_jobtap_reprioritize_all()`:

```c
/*  Start a loop to re-prioritize all jobs. The plugin "priority.get"
 *   callback will be called for each job currently in SCHED or
 *   PRIORITY states.
 */
int flux_jobtap_reprioritize_all (flux_plugin_t *p);
```

The plugin should probably have a way to look at all associations' jobs in `DEPEND` and see if they meet the requirements to have their dependencies lifted after an update.

cmoussa1:
#### Problem

The priority plugin does not consider that jobs held in `DEPEND` state could potentially have their limits removed if an association's or queue's limits have changed after an update. So, a job held in `DEPEND` will still remain in `DEPEND` even if it meets all of the conditions to run after a change to limits.

---

This PR adds an iteration through the plugin's internal map of associations when `flux account-priority-update` is called and checks the held jobs of every association to see if they meet any and all of the requirements to have its dependencies removed.

I've also added some tests that walk through each flux-accounting dependency and makes sure that a job held in `DEPEND` for each of these limits can proceed to run after changing the limits and pushing to the plugin with `flux account-priority-update`.

Fixes #749

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -295,6 +295,7 @@ static int decrement_resources (Association *b,
 static int check_and_release_held_jobs (flux_plugin_t *p, Association *b)
 {
     std::string dependency;
+    flux_jobid_t held_job_id;
     // the Association has at least one held Job; begin looping through
     // held Jobs and see if they satisfy the requirements to be released
     auto it = b->held_jobs.begin ();
@@ -310,6 +311,7 @@ static int check_and_release_held_jobs (flux_plugin_t *p, Association *b)
                                                held_job.id,
                                                D_QUEUE_MRJ) < 0) {
                 dependency = D_QUEUE_MRJ;
+                held_job_id = held_job.id;
                 goto error;
             }
             held_job.remove_dep (D_QUEUE_MRJ);
@@ -322,6 +324,7 @@ static int check_and_release_held_jobs (flux_plugin_t *p, Association *b)
                                                held_job.id,
                                                D_QUEUE_MRES) < 0) {
                 dependency = D_QUEUE_MRES;
+                held_job_id = held_job.id;
                 goto error;
             }
             held_job.remove_dep (D_QUEUE_MRES);
@@ -344,6 +347,7 @@ static int check_and_release_held_jobs (flux_plugin_t *p, Association *b)
                                                held_job.id,
                                                D_ASSOC_MRES) < 0) {
                 dependency = D_ASSOC_MRES;
+                held_job_id = held_job.id;
                 goto error;
             }
             held_job.remove_dep (D_ASSOC_MRES);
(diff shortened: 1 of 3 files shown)


src/plugins/mf_priority.cpp
```diff
@@ -368,9 +368,10 @@ static int check_and_release_held_jobs (flux_plugin_t *p, Association *b)
                                  held_job_id,
                                  "mf_priority",
                                  0,
-                                 "job.state.inactive: failed to remove %s "
-                                 "dependency from job",
-                                 dependency.c_str ());
+                                 "check_and_release_held_jobs: failed to "
+                                 "remove %s dependency from job %lld",
```
grondo, line 372: Suggestion: use `%ju` for `flux_jobid_t` format spec to avoid displaying negative numbers.

In the future you could look into pulling in `idf58.h` from flux-core's `src/common/libjob` which allows f58 jobids to easily be printed.

src/plugins/mf_priority.cpp
```diff
@@ -768,6 +768,16 @@ static void reprior_cb (flux_t *h,
...
     if (flux_respond (h, msg, NULL) < 0)
         flux_log_error (h, "flux_respond");
+
+    // iterate through map that stores associations and held job IDs; check to
+    // see if any previously-held jobs can now be released with the update
+    for (auto &entry: users) {
+        auto &banks = entry.second;
+
+        for (auto &bank_entry : banks) {
+            check_and_release_held_jobs (p, &bank_entry.second);
```
grondo, line 778: Should error code be checked here? If purposefully ignoring error code, I suggest casting the function to `void` to indicate this is intentional:

```c
    (void) check_and_release_held_jobs (p, &bank_entry.second);
```

t/t1070-issue749.t
```diff
@@ -0,0 +1,258 @@
...
+
+test_expect_success 'ensure previously held job proceeds to RUN' '
+	flux job wait-event -vt 5 \
+		--match-context=description="max-resources-queue" \
+		${job2} dependency-remove
+'
+
+test_expect_success 'cancel job(s)' '
+	flux cancel ${job1} &&
+	flux cancel ${job2}
```
grondo, line 251: No need to make a change here, but just fyi `flux cancel` can take multiple jobids on the command line:
```sh
    flux cancel ${job1} ${job2}
```

grondo (approved):
LGTM! Minor suggestions inline.

grondo:
Manually merged since mergify seemed to be confused.

PR #755, opened 2025-09-24
plugin: remove job from `held_jobs` if cancelled

### LINKED ISSUE #754: plugin: held jobs still kept in `Association` object even if held jobs are cancelled
I just ran a quick test in my Docker environment and I noticed that if a job is held due to a flux-accounting dependency and the held job is cancelled, its ID and dependency are still tracked in the `Association` object in the plugin. So, when a currently-running job is finished running, the plugin will still attempt to remove any dependencies from that held job, which is incorrect:

```
2025-09-24T20:10:29.429554Z job-manager.debug[0]: jobtap: mf_priority.so: job.state.inactive: rc=-1
2025-09-24T20:10:29.429559Z job-manager.err[0]: jobtap: job.state.inactive: callback returned error
```

If a held job is cancelled, it should also be removed from the `Association` object's list of held jobs.

cmoussa1:
#### Problem

The plugin does not remove a held `Job` object from an `Association`'s list of held jobs if that held job is cancelled. So, when a currently running job finishes and the association's held jobs are iterated through to see if they meet all requirements to have their dependencies removed, this fails (because the job doesn't exist anymore) and the plugin reports that `job.state.inactive` returns `-1`.

---

This PR adds a check in `job.state.inactive` to see if the held job that is cancelled also exists in the `Association`'s list of held jobs, and if it does, it removes it from the list of held jobs.

Basic tests are added to ensure that a held job that is cancelled no longer shows up in an `Association`'s list of held jobs.

Fixes #754

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -73,6 +73,7 @@ TESTSCRIPTS = \
 	t1068-mf-priority-queues-projects-query.t \
 	t1069-duration-delta.t \
 	t1070-issue749.t \
+	t1071-issue754.t \
 	t5000-valgrind.t \
 	python/t1000-example.py \
 	python/t1001_db.py \
--- a/t/t1071-issue754.t
+++ b/t/t1071-issue754.t
@@ -1,0 +1,173 @@
+#!/bin/bash
+
+test_description='test ensuring held jobs are removed from Association object after cancellation'
+
+. `dirname $0`/sharness.sh
+
+mkdir -p config
+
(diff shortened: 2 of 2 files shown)


cmoussa1:
Thanks @grondo! Setting MWP here. Here goes nothing 😆

cmoussa1:
Going to try setting MWP here after rebasing on top of #762.

grondo:
@mergifyio refresh

cmoussa1:
I just removed the merge-when-passing label but will hold off on closing this PR in case mergify support ends up looking at this PR.

PR #767, opened 2025-09-30
mergify: remove `queue_conditions`, `status-success` checks

cmoussa1:
####  Problem

The current Mergify configuration uses a two-step CI, which now always creates a temporary branch. This has put flux-accounting's Mergify configuration out-of-date, as approved PRs now fail to merge as a result of the `"validate commits"` check in merge_conditions, which fails because it tries to include a merge commit.

---

This PR drops the list of `merge_conditions` checks in favor of using branch protections directly within GitHub and copies the Mergify configuration in flux-core.

This will need a manual merge due to the updated `.mergify.yml` config.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,35 +1,8 @@
 queue_rules:
   - name: default
     batch_size: 1
-    queue_conditions:
-      - base=master
-      - label="merge-when-passing"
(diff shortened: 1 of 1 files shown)


grondo (approved):
Ok, lets try it! I'll press the button.

PR #768, opened 2025-09-30
python: explicitly close `Cursor` objects after use in user commands

cmoussa1:
#### Problem

`Cursor` objects are not explicitly closed in a number of the functions that are called that read from/write to the SQLite database. While not totally necessary (`Cursor` objects *should* be cleaned up by Python's garbage collectors), if there are lots of parallel processes that are reading and writing to the database (such as the flux-accounting service and the `priority-update` script), these `Cursor` objects can keep transactions alive and potentially lead to a database lock. It also would provide better code clarity if these functions actually explicitly cleaned up the use of these objects when the function completes.

---

This PR introduces the use of a context manager for the `Cursor` objects created in the Python bindings that interact with the flux-accounting database.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/user_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/user_subcommands.py
@@ -150,7 +150,7 @@ def update_default_bank(conn, cur, username):
         edit_user(conn, username, default_bank=new_default_bank)
 
 
-def update_mod_time(conn, username, bank):
+def update_mod_time(conn, cur, username, bank):
     mod_time_tup = (
         int(time.time()),
         username,
@@ -162,36 +162,36 @@ def update_mod_time(conn, username, bank):
     else:
         update_stmt = "UPDATE association_table SET mod_time=? WHERE username=?"
 
-    conn.execute(update_stmt, mod_time_tup)
+    cur.execute(update_stmt, mod_time_tup)
 
 
-def clear_queues(conn, username, bank=None):
+def clear_queues(conn, cur, username, bank=None):
     if bank is None:
-        conn.execute(
+        cur.execute(
             "UPDATE association_table SET queues='' WHERE username=?", (username,)
         )
     else:
-        conn.execute(
+        cur.execute(
             "UPDATE association_table SET queues='' WHERE username=? AND bank=?",
             (
                 username,
                 bank,
             ),
         )
-        update_mod_time(conn, username, bank)
+        update_mod_time(conn, cur, username, bank)
 
     conn.commit()
 
     return 0
 
 
-def clear_projects(conn, username, bank=None):
+def clear_projects(conn, cur, username, bank=None):
     update_stmt = "UPDATE association_table SET projects='*' WHERE username=?"
     if bank is None:
-        conn.execute(update_stmt, (username,))
+        cur.execute(update_stmt, (username,))
     else:
         update_stmt += " AND bank=?"
-        conn.execute(
+        cur.execute(
             update_stmt,
             (
                 username,
@@ -199,7 +199,7 @@ def clear_projects(conn, username, bank=None):
             ),
         )
 
-    update_mod_time(conn, username, bank)
+    update_mod_time(conn, cur, username, bank)
     conn.commit()
 
     return 0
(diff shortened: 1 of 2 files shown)


jameshcorbett (commented):
I wonder if it might be simpler to use the cursor as a context manager with [this](https://docs.python.org/3.13/library/contextlib.html#contextlib.closing)?

cmoussa1:
OK, I've reworked this PR to use a context manager instead of explicitly calling `.close()` in each of the functions.

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -460,31 +468,32 @@ def delete_user(conn, username, bank, force=False):
...
-    conn.commit()
+        conn.execute(
+            sql_stmt,
+            (
+                username,
+                bank,
+            ),
+        )
+        # commit changes
+        conn.commit()
```
jameshcorbett, line 485: This is not something introduced by this commit, but I find it a little puzzling that a cursor is created in this function but then it isn't used, and instead queries are run directly against the `Connection` object. I don't think it's a problem, just surprising.

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -346,103 +352,105 @@ def add_user(
...
+            """
+            INSERT OR IGNORE INTO job_usage_factor_table (username, userid, bank)
+            VALUES (?, ?, ?)
+            """,
+            (
+                username,
+                uid,
+                bank,
+            ),
+        )
```
jameshcorbett, line 448: Noticed another `Connection` object being used instead of a `Cursor`.

jameshcorbett (approved):
Looks reasonable to me. Do you prefer this to your previous version with `try/finally`?

I noticed the diff is really big because of all the indentation changes. If you're interested, I think it could be reduced with a decorator. If you added this decorator into a common location:

```python
import functools
import contextlib

def get_and_close_cursor(func):
    @functools.wraps(func)
    def wrapper(conn, *args, **kwargs):
        with contextlib.closing(conn.cursor()) as cursor:
            return func(cursor, *args, **kwargs)

    return wrapper


@get_and_close_cursor
def some_function(cursor, foo, bar):
    ...


some_function(sqlite3.connect(":memory:"), "foo", "bar")
```
You could just tag any function that starts by creating a cursor and ends by closing the cursor with this decorator. If some functions need the `Connection` object as well, that could be provided too. Just an idea. 

Drawback of this approach is that it could make code a bit harder to read, because functions would be invoked with a `Connection` object but the function signature would say it takes a `Cursor` object instead (or in addition, if you went that route).

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: No, you're absolutely right. This is a good catch! I should probably drop these extra creations of `cursor` objects, especially if they're not going to be used anywhere. Probably best in a short cleanup PR.

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -149,7 +149,8 @@ def update_default_bank(conn, cur, username):
     if len(result) > 0:
         # update user rows to have a new default bank (the next earliest user/bank created)
         new_default_bank = result[0][0]
-        edit_user(conn, username, default_bank=new_default_bank)
+        # pylint: disable=no-value-for-parameter
+        edit_user(conn, username=username, default_bank=new_default_bank)
```
jameshcorbett, line 150: Does the decorator confuse pylint about what the arguments are? That's unfortunate. 

Does the decorator force to provide `username` here as a keyword argument or did you make that change for pylint?

jameshcorbett (commented):
Looks fine to me. I think the choice of whether to use a context manager or `try/finally` and whether or not to use a decorator is mostly stylistic so if you like it and think it would make the code base easier to read and maintain you should proceed with it.

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: Yeah, unfortunately pylint reports an error here unless explicitly suppressed. I went ahead and provided `username` as a keyword argument to hopefully make the function call a little clearer, but I don't think it's entirely necessary to pass it as a keyword argument. Do you have a preference?

jameshcorbett:
🎉

Issue #650, opened 2025-05-30
limits: look into adding support for quota/queue limits based on compute hours


vsoch:
In cloud, it's common to want to limit a user's ability to run jobs on a specific instance type (e.g., one using GPUs). Speaking with @cmoussa1, we think the core metrics are already collecting / existing, and a little bit of extra exposure of configuration might make it work. Here are notes from our discussion:

- flux-accounting currently tracks job usage for a user by taking the product of the number of nnodes a job consumes and the duration of the job. This is similar to how compute hours are measured.
- flux-accounting measures and tracks this for both users and the banks they belong to
-  The total accumulated value is stored per-user in the `association_table` of the flux-accounting database (viewable with `flux account view-user`) and per-bank in the `bank_table` (viewable with `flux account view-bank`).

Thinking through design:

- When a user submits a job, with `--time` they can either explicitly set a duration, or it defaults to 0 (which is the cluster default, usually one hour)
- The time is converted into seconds and added to the jobspec at `.attributes.system.duration`
- When the job comes in, flux-accounting can assume that the job will use the total of the duration that is requested. There would need to be two values stored - the total _actual_ usage (the nnodes multipled by actual duration) and then _anticipated_ duration (nnodes multiplied by expected maximum time). When job completes, we would need to subtract the initial anticipated time, and add the actual time. This would be done so a user couldn't get away with submitting a huge number of jobs that would (eventually) go over their total quota.
- I am assuming we would add some values for limits to the plugin configuration, and it likely would be based on user, queue, or a combination of user/queue (e.g., "User X can only use this many GPU hours on Cluster X"
- I am assuming there is some cleanup / post-process at the end that could perform the update to the actual vs. estimated hours used.

If we can get this working, I want to test this in a cloud environment, and have flux-accounting running as a service for multiple flux clusters (MiniClusters) to use. That is the strongest cloud use case, for HPC in Kubernetes, as I see it.

PR #770, opened 2025-10-08
[WIP] plugin: add new jobtap plugin to track job usage across associations' jobs

cmoussa1:
#### Problem

As mentioned by @vsoch in #650, there is currently no way to enforce a limit on an association's ability to run jobs on a certain instance type, which flux-accounting already tracks in its SQLite database by calculating job usage (which is a product of the number of nodes allocated to a job and its duration). However, there is currently no way to enforce this using flux-accounting's mf_priority jobtap plugin.

---

This PR begins to lay the groundwork for this "usage" limit enforcement by adding a new jobtap plugin (called `compute_hours_limits`) which, for now, just tracks an association's current usage across all of their running jobs (by calculating the job's anticipated usage by multiplying the job's `nnodes` by its requested duration) and computes the jobs _actual_ usage when the job completes. The workflow looks something like this:

The user submits a job and specifies its size and duration (or a default duration is set on the job):

```console
$ flux submit -N4 -S duration=3600 my_job
```

The plugin takes these resource specifications and calculates an _expected usage_ by multiplying the both of these numbers together:

```c++
job->expected_usage = counts.nnodes * duration; 
```

When the job transitions to `RUN`, the `expected_usage` is added to the association's `current_usage` attribute:

```c++
current_usage += job->expected_usage;
```

When the job completes, the job's _actual usage_ is calculated and added to the association's `total_usage` attribute:

```c++
total_usage += (job->nnodes * (t_inactive - job->t_run));
```

and the _expected usage_ from the job is subtracted from the association's `current_usage` attribute:

```c++
current_usage -= job->expected_usage;
```

The associations' `total_usage` attributes can be reset to `0.0` by sending a `"clear"` rpc to the plugin:

```python
flux.Flux().rpc("job-manager.compute_hours_limits.clear")
```

To avoid making the plugin code a lot to review, I've only added tracking of an association's job usage in this PR. If this looks like the right way we want to track an association's usage (and eventually enforce a limit), I can submit follow-up PRs to add that functionality.

I've added some basic tests to showcase how the plugin tracks job usage for a given association by submitting one or multiple jobs and ensuring the `current_usage` and `total_usage` values are calculated correctly.

---

#### still some things to take care of

- [x] not tracking a job's usage if it is cancelled before it ever runs
- [x] removing a `Job` object after it has transitioned to `job.state.inactive`
- [x] updating (resetting?) an association's `total_usage` value so it doesn't just increase indefinitely

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -74,6 +74,7 @@ TESTSCRIPTS = \
 	t1069-duration-delta.t \
 	t1070-issue749.t \
 	t1071-issue754.t \
+	t1072-compute-hours-limits-basic.t \
 	t5000-valgrind.t \
 	python/t1000-example.py \
 	python/t1001_db.py \
--- a/t/t1072-compute-hours-limits-basic.t
+++ b/t/t1072-compute-hours-limits-basic.t
@@ -1,0 +1,252 @@
+#!/bin/bash
+
+test_description='test loading compute hours limit plugin'
+
+. `dirname $0`/sharness.sh
+
+COMPUTE_HOURS_LIMITS=${FLUX_BUILD_DIR}/src/plugins/.libs/compute_hours_limits.so
+DB_PATH=$(pwd)/FluxAccountingTest.db
+SUBMIT_AS=${SHARNESS_TEST_SRCDIR}/scripts/submit_as.py
+
+mkdir -p config
+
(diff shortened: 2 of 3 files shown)


cmoussa1:
Going to close this PR in light of some really great work being done elsewhere to support custom resource type configuration in both TOML files and in another custom jobtap plugin.

PR #774, opened 2025-10-22
`fetch-job-records`: check if `t_run` exists

### LINKED ISSUE #773: `fetch-job-records`: handle case where job does not have `t_run` timestamp
There's a line in `fetch-job-records` that tries to compute the actual duration for a job that may not have run, and thus, raises an Exception that it can't perform a math operation on a `NoneType` object:

```python
# compute actual job duration
single_record["actual_duration"] = single_job.get("t_inactive") - single_job.get("t_run")

...

"TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'"
```

This script needs to be able to handle the case where a job does not have `t_run` when trying to compute the actual duration for the job.

cmoussa1:
#### Problem

The `fetch-job-records` script tries to compute the actual duration for a job by taking the difference between the `t_inactive` and `t_run` timestamps. If a job never receives a `t_run` timestamp (i.e. never runs), then the calculation of the actual_duration fails because `job.get("t_run")` returns `None`, which causes the script to raise a `TypeError` and halt execution, preventing job records from being inserted into the `jobs` table.

---

This PR initializes the `"actual_duration"` key-value pair to `0.0`. Check that both `t_inactive` and `t_run` are not `None` before calculating the `actual_duration` value for the job.

Fixes #773

### DIFF
--- a/t/t1011-job-archive-interface.t
+++ b/t/t1011-job-archive-interface.t
@@ -81,6 +81,40 @@ test_expect_success 'check that no jobs show up under user' '
 	grep -f no_jobs.expected no_jobs.test
 '
 
+# This set of four tests ensures that a job has an R in the KVS but no "alloc"
+# event, and therefore the job did not enter the RUN state and has no t_run
+# timestamp. We put a fake R in the KVS after cancelling a job to simulate this
+# condition, which may occur in practice when a job is cancelled between after
+# the KVS commit of R is started but before it completes and the "alloc" event
+# is posted (see https://github.com/flux-framework/flux-core/issues/7172).
+test_expect_success 'submit a job and update its R to have no t_run timestamp' '
+	job=$(flux submit --urgency=0 hostname) &&
+	flux job wait-event -vt 10 ${job} priority &&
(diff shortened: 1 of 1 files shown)


t/t1011-job-archive-interface.t
```diff
@@ -81,6 +81,37 @@ test_expect_success 'check that no jobs show up under user' '
 	grep -f no_jobs.expected no_jobs.test
 '
 
+# This set of four tests ensures that a job that has no 'alloc' event and
+# therefore no 't_run' timestamp is properly handled by the fetch-job-records
+# command.
```
grondo: Suggestion: Amend the comment to be clear we're testing the case where a job has an _R_ in the KVS but no `alloc` event, and therefore the job did not enter the RUN state and has no `t_run` timestamp. We put a fake R in the KVS after canceling a job to simulate this condition, which may occur in practice when a job is canceled between after the KVS commit of R is started but before it completes and the `alloc` event is posted.
(See https://github.com/flux-framework/flux-core/issues/7172)

PR #775, opened 2025-10-23
commands: add `sync-userids` command

cmoussa1:
#### Problem

There is a bug in flux-accounting < `v0.50.0` where the userid for an association is different in the `association_table` and in the `job_usage_factor_table`, which prevents the association's job usage values from being updated. This can quickly be remedied by running a quick query to see which userids are inconsistent and then another one to make them consistent across both tables, but it involves launching into an interactive SQLite shell and executing the queries. It would be useful if this could be done from the command line for an admin.

---

This PR adds a new command to the flux-accounting command suite to make the userids in the `job_usage_factor_table` consistent with what is defined in the `association_table` by executing an `UPDATE` statement.

I've also added some basic tests to check that the command will accurately update the userids to make them consistent across both tables.

### DIFF
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -74,6 +74,7 @@ TESTSCRIPTS = \
 	t1069-duration-delta.t \
 	t1070-issue749.t \
 	t1071-issue754.t \
+	t1072-sync-uids.t \
 	t5000-valgrind.t \
 	python/t1000-example.py \
 	python/t1001_db.py \
--- a/t/t1072-sync-uids.t
+++ b/t/t1072-sync-uids.t
@@ -1,0 +1,162 @@
+#!/bin/bash
+
+test_description='test synchronizing userids across multiple tables'
+
+. `dirname $0`/sharness.sh
+
+mkdir -p config
+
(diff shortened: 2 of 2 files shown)


jameshcorbett (approved):
Looks good to me, how does the inconsistency happen in the first place though?

cmoussa1:
That is what I am trying to figure out for the moment 😓 this has been reported only a couple of times, and the debugging I've done hasn't revealed anything obvious to me (yet). This command is just a quick fix for an admin to be able to run in the case they are not comfortable with logging into an interactive `sqlite3` shell and executing SQL queries while I continue to explore a potential cause for the inconsistency.

cmoussa1:
Actually, I think I might have just discovered it 😆 before `v0.50.0`, the `set_uid()` function (now deprecated with #731) caused associations who were added without specifying a userid to have a default userid of `65534` in `job_usage_factor_table`. #731 fixed this bug, but associations added to the flux-accounting DB on clusters before `v0.50.0` was installed were still affected by this issue. So here's to hoping we don't see it anymore. Setting MWP here!

PR #781, opened 2025-11-10
plugin: add ability to initialize plugin with database information on load

### LINKED ISSUE #772: plugin: jobs held in DEPEND not checked when plugin is reloaded
I got a report a couple days ago that some jobs were stuck in `DEPEND` state with some flux-accounting limits even when the association was well under their limits for the queue. After some digging and some offline conversations, I believe one of the reasons these jobs can get stuck in `DEPEND` is if the jobtap plugin is reloaded (and/or Flux is restarted?), the previously-held jobs are not rechecked to see if they meet any of the requirements to have their dependencies removed.

There probably should be a check in the callback for `job.new` to see if the job is in `DEPEND` state, and if so, check to see if it meets any or all of the requirements to have its dependencies removed and be allowed to transition to `RUN`.

cmoussa1:
#### Problem

The priority plugin relies on keeping track of an association's active job and resource counts by associating an aux item with each job. This aux item gets removed when the plugin is unloaded and has to be re-added when the plugin is loaded back in, and the flux-accounting database information isn't immediately sent and unpacked when the plugin itself is loaded; if this is the case, the plugin sets a temporary aux item on the active job while it waits for information from the database, which can make keeping count of an association's active and running jobs difficult, as well as managing dependencies on jobs while it waits for database information to come through.

---

This PR adds the ability for the priority plugin to accept an optional `config` key-value pair to initialize the priority plugin with the necessary information from the flux-accounting database _when it is loaded_, therefore enabling it to accurately track an association's active jobs _before_ it rediscovers all of the active jobs in `job.new` and no longer needing a temporary aux item on those active jobs while it waits for information from the database.

To configure the plugin with this accounting database information, the output of a new command, `flux account export-json` can be passed as the value to the `config` key. Helper functions are added to `accounting.cpp` to extract each of the key-value pairs within `config` and initialize the plugin's internal data structures used to keep track of job and resource counts per-association and per-queue.

Basic tests are added to check that the priority plugin can either be loaded with or without the `config` key and can accurately keep track of an association's active jobs when it is loaded with the `config` key without needing a separate update via `flux account-priority-update`.

I have not added support for managing dependencies for rediscovered jobs on a plugin reload just yet in order to first make sure that this approach is sound. But, now that flux-framework/flux-core#7193 has landed, I can either work on a separate PR to add that support (and close #772) or add on additional commits here.

### DIFF
--- a/doc/guide/accounting-guide.rst
+++ b/doc/guide/accounting-guide.rst
@@ -130,8 +130,17 @@ configure the ``job-manager`` to load this plugin.
 
 See also: :core:man5:`flux-config-job-manager`.
 
-The plugin can also be manually loaded with ``flux jobtap load``. Be sure to
-send all flux-accounting data to the plugin after it is loaded:
+The plugin can also be manually loaded with ``flux jobtap load``. The data
+required by the priority plugin to track jobs can be sent when the plugin is
+first loaded by setting a ``config`` key to the output of the ``flux account
+export json`` command:
+
+.. code-block:: console
+
+ $ flux jobtap load mf_priority.so config="$(flux account export-json)"
+
+Otherwise, be sure to send all flux-accounting data to the plugin after it is
+loaded:
 
 .. code-block:: console
 


grondo:
What is the expected maximum size of the `conf` object generated by the `export-json` command? There's an upper limit on the size of command line arguments so I'm afraid on systems with a lot of associations, we could overflow that. I wonder if the `flux jobtap load` should allow the config to be supplied on stdin or a file as well as command line args (which are really meant for one or two config items...)

src/plugins/accounting.cpp
```diff
@@ -404,3 +404,253 @@ json_t* convert_projects_to_json (const std::vector<std::string> projects)
...
+                            "max_running_jobs", &max_running_jobs,
+                            "max_active_jobs", &max_active_jobs,
+                            "queues", &assoc_queues,
+                            "active", &active,
+                            "projects", &assoc_projects,
+                            "def_project", &def_project,
+                            "max_nodes", &max_nodes,
+                            "max_cores", &max_cores) < 0) {
+            return -1;
+        }
```
grondo: The `json_error_t` is unused here - should any jansson error be returned to the caller or logged here? (Same goes for the other unpacks)

src/plugins/accounting.cpp
```diff
@@ -404,3 +404,253 @@ json_t* convert_projects_to_json (const std::vector<std::string> projects)
 
     return known_projects;
 }
+
+int load_associations (
+                    json_t *data,
+                    std::map<int, std::map<std::string, Association>> &users,
+                    std::map<int, std::string> &users_def_bank)
+{
+    char *bank, *def_bank, *assoc_queues, *assoc_projects, *def_project = NULL;
```
grondo: These should all be `const char *`?

src/plugins/accounting.cpp
```diff
@@ -404,3 +404,253 @@ json_t* convert_projects_to_json (const std::vector<std::string> projects)
...
+        users_def_bank[uid] = def_bank;
+    }
+
+    return 0;
+}
+
+
+int load_queues (json_t *data, std::map<std::string, Queue> &queues)
+{
+    char *queue = NULL;
```
grondo: Should be `const char *`?

src/plugins/accounting.cpp
```diff
@@ -404,3 +404,253 @@ json_t* convert_projects_to_json (const std::vector<std::string> projects)
...
+        q->max_nodes_per_assoc = max_nodes_per_assoc;
+    }
+
+    return 0;
+}
+
+
+int load_projects (json_t *data, std::vector<std::string> &projects)
+{
+    char *project = NULL;
```
grondo: `const char *project`?

src/plugins/accounting.cpp
```diff
@@ -404,3 +404,253 @@ json_t* convert_projects_to_json (const std::vector<std::string> projects)
...
+    }
+
+    return 0;
+}
+
+
+int load_priority_factors (json_t *data,
+                           std::map<std::string, int> &priority_weights)
+{
+    char *factor = NULL;
```
grondo: another `const char *`? :)

src/plugins/accounting.cpp
cmoussa1: Yeah, this is a good call. Thanks for pointing this out; I think maybe I should try to just log an error with `flux_log_error`, perhaps to `stderr` since this error would be occurring during plugin load - what do you think?

grondo: I don't think logging to `stderr` is appropriate. Logging an error to the flux log seems right.

PR #782, opened 2025-11-14
test: add `el9` image to CI

cmoussa1:
#### Problem

The flux-accounting project does not have `el9` in its matrix of images built when running through CI.

---

This PR adds it.


cmoussa1:
Thanks @jameshcorbett!

PR #785, opened 2025-11-25
plugin: set load order of active jobs during initialization

### LINKED ISSUE #772: plugin: jobs held in DEPEND not checked when plugin is reloaded

cmoussa1:
#### Problem

The order in which jobs are seen when the plugin is loaded is not guaranteed *unless* a sort mode is passed when initializing the plugin (see [`flux-jobtap-plugins(7)`](https://github.com/flux-framework/flux-core/blob/master/doc/man7/flux-jobtap-plugins.rst)), which is required in order to accurately track active jobs and their flux-accounting dependencies after the priority plugin is reloaded.

---

This PR unpacks an optional `"sort"` key when loading the plugin, and if unpacked, calls, `flux_jobtap_set_load_order ()` to see jobs by reverse state so that it can accurately track and enforce limits and job/resource counts per-association and per-queue.

Basic tests are added that walk through each flux-accounting dependency and makes sure that the order correctly affects how jobs are tracked and held across a restart.

Fixes #772

### DIFF
--- a/config/ax_flux_core.m4
+++ b/config/ax_flux_core.m4
@@ -61,7 +61,7 @@ AC_DEFUN([AX_FLUX_CORE], [
   saved_PKG_CONFIG_PATH=$PKG_CONFIG_PATH
   PKG_CONFIG_PATH=${prefix}/lib/pkgconfig:${PKG_CONFIG_PATH}
   export PKG_CONFIG_PATH
-  PKG_CHECK_MODULES([FLUX_CORE], [flux-core],
+  PKG_CHECK_MODULES([FLUX_CORE], [flux-core >= 0.81.0],
     [
       FLUX_PREFIX=`pkg-config --variable=prefix flux-core`
       LIBFLUX_VERSION=`pkg-config --modversion flux-core`


src/plugins/mf_priority.cpp
```diff
@@ -1770,6 +1771,13 @@ extern "C" int flux_plugin_init (flux_plugin_t *p)
         };
     }
 
+    if (flux_plugin_conf_unpack (p, "{s?s}", "sort", &sort_mode) == 0
+        && sort_mode) {
+        if (flux_jobtap_set_load_sort_order (p, sort_mode) < 0) {
+            return -1;
```
grondo: Log an error here if set_load_sort_order fails?

cmoussa1:
Thanks for the review here @grondo. We had an offline conversation on just removing the need to unpack a `sort_mode` key-value and just making it the default behavior of the plugin during initialization, so I'll go ahead and make that change to this PR, which should alleviate the need add testing for invalid `sort_mode` strings.

PR #789, opened 2025-12-08
database: add `max_sched_jobs` property for associations in flux-accounting DB

cmoussa1:
#### Problem

The flux-accounting database needs a way to track and configure a `max_sched_jobs` property for an association, i.e. set a limit on how many jobs can be in SCHED state for an association at any given time.

---

This PR adds a new column to `association_table` called `max_sched_jobs`, an integer representing the max number of jobs in SCHED state an association can have at any given time. This allows a `max_sched_jobs` property to be defined per-association.

This change requires incrementing the `DB_SCHEMA` version number with the addition of the new column. A new `--max-sched-jobs` optional argument is added to the commands that deal with adding, editing, and viewing associations in `association_table`. The corresponding man(1) pages are also updated to include the new optional argument.

Basic tests are added for those commands to include configuring `max_sched_jobs` for associations in the database.


jameshcorbett (approved):
Nice! But the new property isn't checked anywhere yet right? Is that for a future PR?

cmoussa1:
Thanks @jameshcorbett! Yes, the actual implementation for checking and setting a `max-sched-jobs` property is proposed in #790 to keep the PR from growing way too cumbersome to review. I will probably hold off on merging this one until after `v0.55.0` has been released to land both this and #790 in one release.

PR #797, opened 2025-12-23
doc: fix use of `-J` option in job usage factor example

cmoussa1:
#### Problem

The example of showing the different job usage factors for an association uses an incorrect optional argument: it should be `-J`, but the documentation shows `-j`.

---

This PR fixes the example in the documentation.


garlick (approved):
LGTM!

PR #800, opened 2026-01-07
bindings: use `FLUX_USERID_UNKNOWN` as default

cmoussa1:
#### Problem

The Python bindings use `65534` as the default userid for a user when one cannot be found, but this userid is technically valid (it is the userid for the `nobody` user).

---

This PR changes the default behavior to use `flux.constants.FLUX_USERID_UNKNOWN` as the default userid for a user when one cannot be found.


jameshcorbett (approved):
Nice!

PR #804, opened 2026-01-13
commands: add new `clear-usage` command to reset a bank's job usage

### LINKED ISSUE #801: feature request: ability to "reset" bank usage
Greg recently asked me about something that I'm pretty sure we don't have the ability to do currently. New grant cycles (LDRD, Grand Challenge, etc.) generally have a mix of new and returning / continuing projects. For returning / continuing projects we tend to just keep the existing bank name, but Greg would like to be able to zero out / reset the usage for those banks at the start of the new grant period so that they start out on equal footing with new projects.

I was thinking that something like a start date field could be used to tell update-usage to disregard usage before that date, but there are probably other ways to implement something like this.

cmoussa1:
#### Problem

There is no way to clear and reset the usage for a given bank in `bank_table` without fully deleting the bank and re-adding it. New grant cycles generally have a mix of new and returning / continuing projects. For returning / continuing projects, the same bank name tends to be kept while *also* zeroing out and resetting the usage for the bank at the start of the new grant.

---

This PR adds a new `clear-usage` command to the flux-accounting command suite. `clear-usage` takes one positional argument: a list of banks to have their usage reset. When a bank's usage is reset, their `job_usage` value is reset to 0. Any users under this bank will also have their `job_usage` value reset to 0.

Example usage:

```console
flux account clear-usage my_bank
```

When the bank's usage is reset, the job usage for the rest of the hierarchy is updated to account for the change in overall usage.

Even though a bank's usage is reset to 0, the old jobs submitted under this bank are still kept in the `jobs` table. These jobs are ignored in subsequent usage updates by filtering the table against an `ignore_older_than` timestamp, a new column added to the `bank_table` which represents a timestamp for which all jobs that have completed before said timestamp are ignored:

```sql
SELECT ...,b.ignore_older_than
FROM jobs ...
LEFT JOIN bank_table b
...
AND r.t_inactive > b.ignore_older_than
```

If the `clear-usage` command is run without any additional arguments, the `ignore_older_than` timestamp is set to the time in which the command is run, i.e. that bank will begin to ignore any jobs older than when `clear-usage` was run. An optional `--ignore-older-than` optional argument has also been added to `clear-usage` (as well as `add-bank` and `view-bank`) to allow further customization for banks in relation to which jobs are or are not ignored. The format can be any timestamp accepted by Flux's [`parse_datetime()`](https://github.com/flux-framework/flux-core/blob/master/src/bindings/python/flux/util.py#L385) function.

Some basic unit tests and sharness tests are added which test clearing usage for a bank and ensuring that all of its users' usage also gets cleared along with any and all parent banks.

Fixes #801


src/bindings/python/fluxacct/accounting/job_usage_calculation.py
```diff
@@ -546,3 +547,78 @@ def view_usage_report(
...
+    """
+    Reset job usage for one or more banks in the flux-accounting database.
+
+    Args:
+        conn: The SQLite Connection object.
+        banks: One or more banks to have its usage cleared.
+        ignore_older_than: The timestamp in which all older jobs will not be considered
+            towards job usage.
+    """
+    if len(banks) > 0:
```
jameshcorbett: This could be `if banks:`

jameshcorbett (approved):
Looks great! Just a couple of small things

PR #808, opened 2026-01-22
`delete_bank()`: update job usage after bank is removed with `--force`

### LINKED ISSUE #802: `delete-bank` with `--force` doesn't propagate usage up to parent bank(s)
While working on #801, I noticed that when a bank is deleted with `--force`, the change in job usage reflected in the rest of the hierarchy is not immediately affected. For example, if I have the following hierarchy:

```console
Bank                            Username           RawShares            RawUsage           Fairshare
root                                                       1                0.27
 A                                                         1             0.20422
  A                                user1                   1             0.20422                 1.0
 B                                                         1             0.06321
  B                                user1                   1             0.06321                 1.0
```

And I delete bank `A` with `--force`, the usage for the `root` bank stays the same:

```console
$ flux account delete-bank A --force
Bank                            Username           RawShares            RawUsage           Fairshare
root                                                       1                0.27
 B                                                         1             0.06321
  B                                user1                   1             0.06321                 1.0
```

I need to add a recursive traversal to update the usage for any and all parent banks with which the deleted bank sits under to reflect the removal of the bank.

cmoussa1:
#### Problem

Parent banks inadvertently keep the usage of a bank that is deleted with `--force`, which leads to inconsistent reporting.

---

This PR just adds an update for the job usage for the bank hierarchy when a bank is deleted with `--force`. A quick test is also added to ensure that parent banks' usage is also updated after a bank is removed with `--force`.

Fixes #802

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -15,6 +15,7 @@ import fluxacct.accounting
 from fluxacct.accounting import user_subcommands as u
 from fluxacct.accounting import formatter as fmt
 from fluxacct.accounting import sql_util as sql
+from fluxacct.accounting import job_usage_calculation as jobs
 from fluxacct.accounting.util import with_cursor
 
 ###############################################################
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
LGTM!

PR #810, opened 2026-01-27
`add-bank`/`add-user`: add check for existing DB structure when trying to add a bank or association

### LINKED ISSUE #809: support for banks that have users and banks as children
On tuolumne, there is a bank called 'bioinfo' that has one user and a sub-bank ('reclaim') with multiple users. It looks like the usage from the one user has not been added to the total usage of the 'bioinfo' parent (the usage of the bioinfo bank is significantly smaller than the usage from the one user). Additionally, the tree view from 'flux account view-bank -t ...' does not show the one user.

cmoussa1:
#### Problem

As noted in #809, flux-accounting requires a bank hierarchy structure in which sub-banks cannot be added to a parent bank that already has associations in it (and vice versa; associations cannot be added to a bank that
already has sub-banks in it), but there is no check for this hierarchy structure when calling `add-bank` or `add-user`.

---

This PR adds a check in `add_bank()` and `add_user()` that the bank or association being added is valid and complies with the hierarchy structure for flux-accounting; if not, it raises an error. Existing tests that did _not_ follow this structure have also been adjusted to comply and are adjusted in the first commit. Documentation has also been updated to leave a warning to readers about the assumptions of the structure of banks and associations in flux-accounting.

Fixes #809

### DIFF
--- a/doc/components/database-administration.rst
+++ b/doc/components/database-administration.rst
@@ -59,6 +59,13 @@ there is one root bank which all other banks and associations descend from:
 Multiple levels of banks can be defined under this root bank. Users can belong
 to more than one bank and will have at most one default bank.
 
+.. warning::
+    You are not allowed to have both an association and a bank under the same
+    parent bank, as this conflicts with how fair-share and job usage are
+    calculated. Trying to add a sub-bank to a parent bank which already has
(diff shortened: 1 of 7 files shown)


jameshcorbett (approved):
nice!

PR #816, opened 2026-02-13
`apply_decay_factor`: change decay application to incremental instead of cumulative

### LINKED ISSUE #815: `apply_decay_factor()`: decay is cumulative instead of incremental
@ryanday36 pointed out that some of the job usage values for some users looked kinda off. After some digging, I think I found the culprit.

The `apply_decay_factor()` function applies _cumulative_ decay factors (`0.5^1`, `0.5^2`, `0.5^3`, etc.) to each usage period every time the function runs. However, each period already contains values that have been decayed in previous cycles. This causes compound decay - for example, a value that should have been decayed by 0.5^3 total was actually being decayed by much more (0.5^5 or higher).

Take, for example, a historical job usage value of 100. The current `apply_decay_factor()` function would bring this to 50, and then to 12.5, and then to 1.5.

Changing the decay application from cumulative to incremental should relieve this issue. Now when shifting periods, we simply multiply each value by 0.5 once (representing one half-life period), rather than raising 0.5 to increasing powers. Since the periods already contain their historical decay from previous cycles, we only need to apply one additional decay factor when shifting everything down.

That same historical job usage value of 100 would then go down to 50, and then to 25, and then to 12.5.

cmoussa1:
#### Problem

As mentioned in #815, the `apply_decay_factor()` function applies cumulative decay factors (`0.5^1`, `0.5^2`, `0.5^3`, etc.) to each usage period every time the function runs. However, each period already contains values that have
been decayed in previous cycles. This causes compound decay - for example, a value that should have been decayed by `0.5^3` total was actually being decayed by much more (`0.5^5` or higher).

---

This PR changes the decay application from cumulative to incremental, i.e. just multiplies each value by 0.5 once (representing one half-life period) rather than raising 0.5 to increasing powers.

As a result, some of the expected job usage values in `t1006_job_archive.py` are changed with the new decay application. I've also added some new tests to simulate multiple half-life periods and checks to make sure that **all** job usage period values are updated properly.

Fixes #815

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
+++ b/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
@@ -10,7 +10,6 @@
 # SPDX-License-Identifier: LGPL-3.0
 ###############################################################
 import time
-import math
 import logging
 import sqlite3
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


jameshcorbett (approved):
Nice!

PR #817, opened 2026-02-17
plugin: minor improvements to logic, variable initialization

cmoussa1:
#### Problem

There are few minor issues in the priority plugin:

- In `check_and_release_held_jobs ()`, `held_job_id` is not set in the case where removing the `max_sched_jobs` dependency fails, which is inconsistent with the other checks throughout this function.
- The `goto error` block is unconditionally executed at the end of `check_and_release_held_jobs ()` because there is no `return` statement before the label, even on success. This incorrectly populates the journal with error messages, even if the function succeeds.
- In `run_cb ()`, the `held_jobs` attribute for an Association object is not checked to see if it is empty before calling `check_and_release_held_jobs ()`.

---

This PR includes minor fixes for all three problems listed above.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1478,10 +1478,12 @@ static int run_cb (flux_plugin_t *p,
     b->cur_sched_jobs--;
     // check to see if any jobs held due to max_sched_jobs limit can now
     // have their dependency removed
-    if (check_and_release_held_jobs (p, b) < 0) {
-        flux_log_error (h,
-                        "job.state.run: error checking and releasing "
-                        "held jobs for association");
+    if (!b->held_jobs.empty ()) {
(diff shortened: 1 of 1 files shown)


src/plugins/mf_priority.cpp
jameshcorbett: I'm surprised this issue wasn't noticed before. Without the `return 0` added here, wouldn't every job have this exception raised?

jameshcorbett (approved):
LGTM! One question

src/plugins/mf_priority.cpp
cmoussa1: Yes, exactly 🤦 this floods the journal with these exception messages. I think I've mentioned it in a meeting or two a while ago that inspecting the log showed so many of these messages, even though nothing was wrong. Now I know why 😅

jameshcorbett: Oh got it! Makes sense.

PR #818, opened 2026-02-18
t: make Python test DB names unique

cmoussa1:
#### Problem

Many of the unit tests in the `t/python/` directory share the same database name, which can cause errors in the unit tests if they are running at the same time and in the same directory, since they would essentially be trying to create and connect to the same database, when they really should each be testing against their own DB.

---

This PR makes the test database names for each of the unit tests in `t/python/` unique. I've also snuck in a minor commit to set the `log-stderr-level` attribute to match the rest of sharness tests that were fixed in #812.

### DIFF
--- a/t/python/t1003_bank_cmds.py
+++ b/t/python/t1003_bank_cmds.py
@@ -12,6 +12,7 @@
 import unittest
 import sys
 import os
+import time
 import sqlite3
 
 from fluxacct.accounting import bank_subcommands as b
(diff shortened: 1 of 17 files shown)


t/python/t1003_bank_cmds.py
```diff
@@ -22,11 +22,13 @@ class TestAccountingCLI(unittest.TestCase):
     # create test flux-accounting database
     @classmethod
     def setUpClass(self):
-        c.create_db("TestBankSubcommands.db")
+        c.create_db("FluxAccounting_t1003.db")
```
grondo: Were this and other db names already unique?

grondo (approved):
LGTM. I tested on this branch and it resolves the issue I was seeing with `make -j N check`.

I did add some suggestions inline.

t/python/t1003_bank_cmds.py
cmoussa1: Yes they were. I figured I'd try to standardize the test DB names across all of the unit tests - it might be nice since I am going to take your suggestion below and just create a `dbname` member for the classes in each unit test file. Would that be okay?

grondo (approved):
LGTM!

PR #824, opened 2026-03-09
`view-job-records`: change default job ID format to f58

cmoussa1:
#### Problem

The `view-job-records` command displays job IDs in decimal format instead of f58 format, which is what the rest of the Flux command suite uses. It would be more consistent for this command to follow the same format as other job-related Flux commands.

---

This PR changes the default job ID format for the `view-job-records` command to be in [f58](https://github.com/flux-framework/rfc/blob/master/spec_19.rst) format. It uses flux-core's `JobID` class to convert the decimal version of the job ID (which is how it is stored in the `jobs` table in the flux-accounting DB) to any of the available formats provided by the class.

It adds an optional argument to the view-job-records command called `--jobid-format`, which will allow the user to specify a number of allowed job ID formats when displaying job records.

A new set of tests are added that specify the `--jobid-format` optional argument in a mixture of both default and custom formatting for the `view-job-records` command.

I've also snuck in a commit to add a missing `|` between two fields in the header for `view-job-records`.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/util.py
+++ b/src/bindings/python/fluxacct/accounting/util.py
@@ -16,6 +16,7 @@ import contextlib
 
 from flux.constants import FLUX_USERID_UNKNOWN
 from flux.util import parse_datetime
+from flux.job.JobID import JobID
 import fluxacct.accounting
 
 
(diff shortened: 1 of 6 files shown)


src/bindings/python/fluxacct/accounting/util.py
```diff
@@ -146,3 +147,29 @@ def update_parent_bank_usage(conn, bank):
...
+def format_jobid(jobid_dec, jobid_format="f58"):
+    """
+    Convert a decimal jobid to a requested Flux JobID string format.
+    """
+    fmt = (jobid_format or "f58").lower()
+    if fmt not in JOBID_FORMATS:
+        raise ValueError(
+            f"invalid jobid format: {jobid_format} "
+            f"(choose from: {', '.join(sorted(JOBID_FORMATS))})"
+        )
```
jameshcorbett: Do you need this check? Couldn't you just count on the `getattr()` below raising an error? Especially since the `view-job-records` command does some client-side validation (with `choices`).

jameshcorbett (approved):
This looks good but I have a suggestion. Instead of adding a new `--jobid-format` flag, you could allow users to specify the output format in their format string. Like allow users to write a format string like `--format="{jobid.emoji:<20} | {nnodes:<8}"`. Python's format strings allow you to directly look up attributes on objects being formatted:

```
>>> x
JobID(123)
>>> "{jobid.f58}".format(jobid=x)
'ƒ38'
>>> "{jobid.emoji}".format(jobid=x)
'👼'
>>> "{jobid}".format(jobid=x)
'ƒ38'
```

jameshcorbett (approved):
I think this looks great, the implementation looks like it took a little more code than I was hoping it would, but so it goes.

PR #830, opened 2026-03-24
edit-bank: fix `ValueError` message when trying to edit a bank's shares to `<= 0`

cmoussa1:
##### Problem

The `ValueError` message that is returned when a shares value of less than or equal to 0 is passed is incorrect; it should specify that the new shares value must be greater than 0, not greater than or equal to 0.

---

This PR just corrects the error message to specify that the new shares value must be greater than 0.

It also updates one of the tests in `t1023-flux-account-banks.t` that `grep`s for this error message as a result of the message changing.


jameshcorbett (approved):
👍

PR #832, opened 2026-03-26
docker: add `noble` Dockerfile to CI

cmoussa1:
#### Problem

flux-accounting does not have a `noble` Docker image, which is useful for testing things like the bindings against Python 3.12.

---

This PR adds a `noble` Dockerfile to the repository and adds it to the matrix of builders that run during CI.


jameshcorbett (approved):
LGTM

cmoussa1:
Thanks @jameshcorbett - setting MWP here

PR #837, opened 2026-03-30
database: add `max_sched_jobs` property for queues in flux-accounting DB

cmoussa1:
#### Problem

The flux-accounting database needs a way to track and configure a `max_sched_jobs` property for a queue, i.e. set a limit on how many jobs can be in SCHED state for an association in a particular queue at any given time.

---

This PR adds a new column to `queue_table` called `max_sched_jobs`, an integer representing the max number of jobs in SCHED state an association can have in a queue at any given time. This allows a `max_sched_jobs` property to be defined per-queue.

This change requires incrementing the `DB_SCHEMA` version number with the addition of the new column. A new `--max-sched-jobs` optional argument is added to the commands that deal with queues in `queue_table`. The corresponding `man(1)` pages are also updated to include the new optional argument.

Basic tests are added for those commands to include configuring `max_sched_jobs` for queues in the database.


jameshcorbett (approved):
LGTM

PR #840, opened 2026-04-01
plugin: add limit of max number of jobs in SCHED state per-queue

cmoussa1:
#### Problem

The priority plugin does not enforce a limit on an association's job(s) if they have a certain number of jobs in SCHED state in a particular queue already.

---

This PR (built on top of #837) adds enforcement of a per-queue max SCHED jobs limit in the priority plugin. It adds a new property to the `QueueUsage` class to keep track of how many jobs in SCHED state an association has in a particular queue as well as a new method to the `Association` class to compare the association's number of jobs in SCHED state versus the queue's `max_sched_jobs` property. If the association is at the limit, the job is held with a dependency called `"max-sched-jobs-queue-limit"`.

The current number of jobs in SCHED state for a particular queue is incremented when a new job enters `job.state.sched` and is decremented when a job enters `job.state.run`. After decrementing the count in `job.state.run`, the association's set of held jobs (if any) is re-checked to see if any jobs held due to the `"max-sched-jobs-queue-limit"` can also be released.

Some basic tests are added to check the enforcement of the `"max-sched-jobs-queue-limit"` limit both with the plugin fully loaded and after a restart.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
@@ -164,6 +164,7 @@ def export_as_json(conn, cursor):
             "priority": int(row["priority"]),
             "max_running_jobs": int(row["max_running_jobs"]),
             "max_nodes_per_assoc": int(row["max_nodes_per_assoc"]),
+            "max_sched_jobs": int(row["max_sched_jobs"]),
         }
         queues.append(queue)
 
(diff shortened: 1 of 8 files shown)


grondo (commented):
A couple comments inline after a first pass.

Also, I only took a quick look at the sharness test. Does it check that with two queues a job held in queue A due to a max SCHED limit is not released when a job in B transitions to RUN state?

grondo (approved):
LGTM! Though I did point out a few minor things that are easy fixes.

PR #841, opened 2026-04-01
t: use format string for comparing output in `t1043-view-jobs-by-bank.t`

cmoussa1:
#### Problem

When `grep` finds no matches, `grep -c` prints `0` but exits with status `1`. Since the test bodies in `t1043-view-jobs-by-bank` are chained with `&&`, that causes the chain to stop before the later test commands can run, and subsequently the tests to fail.

---

This PR changes those tests to just use a format string to get the output of `view-job-records` so it's easier to look for and compare expected output in the sharness tests in `t1043-view-jobs-by-bank.t`.


jameshcorbett (approved):
LGTM

PR #844, opened 2026-04-08
generate-matrix: change release builder from RHEL8 to RHEL9

cmoussa1:
#### Problem

The builder that is responsible for creating a release tarball for flux-accounting is based off of RHEL8, which ships an extremely old version of `automake`, which subsequently fails when building TOSS packages because it is using a very old version of `py-compile` that tries to import a now deprecated Python module (`imp`).

---

Change the builder that is responsible for generating a release tarball from RHEL8 to RHEL9.

### DIFF
--- a/src/test/generate-matrix.py
+++ b/src/test/generate-matrix.py
@@ -148,11 +148,15 @@ matrix.add_build(
     docker_tag=True,
 )
 
-# el9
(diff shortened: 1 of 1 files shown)


grondo (approved):
LGTM!

PR #846, opened 2026-04-09
github: add dependabot support

cmoussa1:
#### Problem

GitHub Actions are pinned to older versions and there is no automated process to keep them up to date.

---

This PR adds a `dependabot.yml` to automatically check for GitHub Actions updates monthly and group them into a single PR, matching the configuration in flux-core.


jameshcorbett (approved):

cmoussa1:
Thanks @jameshcorbett!

PR #851, opened 2026-04-15
plugin: clear internal data structures during initialization

### LINKED ISSUE #849: multi-factor priority plugin not clearing data when reloade on RHEL10
While working on adding an `el10` builder to the flux-accounting testsuite, I noticed that some of the sharness tests, particularly the ones that reloaded the plugin like this:

```shell
test_expect_success 'reload priority plugin' '
	flux jobtap remove mf_priority.so &&
	flux jobtap load ${MULTI_FACTOR_PRIORITY} &&
	flux jobtap list | grep mf_priority
'
```

didn't seem to actually clear the database data that's loaded into the plugin, even if there is a delay between unloading and reloading the plugin.

cmoussa1:
#### Problem

On RHEL10 systems, unloading and reloading the plugin doesn't actually clear the data in any of the internal data structures in the plugin. Thus, a number of the tests in the testsuite fail that deal with reloading the plugin because the test expects a freshly-loaded plugin with no data in it.

It also raises a more general concern that when the plugin *is* loaded, it should clear any previously-loaded data. Data sent from the flux-accounting database could be removed, and if the data structures in the plugin are not properly cleared, then leftover data can remain in the plugin and unexpected behavior can occur, such as users who do not have a valid entry in the flux-accounting database being able to submit jobs.

---

This PR adds a clear of all the internal data structures in the plugin during `plugin_init ()` so that leftover data does not remain after a plugin reload. I've also added some tests to check that data in the plugin does not remain after a reload.

Fixes #849

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -1864,6 +1864,13 @@ static const struct flux_plugin_handler tab[] = {
 
 extern "C" int flux_plugin_init (flux_plugin_t *p)
 {
+    // explicitly reset all global state of internal data structures
+    users.clear ();
+    queues.clear ();
+    banks.clear ();
+    users_def_bank.clear ();
+    projects.clear ();
+    priority_weights.clear ();
 
     json_t *config_obj = NULL;
     flux_t *h = flux_jobtap_get_flux (p);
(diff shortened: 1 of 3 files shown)


grondo (approved):
LGTM!

cmoussa1:

PR #852, opened 2026-04-15
ci: add `el10` builder

cmoussa1:
#### Problem

There is no `el10` builder in the flux-accounting testsuite.

---

This PR adds one. This is built on top of #851 as a result of discovering some test failures that were run in a `el10` container that are fixed in that PR.

### DIFF
--- a/src/test/docker/el10/Dockerfile
+++ b/src/test/docker/el10/Dockerfile
@@ -1,0 +1,17 @@
+FROM fluxrm/flux-core:el10
+
+ARG USER=flux
+ARG UID=1000
+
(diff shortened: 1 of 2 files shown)


grondo (approved):
LGTM!

cmoussa1:
Thanks @grondo! I think we'll need to land #851 before I can set MWP here. Would you mind giving that a look as well?

grondo:
Oh, sorry I did review those commits already but missed that this was based on another PR. I'll go approve that one 👍

cmoussa1:
Rebased this to catch up after #851 - setting MWP here!

PR #854, opened 2026-04-16
`update-usage`: change `t_run` filter to `t_inactive`

### LINKED ISSUE #853: `update_job_usage()` may skip jobs due to `t_run` vs `t_inactive` mismatch in SQL query
In `job_usage_calculation.py`, the SQL query inside `update_job_usage()` that fetches new jobs for each association filters on `r.t_run > j.last_job_timestamp`. However, `last_job_timestamp` is always set from a job's `t_inactive` value (via `update_t_inactive()`). This mismatch between the two timestamps means that jobs can be silently skipped during a usage update.

Consider an association with two jobs:

```
Job A: t_run = 1000, t_inactive = 5000  (long-running job)
Job B: t_run = 2000, t_inactive = 3000  (short job that finishes first)
```

If Job B is processed first, `last_job_timestamp` gets set to 3000 (its `t_inactive`). On the next call to `update_job_usage()`, the query filters on `r.t_run > 3000`. Job A has `t_run = 1000`, so it fails that condition and is skipped entirely — even though it finished at `t_inactive = 5000`, well after Job B completed.

I should change the filter in the SQL query from `r.t_run > j.last_job_timestamp` to `r.t_inactive > j.last_job_timestamp` so that both sides of the comparison are working with the same type of timestamp.

cmoussa1:
#### Problem

As mentioned in #853, the query inside `update_job_usage()` that fetches new jobs for each association filters on `r.t_run > j.last_job_timestamp`. However, `last_job_timestamp` is always set from a job's `t_inactive` value (via
`update_t_inactive()`). This mismatch between the two timestamps means that jobs can be silently skipped during a usage update.

---

This PR changes the filter in the SQL query from `r.t_run > j.last_job_timestamp` to `r.t_inactive > j.last_job_timestamp` so that both sides of the comparison are working with the same type of timestamp.

I've also added some unit tests to check and see that a long running job is not skipped when updating the job usage value for an association.

Fixes #853

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
+++ b/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
@@ -289,7 +289,7 @@ def update_job_usage(acct_conn, pdhl=1):
             FROM jobs r LEFT JOIN job_usage_factor_table j
             ON r.userid = j.userid AND r.bank = j.bank
             LEFT JOIN bank_table b
-            ON r.bank = b.bank WHERE r.t_run > j.last_job_timestamp
+            ON r.bank = b.bank WHERE r.t_inactive > j.last_job_timestamp
             AND r.t_inactive > b.ignore_older_than
         """
         cur.execute(s_new_jobs)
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):

cmoussa1:
Thanks @jameshcorbett! I'll set MWP here

PR #857, opened 2026-04-17
`view-job-records`: add dynamic width sizing for columns

cmoussa1:
#### Problem

The column widths for the output of the `view-job-records` command is statically defined, which can cause problems if the field
itself is very long.

---

This PR adds dynamic sizing to the width of the different columns in the output of the `view-job-records` command by comparing the max length of the longest field and its corresponding header to create cleaner and more uniform output. Add special formatting for the floating point values to only go to 2 decimal places when printing those values.

It also includes an adjustment a couple of the sharness tests in `t1011-job-archive-interface.t` that look at the output of the headers when there are no job records to account for the new dynamic sizing. I've added a simple set of tests to check that the output also adjusts when there are two rows with drastically different text lengths and making sure that the width adjusts to the longer string.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
@@ -16,6 +16,7 @@ from flux.job.JobID import JobID
 from flux.constants import FLUX_USERID_UNKNOWN
 from fluxacct.accounting import formatter as fmt
 from fluxacct.accounting import util
+from fluxacct.accounting import JOB_RECORD_FIELDS, JOB_RECORD_FLOAT_FIELDS
 
 
 class JobRecord:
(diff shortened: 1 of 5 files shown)


src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
```diff
@@ -69,18 +70,38 @@ def convert_to_str(job_records, fmt_string=None):
...
-        )
-    output = fmt.JobsFormatter(fmt_string)
-    job_record_str = output.build_table(job_records)
+    if fmt_string:
+        output = fmt.JobsFormatter(fmt_string)
+        return output.build_table(job_records)
+
+    # if not job_records:
+    #     return ""
+
```
jameshcorbett: delete?

jameshcorbett (approved):
Looks reasonable, one nit

src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
cmoussa1: Good catch! Sorry for leaving that in there. Just fixed this and force-pushed up the commit

cmoussa1:
Thanks @jameshcorbett! I addressed the nit you caught. Will set MWP here

PR #860, opened 2026-04-27
`docker-run-checks.sh`: use `el10` by default

cmoussa1:
#### Problem

`docker-run-check.sh` uses `el8` as the default image when launching into an interactive container, but there are newer images that can be used.

---

This PR updates the default image to be used to `el10`.


jameshcorbett (approved):

cmoussa1:

PR #862, opened 2026-04-29
`view-bank`: add `active` column, optional argument to filter for only active users

### LINKED ISSUE #861: flux account view-bank -t includes deleted users in output
I use 'flux account view-bank -t' in a couple of tools to grab full bank hierarchies and usages. Both 'flux account view-bank -t' and 'flux account view-bank -u' display deleted (active=0) users so I find myself needing to run a second 'list-users' command to figure out which users can actually currently use a given bank. Would it be possible to either add an 'active' column to the output of those commands or otherwise indicate which associations are inactive (maybe just a '*' next to the username)?

cmoussa1:
#### Problem

Both `flux account view-bank -t` and `flux account view-bank -u` display deleted (`active=0`) users, causing the need to run a second `list-users` command to figure out which users can actually currently use a given bank.

---

This PR adds the `active` column for banks and associations in the default output of the `view-bank` command and adds an optional argument to the `view-bank` command to *only* show users who are active under a given bank in the `view-bank` command.

I've also updated all of the tests that look for specific output of the `view-bank` command to account for the addition of the new `active` column.

Fixes #861

### DIFF
--- a/src/bindings/python/fluxacct/accounting/bank_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/bank_subcommands.py
@@ -163,6 +163,7 @@ def view_bank(
     cols=None,
     format_string="",
     concise=False,
+    active=False,
 ):
     if tree and cols is not None:
         # tree format cannot be combined with custom formatting, so raise an Exception
@@ -186,10 +187,10 @@ def view_bank(
         return formatter.as_format_string(format_string)
     if tree:
         if parsable:
-            return formatter.as_parsable_tree(bank, concise)
-        return formatter.as_tree(concise)
+            return formatter.as_parsable_tree(bank, concise, active)
+        return formatter.as_tree(concise, active)
     if users:
-        return formatter.with_users(bank, concise)
+        return formatter.with_users(bank, concise, active)
     return formatter.as_json()
 
 
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -32,6 +32,16 @@ class AccountingFormatter:
             # the SQL query didn't fetch any results; raise an Exception
             raise ValueError(error_msg)
 
+    def _transform_row(self, row):
+        """Transform display values in a row before formatting."""
+        transformed = []
+        for col_name, value in zip(self.column_names, row):
+            if col_name == "active":
+                transformed.append("true" if value == 1 else "false")
+            else:
+                transformed.append(u.format_value(value))
+        return transformed
+
     def get_column_names(self):
         """
         Return the column names from the query result.
@@ -57,29 +67,25 @@ class AccountingFormatter:
         Returns:
             table: the data from the query formatted as a table.
         """
-        # fetch column names and determine the width of each column
-        col_names = [description[0] for description in self.cursor.description]
+        transformed_rows = [self._transform_row(row) for row in self.rows]
+
         col_widths = [
             max(
-                len(str(u.format_value(value)))
-                for value in [col] + [row[i] for row in self.rows]
+                len(str(value))
+                for value in [col] + [row[i] for row in transformed_rows]
             )
-            for i, col in enumerate(col_names)
+            for i, col in enumerate(self.column_names)
         ]
 
         # format a row of data
         def format_row(row):
             return " | ".join(
-                [
-                    f"{str(u.format_value(value)).ljust(col_widths[i])}"
-                    for i, value in enumerate(row)
-                ]
+                f"{str(value).ljust(col_widths[i])}" for i, value in enumerate(row)
             )
 
-        # format the header, separator, and data rows
-        header = format_row(col_names)
+        header = format_row(self.column_names)
         separator = "-+-".join(["-" * width for width in col_widths])
-        data_rows = "\n".join([format_row(row) for row in self.rows])
+        data_rows = "\n".join([format_row(row) for row in transformed_rows])
 
         table = f"{header}\n{separator}\n{data_rows}"
 
--- a/src/bindings/python/fluxacct/accounting/formatter.py
+++ b/src/bindings/python/fluxacct/accounting/formatter.py
@@ -153,24 +153,30 @@ class BankFormatter(AccountingFormatter):
             cursor, error_msg=f"bank {self.bank_name} not found in bank_table"
         )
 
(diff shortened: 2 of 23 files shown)


t/expected/flux_account/A_bank.expected
```diff
@@ -2,7 +2,7 @@ bank_id | bank | active | parent_bank | shares | job_usage | priority | ignore_o
...
 2       | A    | 1      | root        | 1      | 0.0       | 0.0      | 0                
 
-username | default_bank | shares | job_usage | fairshare
----------+--------------+--------+-----------+----------
-user5011 | A            | 1      | 0.0       | 0.5      
-user5012 | A            | 1      | 0.0       | 0.5      
+username | active | default_bank | shares | job_usage | fairshare
+---------+--------+--------------+--------+-----------+----------
+user5011 | 1      | A            | 1      | 0.0       | 0.5      
+user5012 | 1      | A            | 1      | 0.0       | 0.5
```
jameshcorbett: I think my preference would be to display `active` as `true` or `false` rather than `1` or (I presume) `0`

t/expected/print_hierarchy/small_tie_parsable.txt
```diff
@@ -1,14 +1,14 @@
...
-  account2|leaf.2.2|1000|1.0|0.5
-  account2|leaf.2.3|100000|1.0|0.75
+  account2|leaf.2.1|1|10000|10.0|0.5
+  account2|leaf.2.2|1|1000|1.0|0.5
+  account2|leaf.2.3|1|100000|1.0|0.75
  account3||10|1.0
-  account3|leaf.3.1|100|0.0|1.0
-  account3|leaf.3.2|10|1.0|0.875
+  account3|leaf.3.1|1|100|0.0|1.0
+  account3|leaf.3.2|1|10|1.0|0.875
```
jameshcorbett: It looks like these two tables are becoming misaligned. The top line here has just three fields but the bottom two lines have six. I think at least another `|` for the `Active` field would help?

src/bindings/python/fluxacct/accounting/formatter.py
jameshcorbett: Add an extra `|` character between the bank name and the shares in the two format strings here to account for the missing `Active` field?

src/cmd/flux-account.py
```diff
@@ -664,6 +664,13 @@ def add_view_bank_arg(subparsers):
         const=True,
         help="only list associations that have a job usage value > 0",
     )
+    subparser_view_bank.add_argument(
+        "-A",
+        "--active",
+        action="store_const",
+        const=True,
```
jameshcorbett: why `store_const` rather than `store_true`?

src/bindings/python/fluxacct/accounting/bank_subcommands.py
```diff
@@ -186,10 +187,10 @@ def view_bank(
...
     if tree:
         if parsable:
-            return formatter.as_parsable_tree(bank, concise)
-        return formatter.as_tree(concise)
+            return formatter.as_parsable_tree(bank, concise, active)
+        return formatter.as_tree(concise, active)
     if users:
-        return formatter.with_users(bank, concise)
+        return formatter.with_users(bank, concise, active)
     return formatter.as_json()
```
jameshcorbett, line 194: `active` isn't passed here?

t/t1023-flux-account-banks.t
```diff
@@ -191,6 +191,119 @@ test_expect_success 'edit the priority of a bank' '
...
+
+test_expect_success 'view-bank --users --active only shows active users under that bank' '
+	flux account view-bank --users Z > view_bank_users3.test &&
+	cat <<-EOF > view_bank_users3.expected &&
+	username  | active | default_bank | shares | job_usage | fairshare
+	----------+--------+--------------+--------+-----------+----------
+	user90001 | 1      | Z            | 1      | 0.0       | 0.5      
+	EOF
+	grep -f view_bank_users3.test view_bank_users3.expected
+'
```
jameshcorbett, line 305: This test title and what actually runs seem mismatched? There's no `--active` flag for instance.

jameshcorbett (commented):
Looks good generally but I have some minor suggestions! Also I think my preference would be to put all the updates to the expected output into a separate commit rather than bundling it into the first commit. But I'm not picky, happy to approve it without that change too.

src/bindings/python/fluxacct/accounting/bank_subcommands.py
jameshcorbett: To `as_json()` I mean

t/expected/flux_account/A_bank.expected
cmoussa1: This is a good suggestion, thanks. It gets especially hard to read when it's right next to the `RawShares` column as well. I'll change this to display `true` or `false`.

t/expected/print_hierarchy/small_tie_parsable.txt
cmoussa1: Yeah, I think this is because I forgot to add the extra `|` for the `Active` column for the root bank. Thanks for catching! I will fix this

src/cmd/flux-account.py
cmoussa1: Because I didn't know `store_true` was an action until just now! ;) Thanks, I will change this to use `store_true`

src/bindings/python/fluxacct/accounting/bank_subcommands.py
cmoussa1: `active` is only relevant for the various optional arguments of `view-bank` that list any users under that particular bank. Since `as_json()` is only for looking at one bank at a time, the `active` argument doesn't need to be passed here.

t/expected/flux_account/A_bank.expected
jameshcorbett: I notice it's still `1` and not `true` in just this one file ^

jameshcorbett (approved):
LGTM, one comment

cmoussa1:
Thanks @jameshcorbett - I believe I addressed your comment above. It ended up being a bit more of a change than I had anticipated (because it was using a general class method to format that table output), but I think it was more for the better, as now the `active` column displays more readable output in general and it will be easy to add more transformations in the future. I added a couple of additional commits as to not clutter the first couple. I'll go ahead and set MWP here shortly

Issue #864, opened 2026-05-06
max-sched-nodes/cores per association queue limit


ryanday36:
related to #838, it would be useful to have max-sched-nodes and max-sched-cores per association limits on queues so that we can limit the total fraction of a queue that can be used by one user at a time.

cmoussa1:
Alright, this should now be fixed by #871, which will go into the next release. We should probably test this on one of the testbed clusters before starting to enforce these limits in our production environments.

I'll close this and can re-open any more specific issues if/as they come up.

PR #869, opened 2026-05-13
database: add new per-queue property for max resources in SCHED state

cmoussa1:
#### Problem

As mentioned in #864, there needs to be a way to limit the max number of nodes/cores in SCHED state for an association in a given queue.

---

This PR looks to begin to add support for this by adding two new properties to `queue_table`: `max_sched_nodes_per_assoc` and `max_sched_cores_per_assoc`, which represents the max number of nodes and cores an association can have in SCHED state across all of their jobs in a queue at any given time.

These properties are also added as optional arguments to the `add-queue` and `edit-queue` commands so that they can be configured on the command line.

Some basic tests are also added that go through viewing, editing, and resetting these properties for a given queue.


src/cmd/flux-account.py
```diff
@@ -846,6 +847,28 @@ def add_add_queue_arg(subparsers):
...
+        default=INTEGER_MAX,
+        metavar="NNODES",
+    )
+    subparser_add_queue.add_argument(
+        "-msc",
+        "--max-sched-cores-per-assoc",
+        type=int,
+        help=(
+            "max number of cores in SCHED state an association can have across all of "
+            "in this queue at any given time"
```
jameshcorbett: I think you mean "...across all of _their active jobs_ in this queue..." in both of these help messages?

src/cmd/flux-account.py
```diff
@@ -941,6 +964,28 @@ def add_edit_queue_arg(subparsers):
         default=2147483647,
```
jameshcorbett, line 964: Should these be converted to use `INTEGER_MAX` as well?

t/t1087-max-sched-resources-queue-basic.t
```diff
@@ -0,0 +1,93 @@
...
+	grep "\"max_sched_nodes_per_assoc\": 9999" edit_queue1.out &&
+	grep "\"max_sched_cores_per_assoc\": 9999" edit_queue1.out
+'
+
+test_expect_success 'edit-queue: max_sched_*_per_assoc properties can be reset' '
+	flux account edit-queue q1 -msn=-1 -msc=-1 &&
+	flux account view-queue q1 > edit_queue2.out &&
+	grep "\"max_sched_nodes_per_assoc\": \"unlimited\"" edit_queue2.out &&
+	grep "\"max_sched_cores_per_assoc\": \"unlimited\"" edit_queue2.out
+'
```
jameshcorbett, line 87: What happens with `-2` and other negative values?

jameshcorbett (approved):
LGTM some minor things

src/cmd/flux-account.py
jameshcorbett: nvm, https://github.com/flux-framework/flux-accounting/pull/870

src/cmd/flux-account.py
cmoussa1: 🤦 I never filled in the rest of it. Thanks for catching this!

t/t1087-max-sched-resources-queue-basic.t
cmoussa1: With `-2` and other negative values, you could just set the property to that value. I'll go ahead and add a check for a majority of the properties that can be edited in `queue_table` that if they are less than `-1`, a `ValueError` will be raised.

cmoussa1:
Thanks for reviewing @jameshcorbett! I've addressed your great suggestions above and force-pushed. I'll set MWP here

PR #871, opened 2026-05-14
plugin: add per-association dependency for max resources in SCHED state in a queue

cmoussa1:
#### Problem

As mentioned in #864, the priority plugin does not enforce a limit on the max number of resources (nodes or cores) an association can have in SCHED state across all of their active jobs in a queue.

---

This PR adds support for such a limit. To do this, it unpacks both the `max_sched_nodes_per_assoc` and `max_sched_cores_per_assoc` properties for each queue in the plugin and stores it in each Queue object. New methods to check an association's current number of resources in SCHED state in a particular queue and whether a new job would put them over this limit are added and called when a job enters `job.state.depend`. These counters are incremented and subsequently decremented in `job.state.sched` and `job.state.run`, respectively. New members to keep track of the association's *current* number of resources in SCHED state in a particular queue are added to the QueueUsage class.

Some tests that work through hitting certain resource limits in one queue and multiple queues are also added.

### DIFF
--- a/src/plugins/accounting.hpp
+++ b/src/plugins/accounting.hpp
@@ -80,6 +80,8 @@ public:
     int max_running_jobs = std::numeric_limits<int>::max ();
     int max_nodes_per_assoc = 2147483647;
     int max_sched_jobs = 2147483647;
+    int max_sched_nodes_per_assoc = std::numeric_limits<int>::max ();
+    int max_sched_cores_per_assoc = std::numeric_limits<int>::max ();
 };
 
 // a class to track an association's usage in a particular queue
--- a/src/plugins/accounting.hpp
+++ b/src/plugins/accounting.hpp
@@ -54,6 +54,8 @@ extern "C" {
 #define D_QUEUE_MRES "max-resources-queue"
 #define D_ASSOC_MSJ  "max-sched-jobs-user-limit"
 #define D_QUEUE_MSJ  "max-sched-jobs-queue-limit"
+#define D_QUEUE_MSN  "max-sched-nodes-queue-limit"
+#define D_QUEUE_MSC  "max-sched-cores-queue-limit"
 
 // error messages for flux-accounting-specific validation messages
 #define MSG_INVALID_QUEUE \
@@ -90,6 +92,8 @@ public:
     int cur_run_jobs = 0;   // number of running jobs in queue
     int cur_nodes = 0;      // number of nodes across all running jobs in queue
     int cur_sched_jobs = 0; // number of jobs in SCHED state in queue
+    int cur_sched_nodes = 0;// number of nodes in SCHED state in queue
+    int cur_sched_cores = 0;// number of cores in SCHED state in queue
 };
 
 // all attributes are per-user/bank
@@ -148,6 +152,20 @@ public:
     bool under_queue_max_sched_jobs (const std::string &queue,
                                      std::map<std::string, Queue> &queues,
                                      int pending);
+    bool under_queue_max_sched_nodes (const Job &job,
+                                      const std::string &queue,
+                                      std::map<std::string, Queue> &queues);
+    bool under_queue_max_sched_cores (const Job &job,
+                                      const std::string &queue,
+                                      std::map<std::string, Queue> &queues);
+    bool under_queue_max_sched_nodes (const Job &job,
+                                      const std::string &queue,
+                                      std::map<std::string, Queue> &queues,
+                                      int pending);
+    bool under_queue_max_sched_cores (const Job &job,
+                                      const std::string &queue,
+                                      std::map<std::string, Queue> &queues,
+                                      int pending);
 };
 
 class Bank {
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -692,6 +692,7 @@ static void rec_q_cb (flux_t *h,
     char *queue = NULL;
     int min_nodes_per_job, max_nodes_per_job, max_time_per_job, priority;
     int max_running_jobs, max_nodes_per_assoc, max_sched_jobs;
+    int max_sched_nodes_per_assoc, max_sched_cores_per_assoc;
     json_t *data, *jtemp = NULL;
     json_error_t error;
     int num_data = 0;
@@ -714,7 +715,8 @@ static void rec_q_cb (flux_t *h,
         json_t *el = json_array_get(data, i);
 
         if (json_unpack_ex (el, &error, 0,
-                            "{s:s, s:i, s:i, s:i, s:i, s:i, s:i, s:i}",
+                            "{s:s, s:i, s:i, s:i, s:i, s:i, s:i, s:i, "
+                            "s:i, s:i}",
                             "queue", &queue,
                             "min_nodes_per_job", &min_nodes_per_job,
                             "max_nodes_per_job", &max_nodes_per_job,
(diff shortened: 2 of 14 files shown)


src/plugins/accounting.hpp
```diff
@@ -80,6 +80,8 @@ class Queue {
     int max_running_jobs = std::numeric_limits<int>::max ();
     int max_nodes_per_assoc = 2147483647;
     int max_sched_jobs = 2147483647;
+    int max_sched_nodes_per_assoc = 2147483647;
+    int max_sched_cores_per_assoc = 2147483647;
```
grondo: Just curious why `std:numeric_limits<int>::max ()` isn't used for max_*nodes/cores as it is for `max_running_jobs`?

src/plugins/mf_priority.cpp
```diff
@@ -1804,6 +1888,20 @@ static int inactive_cb (flux_plugin_t *p,
...
+        // count resources requested for the job
+        if (job.count_resources (jobspec) < 0) {
+            flux_jobtap_raise_exception (p,
+                                         FLUX_JOBTAP_CURRENT_JOB,
+                                         "mf_priority",
+                                         0,
+                                         "job.state.run: unable to count " \
+                                         "resources for job");
+            return -1;
+        }
```
grondo: The error message here references `job.state.run`, but this is `inactive_cb`. (copy-paste error)

Actually, the `char *topic` is available in each callback, a suggestion would be to use that for the error prefix, e.g. `"%s: unable to count resources for job", topic);`

src/plugins/mf_priority.cpp
```diff
@@ -1804,6 +1888,20 @@ static int inactive_cb (flux_plugin_t *p,
         // decrement cur_sched_jobs for association since it never ran
         b->cur_sched_jobs--;
         b->queue_usage[queue_str].cur_sched_jobs--;
+        // count resources requested for the job
+        if (job.count_resources (jobspec) < 0) {
```
grondo: A question: The plugin currently calls `job.count_resources (jobspec)` in every callback. Is this necessary? Shouldn't the counts be calculated once then reused? This would save unpacking the whole jobspec on every callback.

There may be a subtlety I'm missing though.

grondo (commented):
Just made a quick first pass. One thing I'm wondering is if cur_sched_nodes and cur_sched_cores are adjusted on reload/restart? Looking at the diff on github it is difficult to tell, but I didn't see any tests that ensured counts are correct after a restart.

src/plugins/mf_priority.cpp
```diff
@@ -1444,6 +1497,29 @@ static int sched_cb (flux_plugin_t *p,
...
     a->queue_usage[queue_str].cur_sched_jobs++;
 
+    if (jobspec == NULL) {
+        flux_jobtap_raise_exception (p, FLUX_JOBTAP_CURRENT_JOB, "mf_priority",
+                                     0, "job.state.sched: failed to unpack " \
+                                     "jobspec");
+        return -1;
+    } else {
+        // if a queue cannot be found, just set it to ""
+        queue_str = queue ? queue : "";
```
grondo: Note: this line is redundant with 1497 above.

src/plugins/mf_priority.cpp
cmoussa1: Yeah, this would be a clean solution if I could just count it once (maybe in `job.new`?) and then refer to it throughout the rest of the callbacks. That could probably also apply even to plugin reloads when it is rediscovering jobs for the first time.

Maybe I should try just packing a `Job` object in `job.new` that can be referenced throughout the rest of the plugin.

cmoussa1: Thanks for your advice here - I think I was actually able to get somewhere here! It ended up being a much cleaner solution. I just had to pack the a `Job` object during `job.new` (as well as in `job.state.priority` in the case where the plugin is reloaded on an active job and the aux items are cleared) and could just unpack it in the later callbacks and reference the node and core counts.

One of the things I'm going back and forth on is if on an unpack failure (i.e. if `flux_jobtap_job_aux_get ()` returns a null pointer), the callback should fall back to re-parsing jobspec. In the commit I force-pushed, I resorted to just raising an exception on the job, but perhaps just re-parsing jobspec might be a good failsafe? What do you think?

cmoussa1:
Thanks a lot for your feedback here @grondo so far. I've rebased and force-pushed up some changes to this PR to hopefully make it more straightforward how this new dependency is implemented in the plugin.

Attributes about the job (and in particular, its scheduled nodes and core count) are stored in the job as an `aux` item in `job.new`, where it can just be unpacked and referenced throughout the various callbacks that require it instead of having to unpack and re-parse jobspec throughout every callback.

The `job.state.inactive` callback has also been reworked to add an additional check for cancelled jobs that were specifically in `SCHED` state before decrementing various counters to avoid double-counting the active counters.

grondo (approved):
This LGTM!

One thing to check that falls outside this PR: When held jobs are released and the speculative counter is incremented for the pending offset, is the counter incremented even if there are still _other_ dependencies placed on the job by accounting? This would result in the opposite issue you fixed before: the speculative count is incremented when a job is not really yet released to the SCHED state, so subsquent jobs that could be eligible to be released are not. Just something to check and add a test for.

cmoussa1:
Thanks so much @grondo! Your suggestion actually brought up a use case that broke this existing PR 😅 the change to add a speculative counter for both the nodes and cores in SCHED state actually did not hold for the case where more than one held job _is_ eligible for release because the speculative counter was being double-counted with the association's actual counter by the time a subsequent held job is checked for eligible release, and I _believe_ the reason is because `sched_cb ()` fires synchronously (and already increments the number of resources in SCHED state). So, keeping a speculative counter for the resources requested does not behave as expected. I've gone ahead and removed it from `under_queue_max_sched_nodes ()` and `under_queue_max_sched_cores ()` (as well as their overloaded functions).

I've added sharness tests for the scenario I outlined above and just appended them to the new test file proposed in this PR.

I think the original piece of feedback that "releasing a held job in the loop doesn't update counters speculatively" was under the assumption that `sched_cb ()` didn't fire synchronously? Your expertise could be useful here. If it's not too much trouble, could you give the last two commits one last look and let me know if what I've described here makes sense?

grondo:
>  I believe the reason is because sched_cb () fires synchronously (and already increments the number of resources in SCHED state). So, keeping a speculative counter for the resources requested does not behave as expected. I've gone ahead and removed it from under_queue_max_sched_nodes () and under_queue_max_sched_cores () (as well as their overloaded functions).

Yes, I think you are right. I hadn't looked back at this carefully and had made the assumption that events posted to _any_ job are deferred until after the current callback returns via the `job->event_queue`. However, the event queue is not used when an event is posted to a different job or there's no currently processing job which I think is the case here? So the `dependency-remove` event is posted immediately, and _if that's the last remaining dependency_ the job is released and proceeds immediately to PRIOIRTY, if the `job.state.priority` callback can immediately provide a priority, then the job proceeds immediately to SCHED and `job.state.sched` is called. I assume this is what you were seeing in testing?

This is nice for this case because you don't need the speculative counter: jobs proceed to SCHED as they are released and counters are updated before `flux_jobtap_dependency_remove()` returns.

However, there's at least a couple cases where the jobs might not proceed immediately to SCHED, so the counter would not update immediately:
- there are other dependencies placed on the job, either managed by flux-accounting or not. The mf_priority plugin releases the job, but it stays in DEPEND state and counters are not updated. Therefore other jobs that should not be eligible are released.
- The job does not move immediately from PRIORITY->SCHED. I'm not sure this could happen now, but something to think about.

I wonder if the plugin could check the current state of the job after removing the dependency, and if it is not in the SCHED state then a speculative counter is required?

A specific test case here might be to submit a series of jobs over the associations max-sched-jobs limit, all with afterany dependencies on the same running job. Remove one job in SCHED state such that *one* job should have its max-sched-job dependency removed and be eligible to run. Finally, cancel the antecedent job on which all the other jobs depend. Only one job should run, not all of them.

cmoussa1:
I've rebased to catch up after a number of approved PRs. I know this has already been approved, but I believe that was before we noticed a couple of issues fixed by #880. @grondo, I'm not sure if you would want to give this another a look before I officially merge this!

cmoussa1:
I'm going to set MWP here shortly unless there are any objections!

Issue #858, opened 2026-04-24
job usage: redesign job usage decay to make it more configurable post-DB deployment


cmoussa1:
I had an offline conversation with @ryanday36 about the possibility of redesigning how job usage is decayed in flux-accounting to make it more configurable post-deployment.

Currently, the number of job usage "bins" is baked in at **database creation time** using the `PriorityDecayHalfLife` and `PriorityUsageResetPeriod` parameters. These values become hardcoded as actual schema columns in `job_usage_factor_table`. The decay logic then multiplies each value in a bin by 0.5. The drawback here is that the schema **is** the configuration - the only way to figure out what the configuration is is to actually look at how many columns are in the table.

While there are going to be a number of steps required to change this design to be more extensible, there are a couple of features I could probably implement to make this feature a) easier to read and understand at a high-level and b) configurable after the database has been deployed:

- adding a configuration table to the database would allow parameters like the decay factor, the number of usage bins, the reset period, and any other potentially relevant information easily documentable and accessible.
- Use a row-per-period design instead of columns. Instead of `usage_factor_period_0`, `usage_factor_period_1`, etc. as columns, we could introduce a new table that stores the periods as rows:

```sql
CREATE TABLE IF NOT EXISTS job_usage_factor_table (
    username   tinytext NOT NULL,
    userid     int(11)  NOT NULL,
    bank       tinytext NOT NULL,
    period     int(11)  NOT NULL,
    value      real     DEFAULT 0.0,
    PRIMARY KEY (username, bank, period)
);
```

This design allows for the number of bins to be changed as simply as just adding or deleting rows instead of completely altering the schema. The `apply_decay_factor()` method can just query `WHERE period < num_periods` instead of iterating over all of the constructed columns.

There are probably a bunch of specifics to be worked through that I haven't thought about yet, but this issue can be used as a feature tracking one to organize sub-issues as I work through this design.

ryanday36:
Thanks for opening this issue and starting to work through this @cmoussa1! In addition to letting us change the number of periods, would the row-per-period design also allow us to change how long each period is and how much we decay the usage in each period (i.e. rather than decaying by 0.5 every week, decay by 0.9 (ish) every day)?

cmoussa1:
[...]

PR #872, opened 2026-05-19
database: add `config_table` to database schema

### LINKED ISSUE #859: database: add a `config_table` to the flux-accounting database
Currently, the configuration parameters that dictate how job usage and its decay is calculated (`PriorityDecayHalfLife`, `PriorityUsageResetPeriod`, and the decay factor) are either implied by the schema structure (column count in job_usage_factor_table) or hard-coded in `job_usage_calculation.py`. There is no way to inspect a live database and know what parameters it was created with without counting columns or reading source.

As a first step toward making the decay system more configurable post-deployment, I should introduce a `config_table` that stores these values explicitly as query-able data.

On database creation, populate it with the relevant parameters. At minimum:

- `priority_decay_half_life`: the number of weeks per half-life period
- `priority_usage_reset_period`: the number of weeks before usage fully resets
- `decay_factor`: currently always 0.5, but should be stored explicitly

Other potentially relevant factors can be added in the future as well, maybe with a nice command-line interface.

cmoussa1:
#### Problem

As mentioned in #859, the configuration parameters that dictate how job usage and its decay is calculated (`PriorityDecayHalfLife`, `PriorityUsageResetPeriod`, and the decay factor) are either implied by the schema structure (column count in job_usage_factor_table) or hard-coded in `job_usage_calculation.py`. There is no way to inspect a live database and know what parameters it was created with without counting columns or reading source code.

---

This PR makes the first step for the design outlined in #858 by introducing a `config_table` to the flux-accounting DB that stores these values explicitly as query-able data, as well as allows any generic key-value pair to be added. At a minimum, the table stores three values: `priority_decay_half_life`, `priority_usage_reset_period`, and `decay_factor`. 

When the flux-accounting database is first set up, these values are `INSERT`-ed into the table and populated either with custom Flux Standard Duration (FSD) values for the half-life and reset period parameters (which can be passed on the command line during `create-db`), or they are set to their previous default values (1 week for PriorityDecayHalfLife and 4 weeks for PriorityUsageResetPeriod). A side effect of this change is that the unit for the `--priority-decay-half-life` and `--priority-usage-reset-period` optional arguments for the `create-db` command have been changed from weeks to FSD.

On an already-existing database, these values are `INSERT`-ed while the database is being upgraded with `flux account-update-db` (which will occur automatically if the flux-accounting cron jobs are set up).

Note that no Python bindings are added to actually interact with `config_table` in this PR as they will be added in a subsequent PR in hopes that this is simpler to review.

#### TODO

- [x] add `--decay-factor` optional argument to `create-db` command
- [x] change `flux-account-create-db(1)` to reflect the change in expected type for `--priority-decay-half-life` and `--priority-usage-reset-period`

Closes #859


src/bindings/python/fluxacct/accounting/create_db.py
```diff
@@ -262,4 +270,40 @@ def create_db(
...
+
+    conn.execute(
+        f"INSERT INTO config_table VALUES "
+        f"('priority_usage_reset_period', {priority_usage_reset_period})"
+    )
+    conn.execute(
+        f"INSERT INTO config_table VALUES "
+        f"('priority_decay_half_life', {priority_decay_half_life})"
+    )
+    conn.execute(f"INSERT INTO config_table VALUES " f"('decay_factor', '0.5')")
```
jameshcorbett: I don't think it's a huge threat in this case but you could avoid SQL injection problems with by using the `?` parameterization:

```
  conn.execute(
      "INSERT INTO config_table VALUES (?, ?)",                                                                                                  
      ('priority_usage_reset_period', priority_usage_reset_period)                                                                               
  )                                                               
  conn.execute(                                                                                                                                  
      "INSERT INTO config_table VALUES (?, ?)",
      ('priority_decay_half_life', priority_decay_half_life)                                                                                     
  )                                                         
  conn.execute(                                                                                                                                  
      "INSERT INTO config_table VALUES (?, ?)",
      ('decay_factor', str(decay_factor))                                                                                                        
  )     
```

src/bindings/python/fluxacct/accounting/create_db.py
```diff
@@ -302,7 +305,9 @@ def create_db(
         f"INSERT INTO config_table VALUES "
         f"('priority_decay_half_life', {priority_decay_half_life})"
     )
-    conn.execute(f"INSERT INTO config_table VALUES " f"('decay_factor', '0.5')")
+    conn.execute(
+        f"INSERT INTO config_table VALUES " f"('decay_factor', '{decay_factor}')"
+    )
     conn.commit()
```
jameshcorbett, line 324: should use `?`

src/cmd/flux-account.py
```diff
@@ -577,6 +577,13 @@ def add_create_db_arg(subparsers):
         ),
         metavar=("DURATION"),
     )
+    subparser_create_db.add_argument(
+        "--decay-factor",
+        help="the amount of decay to apply to historical usage",
+        type=float,
+        default=0.5,
+        metavar="DECAY_FACTOR",
+    )
```
jameshcorbett, line 586: Should you validate that the value is between 0 and 1? And then probably have some tests for that as well.

jameshcorbett (approved):
Looks good, a few things.

cmoussa1:
Thanks @jameshcorbett! These were all great suggestions. I've gone ahead and force-pushed up some changes to this PR based on your feedback. I will set MWP here shortly

PR #873, opened 2026-05-21
commands: add missing shebangs

sam-maloney:
Problem: `flux-account.py` and `flux-account-service.py` are executable but have no shebangs. This can lead to the exceutable bit being stripped during RPM builds.

Add shebangs to match other Python command scripts.

Unless there is a reason why all of the Python scripts in `src/cmd/` have shebangs except for these two?


cmoussa1 (approved):
This LGTM! I must have never added shebangs for these Python files. Thanks for catching this!

PR #875, opened 2026-05-29
plugin: add pending-offset limit checks when potentially releasing jobs

### LINKED ISSUE #874: plugin: `check_and_release_held_jobs ()` over-releases held jobs for count-based sched/run limits
When an association has more than one held job sharing the same count-based limit (`max_sched_jobs` and `max_run_jobs`), a single `job.state.run` transition can release _all_ of them at once instead of just the one slot that actually opened. The result is that an association can briefly exceed its configured `max_sched_jobs` or `max_run_jobs` limits because held jobs are released without the loop accounting for the slots consumed by jobs it has already released earlier in the same loop.

This issue can be reproduced by:

* setting `max_sched_jobs` to 2 for an association
* submitting a job that consumes all resources
* submitting jobs to be placed in SCHED to take up the `max_sched_jobs` limit and then submitting two more to actually trigger the limit
* cancelling the running job

Instead of just one job being released, _both_ jobs get released in the same loop because `under_max_sched_jobs ()` returns `true` for both held jobs, thereby giving the association more jobs in SCHED than its configured limit.

The fix seems to be to track within a single `check_and_release_held_jobs ()` pass, how many held jobs have already been released against each count-based limit, and subtract that running tally from the available headroom when evaluating the next held job.

cmoussa1:
#### Problem

`check_and_release_held_jobs ()` gates the release of held jobs on the current value of an association's count-based limits, but releasing a held job within the loop doesn't update the counter the next iteration reads. As a result, when an association has more than one job held on the same count-based limit (`max_sched_jobs` or `max_run_jobs`, per-association or per-queue), a single `job.state.run` transition can release all of them at once instead of just the one slot that actually opened, which can put an association over their defined limit.

---

This PR adds an internal tally counter to the four count-based limit checks and updates `check_and_release_held_jobs ()` to track how many held jobs it has released against each limit within a single pass and feeds that tally into the checks. It uses overloaded functions to compare how many jobs have potentially already been released within the loop with the association's actual limit since the `cur_sched_jobs` and `cur_run_jobs` counters (both per-association and per-queue) are not changed until the job actually enters `job.state.run`.

A new sharness test file is also added to check the various counters and ensure that only one held job is released at a time when `check_and_release_held_jobs ()` is iterating through an association's set of held jobs.

Fixes #874

### DIFF
--- a/src/plugins/accounting.hpp
+++ b/src/plugins/accounting.hpp
@@ -120,17 +120,32 @@ public:
     json_t* to_json () const;    // convert object to JSON string
     // check to see if a job can be released from all flux-accounting
     // dependencies
+    //
+    // the "pending" parameter in the overloaded functions is used to account
+    // for jobs that have been released in the current pass of the held jobs
+    // loop but have not yet been reflected in the association's persistent
+    // counters; by passing in the number of released jobs against a particular
+    // limit, we ensure that subsequent jobs in the held jobs vector are
+    // evaluated against the correct headroom for that limit
     bool under_max_run_jobs ();
+    bool under_max_run_jobs (int pending);
     bool under_queue_max_run_jobs (const std::string &queue,
                                    std::map<std::string, Queue> queues);
+    bool under_queue_max_run_jobs (const std::string &queue,
+                                   std::map<std::string, Queue> queues,
+                                   int pending);
     bool under_max_resources (const Job &job);
     bool under_queue_max_resources (
                                   const Job &job,
                                   const std::string &queue,
                                   const std::map<std::string, Queue> &queues);
     bool under_max_sched_jobs ();
+    bool under_max_sched_jobs (int pending);
     bool under_queue_max_sched_jobs (const std::string &queue,
                                      std::map<std::string, Queue> &queues);
+    bool under_queue_max_sched_jobs (const std::string &queue,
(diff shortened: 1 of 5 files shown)


src/plugins/accounting.hpp
```diff
@@ -121,16 +121,24 @@ class Association {
...
                                   const std::string &queue,
                                   const std::map<std::string, Queue> &queues);
     bool under_max_sched_jobs ();
+    bool under_max_sched_jobs (int pending);
     bool under_queue_max_sched_jobs (const std::string &queue,
                                      std::map<std::string, Queue> &queues);
+    bool under_queue_max_sched_jobs (const std::string &queue,
+                                     std::map<std::string, Queue> &queues,
+                                     int pending);
 };
```
grondo, line 149: I think after this change it might be helpful to have comments for each of these functions that now take a `pending` argument that describe the function, what it does, and what the pending argument is used for.

grondo (approved):
LGTM! Just a suggestion of some documentation for the methods that now take a running count. I think that will help future developers looking at the code.

cmoussa1:
Thanks @grondo! I've gone ahead and force-pushed up a fix to add a comment to `accounting.hpp` for those overloaded functions to explain their subtle difference from the others. I'll go ahead and set MWP here shortly

PR #876, opened 2026-05-29
plugin: `goto error` on unpack fail instead of continuing to initialize `queues` map

cmoussa1:
#### Problem

If `json_unpack_ex ()` fails (e.g. due to a protocol version mismatch), then the `queues` map is still viable to be initialized with potentially garbage values.

---

This PR just updates the callback to `goto error` if `json_unpack_ex ()` fails unpacking information from the RPC containing queue data from the flux-accounting database. As a result, some of the tests in the testsuite that generate this payload manually have also been updated to contain the correct information that the flux-accounting database would be sending to the plugin via RPC.

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -684,8 +684,10 @@ static void rec_q_cb (flux_t *h,
                             "priority", &priority,
                             "max_running_jobs", &max_running_jobs,
                             "max_nodes_per_assoc", &max_nodes_per_assoc,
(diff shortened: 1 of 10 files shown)


jameshcorbett (approved):
LGTM, could you maybe add a test for a malformed payload though, one that's missing some fields?

cmoussa1:
Thanks @jameshcorbett! I've gone ahead and added a third commit to add a simple test of expecting a logged error when a malformed payload is sent to this callback. I'll go ahead and set MWP here

PR #880, opened 2026-06-04
plugin: miscellaneous fixes to releasing jobs in SCHED state

cmoussa1:
#### Problem

If a job is cancelled before it ever runs, the association's overall `cur_sched_jobs` and `cur_sched_jobs` for that queue are never
decremented.

The priority plugin currently keeps track of SCHED-related dependencies removed from jobs in `check_and_release_held_jobs ()` by
using a speculative counter to bump the number of jobs whose dependencies have been removed. However, jobs that are held by SCHED-related dependencies increment the association's number of jobs in SCHED state synchronously (i.e. the `job.state.sched` callback is called synchronously in the case where there is more than one held job), and thus, using a speculative counter double-counts the total number of current jobs in SCHED state, which can incorrectly hold a job from being released of its dependencies.

In a similar vein, the speculative counters for the running job limits (both per-association and per-queue) are bumped regardless of whether a job is fully eligible to be released or not (i.e. it has had all of its dependencies cleared or not), which can result in incorrectly holding jobs when they should be released.

---

The first thing this PR does is add decrements in the callback for `job.state.inactive` for jobs that were in SCHED state but never ran.

This PR adds a check for the state of the job when it no longer has any flux-accounting dependencies placed on it to see if it is in SCHED state or not. If it is not, *then* use speculative counters to keep track of the job that was released so that other jobs do not incorrectly get released due to limits that may not have been updated yet.

Lastly, it moves the bump of the speculative counters for the running job limits to a conditional where they are only bumped *if* the job has met all requirements to be released to SCHED state; if not, it does not bump the speculative counter and move onto the next eligible job.

Tests are added for accurately checking and working through multiple held jobs due to max-sched-jobs and max-running-jobs limits (both per-association and per-queue).

### DIFF
--- a/src/plugins/mf_priority.cpp
+++ b/src/plugins/mf_priority.cpp
@@ -303,6 +303,16 @@ static int check_and_release_held_jobs (flux_plugin_t *p, Association *b)
     // fire, so without this local counter, a second held job would observe the
     // same headroom as the first and be released even though the limit no
     // longer permits it.
+    //
+    // A held job may carry more than one flux-accounting dependency, and it is
+    // only truly released to SCHED state once *all* of them are removed. We
+    // must therefore not reserve headroom for a job that had one dependency
+    // removed but still retains another: doing so would consume limit headroom
+    // on behalf of a job that is not actually entering SCHED state, wrongly
+    // keeping subsequent eligible jobs held. To avoid this, each iteration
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
Nice!

PR #881, opened 2026-06-11
add bindings, commands for interacting with `config_table`

### LINKED ISSUE #865: commands: add commands to interface with new `config_table` in database
Like many of the other tables in the flux-accounting database, bindings and commands should be added to interface with all of the data in `config_table`.

cmoussa1:
#### Problem

There is no way to interact with `config_table` to view, add, edit, or remove key-value pairs.

---

This PR adds Python bindings and commands to add, edit, view, and remove key-value pairs from `config_table`. Some basic unit and sharness tests are also added for each of the functions as well as their front-end callers.

Closes #865

### DIFF
--- a/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
@@ -13,7 +13,11 @@ import csv
 import sqlite3
 import json
 
+import fluxacct
 from fluxacct.accounting.util import with_cursor
+from fluxacct.accounting import formatter as fmt
+from fluxacct.accounting import sql_util as sql
+from flux.util import parse_fsd
 
 
 def export_db_info(conn):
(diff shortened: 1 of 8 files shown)


src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -202,3 +206,108 @@ def export_as_json(conn, cursor):
 
     # return a single JSON object containing the above DB information
     return json.dumps(config)
+
+
+@with_cursor
+def add_config(conn, cursor, key_value_string):
+    """
+    Add a key-value pair to config_table.
+    """
```
jameshcorbett, line 220: This could be `"""Add a key-value pair to config_table."""`. I think that looks nicer but up to you of course.

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -202,3 +206,108 @@ def export_as_json(conn, cursor):
...
+    key, value = key_value_string.split("=")
```
jameshcorbett, line 223: Is it ok not to check that there is exactly one `=` in each string?

t/python/t1017_config_cmds.py
```diff
@@ -0,0 +1,199 @@
...
+        d.edit_config(conn, key_value_strings=["priority_usage_reset_period=1d"])
+        test = cur.execute(
+            "SELECT value FROM config_table WHERE key='priority_usage_reset_period'"
+        ).fetchone()
+        self.assertEqual(test["value"], "86400.0")
+
+    # trying to set PriorityDecayHalfLife to a non-integer will raise a ValueError
+    def test_07_edit_priority_decay_half_life_bad(self):
+        with self.assertRaises(ValueError):
+            d.edit_config(conn, key_value_strings=["priority_decay_half_life=foo"])
```
jameshcorbett, line 76: Maybe add a test for a key_value_string that has no `=` and one that has multiple?

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -202,3 +206,108 @@ def export_as_json(conn, cursor):
...
+            if (float(value) < 0) or (float(value) > 1):
+                raise ValueError(
+                    "decay_factor must be a floating-point value between 0 and 1"
+                )
+        cursor.execute(
+            "UPDATE config_table SET value=? WHERE key=?",
+            (value, key),
+        )
+
+    conn.commit()
```
jameshcorbett: This `commit` call isn't present in `add_config` and I wonder whether it is necessary at all?

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -202,3 +206,108 @@ def export_as_json(conn, cursor):
...
+            value = parse_fsd(str(value))
+        if key == "decay_factor":
+            if (float(value) < 0) or (float(value) > 1):
+                raise ValueError(
+                    "decay_factor must be a floating-point value between 0 and 1"
+                )
+        cursor.execute(
+            "UPDATE config_table SET value=? WHERE key=?",
+            (value, key),
+        )
```
jameshcorbett, line 261: Maybe check if any rows were actually updated?
```  
if cursor.rowcount == 0:                                                   
      raise ValueError(f"key {key} not found in config_table")      
```

And add a test?

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -202,3 +206,108 @@ def export_as_json(conn, cursor):
...
+
+    conn.commit()
+    return 0
```
jameshcorbett, line 265: Why return anything?

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -202,3 +206,108 @@ def export_as_json(conn, cursor):
...
+    key, value = key_value_string.split("=")
+    cursor.execute(
+        "INSERT INTO config_table (key, value) VALUES (?, ?)",
+        (
+            key,
+            value,
+        ),
+    )
+
+    return 0
```
jameshcorbett, line 232: why return anything?

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -202,3 +206,108 @@ def export_as_json(conn, cursor):
...
+        key, value = key_value_string.split("=")
+
+        if key in bin_config_keys:
+            # parse value as Flux Standard Duration (FSD)
+            value = parse_fsd(str(value))
+        if key == "decay_factor":
+            if (float(value) < 0) or (float(value) > 1):
+                raise ValueError(
+                    "decay_factor must be a floating-point value between 0 and 1"
+                )
```
jameshcorbett, line 257: should this validation go in `add_config` as well?

jameshcorbett (approved):
LGTM but a bunch of little things!

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
cmoussa1: This is a great suggestion! I think I might actually update this docstring to also contain the arguments to the function, which matches the others.

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
cmoussa1: I `return 0` here because the front-end that calls these functions looks to print anything returned _unless_ `0` is returned:

```python
if list(return_val.values())[0] != 0:
    print(list(return_val.values())[0])
```

Although, I could just check for `None` instead and not have to have all of these `return 0` calls all throughout the bindings. That will affect most, if not all, of the other Python bindings, however, and might fall out of the scope of this PR. I can open an issue to track this

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
cmoussa1: Ah, yeah, since the functions use a context manager to handle the SQLite connection, the changes are automatically committed, so yes, I don't believe this is necessary. I'll go ahead and remove them. Thanks for catching this!

PR #882, opened 2026-06-15
database: add new table to store job usage factors

### LINKED ISSUE #866: database: add new job usage per-association table
To use the new row-per-period design outlined in #858, the database should use a new table to store job usage factors as _rows_ in a table instead of as columns. Functionality to migrate data from the old `job_usage_factor_table` to this new table should exist in the cases where there is already a flux-accounting database in production on a system.

cmoussa1:
#### Problem

`job_usage_factor_table` is rigid in design in that its configuration for how many usage periods there are (or how long each
half-life for each usage period is) cannot be changed once the database is created. To make this more flexible, this kind of data should be stored on a *per-row* basis instead of *per-column*.

---

This PR adds a new table to the flux-accounting database: `job_usage_per_association_table`, which will serve as a new-and-improved version of `job_usage_factor_table`. The key difference here is that `job_usage_per_association_table` stores each job usage period per-association as *rows*, which means that in the future, the number of "job usage bins" or how long each job usage period is can be changed by just inserting or deleting rows.

Since this PR does not contain the conversion of all of flux-accounting's job usage calculation code to make use of this new table, only one test is added to make sure that when a new association is added, the new table is also populated with the correct amount of rows.

Closes #866


src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -207,6 +208,44 @@ def clear_projects(conn, cur, username, bank=None):
     return 0
 
 
+def insert_per_assoc_usage_rows(conn, cur, username, uid, bank):
```
jameshcorbett, line 211: Should this whole function run in a transaction?

src/bindings/python/fluxacct/accounting/user_subcommands.py
```diff
@@ -207,6 +208,44 @@ def clear_projects(conn, cur, username, bank=None):
...
+    cur.execute(
+        "SELECT value FROM config_table WHERE key='priority_usage_reset_period'"
+    )
+    reset_period = cur.fetchone()
+
+    cur.execute("SELECT value FROM config_table WHERE key='priority_decay_half_life'")
+    half_life = cur.fetchone()
+
+    if reset_period and half_life:
+        num_periods = math.ceil(float(reset_period[0]) / float(half_life[0]))
```
jameshcorbett, line 233: Should there be a check that this is > 0?

jameshcorbett (approved):
LGTM two things!

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: I believe that since this function is called inside `add_user()`, it already runs inside of a transaction (one that starts from the first `INSERT` statement from `add_user()`) that doesn't complete and get committed until the final `.commit()` call at the end of `add_user()`.

src/bindings/python/fluxacct/accounting/user_subcommands.py
cmoussa1: Yes, this is a great suggestion! Thanks. I'll add a check for this and fallback to default behavior (setting `num_periods` to 4)

cmoussa1:
Thanks for reviewing this @jameshcorbett! Setting MWP here

PR #883, opened 2026-06-17
job usage: use new table for usage calculation, decay

### LINKED ISSUE #867: job usage: use new row-per-period table for job usage calculation & decay
All of the helper functions and commands that interact with `job_usage_factor_table` should be changed to use the new row-per-period table when calculating, decaying, and displaying job usage factors. References to fetching and editing job usage values in `job_usage_factor_table` should probably be dropped at the same time to avoid duplicate reads/writes to the database.

cmoussa1:
#### Problem

The code that touches job usage calculation and its decay still reference `job_usage_factor_table` when it needs to use the new `job_usage_per_association_table`.

---

This PR reworks the job usage calculation and decay code to use the new `job_usage_per_association_table`. The job usage calculation pipeline has been refactored in the following ways:

* `update_job_usage()` now fetches userid from `association_table` at the start of each update cycle (because it is required as a key in the new table) and passes it through to downstream functions
* Removed the old logic that fetched job usage bins from `job_usage_factor_table`, as this data is now retrieved later in the pipeline from the new table.
* `calc_usage_factor()` is refactored to query job usage bins (the per-period usage values) from `job_usage_per_association_table` at the beginning of the calculation, replacing the previous approach where bins were passed in from the caller.
* `update_curr_usage_col()` is updated to `INSERT`/`UPDATE` rows in `job_usage_per_association_table` using the `(username, userid, bank, period)` schema instead of updating specific columns in the old table.
* `apply_decay_factor()` is refactored to fetch existing period values from the new table, apply decay calculations, and update the corresponding rows (by username, bank, and period) rather than updating multiple columns in a single row.

The `-J/--job-usage` optional argument for the `view-user` command has also been modified to refer to the new `job_usage_per_association_table` instead of the old one. The same modification has been made to the `clear-usage` command.

Existing tests that check job usage values only needed to be modified to query the new `job_usage_per_association_table`, and no changes to their actual expected values have needed to be made in this PR.

Closes #867

### DIFF
--- a/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
+++ b/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
@@ -588,15 +588,9 @@ def clear_usage_period_columns(cur, bank):
         cur: The SQLite Cursor object.
         bank: The bank being cleared.
     """
-    cur.execute("PRAGMA table_info('job_usage_factor_table')")
-    result = cur.fetchall()
-    for column in result:
-        # column[1] accesses just the column name
-        if column[1].startswith("usage_factor_period_"):
-            cur.execute(
-                f"UPDATE job_usage_factor_table SET {column[1]}=0 WHERE bank=?", (bank,)
-            )
-    # clear last_job_timestamp
+    cur.execute(
+        "UPDATE job_usage_per_association_table SET value=0.0 WHERE bank=?", (bank,)
+    )
     cur.execute(
         "UPDATE job_usage_factor_table SET last_job_timestamp=0 WHERE bank=?", (bank,)
     )
--- a/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
+++ b/src/bindings/python/fluxacct/accounting/job_usage_calculation.py
@@ -59,65 +59,83 @@ def update_hist_usg_col(acct_conn, usg_h, user, bank):
     )
 
 
-def update_curr_usg_col(acct_conn, usg_h, user, bank):
+def update_curr_usg_col(acct_conn, usg_h, user, bank, userid):
     """
     Write the current job usage factor for the association to the
     job_usage_factor_table.
     """
-    u_usg_factor = """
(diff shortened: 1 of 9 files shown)


src/bindings/python/fluxacct/accounting/job_usage_calculation.py
```diff
@@ -59,76 +59,107 @@ def update_hist_usg_col(acct_conn, usg_h, user, bank):
...
+        # the oldest period just gets dropped off the end since it no longer affects
+        # historical usage
+        next_period = period + 1
+        cur.execute(
+            """
+            UPDATE job_usage_per_association_table SET value=?
+            WHERE username=? AND userid=? AND bank=? AND period=?
+            """,
+            (value * decay, user, userid, bank, next_period),
         )
```
jameshcorbett, line 117: Should this not be an `INSERT ... ON CONFLICT UPDATE` statement? Otherwise how are new periods added to the table?

jameshcorbett (approved):
LGTM one question!

src/bindings/python/fluxacct/accounting/job_usage_calculation.py
cmoussa1: Are you referring to whenever the job usage parameters get reconfigured to have a different number of periods? If so, this function is only ever responsible for decaying and updating _existing_ columns that contain job usage values. There isn't any code to change the number of job usage periods (yet), but when it does get added, new periods will be inserted into the table in a function separate from this. Maybe I am misunderstanding your question though!

jameshcorbett: No I think you got it! I was thinking that if the table didn't have any periods here, this function would silently have no effect. I was wondering whether new rows for usage would be added as time passed and new periods were added, but if all the rows are present from the beginning this seems correct!

cmoussa1:
Thanks a bunch for the feedback here @jameshcorbett! Setting MWP here

Issue #885, opened 2026-06-22
flaky test: t1011-job-archive-interface.t 19


sam-maloney:
I am getting intermittent failures:
```
FAIL: t1011-job-archive-interface.t 19 - check that job usage and fairshare values get updated
```
of this test: https://github.com/flux-framework/flux-accounting/blob/master/t/t1011-job-archive-interface.t#L152

I also note that there is another test a few down with an identical description but a different test script pattern, which I've never had fail: https://github.com/flux-framework/flux-accounting/blob/master/t/t1011-job-archive-interface.t#L175

I haven't had time to investigate further, and unfortunately my CI pipeline didn't save the detailed logs (I've now modified it to hopefully save the logs in case it happens again) but I thought I'd open an issue now in case you can see any potential race condition or something?

cmoussa1:
Ah, yeah - this test has been known to be a little flaky for me too. I think I had written this set of tests a long while back, and the job archive interface has long since evolved to be a little more robust. I think I should try to figure out a more reliable way to test for these particular job usage values.

It's quite possible that there is a race between writing the job usage/fair-share values to the database and subsequently reading from the database to check for certain values, since they occur in back-to-back tests. If this test fails for you again, please feel free to dump the logs here!

cmoussa1:
@sam-maloney just wanted to follow up on this - are you seeing this test file acting up at all in the last couple weeks since #886 landed? If not, I might close this issue for now.

sam-maloney:
@cmoussa1 I haven't actually run a build since, but once v0.59.0 releases I will rebuild my RPMs and report back here on the outcome!

sam-maloney:
Everything seems to be passing well in my latest builds! Many thanks @cmoussa1 🙏

cmoussa1:
No problem at all! I'm glad to hear it 😁 thanks for opening this and letting me know

PR #886, opened 2026-06-22
t1011: improve job usage check in tests

cmoussa1:
#### Problem

As mentioned in #885, there are intermittent failures in one of the tests in `t1011-job-archive-interface.t` because of a potential race in the write to the database from a job-usage/fair-share update and read from the database.

---

This PR makes some minor and general improvements to `t1011-job-archive-interface.t` by using the `view-user` command from the flux-acounting service (i.e. dropping the `-p` argument) that is started before the tests run. Some general improvements are also made to the subsequent tests that check job usage values by wrapping the bash variable `username`
in curly brackets, renaming the JSON files that those tests create to incrementally follow the first test, and also dropping the `-p ${DB}` argument in front of the `view-user` command.

Even after this PR lands, let's keep #885 open for a little while longer to make sure that the intermittent failures that @sam-maloney was seeing has truly been resolved. :-)

### DIFF
--- a/t/t1011-job-archive-interface.t
+++ b/t/t1011-job-archive-interface.t
@@ -176,10 +176,13 @@ test_expect_success 'run update-usage and update-fshare commands' '
 	flux account-update-fshare -p ${DB_PATH}
 '
 
+# The association has now submitted jobs under their secondary bank, so we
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
LGTM!

PR #887, opened 2026-06-22
job usage: add reconfiguration support

### LINKED ISSUE #868: job usage: add reconfiguration support
If a sys admin decides to change the configuration for how job usage is binned and decayed in the flux-accounting database, there should be some mechanism to make note of when the reconfiguration happened and recalculate usage per-association/per-bank. Details of _how_ to recalculate this job usage (i.e. either starting fresh from the reconfiguration date, or maybe trying to calculate job usage backwards using the new bin size) are still pending here.

cmoussa1:
#### Problem

If an admin decides to change how flux-accounting applies decay to jobs as they get older or how long to keep jobs relevant towards calculating total usage (i.e. changing any of the `priority_usage_reset_period`, `priority_decay_half_life`, or `decay_factor` parameters), the usage bins per-association need to also be changed to reflect the new sizes of the bins.

---

This PR adds a helper function that will reconstruct the `job_usage_per_association_table` according to the new parameters that are defined in `config_table`. If any of the `priority_usage_reset_period`, `priority_decay_half_life`, or `decay_factor` parameters are changed, this function is called (with a warning message that requires additional input
since it is completely reconstructing the table according to brand new parameters) and the table is reconstructed to use the new parameters.

Once the update to `job_usage_per_association_table` is complete, a `reconfigure_time` key-value pair in `config_table` is inserted or updated with a timestamp that represents when `job_usage_per_association_table` was last reconfigured. This key-value pair will potentially be referenced during a job usage update to know which jobs to ignore and which ones to include in job usage calculation.

Some minor cleanup of the now soon-to-be-retired `job_usage_factor_table` is also included in this PR. I've also added some unit tests for the new helper function as well as for calculating job usage in general after a reconfiguration.

Closes #868

### DIFF
--- a/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
@@ -12,12 +12,98 @@
 import csv
 import sqlite3
 import json
+import math
+import time
 
 import fluxacct
 from fluxacct.accounting.util import with_cursor
 from fluxacct.accounting import formatter as fmt
 from fluxacct.accounting import sql_util as sql
 from flux.util import parse_fsd
+
+
+@with_cursor
+def reconfigure_usage_bins(conn, cursor):
+    """
+    Update the usage bin configuration for all associations from scratch. This function
+    should be called *after* the admin has updated priority_decay_half_life,
+    priority_usage_reset_period, and/or decay_factor in config_table.
+
+    Args:
+        conn: The SQLite Connection object.
+        cursor: The SQLite Cursor object.
+    """
+    # read the new configuration from config_table
+    cursor.execute(
+        "SELECT value FROM config_table WHERE key='priority_decay_half_life'"
+    )
+    half_life = cursor.fetchone()
+
+    cursor.execute(
+        "SELECT value FROM config_table WHERE key='priority_usage_reset_period'"
+    )
+    reset_period = cursor.fetchone()
+
+    if half_life is None or reset_period is None:
+        raise ValueError(
+            "priority_decay_half_life and priority_usage_reset_period must be set in "
+            "config_table before reconfiguring usage bins"
+        )
+
+    new_half_life = float(half_life[0])
+    new_reset_period = float(reset_period[0])
+    new_num_periods = math.ceil(new_reset_period / new_half_life)
+
+    # fetch all associations
+    cursor.execute(
+        "SELECT DISTINCT username, userid, bank FROM job_usage_per_association_table"
+    )
+    associations = cursor.fetchall()
+
+    try:
+        # delete all existing period rows for every association
+        cursor.execute("DELETE FROM job_usage_per_association_table")
+
+        # re-insert the correct number of period rows for every association
+        # initialized to 0.0 under the new configuration
+        for username, userid, bank in associations:
+            for period in range(new_num_periods):
+                cursor.execute(
+                    """
+                    INSERT INTO job_usage_per_association_table
+                    (username, userid, bank, period, value)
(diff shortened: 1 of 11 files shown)


src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -20,6 +22,90 @@
...
+        "SELECT value FROM config_table WHERE key='priority_decay_half_life'"
+    )
+    half_life = cursor.fetchone()
+
+    cursor.execute(
+        "SELECT value FROM config_table WHERE key='priority_usage_reset_period'"
+    )
+    reset_period = cursor.fetchone()
+
+    if not half_life or not reset_period:
```
jameshcorbett: Should these checks maybe be `is None` instead?

t/python/t1018_job_usage_custom_bins.py
```diff
@@ -0,0 +1,117 @@
...
+    # change the configuration of the bins to have a half-life period of 2 weeks
+    # instead of 1 week and a usage reset period of 6 weeks instead of 4 weeks, which
+    # will result in a total of 3 bins per-association
+    @patch("builtins.input", return_value="y")
+    def test_03_edit_configs_1(self, mock_input):
+        d.edit_config(conn, ["priority_usage_reset_period=42d"])
+        d.edit_config(conn, ["priority_decay_half_life=14d"])
+        result = d.list_configs(conn)
+        self.assertIn("priority_usage_reset_period | 3628800", result)
+        self.assertIn("priority_decay_half_life    | 1209600", result)
```
jameshcorbett, line 64: Should there be some tests with various inputs other than `y` and then make sure the changes are rolled back?

src/bindings/python/fluxacct/accounting/job_usage_calculation.py
```diff
@@ -339,6 +339,13 @@ def update_job_usage(acct_conn, pdhl=1):
             key = (job.userid, job.bank)
             association_jobs[key].append(job)
 
+        # get PriorityDecayHalfLife
+        pdhl = float(
+            cur.execute(
+                "SELECT value FROM config_table WHERE key='priority_decay_half_life'"
+            ).fetchone()[0]
+        )
+
```
jameshcorbett, line 358: why this change in this commit?

jameshcorbett (commented):
Generally looks good! There are some stray changes that I have some questions about though.

t/python/t1018_job_usage_custom_bins.py
cmoussa1: Yes, this is a good suggestion. I've gone ahead and added a test to make sure that the changes do not go through with an input other than `y` and ensure that the parameters were the same as before (as well as added a set of unrelated tests to change `decay_factor` and make sure that job usage gets decayed with different values).

src/bindings/python/fluxacct/accounting/job_usage_calculation.py
cmoussa1: Ah, sorry, I think I probably messed this up during a rebase. This change was meant to be a part of a larger change to just remove having to pass the `priority_decay_half_life` parameter to the `update_usage()` function itself in favor of just reading it from `config_table` directly (since that would be more consistent). I've gone ahead and added this change to a separate commit (03d5902) that just removes the parameter from the function header as well as any references to it. Let me know if that makes more sense

jameshcorbett: Yup!

jameshcorbett (approved):

cmoussa1:

PR #891, opened 2026-07-02
user_subcommands: don't catch and re-raise `ValueError` for `validate_queue()`, `validate_project()`

cmoussa1:
#### Problem

The `validate_queue()` and `validate_project()` helper functions raises a `ValueError` when it cannot find a specified queue or project in the flux-accounting database, but the callers of these helper functions also wrap their respective calls in a `try/except` just to re-raise a `ValueError` with a more specific error message.

---

This PR moves the specific error message to the `validate_queue()` and `validate_project()` helpers and removes the `try/except` block around the various calls to `validate_queue()` throughout `user_subcommands.py`.


jameshcorbett (approved):
yupppp

PR #892, opened 2026-07-02
edit-user: add incremental queue management to command

### LINKED ISSUE #889: `edit-user --queues` requires full knowledge of association's queue access list
Currently, adding a queue to a user's association requires passing the complete list of all queues they should have access to via `flux account edit-user --queues=A,B,C`. This is inconvenient because:

* the `--queues` argument completely overwrites the existing queues value rather than adding to it
* to add a single queue, admins must first query the current queues, manually construct the full comma-delimited list, then pass it all to `edit-user`
* it's easy to accidentally remove existing queue access by forgetting to include all current queues

One idea for improvement could potentially be adding incremental operations for managing queues, such as:

```console
$ flux account edit-user user1 --add-queue=C <-- (append without replacing)
$ flux account edit-user user1 --remove-queue=B <-- (remove specific queue)
```

cmoussa1:
#### Problem

Currently, adding a queue to a user's association requires passing the complete list of all queues they should have access to via `--queues`, which requires querying the user's current queues, manually constructing the full comma-delimited list, and passing the entire list to `edit-user`.

---

This PR adds incremental queue operations to the `edit-user` command:

* `--add-queue=<queue>` adds a single queue to existing queues
* `--delete-queue=<queue>` removes a single queue from existing queues

These options work across all of a user's associations when no bank is specified, or on a specific association when used with `--bank`. The existing `--queues` option remains available for full replacement when needed, but cannot be used simultaneously with the new incremental options.

The `edit-user(1)` man page has also been updated to include the new options.

Closes #889

### DIFF
--- a/t/python/t1020_edit_user_properties.py
+++ b/t/python/t1020_edit_user_properties.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2026 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import unittest
+import os
+import sqlite3
+import time
+from unittest.mock import patch
+
+from fluxacct.accounting import create_db as c
+from fluxacct.accounting import bank_subcommands as b
+from fluxacct.accounting import user_subcommands as u
+from fluxacct.accounting import queue_subcommands as q
+
+
+class TestAccountingCLI(unittest.TestCase):
+    @classmethod
+    def setUpClass(self):
+        # create test accounting database
+        self.dbname = f"TestDB_{os.path.basename(__file__)[:5]}_{round(time.time())}.db"
+        c.create_db(self.dbname)
+        global conn
+        global cur
+
+        conn = sqlite3.connect(self.dbname, timeout=60)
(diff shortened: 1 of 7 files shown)


t/python/t1020_edit_user_properties.py
```diff
@@ -0,0 +1,146 @@
...
+        self.assertIn(
+            "cannot specify --queues with --add-queue", str(context.exception)
+        )
+
+    def test_07_validate_nonexistent_queue(self):
+        # try to add a queue that doesn't exist in queue_table
+        with self.assertRaises(ValueError) as context:
+            u.edit_user(conn, username="user1", bank="A", add_queue="nonexistent_queue")
+        print(str(context.exception))
+        self.assertIn("does not exist in queue_table", str(context.exception))
```
jameshcorbett: using [assertRaisesRegex](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) might simplify these a bit!

jameshcorbett (approved):
LGTM, one suggestion. You might consider adding a test for passing `--add-queue=x --delete-queue=x` at the same time?

t/python/t1020_edit_user_properties.py
cmoussa1: Oh neat! Thanks for linking this - TIL. This simplified these kinds of tests quite a bit

cmoussa1:
Thanks @jameshcorbett for the review and the suggestion! I've gone ahead and added a simple check that the same queue shouldn't be specified to _both_ `--add-queue` and `--delete-queue` at the same time. I'll go ahead and set MWP here

PR #895, opened 2026-07-07
doc: add `*-config` man(1) pages, guide to reconfiguring job usage table

### LINKED ISSUE #894: documentation: add guide on reconfiguring job usage factors
The next flux-accounting release will contain support for reconfiguring how job usage is decayed. As this gets tested, I should create some documentation that contains examples on how to reconfigure these parameters and how it affects job usage and fair-share.

cmoussa1:
#### Problem

There are no man(1) pages for any of the `*-config` commands. Also related, there is no official documentation on how to reconfigure the job usage table in the flux-accounting database and how it affects job usage and fair-share calculation.

---

This PR adds man(1) pages for all of the `*-config` commands as well as adds a section to the ReadTheDocs site on how to reconfigure the job usage table, motivation for why an admin might want to do this, and its effects on the database.

Fixes #894


doc/components/job-usage-calculation.rst
```diff
@@ -118,6 +118,125 @@ flux-accounting to *only* consider jobs newer than said timestamp with the
...
+    Reconfiguring job usage parameters resets all job usage values to 0.0 for
+    every association in the database. This has several important implications:
+
+    - **fair-share values will be affected**: Since fair-share calculations
+      depend on job usage, all fair-share values will also see changes based on
+      the new usage data.
+
+    - **historical usage is rebuilt**: The system will replay all job records
+      from scratch using the new configuration. Depending on the size of your
+      jobs table, this may take some time during the next usage update.
```
wihobbs: Maybe give an indicator of how long "some time" should be? Order minutes? 30s for 10,000 users?

doc/components/job-usage-calculation.rst
```diff
@@ -118,6 +118,125 @@ flux-accounting to *only* consider jobs newer than said timestamp with the
...
+
+- ``priority_usage_reset_period``: The total time span over which historical
+  job usage is tracked. This determines the total number of usage bins.
+
+- ``decay_factor``: The multiplier applied to older usage periods when
+  calculating historical usage. The default value is 0.5, meaning each older
+  period contributes half as much as the previous one.
+
+Values for ``priority_decay_half_life`` and ``priority_usage_reset_period`` can
+be specified using Flux Standard Duration (FSD) format or in seconds. FSD
```
wihobbs: Link to FSD RFC?

doc/components/job-usage-calculation.rst
cmoussa1: This is a good suggestion, thanks! I've reworded this section to be a little clearer about what takes an order of minutes to complete

doc/components/job-usage-calculation.rst
cmoussa1: Done! Thanks for pointing this out

doc/man1/flux-account-list-configs.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   Specify a custom output format using Python's string format syntax. Field
+   names should be enclosed in curly braces (e.g., ``{key}->{value}``).
+
+EXAMPLES
+========
+
+List all configuration parameters in table format:
+
+.. code-block:: console
```
wihobbs, line 43: We have `flux config get` which dumps the configuration for an instance, and with these commands, there are now more flux commands to adjust configuration. I imagine there's a good reason for their existence -- maybe that the accounting database needs to know when a key is updated and this is simpler than watching for an event in the core library to see that the instance config was reloaded?

However, accounting, and also Fluxion (when you load the module on the command line) both have ways outside of `flux-config(1)` of adjusting configuration. I imagine the admins would prefer being able to `flux config get` and be presented with _everything_ they might want, rather than having to poke at accounting for accounting info, in `flux-dmesg(1)` for scheduler configuration, etc. 

I'm not going to hold up this PR, great job on the documentation. I just think this might be an area for discussion/further thought as a team since I've now thought about it on two separate framework projects.

wihobbs (approved):

doc/man1/flux-account-list-configs.rst
cmoussa1: I absolutely see your point here - having separate "config" commands for separate Flux entities is cumbersome and having to remember which config options are managed by which config commands is probably confusing. I wonder if there is a way to sync some of these configuration options with the greater Flux system instance configuration. The tricky part is the options live in different places.

This will absolutely require more thought and perhaps a redesign of how options are stored in flux-accounting. Definitely something to think about.

cmoussa1:
Thanks @wihobbs for your feedback here! Going to set MWP

PR #896, opened 2026-07-07
add new `bank-info` command for viewing normalized shares, usage on a cluster

### LINKED ISSUE #884: banks: provide option for showing normalized shares, usage of a cluster
Viewing information about a bank currently only displays things like its total usage and total allocated shares, but not a normalized version to compare it to other banks in the hierarchy. This might be particularly useful to see how much of a cluster one (or multiple) bank(s) is utilizing, or how much of the cluster its been allocated.

cmoussa1:
#### Problem

Users need a way to view hierarchical bank and user information with normalized and effective share calculations to understand their resource allocations within the bank hierarchy.

---

This PR adds a new `bank-info` command to the flux-accounting command suite, drawn from the work by @ryanday36. This command builds and displays a tree structure of all of the banks and users in the flux-accounting database and calculates normalized and effective shares for banks and users.

Some basic unit and sharness tests are added for the command and its corresponding function. A new page is added to the ReadTheDocs site for flux-accounting where it goes into detail about exactly how it calculates normalized shares for a given user or bank in the flux-accounting database.

Fixes #884

### DIFF
--- a/doc/components/bank-info.rst
+++ b/doc/components/bank-info.rst
@@ -1,0 +1,200 @@
+.. _bank-info:
+
+#######################################
+Understanding The ``bank-info`` Command
+#######################################
+
+********
+Overview
+********
+
+The ``bank-info`` command provides a comprehensive view of fairshare
+and resource allocation information for banks and users within the
+flux-accounting database. It displays normalized shares, usage
+statistics, and fairshare values in a hierarchical tree structure that
+reflects the organization of banks and their associations.
+
+This command serves multiple purposes:
+
+- **Quick status checks**: View current resource allocation and usage at a
+  glance
+- **Hierarchy visualization**: Understand the relationship between banks and
+  users
+- **Resource analysis**: Compare allocated shares versus actual usage
+
+Unlike :man1:`flux-account-view-bank` and :man1:`flux-account-view-user`
+which display detailed database attributes, ``bank-info`` focuses
+specifically on share distribution and usage metrics that *may* affect job
+scheduling priorities.
+
+*****************
+Command Execution
+*****************
+
+When called without options, ``bank-info`` displays information for the
+current association. To view the entire database hierarchy, pass the ``-t``
+flag along with the name of the root bank:
+
+.. code-block:: console
+
+    $ flux account bank-info -t root
+    Name                     Shares  Norm Shares   Norm Usage   Level FS
+    root                          1     1.000000     1.000000        inf
+    A                             1     0.333333     0.000000         --
+      user1                       1     0.166667     0.000000      0.500
+      user2                       1     0.166667     0.000000      0.500
+    B                             1     0.333333     0.000000         --
+      user3                       1     0.333333     0.000000      0.500
+    C                             1     0.333333     0.000000         --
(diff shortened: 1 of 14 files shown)


src/bindings/python/fluxacct/accounting/bank_subcommands.py
```diff
@@ -348,3 +459,110 @@ def list_banks(
...
+
+    Args:
+        tree: bank name to display all children of, including users
+        tree_no_users: bank name to display all children of, excluding users
+        to_root: bank name to display all parents for up to root
+        user: username to query
+        verbose: display detailed usage info
+        parsable: output "|" delimited columns for easy parsing
+        noheader: do not display headers
+        exclude: do not display this bank in output (defaults to 'expired')
```
jameshcorbett: Here and throughout this PR, documentation says this defaults to `"expired"` but the code actually defaults to `None`. I think maybe just remove every occurrence of the `(defaults to 'expired')` statement?

src/bindings/python/fluxacct/accounting/bank_subcommands.py
```diff
@@ -87,6 +89,115 @@ def reactivate_bank(conn, cur, bank, parent_bank):
 ###############################################################
 
 
+def _build_sharetree(cur, root_bank):
```
jameshcorbett: Unused parameter?

doc/components/bank-info.rst
```diff
@@ -0,0 +1,200 @@
...
+  .. math::
+
+    \text{Usage}_{normalized_{node}} = \frac{\text{Usage}_{node}}{\text{Usage}_{root}}
+
+Level FS
+  The fair-share value for the association. For more details on
+  fairshare calculation, see :doc:`fair-share`.
+
+********
+examples
```
jameshcorbett:
```suggestion
Examples
```

jameshcorbett (approved):
LGTM, a few minor things. Also Claude has  a question, not sure if it's valid or not so thought I'd see what you think:

> Normalized-shares sibling pool diverges from the canonical fairshare model (bank_subcommands.py:466-473)                           
>    
>   The new code computes sibling shares segregated by type:                                                                              
>   if sharetree[f"{node['parent']}{child}/"]["isuser"] == node["isuser"]
>   So sub-banks are normalized only against sibling banks, and users only against sibling users. But the existing C++ fairshare tree     
>   (weighted_tree.cpp:74-90, calc_set_children_weight) sums sibling_shares_sum over all children regardless of is_user.                  
>                                                                                                                                         
>   Consequence: for a bank that has both sub-banks and users as direct children, each group independently sums to the parent's nshares,  
>   so the level over-counts (total > parent). All the doc examples happen to have banks-only or users-only at each level, so they don't  
>   exercise this. Is the segregation intentional, or should it match the weighted-tree model (one shared pool per level)? This is the
>   main thing to confirm.

cmoussa1:
Thanks @jameshcorbett! That is a good question you brought up. This segregation between child banks and users _is_ intentional and is not allowed behavior for a hierarchy in flux-accounting. So, no, banks and users cannot exist under the same parent bank, so I think we should be okay here.

Issue #890, opened 2026-07-01
accounting: make job usage calculation configurable to support resource types beyond nodes


cmoussa1:
flux-accounting currently defines an association's raw job usage as a single, hard-coded formula: the sum over an association's jobs of `nnodes * t_elapsed` (node-seconds). This value feeds directly into the fair-share calculation and, by extension, job priority.

There has been an ask whether usage can account for other resources — most immediately **cores**, and further out, potentially including resources such as **GPUs, memory, licenses**, etc. On a cluster where jobs are core-packed rather than node-exclusive, node-seconds paints an incomplete picture for a user's or group of users' actual consumption: a job using 1 core on a node and a job using all 128 cores on the same node accrue identical usage today.

`R` describes the allocated resource set (cores, GPUs), so this is probably just a *calculation* problem rather than a *data-collection* problem. We do not need to change what we record about jobs to get started on cores.

There is also already a `config_table` key/value store in use for tunables like `decay_factor`, `priority_decay_half_life`, and `reconfigure_time`. That is a natural home for a site-configurable usage definition, so we likely don't need a brand-new mechanism just to hold the configuration.

## open questions

These are a couple of design questions some or most of the implementation will hinge on. A follow-up design proposal (and the creation of more-specific issues) can work through these questions, but this issue should be used to track sub-issues/features:

1. **How is usage configured?** A weighted sum of per-resource rates (e.g. `node_weight * node_seconds + core_weight * core_seconds`) is the simplest model and mirrors how the priority factors are already weighted. Does that cover the near-term asks, and does it generalize to more resource types in the future without another schema change?

2. **Deriving per-resource counts from `R`.** How robustly can we extract core and GPU counts from the stored `R`/`jobspec`? Perhaps we can leverage existing helpers in flux-core to calculate resource usage by types, since the `R` is already fetched and stored per-job.

3. **Defaults.** The default should probably remain `nnodes * t_elapsed` so existing deployments see no behavior change on upgrade.

cmoussa1:
This feature has officially been landed now, so I'll go ahead and close this, and can re-open specific issues if they come up.

PR #901, opened 2026-07-13
`JobRecord`: add `ncores`, `ngpus` properties

### LINKED ISSUE #897: extract ncores and ngpus from ResourceSet for JobRecord objects
Currently, `JobRecord` objects only extract `nnodes` from the `R` field (ResourceSet). To support configurable job usage calculations that account for cores, GPUs, and potentially other resources (see #890), we need to extract these resource counts as well.

The `flux.resource.ResourceSet` class already provides `.ncores` and `.ngpus` properties alongside the existing `.nnodes` property, so this is a straightforward extension of how current resources are already extracted.

### acceptance criteria

- [x] `JobRecord` objects have `ncores` and `ngpus` attributes
- [x] values are correctly extracted from ResourceSet
- [x] jobs with malformed/missing R field continue to be skipped gracefully
- [x] no change to existing behavior (usage calculation unchanged)
- [x] tests verify ncores/ngpus extraction for various job configurations

cmoussa1:
#### Problem

As mentioned in #890, job usage eventually needs to track and calculate usage that takes more resources into consideration than just nnodes, e.g. ncores and ngpus. To do this, flux-accounting needs to know how many cores and GPUs are consumed by each job.

---

This PR adds two new properties to the `JobRecord` class: ncores and ngpus, attributes that will store the number of cores and GPUs consumed by the job, respectively.

Some basic tests are also added that check accurate counts for cores and GPUs using the new properties for `JobRecord` objects.

Closes #897

### DIFF
--- a/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/jobs_table_subcommands.py
@@ -37,6 +37,8 @@ class JobRecord:
         bank,
         requested_duration,
         actual_duration,
+        ncores,
+        ngpus,
         jobid_format="f58",
     ):
         self.userid = userid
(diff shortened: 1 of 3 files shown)


jameshcorbett (approved):
LGTM

PR #902, opened 2026-07-14
job usage: extend usage calculation to consider cores, GPUs

### LINKED ISSUE #898: add configurable resource weights for job usage calculation
Currently, job usage is calculated as:
```python
usage = nnodes * elapsed  # job_usage_calculation.py:175
```

On core-packed clusters, this means a 1-core job and a 128-core job on the same node accrue identical usage, which is unfair.

We can probably just replace the hardcoded `nnodes * elapsed` usage calculation with a configurable weighted sum formula that can account for nodes, cores, and GPUs. This addresses the core request in #890 to support fair usage accounting on core-packed clusters.

A new (and potentially extensible) formula could look like:

```python
usage = ((nnodes * node_weight) +
         (ncores * core_weight) +
         (ngpus  * gpu_weight)) * elapsed
```

### acceptance criteria

- [x] new config keys added to fresh databases
- [x] existing databases get new keys on upgrade (via migration script)
- [x] default weights (`1.0, 0.0, 0.0`) preserve current behavior exactly
- [x] non-zero core weight changes usage calculation appropriately
- [x] non-zero GPU weight changes usage calculation appropriately
- [x] mixed weights work correctly
- [x] usage calculation handles missing config keys gracefully (uses defaults)
- [x] tests verify backward compatibility
- [x] tests verify weighted calculations with various weight combinations

cmoussa1:
#### Problem

As mentioned in #898, the `calc_usage_factor()` function hard-codes job usage calculation for every job as a product of its node count (nnodes) and its run time. To extend this calculation to consider more resource types, we need to include both the number of cores/GPUs as well as their respective weights in determining a job's total usage.

---

This PR introduces three new key-value pairs to `config_table`: `node_weight`, `core_weight`, and `gpu_weight`, which are keys that represent how much to weigh each resource in the calculation of job usage for a single job.

In **job_usage_calculation.py**, these weights are extracted and utilized when calculating the usage for a given job. The new formula for calculating a job's usage is now:

```python
usage = ((nnodes * node_weight) +
         (ncores * core_weight) +
         (ngpus  * gpu_weight)) * elapsed
```

To maintain backwards compatibility, `node_weight` is the only factor configured with a weight > 0. An administrator can edit the weights for these factors from the command line with `edit-config` to change how usage is calculated for a job.

Existing tests that look at the contents of `config_table` are updated to include the new keys, and a set of unit tests are also added to check that configuring each of the new weights for the added resource types results in different usage calculations for a given job.

##### TODO

- [x] add checks that the new weights cannot be deleted from `config_table`, similar to the existing keys
- [x] complete **acceptance criteria** checklist in #898

Closes #898


jameshcorbett (approved):
I'm in

cmoussa1:
🫡 setting MWP here

PR #904, opened 2026-07-15
update-db: miscellaneous fixes to improve migration to new job usage table

### LINKED ISSUE #903: migration from `job_usage_factor_table` to `job_usage_per_association_table` fails
Upgrading from older flux-accounting databases with legacy `usage_factor_period_N` columns in `job_usage_factor_table` are not having their respective rows migrated to the new `job_usage_per_association_table`.

### Root Causes

1. **Migration timing issue**: The `migrate_job_usage_to_per_assoc()` function runs *after* `update_columns()`, which drops the legacy columns before the migration can read them.

2. **Composite primary key bug**: `update_tables()` only adds the first column of composite primary keys (checking `column[5] == 1` instead of `column[5] > 0`), causing `job_usage_per_association_table` to be created with only `username` as the primary key instead of `(username, bank, period)`.

The solution looks to be the following:

1. Move `migrate_job_usage_to_per_assoc()` to run before `update_columns()` 
2. Fix `update_tables()` to handle composite primary keys correctly
3. Add integrity check to verify all users in `job_usage_factor_table` have entries in `job_usage_per_association_table`

cmoussa1:
#### Problem

- `update_tables()` only added the first column of composite primary keys (checking `column[5] == 1` instead of `column[5] > 0`), causing `job_usage_per_association_table` to be created with only `'username'` as the primary key instead of `(username, bank, period)`. This resulted in `UNIQUE` constraint errors when migrating usage data.
- `migrate_job_usage_to_per_assoc()` ran _after_ `update_columns()`, which dropped the legacy `usage_factor_period_N` columns before migration could read them. This caused all historical job usage data to be lost during database updates.
- Database updates were silently failing to migrate usage data, leaving associations without entries in `job_usage_per_association_table`.

---

This PR adds some miscellaneous fixes to `flux-account-update-db.py` to fix the problems listed above. It joins _all_ columns that are part of a table's PRIMARY KEY when migrating a table from an older version to a newer version. It also moves the call to `migrate_job_usage_to_per_assoc()` before `update_columns()` so that association data is not lost during the update.

**db_integrity_check.py** is slightly expanded to actually check the contents of `job_usage_per_assoc_table()` to make sure that the update to an older version of a flux-accounting DB actually populates the table like it should.

Some expanded logging is also added to the script. In the future, it might be nice to change this script to actually make use of the `logging` module instead of using `print()` statements, similar to the `update-usage` command.

Closes #903

### DIFF
--- a/src/cmd/flux-account-update-db.py
+++ b/src/cmd/flux-account-update-db.py
@@ -299,17 +299,12 @@ def migrate_job_usage_to_per_assoc(cur):
     """
     Migrate existing usage bin columns from job_usage_factor_table into the
     new row-per-period job_usage_per_association table. Each usage_factor_period_N
-    column becomes a row with period=N. Skips migration if the new table
-    already has data, so this is safe to call multiple times.
+    column becomes a row with period=N. Uses INSERT OR IGNORE to skip users
+    who already have entries, so this is safe to call multiple times.
 
     Args:
         cur: the Cursor object used to interact with the database.
     """
-    cur.execute("SELECT COUNT(*) FROM job_usage_per_association_table")
-    if cur.fetchone()[0] > 0:
-        # migration has already been done; just return
-        return
-
     # find all usage bin columns from job_usage_factor_table
     cur.execute("PRAGMA table_info('job_usage_factor_table')")
     columns = cur.fetchall()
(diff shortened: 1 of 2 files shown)


src/cmd/flux-account-update-db.py
```diff
@@ -284,6 +284,7 @@ def migrate_job_usage_to_per_assoc(cur):
     Args:
         cur: the Cursor object used to interact with the database.
     """
+    print("migrating job usage data to per-association table...")
```
jameshcorbett, line 308: If these were `logging.info` calls instead they would be more configurable in the future! Up to you though.

jameshcorbett (approved):
I'm in!

src/cmd/flux-account-update-db.py
cmoussa1: I completely agree. I need to eventually change these `print()` calls to use `logging` like the other flux-accounting scripts. I'll probably change all of the logging for this script in the near future.

cmoussa1:
Thanks for the review @jameshcorbett! Setting MWP here

PR #907, opened 2026-07-16
`update-db`: add logging infrastructure

cmoussa1:
#### Problem

`flux-account-update-db` uses `print()` statements for output, making it inconsistent with other flux-accounting commands and lacking verbosity control.

---

This PR just adds logging infrastructure to the `update-db` command. A new check is also added to **t1017-update-db.t** to make sure that log messages can be read when the command is called.

### DIFF
--- a/src/cmd/flux-account-update-db.py
+++ b/src/cmd/flux-account-update-db.py
@@ -10,6 +10,7 @@
 # SPDX-License-Identifier: LGPL-3.0
 ###############################################################
 import argparse
+import logging
 import os
 import sqlite3
 import sys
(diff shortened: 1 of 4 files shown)


src/cmd/flux-account-update-db.py
```diff
@@ -20,6 +21,14 @@
 
 import fluxacct.accounting
 from fluxacct.accounting import create_db as c
+from fluxacct.accounting import util
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s: %(levelname)s: %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S",
+)
```
jameshcorbett: should maybe `util.config_logging` do this for you maybe?

jameshcorbett (approved):

cmoussa1:
Thanks @jameshcorbett! I've added an additional commit based on your suggestion above to move `basicConfig()` to the `util` module. I'll go ahead and set MWP here

PR #910, opened 2026-07-20
t: add tests for managing unknown queues

cmoussa1:
#### Problem

flux-accounting does not have a dedicated set of tests for validating jobs submitted to queues it does not know about.

---

This PR adds a set of tests that confirms the required behavior defined in [RFC 33](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_33.html) that an external validator shall not reject a job submitted under a queue unknown to the external validator.


grondo (approved):
LGTM!

cmoussa1:
Thanks! Setting MWP here

PR #911, opened 2026-07-20
flux-accounting guide: change queue access policy note to warning

cmoussa1:
#### Problem

The flux-accounting guide could use more clarification on how it treats queues that it doesn't know about, e.g. queues that are dynamically added to Flux. The guide currently contains a small note, but does not link to the RFC that defines this behavior.

---

This PR changes the note about how flux-accounting treats unknown queues in the guide to a warning, links to RFC 33 as a reference for why flux-accounting implements the behavior it does, and clarifies the difference between queues being dynamically added to Flux versus configuring them in flux-accounting.


grondo (approved):
Great! LGTM.

cmoussa1:
Thanks! Setting MWP here

PR #912, opened 2026-07-21
`edit-config`: remove confirmation prompt

cmoussa1:
#### Problem

The `edit_config()` function calls the `input()` to confirm re-binning operations when changing `priority_usage_reset_period`, `priority_decay_half_life`, or `decay_factor`. This causes an `EOFError` when called from the flux-accounting service, which runs without a controlling terminal.

---

This PR removes the interactive prompt and apply config changes directly and moves the warning text to the command help in **flux-account.py** so users are still informed of the consequences. I also removed @patch decorators from a number of unit tests since `input()` is no longer called.

I've also added a simple sharness test to edit the usage parameters with the `edit-config` command in the flux-accounting service to confirm that no `EOFError` is raised.

### DIFF
--- a/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
+++ b/src/bindings/python/fluxacct/accounting/db_info_subcommands.py
@@ -358,21 +358,6 @@ def edit_config(conn, cursor, key_value_strings):
             raise ValueError(f"key {key} not found in config_table")
 
     if requires_rebin:
-        choice = (
-            input(
-                "WARNING: changing one or more of priority_usage_reset_period, "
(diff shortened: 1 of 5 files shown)


jameshcorbett (approved):
LGTM, although you could provide stdin from the accounting service couldn't you? If that's the only motivation for this change.

cmoussa1:
You could, but I think this command being the only one that requires further input is a little clunky, and since this is an admin-only command, I think it can probably come with the expectation that this command should be executed carefully.

jameshcorbett:
roger roger big boss

PR #913, opened 2026-07-22
plugin: add option to toggle rejecting unknown queues during job validation

### LINKED ISSUE #909: plugin: add an `always-deny` option to deny queues that flux-accounting does not know about
@grondo had mentioned in #831 that perhaps flux-accounting should provide an option to change its behavior on how to treat queues not configured by flux-accounting. Currently, flux-accounting does not reject jobs if it is submitted to a queue that it does not know about, which complies with behavior defined in [RFC 33](https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_33.html):

> Access policy MAY be extended by jobtap plugins that enforce additional access conditions. For example, the flux-accounting multi-factor priority plugin controls access to queues based on the user and bank information from the accounting database. When multiple access validators are active, a job request SHALL be granted access only if every validator grants access, so the effective access policy is the intersection of all configured sources. It is RECOMMENDED that access for a given queue be configured in a single source.

However, some sites may want tighter control over how to handle queues unknown to flux-accounting. If we offered an option to toggle this on and off, flux-accounting could then reject jobs submitted to a queue that it does not know about.

One potential way to support this is to offer an `always-deny` key-value pair that can be toggled on/off in flux-accounting's `config_table`. This could be forwarded to the plugin and stored as a boolean there, and depending on its setting, it can change its behavior on whether to accept or deny jobs submitted to queues that it does not know about.

However, as @grondo mentioned, this conflicts with RFC 33, which also mentions that other validators shouldn't reject jobs not under its management:

> A validator that manages access for a subset of queues SHALL NOT reject a job request solely because the named queue is not under its management. This permits queues to be introduced, for example at runtime, without registration in every validator’s configuration, with access to such queues governed by the remaining validators.

So, either the language there needs to be extended to make an exemption for flux-accounting, or perhaps accounting needs to be responsible for documenting that the `always-deny` option conflicts with the RFC, or something else.

EDIT: [WIP] branch can be found [here](https://github.com/cmoussa1/flux-accounting/tree/refs/heads/add.always-deny.option)

cmoussa1:
#### Problem

There is no way to change the behavior for flux-accounting to reject jobs that are submitted to a queue that it does not know about.

---

This PR adds a new built-in option to flux-accounting (stored in `config_table`) that can toggle this behavior on and off. By default, this feature is turned off, which maintains backwards compatibility and the priority plugin still allows jobs to be submitted to queues it does not know about. However, if turned on, the plugin will begin rejecting jobs during `job.validate` if it is submitted to a queue unknown to flux-accounting.

Toggling this option on and off does not require a plugin reload, but just a push of database information to the plugin via `flux account-priority-update`. However, tests are added that cover both cases: toggling the option and completely reloading the plugin, and toggling the option and updating the already-loaded plugin.

Assisted-By:Claude:sonnet-4.5

Closes #909

### DIFF
--- a/t/python/t1017_config_cmds.py
+++ b/t/python/t1017_config_cmds.py
@@ -192,6 +192,46 @@ class TestAccountingCLI(unittest.TestCase):
         with self.assertRaises(ValueError):
             d.edit_config(conn, key_value_strings=["i_dont_exist=foo"])
 
+    # trying to remove deny_unknown_queues option will raise a ValueError
+    def test_27_delete_deny_unknown_queues(self):
+        with self.assertRaises(ValueError):
+            d.delete_config(conn, key="deny_unknown_queues")
+
+    def test_28_edit_deny_unknown_queues_true(self):
+        d.edit_config(conn, key_value_strings=["deny_unknown_queues=true"])
+        test = cur.execute(
+            "SELECT value FROM config_table WHERE key='deny_unknown_queues'"
+        ).fetchone()
+        self.assertEqual(test["value"], "true")
+
+    def test_29_edit_deny_unknown_queues_false(self):
+        d.edit_config(conn, key_value_strings=["deny_unknown_queues=false"])
+        test = cur.execute(
+            "SELECT value FROM config_table WHERE key='deny_unknown_queues'"
+        ).fetchone()
+        self.assertEqual(test["value"], "false")
+
+    def test_30_edit_deny_unknown_queues_case_insensitive(self):
+        d.edit_config(conn, key_value_strings=["deny_unknown_queues=True"])
+        test = cur.execute(
+            "SELECT value FROM config_table WHERE key='deny_unknown_queues'"
+        ).fetchone()
+        self.assertEqual(test["value"], "True")
+
+    # trying to set deny_unknown_queues to invalid value will raise a ValueError
+    def test_31_edit_deny_unknown_queues_bad_1(self):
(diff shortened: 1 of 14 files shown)


t/python/t1017_config_cmds.py
```diff
@@ -192,6 +192,11 @@ def test_26_edit_config_nonexistent_key(self):
         with self.assertRaises(ValueError):
             d.edit_config(conn, key_value_strings=["i_dont_exist=foo"])
 
+    # trying to remove PriorityUsageResetPeriod will raise a ValueError
+    def test_27_delete_deny_unknown_queues(self):
+        with self.assertRaises(ValueError):
+            d.delete_config(conn, key="deny_unknown_queues")
```
grondo: The comment should mention `deny_unknown_queues` not PriorityUsageResetPeriod?

grondo (approved):
LGTM! One potential cut-and-paste error inline. Also, should the new option be documented somewhere?

t/python/t1017_config_cmds.py
cmoussa1: Thanks for catching this! Indeed a cut-and-paste error.

cmoussa1:
Thanks @grondo! I've force-pushed a fix based on your catch above. I've also added a commit to add some documentation to the accounting guide about toggling this option and off. Since you approved, I'll remove [WIP] from here and then set MWP

PR #914, opened 2026-07-22
`update-db`: improve `update_columns()`, add more logging during database update

cmoussa1:
#### Problem

The `update-db` command does not log a lot of information during a database update, which doesn't make it immediately clear what is being changed when a database is being updated.

---

This PR primarily adds a number of logs throughout **flux-account-update-db.py** to make it clearer its overall progress and what is being updated.

It also slightly improves the `update_columns()` helper function in the script to only update tables _if_ they actually have a difference in columns, and if not, it skips the update. Previously, the helper function used to unconditionally rebuild every table even when the schema didn't change.

The test that checks contents of log messages has been updated to look for a new log message (when the database update is complete), and there is also a slight fix to one of the tests that checks for what tables are present in a database to make it more robust (i.e. ignoring order of the test file).

### DIFF
--- a/t/t1017-update-db.t
+++ b/t/t1017-update-db.t
@@ -121,7 +121,10 @@ for db in ${SHARNESS_TEST_SRCDIR}/expected/test_dbs/*; do
 		test_expect_success 'update old DB: '$(basename $db) \
 			"flux account-update-db -v -p $tmp_db > update.test 2>&1"
 		test_expect_success 'verify INFO logs appear: '$(basename $db) \
-			"grep -E '(checking for new tables|checking for new columns|migration complete)' update.test"
+			"grep -E '(checking for new tables|\
+checking for new columns|\
+migration complete|\
+updated database schema version to)' update.test"
 		test_expect_success 'start flux-accounting service' \
 			"flux account-service -p $tmp_db -t"
 		test_expect_success 'add a bank: '$(basename $db) \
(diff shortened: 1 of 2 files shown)


t/t1017-update-db.t
```diff
@@ -121,7 +121,10 @@ for db in ${SHARNESS_TEST_SRCDIR}/expected/test_dbs/*; do
 		test_expect_success 'update old DB: '$(basename $db) \
 			"flux account-update-db -v -p $tmp_db > update.test 2>&1"
 		test_expect_success 'verify INFO logs appear: '$(basename $db) \
-			"grep -E '(checking for new tables|checking for new columns|migration complete)' update.test"
+			"grep -E '(checking for new tables|\
+				checking for new columns|\
+				migration complete|\
+				updated database schema version to)' update.test"
```
jameshcorbett: I think this may now be grep'ing for `"				checking for new columns"` instead of just `"checking for new columns"` because of all the whitespace introduced here. It may pass though because `"checking for new tables"` matches and so the overall OR chain is true!

jameshcorbett (approved):
LGTM! 🦾 💪 🦾

t/t1017-update-db.t
cmoussa1: Oh yeah, I'll go ahead and fix that. Good catch! Thanks

cmoussa1:
Thanks @jameshcorbett for reviewing! I've fixed the added check to not have a bunch of whitespace in between each message. Going to go ahead and set MWP here

PR #916, opened 2026-07-24
`update-db`: add copy of original DB to a `.backup` file

cmoussa1:
#### Problem

If database migration fails or contains logic bugs that only surface after commit, the original database state is lost with no way to recover.

---

This PR creates a `.backup` copy of the database before any modifications begin. If either `OperationalError` or `IntegrityError` is raised during a database migration, it restores the original database from the backup before exiting.

A simple test is added to **t1017-update-db.t** to check the existence of a `.backup` when the update is first started.

### DIFF
--- a/src/cmd/flux-account-update-db.py
+++ b/src/cmd/flux-account-update-db.py
@@ -343,6 +343,11 @@ def migrate_job_usage_to_per_assoc(cur):
 
 def update_db(path, new_db):
     LOGGER.info("starting database update for %s", path)
+    # create a backup of the database
+    backup_path = path + ".backup"
+    shutil.copy2(path, backup_path)
+    LOGGER.info("created backup at %s", backup_path)
+
     old_conn = est_sqlite_conn(path)
     old_cur = old_conn.cursor()
 
@@ -391,11 +396,11 @@ def update_db(path, new_db):
             new_db,
             exc,
         )
-        shutil.rmtree(tmp_dir_name)
+        old_conn.close()
         sys.exit(1)
     except sqlite3.IntegrityError as exc:
         LOGGER.exception("SQLite integrity error: %s", exc)
-        shutil.rmtree(tmp_dir_name)
+        old_conn.close()
         sys.exit(1)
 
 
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
Looks ok to me big boss. The backups aren't ever "cleaned up" though so the second commit message is a bit misleading when it says "there are no tests to verify that backups are
created or that they are cleaned up properly."

Also, there are no tests for restoring from a backup. Can you force a logic error somehow and check if the backup works correctly? I think as it stands now, this change doesn't accomplish much because if `IntegrityError` or `OperationalError` were raised before `commit`, the changes would have been rolled back by sqlite anyway right?

cmoussa1:
Thanks for catching that issue in the commit message @jameshcorbett - I'll go ahead and remove that from the message altogether. The `.backup` DBs are automatically cleaned up in the testsuite, but you are right, the backup isn't deleted if the update completes successfully. I guess the thinking here was that it would be an admin's responsibility to manually clean this up for this reason:

> I think as it stands now, this change doesn't accomplish much because if IntegrityError or OperationalError were raised before commit, the changes would have been rolled back by sqlite anyway right?

Yes, you are right that in the case of an actual exception during an update, changes are rolled back and the original database is left untouched. The intent with the `.backup` file is that it exists in the case there is a logic error during the migration that is _not_ caught with a Python exception. An example is in `0.59.0`, the `update-db` script had a logic error that ended up removing columns from a table that it couldn't recover from. With a `.backup` file, this would allow us to just revert to that `.backup` before the `update-db` command was called and address any issues before trying again. I hope my thinking makes sense.

src/cmd/flux-account-update-db.py
```diff
@@ -391,11 +396,15 @@ def update_db(path, new_db):
...
+        LOGGER.info("restoring database from backup %s", backup_path)
+        old_conn.close()
+        shutil.copy2(backup_path, path)
         sys.exit(1)
     except sqlite3.IntegrityError as exc:
         LOGGER.exception("SQLite integrity error: %s", exc)
-        shutil.rmtree(tmp_dir_name)
+        LOGGER.info("restoring database from backup %s", backup_path)
+        old_conn.close()
+        shutil.copy2(backup_path, path)
```
jameshcorbett: Based on your comment then, these `shutil.copy2` calls aren't really doing anything and the log message doesn't mean much.

PR #917, opened 2026-07-29
commands: add fair-share emulator

cmoussa1:
#### Problem

There is no way to calculate fair-share values from input data without a flux-accounting database. This makes it difficult
to test fair-share scenarios, preview configuration changes, or just to demonstrate the algorithm.

---

This PR adds a new command to the flux-accounting command suite, `fairshare-emulate`, which emulates fair-share calculation over a hierarchy of banks and users. It introduces a Python implementation of the weighted tree fair-share algorithm in `src/fairness/` to calculate these fair-share value for a given hierarchy and outputs the results in a number of optional formats. It does not require an active connection to the flux-accounting SQLite database, so it is implemented as a command separate from those present in the flux-accounting service.

A number of simple sharness tests are added to verify that the results calculated by the emulator line up with those calculated by the C++ implementation.

A man(1) page is also added with a more detailed description of the command as well as how to format JSON input for the command itself.

Closes #829

### DIFF
--- a/src/bindings/python/fluxacct/accounting/fairshare_emulator.py
+++ b/src/bindings/python/fluxacct/accounting/fairshare_emulator.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+
+###############################################################
+# Copyright 2026 Lawrence Livermore National Security, LLC
+# (c.f. AUTHORS, NOTICE.LLNS, COPYING)
+#
+# This file is part of the Flux resource manager framework.
+# For details, see https://github.com/flux-framework.
+#
+# SPDX-License-Identifier: LGPL-3.0
+###############################################################
+import sys
+import json
+
+
+class FairShareNode:
+    """Represents a node in the fair-share hierarchy."""
+
+    EPSILON = 1e-9
+    MAX_UINT64 = (1 << 64) - 1
+
+    def __init__(self, name, shares, usage, bank=None, is_user=False):
+        self.name = name
+        self.bank = bank
+        self.shares = shares
+        self.usage = usage
+        self.is_user = is_user
+        self.children = []
+        self.weight = 0.0
+        self.fairshare = 0.5
+        self.tie_with_next = False
+
+    def add_child(self, child):
+        self.children.append(child)
+
+    def calculate_children_weights(self):
+        """Calculate weights for all children based on shares/usage ratios."""
+        if not self.children:
+            return
+
+        # sum shares and usage across siblings
+        sibling_shares_sum = sum(c.shares for c in self.children)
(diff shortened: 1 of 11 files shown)


src/bindings/python/fluxacct/accounting/fairshare_emulator.py
```diff
@@ -0,0 +1,357 @@
...
+
+        # For each child bank, sort and prepare its users first
+        for child in node.children:
+            if not child.is_user:
+                child.sort_children_by_weight()
+                child.mark_ties()
+
+        # Then sort and mark ties at this level
+        node.sort_children_by_weight()
+        node.mark_ties()
```
jameshcorbett: This function only deals with a tree of depth at most two, but couldn't the tree be of arbitrary depth?

src/cmd/flux-account-fairshare-emulate.py
```diff
@@ -0,0 +1,98 @@
...
+    else:
+        format_string = args.format_string
+
+    try:
+        output = fairshare_emulator.format_results(
+            users, json_fmt=args.json, format_string=format_string
+        )
+        print(output)
+    except ValueError as exc:
+        print(f"fairshare-emulate: ValueError: {exc}")
```
jameshcorbett: unlike all the other `print` statements, this one goes to stdout instead of stderr.

t/t1098-fairshare-emulate.t
```diff
@@ -0,0 +1,380 @@
...
+	}
+	EOF
+'
+
+test_expect_success 'deeply nested hierarchy calculates correctly' '
+	flux account-fairshare-emulate \
+		-i deeply_nested.json -o "{username},{fairshare}" > deeply_nested.test &&
+	cat <<-EOF >deeply_nested.expected &&
+	deep.user1,1.0
+	deep.user2,0.5
```
jameshcorbett: These two users have the same shares and usage, why is their fairshare different?

jameshcorbett (commented):
Looks reasonable but I have some questions?

src/bindings/python/fluxacct/accounting/fairshare_emulator.py
cmoussa1: Ah, yeah, you're right. Great catch! I think this needs to be a recursive call instead of stopping at the first level. Changing this to a recursive call also fixes the tied fair-share test you pointed out below. I will fix this

cmoussa1:
Thanks @jameshcorbett! I think I've addressed your catch above and have force-pushed up the changes

jameshcorbett (approved):
LGTM thanks for updating it!

PR #918, opened 2026-07-30
`export-db`: add `-F/--fairshare-emulate` option

cmoussa1:
#### Problem

The `fairshare-emulate` command proposed in #917 takes in a JSON-formatted file representing a database hierarchy from a flux-accounting database, but there is no convenient way to export an *actual* database's bank and association information and use it as input to the command.

---

This PR (built on top of #917) adds a new option to the `export-db` command, `-F/--fairshare-emulate`, that exports and formats an actual flux-accounting database into the expected JSON format for the `fairshare-emulate` command. This allows for an administrator or user to export an actual database and edit job usage values and emulate fair-share runs without having to actually rack up usage and run the fair-share calculations on a production database.


src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -130,6 +130,56 @@ def export_db_info(conn):
...
+@with_cursor
+def export_db_as_fairshare_json(conn, cursor):
+    """
+    Export database as hierarchical JSON for the fairshare-emulate command.
+    """
+    # fetch all active banks and associations
+    cursor.execute("SELECT bank,parent_bank,shares,job_usage FROM bank_table")
+    banks = cursor.fetchall()
+    cursor.execute("SELECT username,bank,shares,job_usage FROM association_table")
+    associations = cursor.fetchall()
```
jameshcorbett, line 142: Should this filter for only active entries?

t/t1099-export-db-fairshare.t
```diff
@@ -0,0 +1,81 @@
...
+	grep "\"username\": \"user2\"" export.json &&
+	grep "\"username\": \"user3\"" export.json
+'
+
+test_expect_success 'exported JSON works with fairshare-emulate' '
+	flux account export-db --fairshare-emulate > hierarchy.json &&
+	flux account-fairshare-emulate -i hierarchy.json --json > results.json &&
+	grep "user1" results.json &&
+	grep "user2" results.json &&
+	grep "user3" results.json
```
jameshcorbett: should this check more than just users?

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
```diff
@@ -130,6 +130,56 @@ def export_db_info(conn):
...
+
+    # build hierarchy by linking banks to parents
+    root_node = None
+    for row in banks:
+        bank_name = str(row["bank"])
+        parent_bank = str(row["parent_bank"]) if row["parent_bank"] else ""
+
+        if parent_bank == "":
+            # root bank
+            root_node = bank_nodes[bank_name]
```
jameshcorbett, line 172: maybe raise an error if `root_node` is not None here? in case there are multiple roots.

jameshcorbett (approved):
Looks reasonable, pending answers to my questions about #917.

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
cmoussa1: In the case where a user or bank has accumulated usage within the current usage period and is then deactivated, their usage actually still contributes to fair-share calculation for the rest of the tree, which is why I did not filter for only active entries.

t/t1099-export-db-fairshare.t
cmoussa1: Hm, I guess the point of this test was just to make sure that all three users exported were also included in the output of `fairshare-emulate`. Although, since they don't have any usage for the purposes of this test file, all of their fair-share will be equal, which could be a valuable thing to check. I can go ahead and add that check!

src/bindings/python/fluxacct/accounting/db_info_subcommands.py
cmoussa1: There already exists a check in the [`add-bank`](https://github.com/flux-framework/flux-accounting/blob/master/src/bindings/python/fluxacct/accounting/bank_subcommands.py#L205) command to make sure that only one root bank ever exists in a database hierarchy, so I _think_ this should be safe to keep as is.

cmoussa1:
Thanks for the review @jameshcorbett! I believe I've addressed your suggestions above, so I'll go ahead and set MWP here

